PR #16, opened 2016-01-31
TRO 3058u Source updates

pheonixstorm:
Only found two incorrect dates. All units now have a proper source listed.


HammerGS:
Merged.  Thanks.

PR #29, opened 2016-02-12
removed a redundant import

nderwin:
Checkstyle caught this error when I ran it on my fork.


pheonixstorm:
Something this small might have been better served as a patch, just my 2 cents.

nderwin:
True, but it's easier to hash out the process with a simple change, and the devs don't have to bother with creating a branch, applying the patch file, etc.  

The next one will be much larger.  8-)

arlith:
Yea, pull requests are way easier imo.  Especially for small things like this.  I can do it completely through the web interface if I want...

PR #30, opened 2016-02-13
Tabs to spaces

nderwin:
Based on forum postings (http://megamek.info/forums/index.php?topic=2297.msg15084#msg15084) & (http://megamek.info/forums/index.php?topic=2302.msg15184#msg15184), and running Checkstyle against the source code, multiple files were found to contain tabs instead of spaces for indentation.

This pull request represents running a regex search and replace to turn '\t' into '    ', against all .java files in the source directory.

It's easier to see the differences in the Unified view instead of the Split view.


arlith:
In this case, git diff -w is a huge help.

PR #34, opened 2016-02-20
Shadows

SJuliez:
Hiya,

this branch will add a cast shadow to hex elevations, buildings(!) and trees. In addition, four new checkboxes in the graphics tab of the Client Settings allow switching of cast shadows, hex AO shadows (the ones that could formerly be activated from the advanced settings), the thick line level borders and "floating" isometric mode (that does not draw the hex sides in iso mode). Everything's optional as usual.

The cast shadows use the hex buffer as usual and do not impact performance at all so long as the board stays the same. The shadowmap itself is one board-sized image and thus may use a few MB.

The shadows do not work well in all situations (e.g. when there is a building and a road in the same hex) but I checked almost every map during testing and errors are actually very rare (and not so very visible in the end).

I know that this isn't helping with pilot special ability problems, but it was fun to do and I think it does give some depth, esp. with buildings. Personally I prefer the map to look as un-technical as possible, hurting the eye with hex borders and such stuff only as far as necessary. That's why I dislike the official maps and why I tend to work on these things.

Have fun!
S

### DIFF
--- a/megamek/i18n/megamek/client/messages.properties
+++ b/megamek/i18n/megamek/client/messages.properties
@@ -628,6 +628,10 @@ CommonSettingsDialog.tooltipDismissDelay=Tooltip dismiss delay:
 CommonSettingsDialog.tooltipDismissDelayTooltip:Sets the delay before a tooltip is dismised.  -1 uses the default for your look and feel.
 CommonSettingsDialog.Update=Update
 CommonSettingsDialog.showWpsinTT=Show unit weapons in the tooltip
+CommonSettingsDialog.AOHexSHadows=Show ambient-occlusion-like hex shadows at the base of hills
+CommonSettingsDialog.useShadowMap=Show terrain and building shadows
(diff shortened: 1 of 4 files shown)


SJuliez:
Ah I forgot, the light direction and length of shadows can be changed easily but I hardcoded the values for now. (currently it's { -19,7}) :)

arlith:
I think this ended up looking nice.  What about shadows for units? Seems like you could probably use a similar concept for woods shadows to create AO shadows for units.  It would also be nice to have a slider to adjust direction and distance for the light source.  You might also want to add consideration for light conditions.  The shadows might look a little weird if someone is playing at night.  You could potentially use the collection of illuminated hexes to conditionally draw shadows as well.  That might lead to a really nice looking effect (although it might look odd as well, depending on the light source and light direction).

Those are mostly suggestions on top of what's here though.  I'll merge in what's already here.

PR #35, opened 2016-02-21
Maven

Kurios:
So I've split the project into core and client, so I could make them 2 separate maven projects. 

I also had to strip out tinyXML, as it does not exist in the Maven repos to take a dependency on. I replaced it with JDOM.

So far via hand testing, the game loads saves and loads units, along with configs. So it seems there are no problems there.

This way, things like MekHQ only need to take a dependency on the common project. 

For the most part, nothing changed namespace wise, with the exception of a small handful of files being moved around between client and common domains, and the creation of a few interfaces so there would be no problem with the separation. (the code was coupled in a few places. Thats fixed now.)  And I'm sure there's mountains of questions.

Next on my list is integrating that fancy launch4j thing via maven (https://github.com/lukaszlenart/launch4j-maven-plugin). 

Assuming that's done, the project is a lot closer to being ready for CI/automated testing with something like Jenkins and other madness, like daily dev builds.


Dylan-M:
Arlith and I have already discussed doing this, and we won't be doing it using JDOM. We'll be using the xstream library that we already use for savegames and the MULParser.

Additionally, we'll be continuing to use ant for this project. A multi-jar set can be created using ant as well.

Finally, we're not going to make big changes like this while in a feature freeze trying to reach a stable release.

PR #39, opened 2016-02-26
Configurable data directory

Akjosch:
This commit:

* adds a command line option "-data" to set the data directory used by the game (defaults to "data")
* fixes the `megamek.common.QuirksHandler` class not using `Configuration.dataDir()`

This is useful for developing and keeping multiple (custom) variants of the data directory for different games.


Akjosch:
As a warning: This change will not carry over to the meklab and mekhq projects, since they don't use the same configuration architecture. The changes to make it work there aren't too hard to make, but I can't include them within this pull request.

Dylan-M:
You shouldn't make multiple changes in a single pull request, even if they're related. Can you please close this request and split those changes into 2?

Akjosch:
Done.

Mind you, without the fix in QuirksHandler the configurable data directory feature is as likely to crash on start as not.

PR #44, opened 2016-02-27
Bug: Ultra Rubble terrain handled incorrectly for movement and PSR

pheonixstorm:
Ultra Rubble now checks for the correct level for movement cost and PSR
modifier. By changing this I had to remove a section of checkRubbleMove
so it would not double the modifier.

### DIFF
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -6511,12 +6511,7 @@ public abstract class Entity extends TurnOrdered implements Transporter,
                 || isLastStep)
             && (curHex.terrainLevel(Terrains.RUBBLE) > 0)
             && (this instanceof Mech)) {
-            int mod = 0;
-            if (curHex.terrainLevel(Terrains.RUBBLE) > 5) {
-                mod++;
-            }
-            // append the reason modifier
-            roll.append(new PilotingRollData(getId(), mod, "entering Rubble"));
+            
             adjustDifficultTerrainPSRModifier(roll);
         } else {
             roll.addModifier(TargetRoll.CHECK_FALSE,
--- a/megamek/src/megamek/common/Terrain.java
+++ b/megamek/src/megamek/common/Terrain.java
@@ -296,7 +296,7 @@ public class Terrain implements ITerrain, Serializable {
             return 1;
         case Terrains.GEYSER:
         case Terrains.RUBBLE:
-            if (level == 2) {
+            if (level == 6) {
                 return 1;
             }
             return 0;
@@ -329,6 +329,12 @@ public class Terrain implements ITerrain, Serializable {
             }
             return 0;
         case Terrains.RUBBLE:
+            if (level == 6) {
+                if ((e instanceof Mech) && ((Mech)e).isSuperHeavy()) {
+                    return 1;
+                }
+                return 2;
+            }
             if ((e instanceof Mech) && ((Mech)e).isSuperHeavy()) {
                 return 0;
             }


pheonixstorm:
Note: Once this is accepted I can start on refactoring the terrain psr modifiers for difficult terrain so they will actually show what difficult terrain is being applied rather than the blanket "+n difficult terrain"

arlith:
Hmmm, I'm wondering if line 6519 in Entity should not be removed.  Always leave the mod to zero, so you get the message.

The reason I think this is because, when you get the PSR nag, the "+0 (entering rubble)" bit can be useful.  It's the only real indication as to *why* you need to make the PSR.  If you left the line in there, with the mod always being zero, when entering ultra rubble you'd get a message like: "5 (base piloting) + 0 (entering rubble) +1 (difficult terrain)".

pheonixstorm:
I thought about that but decided not to leave it since it should not take long to get everything else worked out. Since this was a bug I did not know if it should have waited, at least for the movement part of the fix.

I can remove the PSR portion of the fix and just make that part of the terrain PSR fix if you want.

arlith:
Ok, a few issues with this PR.  First, there's some minor code style violations.  The policy in MM is to surround binary operators in conditions with parenthesis.  For instance,  the "game != null" checks should be surrounded by parens.  Don't worry about fixing that for this PR; I've got a fix locally, and there were other formatting changes I wanted to make in the area, however I believe for you to incorporate those, I would have to make a pull request to your fork.  I'd much rather just commit the changes locally and then merge them in when I merge this PR.  However, before I do that, I'd like you to fix this bug I found (I worked in QA too ;).

Here are the steps to reproduce:
- Enter the lounge and turn on TW BA weights
- Add a BA and a transporter (I tested with an Achielus and a Goblin Infantry Support Vehicle (Standard), the one with 3 tons of troop transport)
- Mount the BA into the transporter
- Turn off TW BA weights
- Dismount the unit
- Observe that the transporter now has extra cargo space

The problem is in TroopSpace, particular line 158.  What it's doing is, when a unit is unloaded, it "frees up" troop space equal to the units current weight.  Basically, the unit is loaded at 2 tons, then unloaded at 4 tons, yielding a "free space" of 5 tons after unloading.

What I would do to fix this is add a Map to TroopSpace that maps an Entity ID to a tonnage, and when an entity is loaded, keep track of what weight it was loaded in, so that when the Entity is unloaded, you can free up the loaded amount.  Units that are deleted in the ChatLounge will be unloaded from their loading unit, so we shouldn't run into issues where there's an Entity ID mismatch.

arlith:
Ah, I think I failed to read what you wrote in your second comment... where you said you'd refactor the code to indicate what is causing the difficult terrain.

Regardless, I would still like to see some change.  I'll attach a saved game I used to test (player name is Arlith).

In the saved game, there's an Archer in front of some rubble.  If I move through the rubble prior to the PR, I'll get a Nag about a PSR that says something like "[ 5 (Base Piloting) + 0 (entering Rubble)]." This gives me some indication as to why the PSR is required.  After the PR the message is just "[5 (Base Piloting)]" which doesn't give me any real indication.  Since rubble doesn't have a difficult terrain modifier, you don't see that.  Additionally, the difficult terrain modifier is different from the "entering Rubble" PSR.  The difficult terrain modifier applies to *any* PSR the unit makes while in that hex (for instance, if the unit had to make a PSR from 20+ damage).

My suggestion would be to modify the change to Entity so that there is always a 0 modifier added and it will have a message of "(entering rubble)" (you could have a different message for ultra rubble if you were feeling fancy).  That way you will get PSR detail messages like:
"[ 5 (Base Piloting) + 0 (entering Rubble)]"
and
"[ 5 (Base Piloting) + 0 (entering Rubble) + 1 (\<Difficult Terrain\>)]"
where \<Difficult Terrain\> represents whatever that text is, either the current text or some refactoring you submit a PR for.  Without this, even if you updated the difficult terrain text, for the "entering" PSR, you'd still have a message like
"[ 5 (Base Piloting) + 1 (Ultra Rubble)]"
which still doesn't really indicate *why* the PSR is being made (aka, "entering Rubble").

[test.sav.gz](https://github.com/MegaMek/megamek/files/156093/test.sav.gz)

pheonixstorm:
Yeah, getting a fix so that a normal PSR would only have +1 rubble or +n entering rubble was to be part of the overall fix.

When I looked into the PSR checks I noticed it was calling one function (in IHex I think) that called the move modifier in terrain. The extra function call from entity really isn't needed but I don't know if the call to IHEX is required or not (doesn't look that way).

The PSR changes in this PR were more quick fix on the off chance a new release was made before I could get the rest of the work done. 

So now its more of "what would you prefer". Should I uncomment line 6514 & 6519, the entering rubble, or remove the PSR fix and just have the fix to movementCost?

pheonixstorm:
Closing this as the Move cost fix is a separate PR. Beginning work on the pilotingModifier fix/rework

PR #45, opened 2016-02-27
SPA Organization

pheonixstorm:
This should now include all official and unofficial SPAs in (mostly)
alphabetical order.

All SPAs not currently coded are commented out. They are also sorted based on type such as Piloting, Gunnery, and Misc

### DIFF
--- a/megamek/src/megamek/common/options/PilotOptions.java
+++ b/megamek/src/megamek/common/options/PilotOptions.java
@@ -34,31 +34,69 @@ public class PilotOptions extends AbstractOptions {
     @Override
     public void initialize() {
         IBasicOptionGroup adv = addGroup("adv", LVL3_ADVANTAGES); //$NON-NLS-1$
+        
+        // Piloting Abilities - Just uncomment and run with it
+        // addOption(adv, "animal_mimic", false); //$NON-NLS-1$
+        // addOption(adv, "cross_country", false); //$NON-NLS-1$
         addOption(adv, "dodge_maneuver", false); //$NON-NLS-1$
+        // addOption(adv, "dust_off", false); //$NON-NLS-1$
+        // addOption(adv, "hvy_lifter", false); //$NON-NLS-1$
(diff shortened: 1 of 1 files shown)


arlith:
I don't really like the idea of commenting out SPAs that have already been in a release, even if they aren't implemented.  It's not clear to me how MULs and MHQ campaigns that have these SPAs on will be handled if the SPAs are commented out.  

I understand it's nice to know which ones are and are not implemented, but I think a better way to handle this would be via i18n entries.  We could update the name entry to have an asterisk, and the tool tip to indicate that the asterisk means it's not implemented.

arlith:
I'm not sure if you can update the pull request, or if you need to make a new pull request...

djmbm:
You could also disable them so the options can't be changed...

pheonixstorm:
Depends, not sure I follow though as all the new SPA entries should not even load into a MUL or HQ. 

In fact I got this idea from what you guys did with the commented out quirks. (though don't know which devs or how far back those really go).

The fear of having working SPAs commented out is incorrect. None of the addition have ever been in MegaMek and are only a full listing of what **will** be included into MM over time such as Golden Goose, Ground Hugger, Shaky Stick, and Terrain Master (3 so far). All of which I have working in some form but still tinkering and debugging them.

This was just to make future SPA work easier for everyone to track which SPAs still needed to be worked in (just like the commented out bits in quirks) so its not an important pull request. Just let me know which way I should take this. If I have to rework the branch I don't mind, gives me reason to continue playing with and learning git.

arlith:
Ah, I didn't pay attention to the names of the commented out quirks.  I understand now.  I thought you were commenting out quirks that were already there, but unimplemented (of which there are a few).  Instead, you've added in lines for quirks that are not yet there, but are in official sources.

arlith:
Another additional thing that could be useful with this would be removing all of the string literals.  Like with the natural aptitude entries, instead of having strings use a constant, such as OptionsConstants.PILOT_APTITUDE_GUNNERY.

Issue #42, opened 2016-02-26
Battle Armor Weights Not Adjusted for Weight Class/Damage?


mkerensky:
Aren't Battle Armor squads supposed to weigh different amounts depending on the class of suit?  Right now it seems that all squads weigh 1 ton per member and retain their full weight values even when squad members are lost.

arlith:
Those are the TW rules.  What you have in mind is an advanced option for TO.  Pretty sure there's an RFE for that optional rule.

mkerensky:
Ok, thanks

pheonixstorm:
Actually I have had that working for awhile MM forum post here http://megamek.info/forums/index.php?topic=2054.msg13357#msg13357

Working branch https://github.com/pheonixstorm/megamek-storm/tree/TacOps-BAWeights

There are a few things that were brought up I still need to work on but for the most part it works. Once it is complete I will put up a pull request for the devs.

arlith:
Meant to close this before.  It's a TO option, and will get implemented soon.  It's not really a bug, and there's a PR, so I don't think we need this issue.

PR #46, opened 2016-02-27
TacOps Alternate BattleArmor weights

pheonixstorm:
UnitSelectorDialog, Lobby, and in game all show correct tonnage for BA

Satisfies #42 as well as duplicates on SF.net and completes pheonixstorm/megamek-storm#3

Can now select between using Total Warfar tonnage (1t per trooper) or the Tactical Operations Alternate Battle Armor Weights p186/187


arlith:
Ok, a few issues with this PR.  First, there's some minor code style violations.  The policy in MM is to surround binary operators in conditions with parenthesis.  For instance,  the "game != null" checks should be surrounded by parens.  Don't worry about fixing that for this PR; I've got a fix locally, and there were other formatting changes I wanted to make in the area, however I believe for you to incorporate those, I would have to make a pull request to your fork.  I'd much rather just commit the changes locally and then merge them in when I merge this PR.  However, before I do that, I'd like you to fix this bug I found (I worked in QA too ;).

Here are the steps to reproduce:
- Enter the lounge and turn on TW BA weights
- Add a BA and a transporter (I tested with an Achielus and a Goblin Infantry Support Vehicle (Standard), the one with 3 tons of troop transport)
- Mount the BA into the transporter
- Turn off TW BA weights
- Dismount the unit
- Observe that the transporter now has extra cargo space

The problem is in TroopSpace, particular line 158.  What it's doing is, when a unit is unloaded, it "frees up" troop space equal to the units current weight.  Basically, the unit is loaded at 2 tons, then unloaded at 4 tons, yielding a "free space" of 5 tons after unloading.

What I would do to fix this is add a Map to TroopSpace that maps an Entity ID to a tonnage, and when an entity is loaded, keep track of what weight it was loaded in, so that when the Entity is unloaded, you can free up the loaded amount.  Units that are deleted in the ChatLounge will be unloaded from their loading unit, so we shouldn't run into issues where there's an Entity ID mismatch.

pheonixstorm:
> First, there's some minor code style violations
There is no policy on the wiki, please post one. :smiley:

As for that one in a million bug... Will see what I can do.

arlith:
There's 15+ years of history! ;)
But, I know, there's a lot of policy that needs to go up on the wiki.  Only so much time in the day...

pheonixstorm:
Quick question. So any time there is multiple checks such as game ! null && something != condition && some other check, do all of them need to be in parentheses or just null checks?

I would think the above would not need them but game ! null && (something != condition || something else != condition) would need it. Just trying to understand the reasoning for it.

pheonixstorm:
Another question. Would replacing the Vector<Integer> troops = new Vector<Integer>(); with Map<Integer, Float> troops = new HashMap<>(); work? The vector does not appear to be locked or used by any class or function outside of TroopSpace

arlith:
RE: Code formatting - All binary conditionals (such as X == Y, X != Y, X > Y, etc..) should have parens.  Unary conditions (like !X or X), don't need parens.  If you look around in the code, you should see examples of this... 

For your other question, yes it seems like changing the Vector to a Map would work.  I hadn't noticed the troops Vector when I first looked at the code.

Akjosch:
The "modern" recommended Java Collections version of the `Vector` is an `ArrayList` if you need to keep and manipulate the ordering manually, `HashSet` (with corresponding implementations of both `equals()` and `hashCode()` - *never* implement just one!) if the order isn't important and `TreeSet` if the order is important and depends either on the internal ordering of the items (then the classes should be implementing `Comparable<T>`) or some externally specified `Comparator<T>` .

arlith:
Yes, but in this case we want some type of Dictionary to map Entity Ids to weights.  It's not about changing the Vector to another List, but changing the List to a Dictionary because now we want to associate values.

pheonixstorm:
I have the map partially finished, just have a few more spots to fix up and I will add the commit. Hopefully it will be ready tonight.

pheonixstorm:
Map is finished, testing to find to the simpler bugs (NPEs and the like) and make sure it all works.

Hopefully you won't find any more bugs and this PR can be merged.

pheonixstorm:
@arlith It is ready for a final review when you get a chance

PR #56, opened 2016-03-03
Hidden units (double blind) will now be drawn translucent in Iso mode

SJuliez:


arlith:
I did notice that there's a slight difference in the transparencies between isometric and non-iso for unseen units, but it wasn't major.

HammerGS:
I might have discovered a memory leak relating to the:
+ Feature: Added ambient-occlusion like shadows, see client settings graphics feature.

To reproduce use the following:

1. Use the attached Client Settings
2. Load the 80x80 Map -DevilsTower3078
3. I used an Agrotera AGT-1A mech.
4. Select I'm Done then Done.
5. On my machine I experience a long load time - My CPU spikes to between 25%-30%
6. loading
7. Use the 'S' key to scroll to the bottom of the map (extremely slow for me, chunky - FPS hover around 650)
8. Deploy the mech in Hex 3080
9. Hit Done
10. There is a extremely long freeze on my machine (5-7 seconds) then I can move the map. FPS hovers between 100-150.
11. Click to move the mech straight ahead for maximum movement and hit done. The weapon selector will open but the unit won't have moved for a few seconds.  It then jumps ahead to show its move complete.
12. During the targeting phase the FPS is around 60-70
13. If don't fire and just click done through the phases
14. There is a long pause before the map reloads.
15. Start of the next turn the FPS starts around 900 then gradually speeds up.
16. Anytime I experience a slowdown the task manager shows a spike in the CPU to the low 20% and memory expands top



 around 1100 MB.

Log file attack but doesn't show any errors or anything unusual.

[megameklog.txt](https://github.com/MegaMek/megamek/files/160759/megameklog.txt)

Clientsettings attached (change extension to xml)
[clientsettings.txt](https://github.com/MegaMek/megamek/files/160743/clientsettings.txt)

Save game...needs to be renamed.
[autosave(sav.gz).txt](https://github.com/MegaMek/megamek/files/160762/autosave.sav.gz.txt)

pheonixstorm:
I think we need to lock the refresh rate to the monitors v-sync. An FPS of 900 is insane and a complete waste of CPU cycles since most monitors can only refresh at a fraction of that rate.

arlith:
It's not FPS, it's mislabeled.  It's ms taken to redraw.  Swing is designed to only redraw the scene when necessary.

The mislabeling is my fault because there's an "FPS" indicator, but, as I said, it's not FPS but the time it takes to draw the boardview in ms (actually, averaged over 10 redraws).

SJuliez:
I just tried what you did. Funny I cant find an Agrotera but I guess the Mech wont be the issue.
OK what I found:
a) the load time is long, around 10s on my machine too. I did it once with the shadowmap on and once without, makes no discernible difference (counting seconds in my head). Board prep with large boards is generally slow, I guess. A 16x17 board has around 270 hexes, an 80x80 has 6400, about 25 times as many. I will take a look if the long prep time is all due to drawHex().
b) scrolling down when deploying: Drawing the yellow hex borders during deployment is a direct drawline() call or so, not a sprite and is rather sluggish. Not really noticeable when you limit deployment to a border but for the whole board (or the visible part) it becomes noticeable.
b2) also, scrolling with the keys seems not as smooth as scrolling with the mouse. Maybe boardview1 doesnt like it when key events are fired faster than the board can redraw?
c) moving the unit does take long, that is really weird. Also, unplayable.
d) In all cases though the board redraw when pulling with the right mouse button was completely smooth.

HammerGS:
Thanks for checking, let me know if you need any more information.

pheonixstorm:
that makes more sense arlith, thanks

PR #61, opened 2016-03-05
Extract unit nicknames and (when present) display them instead of Mech

SJuliez:
... type+chassis

see my comment below

### DIFF
--- a/megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
+++ b/megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
@@ -794,6 +794,9 @@ public class CommonSettingsDialog extends ClientDialog implements
         cs.setUseAverageSkills(useAverageSkills.isSelected());
         cs.setGenerateNames(generateNames.isSelected());
         cs.setShowUnitId(showUnitId.isSelected());
+        if ((clientgui != null) && (clientgui.bv != null)) {
+            clientgui.bv.updateEntityLabels();
+        }
 
         cs.setLocale(CommonSettingsDialog.LOCALE_CHOICES[displayLocale
                 .getSelectedIndex()]);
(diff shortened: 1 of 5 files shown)


arlith:
These commits should really be two separate pull requests.  It's  good practice to have separate issues have separate tickets.

SJuliez:
How do I do that without creating a new branch?

pheonixstorm:
I don't think it's possible

arlith:
I'll merge it this time, but please keep it in mind in the future.  It can potentially make hunting down issues easier in the future.

arlith:
So, MekHQ allows users to specify a "fluff" name for units.  Unfortunately, this "fluff" name is only stored in the MHQ Unit class (which wraps an Entity).  Really, it'd be nice to see that fluff name pulled into Entity, and it could be displayed in a similar fashion.  Combine this with a new context menu in the ChatLounge for setting the fluff name... It could be pretty convenient.  That's basically an add-on idea to this PR.

Beyond that and my previous mention about two issues in one PR, everything else is fine.

SJuliez:
OK will do. :) Im getting better at adding branches and pushing them in Eclipse.



________________________________
Von: Nicholas Walczak <notifications@github.com>
Gesendet: Sonntag, 6. März 2016 13:27
An: MegaMek/megamek
Cc: SJuliez
Betreff: Re: [megamek] Extract unit nicknames and (when present) display them instead of Mech (#61)


Merged #61<https://github.com/MegaMek/megamek/pull/61>.

-

SJuliez:
I did take a look at a MUL export file and tried changing things. It seems that MM identifies the unit by its name and chassis, so just adding in a nickname there won't work I guess. The pilot already uses nick="xyz". Might be the way to do it.

@Dylan-M 
I was btw surprised to see that many of the MWO hero mechs don't seem to be in the database although (as a friend of mine said) those are all based on "real" in-universe mechs.

pheonixstorm:
Most hero mechs have fluff entries in various TROs. BUT, they are typically just stock units and nothing special.

PR #65, opened 2016-03-06
Tilesets can use an "include" keyword.

SJuliez:
- the .tileset files can now use the command `include "othertileset.xyz"`;
- MM will recursively read in included files (to a max depth of 100 includes to prevent freezing), so the included file is used as if its contents were at the place of the include command;
- all tilesets are updated to make use of it 

Advantage: tilesets are much more readable and editable.


Dylan-M:
Now, why the hell didn't I think of this? Great idea. I'm all for merging if Arlith gives this a thumbs up also.

arlith:
Yea, I thought it was a good idea too.  I'm having Hammer test some of the changes, just to make sure.  It's a large diff to go through, since a lot of stuff got moved from the tilesets that are there to tileinc files.  This will probably get merged later tonight.

PR #66, opened 2016-03-06
Theme replace

SJuliez:
A new menu item (in View) allows changing the tileset theme for the whole current board. Allows using that battleforce2.board as a grass map or even mars map.


arlith:
Might be convenient to make it a combo box instead of a text box.  You could update the TilesetManager to keep track of the themes that are specified in the current tileset, and then display those as a list instead.  What do you think?

Dylan-M:
I too would prefer the combo-box. Otherwise, great idea!

SJuliez:
Updated!
To be truthful, it's just a dropdown menu, not a combobox...

arlith:
"drop down menu" is more of a colloquial name for a combobox.  I think many UI libraries use some form of "combobox", in Java's swing it's JComboBox.  Looking at the diff, you did pretty much exactly what I expected.  I'll test it out tonight and probably merge it shortly after.

arlith:
After testing, I've got mixed opinions on this.  Themes are something that happen on a per-hex basis.  I can see this menu item being useful, but it can also irreversibly change the current board.  I kind of feel like the hexes should save their "no theme" state and when "no theme" is selected, they should revert to that.  Otherwise, if you have multiple themes in a board, you can't get them back.  On the other hand, this would remove the ability to "turn off" a theme on a map.  There should almost be a "Board  Theme" and "No Theme" option.

It would also be convenient to keep track of the currently selected theme, so that it's selected as the default option in the JOptionPane.

SJuliez:
Hmm given the current board files this seems like a rather theoretical issue. I can see though that the feature should be properly done. 

While talking about philosophy I would rather discourage using different themes in different hexes precisely because it makes a theme changer problematic. Ideally different styles of roads would be entered as roads:1 and roads:2 as is already done. If someone wants to depict some kind of a grasslands isle in an otherwise clear surrounding, that should ideally be done as clear:1 and clear:2 (does that even work?) because then a theme changer can change this to a different setting like moon with lets say a dusty surface and a rocky surface or another planet where the grass is red and clear is grey. And to the game, that is excatly what they are - two different kinds of clear. Using different themes in different hexes will make the theme changer only erase the difference and thus reduce flexibility.

Only my 2 cents.

SJuliez:
Ah well. Woods:1 and :2 have gameplay difference, but (I think) the roads don't. And the roads are actually used much more commonly than any theme, I think :smile:. 
To expand, ideally we'd want a way to give graphical but not gameplay distinction to hexes on a board. We'd also want a way to transplant that board to a completely different surrounding possibly preserving all graphical differences.

arlith:
Yes, the "terrain level" (the first value after the terrain type in the tileset) has a specific meaning (best witnessed in woods, where 1 is light, 2 is heavy, and 3 is ultra).

There are two ways to effect how the hex is drawn for a given terrain type and level: there are themes, and there is fluff.  You can design the tileset to have a specific terrain/fluff combo.  For instance, a hex could have a terrain of "clear:1" and a terrain of "clear:1;fluff:1".  Both of those hexes would function as a terrain lvl 1 clear hex, but the latter entry could have a different image specified.  The fluff terrain level is basically a way to add 'fluff' images to different terrain types.

Without looking at the code, I think themes and fluff terrain type basically accomplish the same thing.  I do kind of think that it's useful to have themes work the way they do now, where a theme does not need to necessarily be globally applied to a board.  A good example of where I think this could be well utilized is with the volcano theme.  I feel like volcano clear hexes are a bit overpowering when the whole board is full of them, but small areas/patches of volcano clear hexes could look neat.

I think that a global theme changer isn't necessarily in conflict with per-hex themes.  For one, most boards won't take advantage of this, and 2 the theme changer could be a way for a player to adjust a map that may not be visually appealing to them ("Ah darn, I'm on this all-volcano map, I hate volcano, lets change the whole map to no theme").  And, as long as there's a way to return the map to the "original" theme, it shouldn't be an issue.

As an aside, I feel like one of the reasons that these features aren't better utilized is because the map editor doesn't really reveal them.  Basically, you have to be intimately familiar with the tileset files to know these things exist.  If the map editor did a better job of presenting what's possible in the tileset, I think there could be better utilization.

SJuliez:
On the practical on-this-earth side, yes lets save the original theme and add an option to reinstate that. I hope it'll be as easy as it sounds.

Currently I think the system is inconsistent (and yes, the map editor is no help). Woods:1 and :2 among others, even worse, water:x, add game effect. Roads:1 and :2 are used as fluff. The other fluff definition clear:1;fluff:1 I hadnt seen or noticed before but would serve as fluff too. Theme serves as hex fluff but then it also serves usefully as a general board gfx effect (and is used as such in many boards).

Volcano is a difficult example for a theme change. Yes the theme from clear volcano hexes could be changed but lava isnt going to change. So the player would end up having lava seas in a grass landscape. Or water in a volcano landscape. This doesnt seem like a theme thing. (The hex tile graphic has room for improvement though).

Therefore any theme (as in "on any planet" or "in any area") would define volcano and lava hexes more or less the same. If a player is annoyed by this they should choose another board (lava seas will stay anyway and a volcano map without lava seas could just as well be defined as a clear map to begin with).

A grassy spot in a barren landscape however could be transported elsewhere and given another color or style. Therefore a clear:1;grass:1 would also not be changed but another theme could easily make it look green, blue or violet (or even really the same as a clear:1 without loss of information).

That's why I would vote for a clear distinction between 
- game relevant (ltwoods:, hvywoods:, ltrubble: etc.)
- hex fluff (ltwoods:1, hvywoods:14 or ltwoods:1;fluff:1)
- board theme (grass)
- board pre-game prep: volcano&lava instead of clear&water (reminds me of popolous), deforest&dehydrate (->moon), deep snow but clear weather during the game.

arlith:
Are there road:2 entries?  The first number after the colon represents terrain level, and I think that in all cases it should be considered to have game impacts (such as woods:1 and woods:2).  Entries of road:2 should not exist to my knowledge (perhaps there's a TO road terrain type that I'm not aware of).

I guess, I'm not really sure what problem we're really trying to solve at this point (including the original pull request).  As it works currently, there's basically two ways for providing non-game effecting graphical changes to a single hex: adding a theme, and adding a fluff entry.  

Themes are tied to a specific terrain feature (given you've got a terrain type of clear:1, you could have many themes for that type).  However, I'm not sure that you can combine themes (lets say I wanted to have a road:1 "boulevard" theme placed on top of a clear:1 "grass" hex, that's not possible).  

The fluff terrain type can be used to effect what images get applied to an image.  The way image matching for tilesets works allows for you to have entries like clear:1;fluff:1 that (if entered in the right order) can get matched before clear:1 or fluff:1, allowing you to have a different hex image, without having an impact on the rules.  This does however change the terrain types of the hex (adding a fluff terrain type, however the fluff type has no in-game impact).

With your addition for a theme selector, I kind of wonder about what the intended purpose is.  Since this is a client option, it can have no game effects (and right now it doesn't).  All it can do is effect how the board is drawn.  It's difficult for me to see the utility of this, as I my initial response would be the same as what you gave above: if you don't like the themes in a board, pick another board.  You can have a grass theme, or a mars theme, or whatever, but that's largely up to the designer of the board.  The purpose that makes sense to me is that, a board that got picked doesn't fit visually, so you want to adjust the theme for purposes of display (volcano is an example of this for me, where the magma hexes with the normal clear hexes would probably look better, imo, than the volcano themed clear hexes).

I feel like the idea of a board-wide theme and a hex-level theme aren't mutually exclusive: you could have a hex know about its preferred theme but that gets overridden by a theme specified by the whole board.  I hadn't thought about this before, but this may be a better way to implement your theme changer.

The idea I had in my head for implementing the ability to "return to the original board theme state" would be for each Hex to track it's "default theme".  Basically, add a new String state variable *defaultTheme*, which would be set at line 80 of Hex.java.  Then you could add a IHex.restoreDefaultTheme() method that sets Hex.theme to Hex.defaultTheme.

The last thing I wanted to address is that, your last bullet point seems to be indicating more that you want something that *does* have a rules effect.  Aka, I have some map, and I want to make it a "volcano" themed map, which would change water hexes to lava hexes and having clear hexes look more volcanic, or I want to take a standard map and make it more "desert-like" by reducing water levels and having a sandy-looking clear hex.  This idea concept seems to go beyond what we've been talking about so far.

Overall, I feel kind of like I'm rambling.  This might be a conversation that's better held via instant messaging.  If you'd like, we could try to setup a time to chat.

Akjosch:
The road:2 entries represent the "road with trees" hexes (fluff/road_trees_*.gif files). As far as I'm aware, they aren't an official road variant (like gravel and dirt road) and they have no in-game effect different from road:1. The map generator uses them, too.

arlith:
Yea, those should probably be road:1;fluff:1 or something similar. Using additional terrain levels like that for purely aesthetic reasons seems dangerous (it relies on road checking to use IHex.containsTerrain vs checking the terrain level via IHex.terrainLevel).  I'm not sure if changing it would be smart, since it's likely been around forever, but I don't think it's a practice that should be recommended.

SJuliez:
OK, updated, will now remember the last choice of theme and can revert the board to its original state.

SJuliez:
@arlith I only now read your comment fully. I probably did exactly what you wrote but named everything differently...

Akjosch:
While we're on the subject of fixing the map format, Chaff (coded as "smoke:5") can't co-exist with any smoke type ("smoke:1" to "smoke:4") in the same hex, and I can't find a rule which would prohibit that. Certainly not in TO (where rules for both were introduced) or its current errata.

@SJuliez I'd suggest using `TreeSet<String>` instead of `HashSet<String>` for the list of themes in the combo box, since the latter isn't sorted.

SJuliez:
@Akjosch yes, thanks :smile:

arlith:
@Akjosch 
I don't think that there are any additional effects of Smoke + Chaff (to be clear, this is the chaff from Chaff  Grenades on TO pg 363, I guess there are also Chaff Pods on TO pg 299).  Chaff functions as light smoke plus some other effects.  I had thought that there was a smoke:6 for Chaff + Heavy Smoke, but I guess not.  It may make sense to have a separate terrain type for chaff as opposed to smoke... I only implemented the ECM effects of chaff, the usage of smoke:5 for chaff was already there before that.

PR #57, opened 2016-03-03
Remove dependency on TinyXML

nderwin:
This pull request converts the following remaining areas that were using TinyXML for unmarshalling and marshalling XML files to using the Java Architecture for XML Binding (JAXB)

- Map settings 
- Game options (mmconf/gameoptions.xml)
- Client preferences (clientsettings.xml)
- Entity verifier configuration (UnitVerifierOptions.xml)
- The Drawing Board mech files (data/mechfiles/.../*.xml)

Unit tests have been added for each to test that the files can be marshalled and unmarshalled as expected.  The format of existing XML files for these areas is **unchanged** - all of these conversions are intended to be backwards compatible with existing files.

Additionally, the build script was modified to copy over test resources to sit along side the compiled test classes so they would function.

Minor code updates appear as well, where the impact was minimal.

A bug was also fixed with the Entity Verifier.  The original code was looking for a tag named "tragcomp" in the configuration file.  This has been fixed to "targcomp", so more entities may now fail the verifier.

Files removed with the PR are

- megamek/lib/TinyXML.jar (no longer needed)
- megamek/lib/tinyXML07-src.zip (no longer needed)
- megamek/src/megamek/common/XMLStreamParser.java (deprecated and not used)
- megamek/src/megamek/test/TinyXMLTest.java (no longer needed)

PR #69, opened 2016-03-08
adjusted default JAXB constructor visibility

nderwin:
removed a method from PreferenceManager.Store that wasn't used

Per request https://github.com/MegaMek/megamek/pull/57#issuecomment-193250609.

### DIFF
--- a/megamek/src/megamek/common/loaders/TdbFile.java
+++ b/megamek/src/megamek/common/loaders/TdbFile.java
@@ -20,6 +20,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+
 import javax.xml.bind.JAXBContext;
 import javax.xml.bind.JAXBException;
 import javax.xml.bind.Unmarshaller;
@@ -32,6 +33,7 @@ import javax.xml.bind.annotation.XmlType;
 import javax.xml.bind.annotation.XmlValue;
 import javax.xml.bind.annotation.adapters.XmlAdapter;
 import javax.xml.bind.annotation.adapters.XmlJavaTypeAdapter;
+
 import megamek.common.BipedMech;
 import megamek.common.Entity;
 import megamek.common.EquipmentType;
@@ -479,7 +481,10 @@ public class TdbFile implements IMechLoader {
         @XmlElement(name = "version")
         String creatorVersion = "Unknown";
 
-        Creator() {
+        /**
+         * Required for JAXB.
+         */
+        private Creator() {
         }
     }
 
@@ -518,7 +523,10 @@ public class TdbFile implements IMechLoader {
         @XmlElement(name = "structural")
         Structure structure;
 
-        BasicInformation() {
+        /**
+         * Required for JAXB.
+         */
+        private BasicInformation() {
         }
 
     }
@@ -550,7 +558,10 @@ public class TdbFile implements IMechLoader {
         @XmlElement(name = "internal")
         String structureType;
 
-        Structure() {
+        /**
+         * Required for JAXB.
+         */
+        private Structure() {
         }
 
         boolean isClanTC() {
@@ -572,7 +583,10 @@ public class TdbFile implements IMechLoader {
         @XmlAttribute(name = "level")
         Integer rulesLevel;
 
-        Technology() {
+        /**
+         * Required for JAXB.
+         */
+        private Technology() {
         }
 
     }
@@ -586,7 +600,10 @@ public class TdbFile implements IMechLoader {
         @XmlAttribute
         String LAMTonnage;
 
-        MovementModifier() {
+        /**
+         * Required for JAXB.
+         */
+        private MovementModifier() {
         }
     }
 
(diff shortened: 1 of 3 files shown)


arlith:
I still seem to be getting warnings in PreferenceManager and GameOptions.

nderwin:
@arlith  Can you elaborate?  I have no such warnings showing up in NetBeans, and I don't use Eclipse.

arlith:
It's strange, I'm still getting the same "The constructor GameOptions.GameOptionsXML() is never used locally" warnings, but only if I go to the source file, not in the list of Errors/Warnings.  This happens in GameOptionsXML and PreferenceManager.Settings, PreferenceManager.Store and PreferenceManager.XmlProperty.  However, I have no warnings in TdbFile.

nderwin:
@arlith Sounds like an Eclipse issue?  It's not very smart about recognizing changes to files that it didn't make.  At least that's what I have noticed while using another product I have to use for work that is based on the Eclipse platform.  You could try the old "turn it off and on again" trick, if you haven't already.  That seems to clear up the problems I run into.

arlith:
yea, I tried all of those, heh.  I don't use git inside of Eclipse; I use it at the command line.  Eclipse isn't very good about refreshing files after I make changes outside of Eclipse, so I'm in the habit of refreshing all the project files after making pulls.  Originally it reported those warnings in the list of warnings, after which I refreshed and they still appeared, then I closed Eclipse and reopened it.  That's when the warnings disappeared from the global list, but if I open the individual source file it still warns me that the constructors aren't used locally.

arlith:
It does look like the warnings will appear in the global error/warning list.  I'm not sure why GameOptions and PreferenceManager get these unused constructor warnings but TdbFile doesn't, as the code seems the same to me.

arlith:
Why is it necessary to explicitly define a private constructor for JAXB?  Doesn't the compiler automatically add a default constructor if no other constructors are defined?  Does the constructor need to be private?  I'm still seeing weird behavior in Eclipse where it warns about some of the private default Constructors, but not all of them.  I could potentially suppress those warnings, but I don't think that is a good solution.

Akjosch:
> Why is it necessary to explicitly define a private constructor for JAXB?

Because it uses reflection to create the instances and only tries the non-args constructor. (specifically, via `cls.getConstructor().newInstance()`). There are ways to avoid it, via XMLAdapters, but it's generally simply easier.

> Doesn't the compiler automatically add a default constructor if no other constructors are defined?

It does, but only when no other constructor is defined. Reference: [JLS 8.8.9](http://docs.oracle.com/javase/specs/jls/se7/html/jls-8.html#jls-8.8.9).

>  Does the constructor need to be private?

For JAXB, it can have any access level.

arlith:
It seems then that the private non-args constructors for `GameOptions.GameOptionsXML`, `PreferenceManager.Settings`, `PreferenceManager.Store`, and `PreferenceManager.XmlProperty`could be removed them.  You could leave a comment indicating that there must be a non-args constructor for JAXB, just in case someone adds a constructor in the future.  It does appear that making the constructors public doesn't remove the warnings (I presume because they are private inner classes).

Akjosch:
It's rather simple: If you remove them, JAXB will not be able to unmarshall those classes.

My solution would be to add `@SuppressWarnings("unused")`. You simply need to do this with JAXB sometimes, for example when you implement the event callbacks:

* `private void beforeUnmarshal(Unmarshaller unmarshaller, Object parent);`
* `private void afterUnmarshal(Unmarshaller, Object parent);`
* `private boolean beforeMarshal(Marshaller marshaller);`
* `private void afterMmarshal(Marshaller);`

All of those, if implemented, give you the same warning. In all of those cases, the warning is wrong; the methods are being used by JAXb.

arlith:
I'm willing to accept that you must have those definitions for JAXB to work, but I just want to better understand why.  If the compiler will add a default constructor if no other constructors exist, this should satisfy JAXB's requirements, no?  It seems like, the problem only exists if you need a constructor with arguments (which seems very unlikely in this case).


I am also willing to accept that we have to add suppress warning annotations, however I am not currently convinced that this won't create problems with different versions of Eclipse.  For instance, my version of Eclipse does **not** warn me about `private Creator()` in line 487 of `TdbFile`.

Ahhhh.... I may have figured it out now.  I had thought that there wasn't a difference between something like `TdbFile.Creator` and `GameOptions.GameOptionsXML`, but there is.  And the difference is why we can't do what I suggested above: GameOptions.GameOptionsXML **does** have two constructors.

It looks like, for  `GameOptions.GameOptionsXML`, `PreferenceManager.Settings`, `PreferenceManager.Store`, and `PreferenceManager.XmlProperty` we will need to add annotations to suppress the warnings.  I now have a better understanding and I'm willing to make this change.

nderwin:
Thanks for the follow-up @Akjosch.

Originally, I had the no-arg constructors at protected or package-private visibility, but then changed them all to private to try to alleviate Eclipse's complaint.  I use NetBeans, which has no problem with those constructors at any visibility.

@arlith If you'd rather I do it, I can deal with it and make a new PR.

arlith:
I made the change in 6352cc5.  It's easy to add the suppress annotations, so it's not a big deal.

PR #71, opened 2016-03-09
Fixing the map resize dialogue having switched "west" and "south" labels

Akjosch:
I'm surprised nobody noticed yet ...


arlith:
Stick around long enough, you'll be surprised by nothing.

PR #72, opened 2016-03-09
Bug terrain object contract violation

Akjosch:
The Java Object contract says (among other things) that two objects which return `o1.equals(o2) == true` for each other *also* have to return the same hashCode(). In practice, this means that you *have* to implement hashCode() if you implement equals(). Doing otherwise is an error (sadly one which the compiler doesn't catch).

In addition, there are transitivity constraints on equals() which - for non-final classes - are easiest solved by making sure the two objects are of the exactly same class. Thus replacing `this instanceof that` with `getClass().equals(that.getClass()`. For in-depth discussion on the topic and the pitfalls see [this article](http://www.drdobbs.com/jvm/java-qa-how-do-i-correctly-implement-th/184405053) and [this StackExchange question](http://stackoverflow.com/questions/27581/what-issues-should-be-considered-when-overriding-equals-and-hashcode-in-java). The comment box has *way* too little space to write it all down.

Finally, all the fields used in hashCode() should be final or effectively final. The setting of `level` to "final" just makes this something the compiler checks for us.

This is a solution for Java 6 (and earlier; down to Java 5 I think). In Java 7, the helper methods `Objects.hash()` and `Objects.equals()` (note: "Objects", not "Object") can also be used for a more concise and less error-prone implementation.

Other classes might also suffer from the same bugs, but I only checked Terrain and TerrainFactory for a bit of reference information, so that's what I'm patching.

### DIFF
--- a/megamek/src/megamek/common/Terrain.java
+++ b/megamek/src/megamek/common/Terrain.java
@@ -35,7 +35,7 @@ public class Terrain implements ITerrain, Serializable {
      * types (ie, Light Woods vs Heavy woods). Not to be confused with Hex
      * levels.
      */
-    private int level;
+    private final int level;
     private boolean exitsSpecified = false;
     private int exits;
     private int terrainFactor;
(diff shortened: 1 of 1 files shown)


Dylan-M:
Since we're actually using Java 7, and will soon be switching to Java 8, do you want to re-write this pull request using the Objects.methods you referenced?

Akjosch:
Sure, no problem. This will make testing easier too.

I also checked to be 100% sure, and getClass() returns flyweight objects - only one instance per given class -  which means it's safe to compare them with `==`.

arlith:
I was planning on merging this, but now you have the ability to merge it as well, so I will leave it to you.

PR #75, opened 2016-03-12
PSR checks rewritten to be more descriptive.

pheonixstorm:
All PSR rolls now use the full terrain rather than difficult terrain. ie
+1 Light Jungle + 1 Swamp

Also fixed the Ultra Rubble level to be 6 and added a missing bog down/unstuck entry for mud.

### DIFF
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -6504,19 +6504,14 @@ public abstract class Entity extends TurnOrdered implements Transporter,
             EntityMovementType moveType, IHex curHex, Coords lastPos,
             Coords curPos, boolean isLastStep) {
         PilotingRollData roll = getBasePilotingRoll(moveType);
-        addPilotingModifierForTerrain(roll, curPos);
+        boolean enteringRubble = true;
+        addPilotingModifierForTerrain(roll, curPos, enteringRubble);
 
         if (!lastPos.equals(curPos)
             && ((moveType != EntityMovementType.MOVE_JUMP)
                 || isLastStep)
             && (curHex.terrainLevel(Terrains.RUBBLE) > 0)
             && (this instanceof Mech)) {
-            int mod = 0;
-            if (curHex.terrainLevel(Terrains.RUBBLE) > 5) {
-                mod++;
-            }
-            // append the reason modifier
-            roll.append(new PilotingRollData(getId(), mod, "entering Rubble"));
             adjustDifficultTerrainPSRModifier(roll);
         } else {
             roll.addModifier(TargetRoll.CHECK_FALSE,
@@ -9764,10 +9759,18 @@ public abstract class Entity extends TurnOrdered implements Transporter,
             return;
         }
         IHex hex = game.getBoard().getHex(c);
-        int modifier = hex.terrainPilotingModifier(getMovementMode());
-        if (modifier != 0) {
-            roll.addModifier(modifier, "difficult terrain");
+        hex.terrainPilotingModifier(getMovementMode(), roll, false);
+    }
+
+    public void addPilotingModifierForTerrain(PilotingRollData roll, Coords c, boolean enteringRubble) {
+        if ((c == null) || (roll == null)) {
+            return;
         }
+        if (isOffBoard() || !(isDeployed())) {
+            return;
+        }
+        IHex hex = game.getBoard().getHex(c);
+        hex.terrainPilotingModifier(getMovementMode(), roll, enteringRubble);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


arlith:
I don't like how `Terrain.pilotingModifier` and `Terrain.getUnstuckModifier` had their default values returned, and how they got changed to magic numbers.  I think it would be better to pass the PilotingRollData into those methods and add the text/modifiers there.

Also, instead of adding string literals, you should pull internationalization entries from the properties files.  I know a lot of stuff in Terrains and Hex aren't internationalized as it is, but adding isn't really a road to fixing the problem.

pheonixstorm:
> I don't like how Terrain.pilotingModifier and Terrain.getUnstuckModifier had their default return values changed

No way around it. If you look at the code pre PR hsTerrains gets looped over every terrain type in the current hex and added to a retVal. Since we want each terrain to give its type we can no longer include everything by defaulting to 0. If we do then we get Unknown Terrain as there is only a limited set of terrain types that have a psr modifier. Anything that doesn't have a modifier such as clear or woods should be ignored. So only those types listed in TacOps (even +0) should have an entry. The same goes for stuck/unstuck. The major difference being stuck/unstuck already has a -1 modifier so I could not use it again.

>  I think it would be better to pass the PilotingRollData into those methods and add the text/modifiers there.

I had originally planned on this but decided on staying as close to the original code as possible.

> you should pull internationalization entries

I really think this should be a separate PR as there is a LOT of work that needs to be addressed outside this PR when it comes to movement, strings, and internationalization.
See...
`roll.append(new PilotingRollData(getId(), mod, "entering Depth " + waterLevel + " Water"));`
`roll.addModifier(1, "superheavy mech avoiding bogging down");`
I have seen more than this but they were the recent I have run across.

pheonixstorm:
I think if I can get a handle on how the internationalization works I can make my new project to actually get terrains/hex and the SPAs all nice and language independent. Probably wouldn't finish till post stable though.

Question though, would I just add all this to messages.properties or create a new set of properties files?

Which leads me to another question. Would you rather have the internationalization done for just whats in this code or all at once?

Akjosch:
Turning the "Non-externalized strings" option in Eclipse (Preferences > Java > Compiler > Errors/Warnings) to "Warning" produces 22327 lines producing a warning (yes, *twenty two thousand three hundred and twenty seven*). I'd say it'd be better if new patches and PRs at least wouldn't *add* to this number.

arlith:
>I had originally planned on this but decided on staying as close to the original code as possible.

You're going to have to do it, because it's the only way (without a larger redesign) to not have strange return values for those methods (you could eliminate the magic numbers by having static final public state variables, but I don't like the default returns being -5 or -1; 0 is the 'correct' default return because, if there's no match, you should just not modify the roll).

> I'd say it'd be better if new patches and PRs at least wouldn't add to this number.

This is exactly the sentiment I tried to express before.  I know  there's a ton of offending string literals.  It'd be nice to get it all internationalized, but I don't expect that to happen for this PR.  What I *do* want to happen for this PSR is to internationalize the string literals it's adding.

You don't need to add a new messages.properties, just add to one that's there.  I think it's going to be the i18n/megamek/common/messages.properties, not the client one.

pheonixstorm:
@Akjosch eeewww didn't know there was THAT many.

pheonixstorm:
Should have a commit ready for pushing roll into terrain done tonight. If that one passes muster I will move on to internationalization and hopefully this PR will then be complete.

pheonixstorm:
Tested the changes out as best as I could but the dice were against me and I could never get the swamp to turn into quicksand. The change to roll.addmodifier for quicksand should work but I am not sure due to quicksand never showing its ugly face.

arlith:
You can use the map editor to create a quicksand hex (or modify the code so that the roll for turning from swamp into quicksand always succeeds).

pheonixstorm:
Forgot I could have used the map editor. Will edit my test map and try again.

pheonixstorm:
Code appears to work correctly.

vizax:
> I know there's a ton of offending string literals. It'd be nice to get it all internationalized, but I don't expect that to happen for this PR. 

I'm going to semi-hijack this PR for a comment about internationalization strings.
If you folks can come up with standards/procedures/something. I'll lend a helping hand in populating these strings. I'm never going to be able to provide any super code, but I can do basic maths and grunt work in my free time. I followed the instructions above and check all 3 projects... *25,730 total* according to my Eclipse count. I'm sure some will need to have `//$NON-NLS-<n>$` added, but still, I imagine that if everyone chips in we can knock this out sooner rather than later. I have some thoughts, but it's not my project, so I'll hold them if/until asked.

Let me know.
`/End Hijacking`

arlith:
It'd probably be best to have this conversation in a separate issue ticket.

pheonixstorm:
@arlith Should I hold off on the internationalization until #103 gets worked up or go ahead and get started?

arlith:
I would get started.  I mean, I think that the i18n keys likely won't change, so you should be able to get started now.

pheonixstorm:
Will do, I spotted the weather/environmental conditions so I will add below those. Shouldn't take long or be very difficult.

PR #79, opened 2016-03-13
Adding proper hashCode() implementation and optimising equals() where needed

Akjosch:
This fixes a bunch of classes which implement equals() to also include hashCode() with the same fields. At the same time, equals() is rewritten to work properly where required, and to use a consistent coding style.

A few classes remain which couldn't be so easily fixed:

`megamek.common.CriticalSlot` compares `megamek.common.Mounted` instances, but that class declares no equals(), which means that the comparison is done by instance - the same as someone would write `mounted1 == mounted2`. Additionally, only *one* instance of that class is included, `mount`, but not the other `mount2` - this looks like a separate bug. Furthermore, the `Mounted` class is a rather big data class, so I can't just whip a quick `equals()` method inside without a longer analysis.

`megamek.common.SpecialHexDisplay` and `megamek.common.Team` have similar problems, but comparing `megamek.common.IPlayer` instances. Those too have no `equals()` method and are complex data types.

`megamek.common.Coords` has an overly complex `hashCode()` which I'd love to simplify, but it also has a method called `public static Coords getFromHashCode(int hash)` which screams "Please don't do this!" to me. Hashes aren't meant to be reversible. It *looks* like it's only used in `megamek.test.TestCoords`, but I'm not feeling comfortable enough touching it until someone with more seniority in team checks it.

### DIFF
--- a/megamek/src/megamek/common/Building.java
+++ b/megamek/src/megamek/common/Building.java
@@ -687,14 +687,21 @@ public class Building implements Serializable {
      *         <code>Building</code>.
      */
     @Override
-    public boolean equals(Object other) {
-        if ((other == null) || !(other instanceof Building)) {
+    public boolean equals(Object obj) {
+        if(this == obj) {
+            return true;
+        }
+        if((null == obj) || (getClass() != obj.getClass())) {
             return false;
         }
-
         // True until we're talking about more than one Board per Game.
-        Building bldg = (Building) other;
-        return (id == bldg.id);
+        final Building other = (Building) obj;
+        return (id == other.id);
+    }
+    
+    @Override
+    public int hashCode() {
+        return id;
     }
 
     /**
--- a/megamek/src/megamek/common/ECMInfo.java
+++ b/megamek/src/megamek/common/ECMInfo.java
@@ -15,6 +15,7 @@
 package megamek.common;
 
 import java.util.Comparator;
+import java.util.Objects;
 
 /**
  * A class that keeps track of information related to an ECM field.
@@ -392,6 +393,7 @@ public class ECMInfo {
         return (owner == null) || (other.getOwner() == null) 
                 || owner.isEnemyOf(other.getOwner());
     }
+
     /**
      * Equality is based on whether position, owner, range and all strengths
      * match.
@@ -400,23 +402,24 @@ public class ECMInfo {
      * @return
      */
     @Override
-    public boolean equals(Object o){
-       if (!(o instanceof ECMInfo)) {
-           return false;
-       }
-       ECMInfo other = (ECMInfo)o;
-       boolean ownersMatch = ((owner == null && other.owner == null) 
-               || owner.equals(other.owner));
-       boolean posMatch = pos.equals(other.pos);
-       boolean strMatch = (strength == other.strength) 
-               && (angelStrength == other.angelStrength) 
-               && (eccmStrength == other.eccmStrength) 
-               && (angelECCMStrength == other.angelECCMStrength);
-       boolean novaMatch = isECMNova == other.isECMNova;
-       boolean rangeMatch = range == other.range;
-       return ownersMatch && posMatch && strMatch && rangeMatch && novaMatch;
+    public boolean equals(Object obj){
+        if(this == obj) {
+            return true;
+        }
+        if((null == obj) || (getClass() != obj.getClass())) {
+            return false;
+        }
+        final ECMInfo other = (ECMInfo)obj;
+        return Objects.equals(owner, other.owner) && Objects.equals(pos, other.pos) && (strength == other.strength) 
+                && (angelStrength == other.angelStrength) && (eccmStrength == other.eccmStrength) 
+                && (angelECCMStrength == other.angelECCMStrength) && (isECMNova == other.isECMNova) && (range == other.range);
     }
 
+    @Override
+    public int hashCode() {
+        return Objects.hash(owner, pos, strength, angelStrength, eccmStrength, angelECCMStrength, isECMNova, range);
+    }
+    
     public int getRange() {
         return range;
     }
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -5859,14 +5859,20 @@ public abstract class Entity extends TurnOrdered implements Transporter,
      * Two entities are equal if their ids are equal
      */
     @Override
-    public boolean equals(Object object) {
-        if (this == object) {
+    public boolean equals(Object obj) {
+        if(this == obj) {
             return true;
-        } else if ((object == null) || (getClass() != object.getClass())) {
+        }
+        if((null == obj) || (getClass() != obj.getClass())) {
             return false;
         }
-        Entity other = (Entity) object;
-        return other.getId() == id;
+        final Entity other = (Entity) obj;
+        return (id == other.id);
+    }
+
+    @Override
+    public int hashCode() {
+        return id;
     }
 
     /**
--- a/megamek/src/megamek/common/EquipmentType.java
+++ b/megamek/src/megamek/common/EquipmentType.java
@@ -28,6 +28,7 @@ import java.util.HashMap;
 import java.util.Hashtable;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Vector;
 
 import megamek.common.options.GameOptions;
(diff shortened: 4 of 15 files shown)


arlith:
I seem to recall looking at the `Coords.hashCode()` method in the past and wondering about it.  I seem to recall that there was code that relied upon the structure of `Coords.hashCode()`, and it seemed really dubious, but I didn't look into fixing it.

Akjosch:
MegaMek (aside of the one test class) doesn't use the specific hash, MekHQ doesn't use it, MegaMekLab doesn't have any reason to care about hex coordinates in the first place. I think it's safe to toss it and make the `hashCode()` method consistent with everything else.

Anybody wants to fill me in which fields `megamek.common.Mounted` and `megamek.common.Player` are using to consider each other to be equals so I can augment this PR accordingly?

arlith:
I think that `Player` only needs to check the ID.  I was somewhat surprised to learn that `Player` doesn't have an equals implementation, but I think it's because we just compare Ids, so in most cases there will be code like `p1.getId() == p2.getId()`.  The only time where I can think of matching player Ids wouldn't be sufficient would be in degenerate cases (like where the Server has screwed up with connecting players).

As for `Mounted`, I honestly don't know.  I think in the past 6 months I came across the fact that `Mounted` doesn't implement equals, and (without too much thought) I figured that reference identity was probably what was intended.

`Mounted` represents a piece of equipment that is mounted on a unit.  `Mounted` keeps track of what the `EquipmentType` is used to represent the piece of equipment that is mounted, and the `Mounted` instance is generally housed in some `CriticalSlot`.  Then `Mounted` contains a whole slew of state information that is pertinent to gameplay (is this `Mounted` hit, destroyed, missing, jammed, linked to another mounted for purposes such as ammo or PPC capacitors, etc., etc., etc...).

If we end up going through each state variable in mounted and compare them, in the end I think we approach something that is basically reference identity.  In fact, if we don't keep track of what `CriticalSlot` a `Mounted` instance is in, we almost have to use reference equality.  If I have two medium lasers on a unit, one in the right arm and one in the left arm, the only difference between those two instances is going to be where on the unit they are located.  Actually, mounted has a location state.  If those two instances are located in the same location, I don't think Mounted has state to distinguish them.

I think I was made aware of this problem because I had an instance where I had a `Mounted` and wanted to know what `CriticalSlot` it was in, but didn't have that state in `Mounted`.  It's possible that keeping track of that information could be useful.

If we knew `CriticalSlot` information, I think that knowing `EquipmentType`, (Entity) location, and `CriticalSlot` would be enough to uniquely identify a `Mounted` instance.

Akjosch:
Ok, added equals() and hashCode() to `IPlayer` (and made sure the implementation in `Player` is up to date), this covers the `Team` and `SpecialHexDisplay` problems.

Now I need to analyse `Mounted` in actual play to see what could be used there ...

nderwin:
If you add the equals() and hashCode() methods to IPlayer, which methods are the implementations then overriding - the methods from IPlayer, or the methods from Object?  That's the kind of multiple inheritance mess that the language was designed to avoid, wasn't it?

How many implementations of IPlayer are there?  Can the requirement of implementing equals() and hashCode() be left to the code review process instead of interfering with the inheritance scheme?

Akjosch:
`IPlayer` is an interface. There is no method there to override (not until Java 8, and then it would only need further explanation if it had a default implementation, which it doesn't). `Player`'s methods override that of `Object`.

The reason why I added the method *signatures* to `IPlayer` is to make sure the compiler enforces the contract of both being correctly overridden by whatever implements it. That's the same reason `java.util.Map` and `java.util.List` have them.

nderwin:
Alrighty then.  Still feels weird, even though `List` and `Map` have them (honestly I'd not noticed that before).

arlith:
In general, I think that MM goes a little overboard on making interfaces.  There are numerous instances where I don't think there will ever be more than one implementation.  IPlayer is an example.  IHex and ITerrain are others I can think of off the top of my head.  I don't see any reason that these should have interfaces, but I also didn't make them interfaces myself, so I've left them there.

Akjosch:
Not sure about the others, but `IHex` could use a few implementations: One for ground map, one for low-altitude hex, one for space ...

arlith:
Yea, that's valid.  I'm not sure it's necessary with the current rules, but you could do things that way.  It might make sense for the high altitude map (which isn't implemented yet), but the low altitude map mostly uses the same terrain features.  I guess it could make code simpler, but that ship sailed long ago (we could refactor the code, but I'm not certain it's worth the effort).  Under the official rules, I don't believe there are any "space terrains".

Akjosch:
Space terrain features include at least the three:

* screen chaff [TW 251]
* asteroid field [SO 44]
* debris [SO 45]

arlith:
Screen I knew about, but I don't know how it's implemented right now.  It may be via Terrains (probably is).  Asteroid field and debris I didn't know about.  It does bring up a deficiency in map selection though: MM doesn't allow users to specify a "space" map.  It just generates an MxN grid of "space" hexes.  I think that there's nothing that would prevent us from having defined space boards.  I've had users report that they can edit saved games by hand to load their desired map as a ground map, then edit the save to make it a space game.  This was used to utilize fluff terrains to make some neat backgrounds for space maps.

pheonixstorm:
Would having an interface for the ground and low altitude map allow for units in a ground hex for AA attacks such as air defense artillery units? Also would it allow for hot drops from the space/atmosphere interface all the way to the ground? If so it may be worth it to flesh these out at some point down the road.

Also a question. Why is water not in terrain.java? It is a base terrain after all.

arlith:
@pheonixstorm 

1) (I presume you don't mean interface as in the programming concept, but rather some type of connection between low altitude and ground maps; at first I thought you were talking about a Java interface).  It's hard to say what an undesigned feature would or would not contain, but it would adhere to the rules in TW, so I presume yes.

2) Your question doesn't make sense to me.  All of the defines for terrains are in Terrains.java, and WATER is one line 25.

pheonixstorm:
terrains has all the terrain types, but terrain does not include water which is a base terrain. Instead water is handle almost entirely in movestep I believe.

As for the interface, I was referring to both a java interface (though my assumption on how this actually work may be incorrect) as well as an actual map connection. But, being a java noob doesn't entirely help my understanding here.

Akjosch:
Interaction between different map layers (ground map, low-altitude map, high-altitude map, space map) require at the very least that you can have more than one board in play at the same time. From what I understood, there are plans to have this feature, but the code looks like it relies too much on "there's only one board" at the moment.

Akjosch:
Okay, I've yet to get MegaMek to even call CriticalSlot.equals() *once*. It might be some MML thing, I don't know.

Regardless, analysing `Mounted` code, I think I can set when two instances are identical on the following values. This would be a bigger change, so I won't touch `Mounted` (or by extension `CriticalSlot`) with this PR and will need to think it over and do a different one for all those changes.

* type
* quirks
* armorValue 
* mount type (detachable weapon pack, anti-personnel weapon mount, squad support weapon, mech turret, sponson turret, pintle turret - those are mutually exclusive, as far as I can tell)

Status codes:
* destroyed
* hit
* jammed
* useless (not sure how this is used though)
* repairable

Specific subclasses:
* originalShots [`MountedAmmo` subclass?]
* shotsLeft [`MountedAmmo` subclass?]
* bayWeapons [`MountedBay` subclass?]
* bayAmmo [`MountedBay` subclass?]
* nweapons [`MountedWeaponGroup` subclass?]
* mineType [`MountedAmmo` subclass?]
* Stuff for shields I don't get yet

They are based upon the idea of "if I looked at that equipment by itself, would it be equal to some other equipment? Could I swap one for without changing what the corresponding critical slots contain, and by that what the characteristics of the unit are?". This leaves out the following information to other classes:

* Placement inside the unit is left to the `crits` structure in `Entity`
* If the equipment is front- or rear-mounted should really be a detail in `CriticalSlot`
* `Entity` already implicitly knows which location the specific part is in, so `Mounted` doesn't have to, but it doesn't hurt to have it in a transient field. Regardless, a ML mounted in left arm is the same type of mounted equipment as a ML mounted in right torso facing backwards. You could swap those two and the unit wouldn't change.

In addition, a few flags (destroyed, hit, missing, armored, repairable) are shared between `CriticalSlot` and `Mounted`. If I was to redesign it, I'd move everything but "missing" to `Mounted`. I wonder if there are any synchronisation problems between those.

arlith:
Naively, I feel like the "backwards" status needs to be in mounted.  This is because of firing arcs.  Most of the firing arc computations happen in Compute, but I believe most of it is based upon `Mounted` instances.  I didn't research this, but off the top of my head, I feel like there might be a problem there if "backwards" status were moved to `CriticalSlot`.

PR #83, opened 2016-03-15
Resource leaks fixage

Akjosch:
This fixes a few resource leaks (mostly FileInputStreams being created but not closed). I didn't have time to test it *and* I didn't fix them all, so just a PR for review and possible augmentation later on.


arlith:
I didn't actually know about the [try-with-resources statement](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html), neat.  It's a lot more compact than using try/finally.

PR #85, opened 2016-03-16
SPA: Terrain Master - code review

pheonixstorm:
**Code review only - not ready for merge**

This is everything I have ready for the Terrain Master SPA. At the moment only Drag Racer remains to be implemented and Nightwalker is still not fully functional.

I am looking for any modifications that can be made to make the code cleaner and to be more efficient if possible. I am also trying to track down why Nightwalker doesn't seem to be working properly.

The second commit, Preparation for new SPAs, is one that will be required for at least one addition SPA not covered here and is easier for me to work multiple SPA branches (same as PR #45). Would it be better as a separate PR or left here?

Forest Ranger - Works as specified by the book

Swamp Beast -  In order to have a functional to-hit bonus when targeted I had to make a compromise between the Alpha Strike and A Time of War rules as discussed in http://megamek.info/forums/index.php?topic=2309.0 

Mountaineer - Works as specified  by the book

Frogman - Works by the book except for the crush depth check since extreme depths are not implemented yet

Nightwalker - Would work by the book if I can figure out why it just doesn't want to work properly. At the moment it calculates everything as if the unit ran rather than walked. The move bonus works but incorrectly for walk/cruise 

Drag Racer - Still working on setting this up. I have not found all the code sections that will need to change.


pheonixstorm:
Closing this to rewrite the branch as this is based off obsolete terrain code

PR #87, opened 2016-03-17
Adding Simplex-based random landscape generator

Akjosch:
This adds a Simplex-based landscape elevation generator to the list (with the ID 3).

The SImplex implementation is a public domain one by Stefan Gustavson and Peter Eastman (see the file for the credits), slightly amended to generate "proper" [-1, 1] ranges for the 2D and 3D case.

The elevation generator is added to the BoardUtilities class as an instance of the (new) ElevationGenerator instance, which will simplify adding or changing them in the future. The actual addition as well as the tooltip are currently still hardcoded though. If there's any interest, I can also replace the current numeric entry with a dropdown menu and transform the three hardcoded generators into ElevationGenerator instances as well, but I guess that's beyond the simple PR.

Submitted as PR for discussion and review.


arlith:
Check line 42 in SimplexGenerator.  Looks like there's a typo in the method name for `public String getTolltip()`.

Other than that, I don't have much to add.  Everything looks reasonable.

Akjosch:
*Jedi voice mode on*

# You didn't see anything.

PR #88, opened 2016-03-18
Board soft centering

SJuliez:
Makes the board center in a smooth motion on wherever it wants go (next unit, next target) instead of jumping there. (Clicking the minimap is unchanged and acts instantly).

Much easier on the eye for me. I also feel that it makes it easier to spot the selected unit while the board moves towards it. Of course, theres no need to wait it out, clicking the board, minimap, scrolling with the keys will stop the motion and selecting another unit will immediately start centering on that.

The centering speed can be adjusted by changing the SOFT_CENTERING_SPEED. I wasn't playing, just testing so maybe a little faster (a value of 7 or 6) would be better. Of course, there's no need to wait it out, clicking the board, minimap or scrolling with the keys will stop the motion and selecting another unit will immediately start centering on that.

### DIFF
--- a/megamek/src/megamek/client/ui/IBoardView.java
+++ b/megamek/src/megamek/client/ui/IBoardView.java
@@ -69,6 +69,8 @@ public interface IBoardView extends MechDisplayListener {
     public void centerOnHex(Coords position);
     public void centerOnPointRel(double xrel, double yrel);
     public double[] getVisibleArea();
+    
+    public void stopSoftCentering();
 
     // it's a hack that the popup is Object, but we use this interface
     // for both AWT and swing, and AWT's PopupMenu and Swing's JPopupMenu
--- a/megamek/src/megamek/client/ui/swing/FiringDisplay.java
+++ b/megamek/src/megamek/client/ui/swing/FiringDisplay.java
@@ -543,7 +543,7 @@ public class FiringDisplay extends StatusBarPhaseDisplay implements
             refreshAll();
         }
         
-        if ((ce() != null) &&ce().isWeapOrderChanged()) {
+        if ((ce() != null) && ce().isWeapOrderChanged()) {
             clientgui.getClient().sendEntityWeaponOrderUpdate(ce());
         }
         
@@ -603,9 +603,6 @@ public class FiringDisplay extends StatusBarPhaseDisplay implements
             refreshAll();
             cacheVisibleTargets();
 
-            if (!clientgui.bv.isMovingUnits() && !ce().isOffBoard()) {
-                clientgui.bv.centerOnHex(ce().getPosition());
-            }
 
             // Update the menu bar.
             clientgui.getMenuBar().setEntity(ce());
@@ -621,6 +618,10 @@ public class FiringDisplay extends StatusBarPhaseDisplay implements
             updateClearTurret();
             updateClearWeaponJam();
             updateStrafe();
+            
+            if (!ce().isOffBoard()) {
+                clientgui.bv.centerOnHex(ce().getPosition());
+            }
         } else {
             System.err.println("FiringDisplay: tried to " + //$NON-NLS-1$
                     "select non-existant entity: " + en); //$NON-NLS-1$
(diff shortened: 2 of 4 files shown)


arlith:
This reminds me of two issues that have been in the back of my head for a while, but I haven't really pursued (and I think you may be well suited for).

One is right-click scrolling.  I would like for this to be a more-used option but as it is, I don't think it works well.  I know that it has a tendency to bring-up the context menu, and also doesn't seem to scroll well.

The other issue is that I feel like there's a redraw problem with arrow key scrolling.  Sometimes it seems choppy, even though the screen refresh should be fast enough so that it isn't choppy.  If you do scroll with the right-mouse button or with the minimap, you don't get this choppiness that you get when scrolling with arrow keys.

Akjosch:
I have a hard time following your `centerOnHexSoftStep()`. The generic method looks something like this:

    double speed = 2.0; /* in hexes per second */
    Coords currentPos;
    Coords targetPos;
    public void centerOnHex(double deltaTime /* in seconds */) {
        double distanceLeft = currentPos.distanceTo(targetPos);
        currentPos = lerp(currentPos, targetPos, Math.min(speed * deltaTime / distanceLeft, 1.0));
        updateMapPos();
        if( distanceLeft <= speed * deltaTime ) {
            weAreDone();
        }
    }

... and that's already the verbose version. What *else* are you doing there?

SJuliez:
@arlith I only use right-click scrolling... but it does bring up the context menu. Added to the todo :) 

@Akjosch Sorry I don't get where that code comes from. It does seem to work tho, doesn't it?

Akjosch:
@SJuliez It's a generic rewrite of your `centerOnHexSoftStep()` (with the added bonus that it doesn't depend on how many dimensions Coords has, because you hide those details inside the [`lerp()` method](https://en.wikipedia.org/wiki/Linear_interpolation)).

SJuliez:
@Akjosch OK, but I'd rather discuss what the actual method does than a rewrite that I'd have to look at first. Also, do you see an actual problem to discuss - does your rewrite suggest I put in useless code?

SJuliez:
Comparing the Softstep() (I suppose that was meant in your post rather than centeronhex()) with the generic I dont see (striking) differences. I dont get the Math.min in that lerp() and I dont know where that speed*deltaTime is supposed to come from but apart from that, it is more or less the same. I don't calculate in hexes tho.

Akjosch:
My example suggested you made your code a bit more complicated that it needs be - to the point where it's not immediately clear how the movement looks like (mine is constant speed by moving by `speed * deltaTime` each time).

`Math.min(val, 1.0)` limits the value to at most 1.0, which is important if we don't want to overshoot.

The idea behind using `lerp(...)` is that it makes it easy to vary how the movement exactly happens. If you have a linear movement like `lerp(min, max, f)` and want to make it "ease out" instead (start up faster, then become gradually slower until it reaches the end, but still taking the same time a linear movement would), you just replace that with `lerp(min, max, Math.sin(f * Math.PI / 2))`. In fact, almost *every* type of movement is applied by varying the third parameter to `lerp()` and *nothing else*.

If I wanted to add an ease-out to your movement code, where would I start, and would I have to change more than a small part of a single line to do so?

SJuliez:
@Akjosch I'm (again) not sure where you're going. Do you want to change the way it behaves? If so, please do, I don't mind, but I'm happy with it. The movement does ease out of course. I did not state that in my first post, I thought people would just test it :)

Akjosch:
Seems I'm having a hard time explaining myself today. *sigh*

I'm not trying to tell you how to code.

I'm trying to show how you could easily halve the code amount, make it clearer to others what your code *does* (like, you say it is easing out - I believe you, but it's not visible from the code, I'd expect an explicit sine or cspline curve there, simply to be able to transform it into a linear or even one where you overshoot a tiny bit *on purpose* before ending on the right spot just by changing a single code line) and make it vastly easier to fine-tune, modify and reuse it.

If your code works for you, it works. If you think the above doesn't matter, it's fine too.

SJuliez:
OK I see. Changed a few things, commented more. Kept the direct calculation though.

PR #90, opened 2016-03-21
Clean up RAT loading; make it load from ZIP files if available

Akjosch:
Cleaning up the RAT loading code and making it work with ZIP files (including taking into account the directory structure inside those files).

Anybody wants to point me to the place where I could most easily make the build/distribution scripts create a ZIP file out of the whole data/rat, or should I just add a goal to `packaging.xml` along the lines of the already existing "unitfiles-zip"?

### DIFF
--- a/megamek/packaging.xml
+++ b/megamek/packaging.xml
@@ -329,6 +329,14 @@
 		<delete dir="${pkgdir}/${datadir}/mechfiles/warship" />
 	</target>
 
+	<target name="rat-zip" description="Create a zipfile of the RAT files">
+		<zip zipfile="${pkgdir}/${datadir}/rat_default.zip" basedir="${pkgdir}/${datadir}/rat" />
+		<delete dir="${pkgdir}/${datadir}/rat" />
+		<sleep seconds="1" /> <!-- give the file system time to catch up -->
+		<mkdir dir="${pkgdir}/${datadir}/rat" />
+		<move file="${pkgdir}/${datadir}/rat_default.zip" tofile="${pkgdir}/${datadir}/rat/default.zip" />
+	</target>
+	
 	<target name="win-zip" description="Create a zipfile of the Windows distribution">
 		<zip zipfile="${basedir}/${windist}.zip" basedir="${windist}" />
 	</target>
@@ -356,7 +364,7 @@
 		<delete dir="${srcdist}" />
 	</target>
 
-	<target name="release" depends="pkgdir-build, unitfiles-zip, mac-package, nix-package, win-package, source-package, pkgdir-clean" description="Build the project from git and package it as Windows, Mac, Unix, and source-only" />
+	<target name="release" depends="pkgdir-build, unitfiles-zip, rat-zip, mac-package, nix-package, win-package, source-package, pkgdir-clean" description="Build the project from git and package it as Windows, Mac, Unix, and source-only" />
 
     <target name="mac-release" depends="pkgdir-build, mac-package, pkgdir-clean" description="Build the project from local source and package it as Mac" />
 
--- a/megamek/src/megamek/client/RandomUnitGenerator.java
+++ b/megamek/src/megamek/client/RandomUnitGenerator.java
@@ -17,17 +17,24 @@ package megamek.client;
 
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
+import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileInputStream;
-import java.io.FileNotFoundException;
+import java.io.IOException;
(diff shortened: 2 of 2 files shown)


Dylan-M:
Do it just like we do the unit files, but don't worry about making it a perfect change there as I'll be rebuilding the build & packaging scripts essentially from scratch over the summer. I'm tired of how messy they are ;)

arlith:
Curiously, my Eclipse is complaining that the BufferedReader instantiated on line 197 of RandomUnitGenerator is a leaked resource.  Since this is a try-with-resources statement, this seems like an erroneous warning.  Any thoughts?

arlith:
I noticed the warning when I may have came across a latent issue in RandomUnitGenerator.  I ended up in somewhat of a deadlock situation on line 316 of RandomUnitGenerator.  It looks like, it was waiting for the MechSummaryCache to finish initializing, but there was no thread initializing the cache.

Akjosch:
The deadlock is ... strange. `MechSummaryCache` should be fully initialised before calling `RandomUnitGenerator`. In the "old" version, it would simply not load any RATs (since it couldn't verify the units). The new one will simply wait until it *has* an initialised `MechSummaryCache`, no matter how long it took. Which should be started via the first `MechSummaryCache.getInstance()`. `MechSummaryCache.done()` (which is one of the places which can call `RandomUnitGenerator.getInstance()`, via a listener set in Client.java) even sets the `initialized` field to `true` prior to calling it.

The only other place would be `RandomArmyDialog`, which just calls `RandomUnitGenerator.getInstance()` - and *that* should initialise `MechSummaryCache` via the `newInstance()` method. I didn't manage to break that yet, but then maybe my drives are simply too fast.

Alternative would be to make this requirement explicit, by changing `getInstance()` and `populateUnits()` to require a fully initialised `MechSummaryCache` as an argument and throwing an `IllegalArgumentException` if it is *null* or not initialised. For compatibility's sake (and speedup if the validation isn't required), this method would then be added:

    public static synchronized RandomUnitGenerator getInstance() {
        return getInstance(new MechSummaryCache(){
            private final MechSummary DUMMY = new MechSummary();
            
            @Override
            public boolean isInitialized() {
                return true;
            }
            
            @Override
            public MechSummary getMech(String name) {
                return DUMMY;
            }
        });
    }

pheonixstorm:
Does HQ automatically use this method to load the rats?

SirMegaV:
Can I still use customized RATs?

pheonixstorm:
If it acts like units you should still be able to have the zip as well as a folder for your rats.

Akjosch:
It works essentially the same as with units, only there is no "rats.cache" created and the tree structure with directories and files *inside* any ZIP file is important and used by MegaMek for the "create random force" dialogue.

And yes, MekHQ uses this method as well.

Akjosch:
I've yet to manage to get the `MechSummaryCache` to not be initialised before the `RandomUnitGenerator` starts loading, but regardless; I patched it so that it waits at most 3 seconds then assumes it will have to work with whatever it has.

We can always add a "Reload RATs from disk" option (if there isn't one already somewhere) to MegaMek and MekHQ, which will call `clear()` and grab a fresh new instance.

arlith:
Yea, I did look into the warning and found [this](http://stackoverflow.com/questions/23294612/strange-resource-leak-stream-is-never-closed-with-try-with-resources-if-excep), which indicates that it's a [bug in Eclipse](https://bugs.eclipse.org/bugs/show_bug.cgi?id=371614).  Apparently, this is fixed in 'Mars' but I'm still on 'Luna'.  

As for the 3 second time limit, do you think that's enough?  If there's no units.cache, loading can take a minute or two.

It may be worth mentioning, I hit the deadlock when loading a saved game after I had already been in a game, so the MechSummaryCache would have been loaded, cleared, and then requested to load again.  However, loading saved games had worked previously, and subsequently.

Akjosch:
Suggestion to make this a bit easier to reason about: `MechSummaryCache` has a listener set to trigger a loading of `RandomUnitGenerator` in `megamek.client.Client`'s `changePhase(...)`. Would changing that to "clean and reload" at this point help making the state of both more predictable? I'm not sure what else clears the `MechSummaryCache`, nor how often it happens or why.

arlith:
I believe that the `changePhase(...)` is the only thing that clears it.  The reason it's cleared here is because it's something that can take up memory and is unnecessary once the game has begun.

pheonixstorm:
@Akjosch you didn't add this to packaging_local.xml

Akjosch:
@pheonixstorm See, and that's why I asked where to add those packaging info to *first* ... :D

PR #94, opened 2016-03-22
Movement Envelope & Walk-Jump-Toggle

SJuliez:
Up for discussion: 

- adds a keybind to switch between walk/run and jumping movement, default J;
- as before with the buttons, J will clear any planned movement;
- if the unit cannot jump J will still clear planned movement as if ESC was pressed (I found that easier than constantly moving my hand between J and ESC);
- if the unit can jump, pressing J twice will effectively act as ESC, removing planned moves and returning to the same move type;
- in addition, the movement envelope menu entry is now a toggle (like the FOV darkening) so that when activated it will always be there (I found constantly having to reactivate the movement envelope annoying and not useful and I could never see a reason for not having it there. Of course, it can still be deactivated by the keybind or in the menu.)

(Note: I do remember that throwing in several changes into one PR is not good style. I think I had a reason for doing it here but can't remember.
Second note: MovementDisplay seems like it would benefit from some cleanup. I might do something there but as a separate thing)

### DIFF
--- a/megamek/i18n/megamek/client/messages.properties
+++ b/megamek/i18n/megamek/client/messages.properties
@@ -1020,6 +1020,8 @@ KeyBinds.cmdNames.udWeapons=Unit Display Weapon Tab
 KeyBinds.cmdDesc.udWeapons=Selects the Weapon tab on the unit display
 KeyBinds.cmdNames.udExtras=Unit Display Extras Tab
 KeyBinds.cmdDesc.udExtras=Selects the Extras tab on the unit display
+KeyBinds.cmdNames.toggleJump=Toggle Jumping/Walking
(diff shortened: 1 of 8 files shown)


arlith:
MoveEnvelope was added prior to the latest path finder rewrite.  Those rewrites made the path planner a bit faster.  Prior to that, running on my old laptop, I noted that the movement envelopes could take a noticeable amount of time to compute for units with a lot of movement (think of Spiders with sprinting enabled).  Because of that, I decided that it might be better to have it be something the user had to initiate manually, so their client was slowed down.

However, with the path planner updates, it's probably fine to have it be a toggle.

arlith:
I did glance through the code for this, and it looks fine to me.

PR #95, opened 2016-03-22
Industrial tsm cost

vizax:
Hopefully I found the correct weight calc for the mech tonnage.

Since I'm new here, please bear with me. I'm getting my bearings on how/where everything is.


arlith:
This might work, but I think we should do it differently.  Look at how costs are computed for regular Triple Strength Myomer and Supercooled Myomer.  These are computed on line 4916 of Mech.java.  I think that changes should be made there and not in MiscType.

arlith:
The change you made is the general change for variable cost MiscType items, but in this case, myomer boosters are special.  It's somewhat confusing.

vizax:
Ahh, I see. I bow to your wisdom in this. You definitely know this code much better than I. Let me read through this and see what I can figure out. Thanks for the info.

vizax:
oh, another thing...  do i need to close this pull request and resubmit, or can i update this? I'm not terribly used to git. :-/

arlith:
I'm pretty sure you need to close this and open a new one.  In theory, you could revert your two previous commits and make more, but I think it make senses to close this and make a new one (which is what you did; I just wanted to update this PR so it wasn't left hanging).

PR #96, opened 2016-03-23
Industrial tsm - redo

vizax:
Added hasIndustrialTSM() check to muscCost

I believe this was how you were expecting me to edit. Very good call., nice and simple and already ready to go with a small change. I completely missed it.


arlith:
Looks okay to me, thanks for the PR.

PR #97, opened 2016-03-23
Right-Click Scrolling and the MapMenu 

SJuliez:
With this UIManager change the board context menu will no longer consume mouse clicks outside it so right mouse button scrolling isn't blocked by it.

Will merge it if no one stops me.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/MapMenu.java
+++ b/megamek/src/megamek/client/ui/swing/MapMenu.java
@@ -28,6 +28,7 @@ import java.util.Vector;
 import javax.swing.JMenu;
 import javax.swing.JMenuItem;
 import javax.swing.JPopupMenu;
+import javax.swing.UIManager;
 
 import megamek.client.Client;
 import megamek.client.ui.Messages;
@@ -94,6 +95,9 @@ public class MapMenu extends JPopupMenu {
         selectedEntity = myEntity = game.getEntity(gui.getSelectedEntityNum());
 
         hasMenu = createMenu();
+        // make this popup not consume mouse events outside it
+        // so board dragging can start correctly when it's open
+        UIManager.put("PopupMenu.consumeEventOnClose", false);
     }
 
     private boolean canSelectEntities() {
@@ -110,7 +114,7 @@ public class MapMenu extends JPopupMenu {
                    || (currentPanel instanceof PhysicalDisplay)
                    || (currentPanel instanceof TargetingPhaseDisplay));
     }
-
+    
     private boolean createMenu() {
         removeAll();
         int itemCount = 0;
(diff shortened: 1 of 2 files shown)


pheonixstorm:
Oh?? Is that a threat?? lol

So this means no more annoying context menu popping up when trying to move around the map?

beerockxs:
Why don't you all use right-click drag scrolling? Best option and no pop up either.

arlith:
It's not clear to me what effect this PR has.  I understand what the property your changes does, but it doesn't appear to have any effect to me.  

With and without this PR, the right-click scroll behavior seems that same to me.

I'm running tests as an observer in some saved game, and when I right-click and drag, I get the "Special Hex Display" pop-up window, allowing me to create a note on whatever hex my mouse was over when I initiated the right-click-to-drag.  I am still able to scroll, but the pop-up menu remains in the location it originated at.  This happens with or without the PR.

SJuliez:
Oh. I really didn't write all that much as an explanation... The problem you mentioned, arlith, at least as I understood it, was that when the context-menu for the map is open, you can't right-click-scroll. Right-clicking and dragging around will not actually move the map but just open the context menu again in another spot. I found this annoying myself and so I assumed this was what you meant.

I had the impression when testing that the IOManager setting changes that behaviour successfully.

So, 
- it changes nothing so long as the board context menu isn't open. 
- But: When you right-click to open the context menu and then just want the menu to go away you can right-click scroll directly. Without this patch - at least thats what I seemed to experience - right-click-dragging did not make the menu go away, it would just pop up again.

I'll test it again.

SJuliez:
@pheonixstorm No threat :) And I have been stopped anyway...

arlith:
Are you working in Windows?  It's possible we have different Look & Feels that may have different behavior.  Just to be clear, before and after this PR, I can scroll with right-click scrolling, but it seems like the right-click is firing two different behaviors: scrolling, and showing a pop-up menu.  I feel like, we should have logic to determine the context of the right-click and then only perform one of these behaviors.  

I know I am partly to blame for the confusion, because I wasn't very clear when I described the issue in the first place.

Also, the code change in the PR makes sense to me, from the perspective that I understand what it's supposed to be doing.  It's just, I haven't experienced the problem it should be fixing (and the problem I have in mind is different).

So, I think you can merge this change in, but I also wanted to make it more clear the issue I had in mind.

SJuliez:
Yes, I'm working in windows. I just checked again: After this PR, when I open the right-click menu, then right-click-drag from somewhere on the map, the menu closes immediately and does not open again. And the drag itself works. So, no double effect.

Unless there is some way I can set Java to give me the behavior on your system, I probably can't work on a solution.

arlith:
Well, good to know that the issue doesn't occur on all operating systems.

PR #99, opened 2016-03-23
Skin editor

arlith:
This is a pretty large feature I've been working on for sometime, but since @BLOODWOLF333 has been using the branch, I've been making more quality of life changes.

The main features of this branch are a skin editor, which now has a button on the main menu (and the splash screen was resized to deal with the slightly larger size).  Additionally, the UnitDisplay can be skinned as well.  I moved the skin specification XML files into mmconf/skins to keep mmconfg from getting too cluttered.

We should decide if this branch should get merged before or after the next stable release.  It should also have some wider testing before getting merged in.


vizax:
I'll be happy to test, if you'd like. How would you like me to do it?

arlith:
You will need to checkout the branch, then run MM and in the main menu there is now a "Skin Editor" button.  Try it out and provide feedback or report bugs you find.

SJuliez:
I think "Path: " should ideally have some kind of adornment (or be a button or something). I only found out that you can click on it to get a file chooser by looking at the code :smile:

SJuliez:
Aaaah yes, the tooltip actually says it... I still kinda think a button would be better :grin:

SJuliez:
When the only change is to disable a border of e.g. the default button this does not enable the "Save Skin" button and I think it also does not trigger an update of the displays.

SJuliez:
Nice work! Adding the colors to the buttons is not self-explanatory, but that's not really important in such a feature. Once you get to know what everything does, its easy to throw together a skin.

Ah one other thing that may have to do with my Eclipse: On my PC skins get saved to /mmconf instead of /mmconf/skins. They do get loaded from /skins though.

SJuliez:
I note there is a difference between the PhaseDisplay and the MechDisplay. The PhaseDisplay will stop showing any border/corner images when one of them is emptied (no file name). The MechDisplay handles this very well, it just shows what it's given. I think this is an important behavior because it makes skinning much easier if you can just leave away corner images or any borders that you don't want to skin.

SJuliez:
In the MechDisplay settings the bottom line path seems to be ignored and the top line used instead.

arlith:
RE: UnitDisplay bottom line: Bug, fixed.

RE: Saving skins to wrong path: Bug, fixed

RE: border enable/disable not updating: Bug, fixed

RE: Path "button"

The primary reason I changed the style for these buttons is to make the UI more compact.  If I used the default button look & feel, each row needs to have more height.  I tested it out, and it's not too big of a difference, so I changed it to look like the default button.

RE: font colors are confusing

One problem is that the list of font colors is designed to be a generic UI component specification.  Different UI elements could use them different.  For the Phase Display, it determines the color of the "status" text, I believe.  For buttons, it uses three values: 1) default display color 2) disabled color 3) highlighted color.  Different UI widgets can use these values differently, so there's no one description I could easily fit in.  Ideally, the description of this would be in the help documentation.

RE: Phase Display border vs UnitDisplay border.

This is trickier... The UnitDisplay skinned images was added a long ago, before I arrived.  I didn't really think about this when I designed the MegaMekBorder class, and so their implementations differ a bit.  I feel like my code is more straight forward, generic, and modifiable but that may just be because I wrote it.  There are some hard-coded limitations in the MegamekBorder class.  Namely, the width or height of the top/bottom or left/right "lines" are determined by the minimum width or height in the "line".  That is, the top "row" can consist of 3 or more icons: a top left corner, a top right corner, and 1 or more icons along the top "line."  The top row will have a height equal to the smallest value of any of these icons.  This prevents it from having some "gray void" areas.  The UnitDisplay is different, and only allows 1 icon for each "line."  I hope that makes sense.

Things could potentially be adjusted in MegamekBorder; I implemented with the idea of what I thought would be useful, but didn't have art assets.  Given artists willing to make assets, it may make more sense to change some of my design decisions.

arlith:
With respect to the Path buttons, I was wondering... Does anything think that actually typing in the file paths is useful?  I could make it so that clicking in the text box just opens up the FileChooserDialog.  

As it is now, every key press fires an update, which loses focus on the text box.  This needs to be fixed, but while thinking about it, I was trying to decide if I should just make it so clicking on the  text box opened up the dialog, then the Path buttons could go back to the old style.

BLOODWOLF333:
Yes that is a better idea, clicking the text box opens up the dialog or remove the text box all together and just leave the path: button which might be the more compact way to go.

Date: Fri, 25 Mar 2016 05:55:10 -0700

BLOODWOLF333:
After thinking about it, it still should list the file path after the path: button just so that you can see that all the images are coming from the correct folder.

Date: Fri, 25 Mar 2016 05:55:10 -0700

SJuliez:
I agree. No need for typing in by hand.

vizax:
I checked out the branch and tried to run it in eclipse, but no go. i press the skin editor button and the the "are you sure" popup, then nothing happens. So, I built the jar and i get to a "please wait" window, then the Mech Display window, but nothing else.  log attached.

[megameklog.txt](https://github.com/MegaMek/megamek/files/190083/megameklog.txt)

SJuliez:
Had the same problem, see above. Open the Client settings, load some other skin, then reset to defaultSkin. Then, try again.

SJuliez:
@arlith 
Thanks for your answers.
RE: font colors are confusing
OK I can see this is difficult to give online help for. Also, it's obvious that one can add colors and when doing this and choosing some obvious ones, it's easy to see what they do. Not worth changing.

RE: Phase Display border vs UnitDisplay border.
Agree, probably worth changing only when really necessary. What I'm not sure I get right is the following: Why would another system lead to gray void areas? Wouldn't the background image fill any emptiness? 

Actually, a smaller or larger corner image probably has no use in a graphical sense. I just immediately tried ditching the corner images altogether to make things easier.

vizax:
got it, thanks, i missed that note earlier....
RE: Path buttons - What about making them (essentially) a hyperlink?  I'm not opposed to clicking into the box to pull the file dialog, but what if you want to copy and paste the same filename in multiple places (and then edit it, if it is so similar)? clicking to paste would open the file dialog every time.

BLOODWOLF333:
If/when I need to use the same image for multiple borders or corners it is sometimes easier just to open up the skin.xml with gedit, highlight the image name with the .gif,.png and then just ctrl+v it over the text for all the other borders.  I strongly believe the drop down box or dialog pop up when clicking either the path: button or text box is needed.  Obviously have it immediately open and begin inside the widgets folder to save time compared to starting at C: but, the way it is right now is a huge improvement over it's state 3-4 weeks ago.

BLOODWOLF333:
Oh, and one tip; if you decide to scrap a skin.xml that your working on and delete it out of mmconfg the skin editor will not start up fully like as in vizax's comment where the board view window does not show up to where you can click view/client settings (though I'm on Windows), you will have to go into the client settings XML under mmconfg and manually type in another skin file that is complete.  Maybe it should revert to the default blue skin in case it cannot find the skin file it was trained to from the last session but, I've never said anything to Arlith about it because the log tells you what is wrong.

vizax:
couple things for clarification:
- The button `Add Skin Specification` isn't doing anything. I assume it is to create a new one?
- I don't see an explanation about Skin vs skinScrollBars. Whats the difference? will it be necessary to create both for every skin?

I'm making some test images to play with. I'll upload them here in case someone wants to mess with them. (nothing fancy)

vizax:
So it needs some work, but whatever, it gets the point across :grin:
[test skin.zip](https://github.com/MegaMek/megamek/files/190315/test.skin.zip)

meh, its a test, even it it all doesn't align perfectly. I've only done the game board border.

SJuliez:
I also seem to have the empty path problem vizax mentioned. I first thought I was clicking too fast, but thats not it. When an image is set and I set it to another image via the file chooser, the path gets (only) emptied and the image does indeed not get loaded. When I use the file chooser again, this time starting with an empty path, it works. So, in effect, I load each image by choosing it twice.

I just checked again by typing in the path field itself and deleting again... it seems that really the file choosing only works when the path starts out empty.

SJuliez:
.... and just now I notice that this is true for the PhaseDisplay but not the MechDisplay where the file chooser works directly.

SJuliez:
Another idea: Do you think its possible to add a font chooser as well as a color chooser (headline color and text color) for the unitDisplay? The font chooser might alos be useful for the PhaseDisplay.

SJuliez:
Ah just noticed the Path buttons. Nice. The tooltip now says "...relative to {0}" tho :)

SJuliez:
I'm not sure the colors in the PhaseDisplay are still working. Added all 3, the first one is for the status text, that one works, but the other two colors have no effect anymore (I tried reloading, even restarting MM).

SJuliez:
aaaaaah I take that back: I have to activate the colors in the default button component... (Im not sure then colors nr. 2 and 3 in the PhaseDisplay do)

SJuliez:
I just noticed that a button background gets stretched or squeezed to the correct height but a border does not. To create suitable border gfx I think it would therefore be interesting give the exact sizes necessary for those elements that have a fixed size (buttons, the height of the PhaseDisplay, the tabs on the UnitDisplay e.g.).

SJuliez:
The Late Heavy Bombardment of comments :)

SJuliez:
The button height is 33 pixels.

SJuliez:
The Tile checkboxes should also update display :)

BLOODWOLF333:
The first color for the phase display is the color of the button font when not pressed or moused over.  The second color is for the font when that action is not available (the done/move button).  And the third color is for the font when you click the next/turn/etc buttons.

Date: Sun, 27 Mar 2016 07:03:35 -0700

SJuliez:
I would request that the Path dialog remembers the last folder one was in. I think I have performed around 500 mouse clicks today to change folder.

SJuliez:
I put together a simple skin I call Ocean. Obviously not very techy...
Overall a good experience with the Skin Editor. It's nice seeing it come together.

Just a few points I noticed:
- After all the clicking I think entering/copypasting the path in that path field may can make quite some sense after all.
- I find borders not very important as they only hide the important stuff. Still I tried a few things like using borders on buttons to give them left and right edges that do not stretch. Doesnt work.
- I tried to create a smooth transition from the background of the PhaseDisplay to the boardview using an image that fades to transparency. Did not work either, because the bv background does not extend beneath the borders. Instead the borders use their own BG which, given the same image, does not align with the bv BG. It would be without any disadvantage I think to have the bv BG extend under the borders. But thats probably more a bv1 issue than a SkinEditor one.
- I can't say if it's going to be worth the effort but after making this (took me a few hours, as simple as it is) I still find the way the MechDisplay uses the borders much more user friedly than the way the buttons or the boardview does. You can just throw in one image and it effortlessly shows up like the bubbles at the side. Everywhere else you have to create a complete set of images that more or less align. It is not difficult to create a few transparent images, rather needlessly tedious.
- If I'm not mistaken the tab images of the MechDisplay are limited to the exact size of the current ones?

Anyway thanks arlith for putting this together.

ocean_screen2

vizax:
> I can't say if it's going to be worth the effort but after making this (took me a few hours, as simple as it is) I still find the way the MechDisplay uses the borders much more user friedly than the way the buttons or the boardview does. You can just throw in one image and it effortlessly shows up like the bubbles at the side. Everywhere else you have to create a complete set of images that more or less align. It is not difficult to create a few transparent images, rather needlessly tedious to create and enter all 8 or so of them.

if you want really simple, you can tile the same 2 images for horizontal and vertical and just add corners (total of 6 images). Or you can just turn the borders off completely. :grinning:

vizax:
Combined thoughts/comments/questions of @BLOODWOLF333 , @SJuliez , and @vizax:

**Screen updates:**
- remove on textbox change (acknowledged by Arlith)
- apply/save button
- If not a button, then tile checkboxes should update display also

**Path buttons:**
- Clicking in the textbox opens the file dialog
- Clicking the path button opens the file dialog
- Clicking the path text (hyperlink style) opens the file dialog
- Tooltip shows "...relative to {0}"

**Path textboxes:**
- show the "full path" to the file (anything inside the widgets dir...  subfolder/filename)
- copy/paste works well/makes sense

**Buttons:**
- Exact dimensions buttons can/should be, for proper sizing (supposedly 33px)

**General:**
- Full write up on what everything does and what each additional path/font affects (maybe rights to the wiki, for group entry/edits?)
- file dialog should remember last folder used
- Do MechDisplay tabs change size or are they static?
- Make BV background(s) go edge to edge (under/behind the borders)
  - Maybe just BV BG 2, and BV BG 1 will be just the play area inside the borders

BLOODWOLF333:
The phase display is going to stretch with whatever the user decides the window size to be and the buttons with stretch as well.  My buttons are 72 pixels high with a transparent edge all the way around them so they don't get all smushed together.  The board view has a background/back drop behind the actual board piece(i think of it like the board view is the screen on a laptop) and then there is a background inside the board piece to go behind the actual MM board.
[img]http://i.imgur.com/e7O1vLy.png[/img]

vizax:
That makes sense now. Thanks for the explanation.

arlith:
Alright, I think I've addressed most of the issues in the last few commits.

I decided that I didn't want to make it so that clicking on the text box makes the file chooser appear.  I think that in some cases copy/paste will be faster then using the file chooser.  So, I want to be able to support typing into the text boxes.  That means I need to fix the issue where typing moves the cursor.

arlith:
Alright, I think that I've fixed the focus issues.  With that, I think this branch is okay to merge into master.  There could be some more work, but it at least should work.  Also, more documentation could be added for users, but that's always the case.

BLOODWOLF333:
Skin editor is not included with the 0.3.30 bundle.  I did not check if it comes with just the latest MegaMek stand alone download but, definitely does not come with the HQ bundles.

PR #100, opened 2016-03-23
Visibility of minimap unit icons

SJuliez:
This will make the tiny tiny unit icons on the minimap much more obvious (I believe):
The selected unit gets a circle in the color that is set in the advanced settings (its not in the screenshot). I have tried to include the color for !done() units but even with the minimap at largest zoom this makes the colors barely distinguishable so I deleted that again. Maybe I have a good idea how to do it in the future.
Opinions welcome.

minimap

### DIFF
--- a/megamek/src/megamek/client/ui/swing/MiniMap.java
+++ b/megamek/src/megamek/client/ui/swing/MiniMap.java
@@ -1097,15 +1097,25 @@ public class MiniMap extends JPanel {
         g.setColor(pColor);
         
         // Fill the interior of the unit marker
-        if (!entity.isSelectableThisTurn()) {
-            // entity has moved (or whatever) already
-            g.setColor(g.getColor().darker());
-        }
         g.fillPolygon(xPoints, yPoints, xPoints.length);
-
+        
         // Create a colored border according to the done() status
         Entity se = clientgui == null ? null : m_game.getEntity(clientgui
                 .getSelectedEntityNum());
+        if (entity == se) {
+            g.setColor(GUIPreferences.getInstance().getColor(
+                    GUIPreferences.ADVANCED_UNITOVERVIEW_SELECTED_COLOR));
+        } else if (entity.isSelectableThisTurn()) {
+            // entity has moved (or whatever) already
+//            g.setColor(g.getColor().darker());
+            g.setColor(GUIPreferences.getInstance().getColor(
+                    GUIPreferences.ADVANCED_UNITOVERVIEW_VALID_COLOR));
+            
+        } else {
+            g.setColor(Color.BLACK);
+        }
+        g.drawPolygon(xPoints, yPoints, xPoints.length);
+/*
         if (entity == se) {
             Color w = new Color(255, 255, 255);
             Color b = new Color(0, 0, 0);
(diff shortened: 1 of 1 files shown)


pheonixstorm:
Much more visible. Hope it makes it in soon. 

A question for all the devs though. At the high zoom levels, would it be possible to use the warfare symbols for single units? It might enhance the flavor of the fight and make it feel more realistic/believable.

SJuliez:
You mean those rectangular ones in Strat Ops? They would be a tight fit even at biggest zoom but I could try. The current symbols could also be changed to resemble the big ones better, like a 5-star for Aero, a rounded rectangle for vehicles...

wildfire142:
If your looking for the Strat Ops symbols check this thread out http://bg.battletech.com/forums/index.php?topic=51270.0 it has a file with all the bits you need to try this, in an easy to use format and size.

pheonixstorm:
@wildfire142 
The symbols wouldn't need to be complicated but they do need to be whole. For the minimap the examples below would work best.

@SJuliez the ones in use by HQ should work in a pinch for testing but for an actual release I think just the border and the center icon with no extras might work at a scale small enough. 

This might work: 
mech

But NOT this:
mmechlance

What do you think?

wildfire142:
Aye, though the parts file has all the bits you need to cherry pick what you want :)

You can even just take the straight unit ID's if you want without the border.

SJuliez:
@wildfire142 Thanks for the link. I'll need the flexibility of Paths here for painting (border, shadow etc...). At least, with Paths I *know* how to do it :)
@pheonixstorm This is how it could look. (It's not completed, and it's not in this branch)
Note that on big maps such as this the minimap is actually prevented from getting this big; I had to comment the code out to even make this screenshot.
It's actually better than I had expected it to be.
mm_symbols

arlith:
While we're on the subject of the minimap, I've been thinking for a while that we should rework the minimap for some time.  I feel like we should allow the minimap to be resized to any size, and instead of drawing the whole map, the minimap should only draw as much of the map as can fit inside the window at a currently selected scale.  Users could then adjust the scale of the minimap (like they can now) but this would be much more flexible in how they size the minimap window.

pheonixstorm:
that doesn't look half bad

SJuliez:
@arlith I'm a bit reluctant on this. For me the minimap should always show all the map. To an extent, having to scroll in it would defeat its purpose of giving me an instant overview of the battlefield. Also there is enough scrolling around in the boardview, Id rather not have to in the minimap. So far I liked the easy way the minimap just zoomed and needed no other interaction.

My (tentative) plans therefore were to add a mouse button4 function to show/hide the UnitDisplay and Minimap together to easily get rid of both when they're in the way and get them back. Also I wanted to remove the window decorations on the minimap if possible, which would even take away the ability to resize.

SJuliez:
@pheonixstorm This is btw how it could look in the smallest minimap zoom level. Nevermind the bad aero star. I would personally make this a toggle in the menu rather than switch at some zoom level. At the very least, when symbols are automatically switched, the symbols in both forms must be very similar.
I don't know. Worth continuing?
mm_icons

pheonixstorm:
Yeah, I think it is worth continuing. One things though, the aero star and other offset symbols (such as the mech must be centered for this to look right which is why I said the parts would not work well. Hence my example. 
Border + centered unit icon = Perfect 
Border + off center unit icon = Something than resembles a flag rather than a warfare symbol

If you need actual military references I can provide a few pdfs to give you a better idea.
Can also check out this from Milsketch http://www.historicalsoftware.com/downloads/MILSketch.jpg

pheonixstorm:
I found an image of stacked unit icons on a hex map https://eltentaculodecthulhu.files.wordpress.com/2008/12/dak-0001.png

It would be interesting if this were possible and to make the images (stacked) spread out into adjacent hexes if the mouse hovered over the hex for more than x seconds. Just something else to think about ;)

PR #104, opened 2016-03-25
Updates to history.txt

SJuliez:


pheonixstorm:
@SJuliez Don't know how the other devs feel but please don't do this kind of PR. Just edit the file and push normally. This kind of PR pollutes the history with excess commits. Unless I'm mistaken you should be able to push here w/o needing a PR

SJuliez:
@pheonixstorm Np.

Akjosch:
I'd second this. Updates to `history.txt` should be just a simple commit, no need for a PR. Same for a quick fix which only affects a few lines, like adding a null pointer check somewhere or fixing a spelling mistake.

Akjosch:
I'd second this. Updates to `history.txt` should be just a simple commit, no need for a PR. Same for a quick fix which only affects a few lines, like adding a null pointer check somewhere or fixing a spelling mistake or data entry error.

SJuliez:
Hmm, I just tried but it got rejected as "non-fast-forward". If I fetch and merge in the current master, the local branch shows 2 additional commits. I suspect that pushing this will again not be ideal. Whats the right way of doing this?

pheonixstorm:
You would probably need to have this project repo on your system and push directly from it rather than trying to push from a fork. **OR** you can edit the file directly on this projects web page and commit.

https://github.com/MegaMek/megamek/blob/master/megamek/docs/history.txt over by the history button is a pencil icon. Click it and edit the page directly then add your commit. NOTE the commit message is all the way at the bottom of the page.

arlith:
@SJuliez The 2 additional commits are going to be 1) your commit and 2) the merge commit.  This is common.  The history will show the merge commit, but that's ok (and can be important; if there were a merge conflict which was resolved improperly, that commit could eventually be revisited).

PR #109, opened 2016-03-26
StratOps/NATO Symbols in the Minimap

SJuliez:
OK, as requested the alternate minimap symbols according to part of StratOps pp 340/341 (more or less NATO standard). Theres a checkbox in the Client settings to switch between these and the normal ones. Completely optional.
Opinions?
minim_strat_1
minim_strat_2
minim_strat_3

### DIFF
--- a/megamek/i18n/megamek/client/messages.properties
+++ b/megamek/i18n/megamek/client/messages.properties
@@ -635,6 +635,7 @@ CommonSettingsDialog.AOHexSHadows=Show ambient-occlusion-like hex shadows at the
 CommonSettingsDialog.useShadowMap=Show terrain and building shadows
 CommonSettingsDialog.levelHighlight=Level highlighting (thick borders at level changes)
 CommonSettingsDialog.floatingIso=Floating isometric (no hex sides in isometric mode)
+CommonSettingsDialog.mmSymbol=Use StratOps unit symbols in the Minimap
 ConfirmDialog.dontBother=Do not bother me again
 
 CustomBattleArmorDialog.title=Build your Battle Armor...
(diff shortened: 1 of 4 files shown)


HammerGS:
I think it looks good.  Are the background colors taken from the player colors?

SJuliez:
Yes, the version without transparency that the tooltips also use.

HammerGS:
Well as I side, I like it.  Wonder what they would look like with a Camo pattern as background.

pheonixstorm:
I don't think camo patterns should be used. Some will bled in with the background and make it more difficult to see, especially considering the size of the icons.

Also, I still think that my icon suggestion from before would work better. Border + unit icon. Adding in the unit size at the smallest icon size will just clutter the image. As for NATO/SO symbols, I would prefer the SO symbols but even so the NATO symbols look good. **IF** the NATO symbology is used then I suggest using the friendly/opfor style. Rectangle for allies, diamond for opfor.

What about the other idea I had on having the icons spread out on mouse hover. Is it possible?

Can't wait to see this in game. Think I'm going to have to try and merge and test on my fork :space_invader:

SJuliez:
On camos: Have to agree with phoenix. It might look good with brightly colored ones, but the dark ones would hide the actual unit type; also, the icons are really tiny and I guess the camo would not be really visible all that well. I actually tried but the TileManager just didn't want to give me anything. It said "Don't do it" :)

NATO: I was talking about the "Unit icons" in 
https://en.wikipedia.org/wiki/NATO_Joint_Military_Symbology
not the more generic ones under "Battle dimension".
The unit icons are more or less the same as in StratOps. No changes planned here.

On the size: I wouldn't have left the unit weight in if it was totally unreadable at the smallest scale, but it's actually ok. I suggest trying the system in "reality" and seeing what we should make better. Maybe adding the weight to more units would make sense (vehicles...) - or - indeed removing it from all. I just went for what StratOps did for now.

HammerGS:
Are there images for this commit?

SJuliez:
Image files? No, all Path2D.
(Not sure if I misunderstand)

HammerGS:
I thought I saw something regarding the Force Icons for MekHQ [http://bg.battletech.com/forums/index.php?topic=51270.0](url) and assumed those would be needed as well.

PR #110, opened 2016-03-27
Bugfix/78 unit number producting invalid xml

### LINKED ISSUE #78: unitNumber producing invalid XML
While researching the feasibility of removing the remaining usage of XStream to replace it with JAXB (https://github.com/MegaMek/megamek/issues/62#issuecomment-195764033), I found an issue in Entity where invalid XML is being generated.

Entity.unitNumber is initialized to a char representation of -1, which turns into character `\uffff`.  When this is marshaled into the save file, it is turned into `&#xffff;`, which is illegal for XML (http://www.xml.com/2000/05/10/conformance/reports/report-xerces-nv.html).  Attempting to un-marshal the file, even without full object mappings, with JAXB will fail due to the illegal character.  If you change the PrettyPrintWriter that XStream uses to generate the file to use strict XML 1.0 or 1.1, that will fail due to the illegal character.  XStream, by default, uses quirks mode, which is allowing it to read/write this value.

So, any change with this attribute will result in a breaking change with existing save files, even if the data type is changed, or the initial value is changed.  A conversion program could be written that can open the gzipped save file and do a search for "&#ffff;" and replace it with "-1", or some other value.

What is the consensus of the development team for how to handle this?

nderwin:
Fixes #78


pheonixstorm:
Can you recreate this with a clean branch w/o all the "clutter"? I think it would be better to not have all the unnecessary merge commits whenever possible.

Just my 2 cents anyway.

nderwin:
Those "unnecessary" merge commits are just me keeping the branch up-to-date with the upstream/master branch, to make sure there are no merge conflicts.  The number of commits really doesn't matter as much as the differences in the files.

arlith:
You tested these changes, right?

Did you look to see if unitNumber got written to MUL files?  Did these changes effect that at all?

The code changes all look reasonable to me.

nderwin:
@arlith Always!  It does not appear that the `unitNumber` is written to MUL files.  It is, however written to the save game file.  Odd.

The getter is used only in `Server`, `CustomMechDialog`, and `GameTurn`.  The setter is only used in `Server` and `CustomMechDialog`.  The property is only referenced from the accessor methods in `Entity`.  +1 for Find Usages in the IDE.

Also, I tested adding 10 protomeks to a side, and noted in the save game file that 5 of them had unit number 0, and 5 had unit number 1.  All other entities were -1.

[muls.zip](https://github.com/MegaMek/megamek/files/194557/muls.zip)
[unitNumberTest.sav.gz](https://github.com/MegaMek/megamek/files/194558/unitNumberTest.sav.gz)
[unitNumberTest2.sav.gz](https://github.com/MegaMek/megamek/files/194559/unitNumberTest2.sav.gz)

arlith:
I'm somewhat surprised that it's not saved to the MUL, but then again, I guess I should have expected it.  Protos seem like they're not as well used, so it's something that could get passed by.

I could potentially see it being tricky to load in.  For instance, you want to have a group of 5 protos be in a unit, so they get a unit number of 1.  What happens if I load that MUL twice?  Ideally, you should get the first group with a unit number, and the second group with a different unit number.

Anyways, that's outside the scope of this PR.

PR #116, opened 2016-03-31
Display distance info in the tooltip

SJuliez:
Changes: 
- If there is a selected unit, the board tooltip will display the distance to that unit. The selected unit is the active unit of a player, not necessarily the one viewed in the unitdisplay.
- In the movement phase, when a planned movement is shown, the distance to the endpoint of that is also shown.

I thought that especially when planning movement, this is basically what we are always looking for (or counting in the case of tabletop...). It's more direct and less of a visual obstruction than the firing arcs. Ideas, comments?

In the screenshot, the mouse was on the hex below the "6" of the movepath.

tt_1


arlith:
I think that this is useful information.  I think that there are multiple ways to get this information (there are two rulers, one you get with ctrl-click and one with alt-click or middle mouse button, there's the "max modifiers" display that functions somwhat like the movement envelope, and during the firing phase the firing solutions), however I think that in this case, redundancy is good.  Users may not always know about all of the options that exist.

PR #112, opened 2016-03-29
SPAs for air units

pheonixstorm:
Basic Golden Goose and Ground Hugger
Shaky Stick

Still working on getting th remainder of GG and GH finished as well as one or two other spas for air units.

PR #123, opened 2016-04-02
SPA Foot Cavalry

pheonixstorm:
This one is ready to go and should make infantry users happy.
Gives foot infantry 1 extra MP allowing 0mp infantry to now shoot and move.

Only thing left for this is the terrain modifiers. Once I have those posted this will be ready to merge. Got overly excited at the MP bonus being so easy I kinda forgot about em.


pheonixstorm:
There actually is another infantry SPA that can be added. Urban Guerrilla
As much as can be added for it though will require a lot more planning than this one. For now though this PR is complete and ready for review/merge.

arlith:
There seems to be superfluous code attacked to `WeaponAttackAction`.  It also seems like it would make sense to limit this SPA only to infantry...

pheonixstorm:
@arlith 
Yeah, I based this off the wrong branch (SPA instead of master). The commit that includes the code from waa is a NPE stopper for several other SPAs, mostly found in #85 and #112 

I can probably rebase that out since this branch is behind megamek/master. It may not be needed for this PR but it will for the others. Let me know if I should remove it from here.

pheonixstorm:
Fixing an incorrect assumption on how the code handles infantry in move cost, will have it up shortly.

pheonixstorm:
Urban Guerrilla is now partially done. It is missing its local support spawn but otherwise works well.

To be able to spawn support I have a few things I need to figure out if anyone thinks it is worth adding
First being able to create a new Foot rifle platoon one the fly. I think I can use a modified reinforce call but it would require a deployment based on the code for artillery radius R2. Next I would have to randomly determine the spawn coords using a modifier scatter function which in turn would have to make sure the coords were not off-board. And finally make sure they did not spawn until the end of the turn.

Does anyone think it is worth it?

Akjosch:
So essentially ... (on the server thread):

```
Infantry infantry = new Infantry();
infantry.setChasis("Guerrilla Local Support");
infantry.setModel("(Rifle)");
// Possibly infantry.setTechLevel(TechConstants.T_IS_EXPERIMENTAL);?
infantry.setSquadSize(7);
infantry.setSquadN(4);
infantry.autoSetInternal();
infantry.setPrimaryWeapon(InfantryWeapon) EquipmentType.get("Auto-Rifle"));
infantry.setId(getFreeEntityId());
game.addEntity(infantry);
send(createAddEntityPacket(infantry.getId()));
```

... to create that unit, then add it to some kind of auto-deployment queue, and when the end-of-turn phase rolls around, find a suitable hex field (only if there are legal ones in range; reroll if you roll one out of board or illegal) and basically call something like (still in server thread) ...

```
infantry.setDeployed(true);
infantry.setPosition(pos);
entityUpdate(infantry.getId());

```

... possibly also checking if you land on a mine field.

Not sure if there's an auto-deploy queue already. If there isn't, it shouldn't be a big problem to add one.

Akjosch:
Side note for @arlith - because he wanted practical examples of event handling being useful, here's how you write that without adding any new data structures to the server and without caring on which thread the decision to spawn guerillas happened:

```
public class GuerillaSpawn {
    private Coords origin;
    private int teamId;

    public GuerillaSpawn(Coords origin, int teamId) {
        this.origin = origin;
        this.teamId = teamId;
    }

    @Subscribe
    public void spawn(EndPhaseEvent event) {
        // do all of the above
        EventBus.unregisterHandler(this); // we're done
    }
}
```

And then, wherever you want:

    EventBus.registerHandler(new GuerillaSpawn(pos, myTeamId));

HammerGS:
@Akjosch 
There are rules for Guerrilla Infantry in Historical: Reunification War, also it includes rules for Mobs.  But @arlith and I have had conversations in the past about fully implementing infantry to TW/TO rules which would technically mean we should add the H:RW infantry as well.

pheonixstorm:
@Akjosch  The entry in both Alpha Strike Companion and A Time of War Companion say to spawn a generic foot platoon (rifle) which can be read from the blk file. From what I found doing so from a mul file would be easiest but I don't think we need to clutter things up with that when I know there is a call for reading from a blk. Just haven't found it yet nor the deployment code. But the other items I addressed are sill valid. The support unit is supposed to show up during the end phase within a 3 hex radius.

Akjosch:
You can see the sequence for spawning a new unit in Server's abandonEntity(), for example.

pheonixstorm:
Not exactly what I was after but it does help.

I think the most useful will be `pilot.setPosition(targetCoords);`
I can put in in a loop to make sure the scatter coords are not off-board. Now just need to find the function so I can load a new infantry entity from a blk and I might be able to work this in today. 

Will also need to create a new button for this.

beerockxs:
MechSummary ms = MechSummaryCache.getInstance().getMech("Foot Platoon (Rifle).blk").
Entity entity = new MechFileParser(ms.getSourceFile(), ms.getEntryName()).getEntity();

should work

Akjosch:
Actually, without the ".blk"

The only problem? MechSummaryCache gets *cleared* before the first turn (specifically, just before entering PHASE_DEPLOYMENT). So does MechFileParser and RandomNameGenerator. Asking it for the MechSummary means it has to load all the unit files again.

You probably can get away with just saving the MechSummary instance for the infantry before the game starts, but I suggested creating the unit from scratch for a reason ...

beerockxs:
I think creating from scratch is better, too, because we don't want to hardcode any filepaths or filenames. Those can change.

pheonixstorm:
Have a question I posted on the forums. http://megamek.info/forums/index.php?topic=2365.0
Found a few commits I am running through so I can make all the menu changes.

Akjosch:
Because I was curious, I checked how much memory the various data tables "waste", to see if it even makes sense clearing them up during gameplay. All data after clicking "Start a New Game ..." and waiting for them to load.

RandomNameGenerator: 113.234.957 bytes
RandomUnitGenerator: 32.453.440 bytes
MechSummaryCache: 1.764.387 bytes (because most Strings used are referenced multiple times, else it would be ten times as much)
MechFileParse: 659.262 bytes (in one static `Vector<String>`)

Given modern RAM sizes, MechSummaryCache and MechFileParse don't even matter. The others are here because of all the duplicated strings. A simple change, interning all of the strings (*not* something I'd suggest as an optimisation under any circumstance, just a quick test) pretty much confirms that by lowering the sizes to the following.

RandomNameGenerator: 26.438.879 bytes
RandomUnitGenerator: 19.615.432 bytes

*Not* throwing this data away, in particular the MSC one, would make loading the unit from disk actually quick and easy as opposed to freezing the game for a few seconds for those players unlucky enough to not run off an SSD.

pheonixstorm:
@Akjosch  @arlith 
Having an issue with the code. When the button is pressed it calls a function in infantry.java which is supposed to set isCallingSupport to true. For some reason this doesn't seem to happen even though the function is being called correctly. Other than that one flaw it works fine. I can debug the code and change the boolean to true when it gets checked in server.java and the new unit pops up fine.

beerockxs:
Without looking at the code: You're probably setting isCallingSupport to true in the Client's copy of the Infantry object and the Server doesn't know about it. You'll need to have the client update the Entity on the Server using Client#sendUpdateEntity

pheonixstorm:
That did it, thanks bee. Will have the final commit up in a minute. The final code is a tad gnarly so it may need some cleanup.

pheonixstorm:
Everything is now finished and fixed.
Review and merge when ready.

Akjosch:
So, random notes:

isInUrbanEnvironment() doesn't need IGame, just pass IBoard. Keeping the methods to use as specific instances as possible helps with the eventual goal of allowing multi-board games (ground, high altitude, space, ...).

"Urban Guerilla" is from aToWC, not aToW.

Why does the isInUrbanEnvironment() do what it does and in particular why doesn't it check the target hex itself? (some rule clarification/errata?)

The support's troop strength needs to be cut down to the calling unit's commander's Leadership skill.

The support troop needs a check to "scatter" (flee) as soon as their numbers are reduced to half or less.

The support troops needs to only be able to attack when they are in a structure (so, I guess just building right now, until we have walls ...)  and only when they are within 3 hexes of the original calling unit.

pheonixstorm:
@Akjosch 
isInUrban is design to make sure the infantry unit is within an urban or suburban environment. W/o the check it just wouldn't work as not all urban maps use pavement and a road can't count since roads are uses in non urban maps. On this it was better to check adjacent and only adjacent as there are times on maps your infantry might be on a road or clear terrain but still adjacent to pavement, rubble, or buildings. If you look at the city_suburbs map it has road/buildings on clear terrain rather than pavement. 

Megamek does not have a leadership skill. In this case I use the full platoon. I could base this off the size of the calling unit (squad or platoon) but in the end it wasn't really worth it. The same with the flee < 50% This is based on the SPA wording from Alpha Strike Companion rather than aToW. I could have them set to crippled at 50%... I saw the code somewhere..

> The support troops needs to only be able to attack when they are in a structure (so, I guess just building right now, until we have walls ...) and only when they are within 3 hexes of the original calling unit.

This is where I diverge from aToW and stick with ASC. They spawn within 3 hexes as a new unit under the players control. aToW is first and foremost an RPG so its rules don't make the best of sense for TW. ASC on the other hand does but is just as limited in some ways. aToW says a green unit while ASC is based on the calling units gunnery +2 which could spawn an elite unit at best or a skill 8 at worst. It is a tradeoff between the two. It's not perfect.

Akjosch:
I'd still check the target hex itself for the isInUrbanEnvironment() in addition to the others, just in case they are in that one pavement hex surrounded by roads on ground or parks or whatnot.

Without pulling the Leadership skill into it, I'd suggest limiting the size of the support troop to something like 10 - (2 * calling infantry's gunnery skill). In any case, they should *never* be more than 10 people strong.

pheonixstorm:
Which is why ASC and aToW differ. aToW is based on RPG values while ASC calls for TT vaules.

I can change the call for isInUrban in a bit once I am done with Eagle's Eyes. Probably base the spawn size on the calling unit as well since I think even a squad by AS rules should do 1 point of damage at short range (I think).

pheonixstorm:
Since battle armor are included in allowed unit types I decided against checking the squad size and adjusting accordingly.

I did tweak isInUrban to include the calling units hex.

So unlike arlith has something to add I consider this complete.

megamek/src/megamek/common/Compute.java
```diff
@@ -5457,6 +5457,34 @@ public static int dialDownHeat(Mounted weapon, WeaponType wtype, int range) {
         return entities;
     }
 
+    public static boolean isInUrbanEnvironment(IGame game, Coords unitPOS) {
```
arlith: You end up checking the unit hex 6 times.  This seems superfluous.  I would like to see the unitHex pulled outside of the loop, so that the loop only checkes the `hextor` IHex (I also feel like `hextor`variable name could be better; it doesn't seem to mean anything to me, I would've gone with `adjHex` or something similar).  Likewise, `tempcoords` could be `adjCoords`.

pheonixstorm:
The loop was modified from another function. Either one about smoke or the swarm missiles and I didn't change the variable names. It would be better to do that. As for unit hex checking I don't really know why I put it in the loop. Just wasn't paying attention I guess. Will get that fixed.

pheonixstorm:
Found a bug last night and testing a fix.

pheonixstorm:
Bug fixed, PR ready to go.

PR #124, opened 2016-04-02
Hidden units

arlith:
Opening this up as a PR to consider merging it.  The one outstanding issue that I can think of is that, if a client is taking a PBS turn and the client disconnects, it can put the game into a stuck state.


pheonixstorm:
I take it that isn't an easy fix?

megamek/src/megamek/client/ui/swing/PointblankShotDisplay.java
```diff
@@ -0,0 +1,1068 @@
...
+            priority = p;
+        }
+
+        public String toString() {
+            return Messages.getString("FiringDisplay." + getCmd());
+        }
+    }
+
+    // buttons
+    protected Hashtable<FiringCommand, MegamekButton> buttons;
```
nderwin: Do we need the synchronization on this, or could it be a plain `Map`?

arlith: It should be a Map.  This was copied from `FiringDisplay` code, which was originally written as a `Hashtable`, from a time when I was less careful about types.  All of the *Display classes should have their buttons `Hashtables` changed to `Map`.

arlith:
I think that there's room for improvement in this feature, but it also needs more feedback and testing, and without pulling it into the dev releases it won't get that, so I'm going to merge this for now.  At the very least, it doesn't effect anything if the hidden units rule isn't turned on.

BLOODWOLF333:
I volunteer to break my MM testing things.

PR #135, opened 2016-04-08
Event system move to MegaMek

Akjosch:
Event system, same as in https://github.com/MegaMek/mekhq/pull/39 (see there for discussion and examples), just moved to megamek.common.event package.

Needs the copyright headers in the new class files. I'll add those as soon as we decided on the exact wording.


arlith:
EventBus line 146 has a reference to HQEvent instead of MMEvent in a comment.

Other than that, looks fine to me.

Akjosch:
It also had one in a system error message on line 67. :) The things refactoring can't catch, basically.

Issue #149, opened 2016-04-16
IS Small Laser firing at LOS Range


ObviousTech:
With LOS Range enabled, units are currently able to fire with IS Small Laser (or any other non-pulse energy weapon).

Small Laser has a long range of 3, which is less than 7 (direct-fire energy weapon minimum to use LOS range), thus is not allowed to fire at LOS range.

ObviousTech:
I suspect MG is because is because MGWeapon is missing the F_DIRECT_FIRE flag

ObviousTech:
For the small laser, the current logic is:
if (PULSE and (ENERGY or DIRECT_FIRE))


It should probably be:
if (DIRECT_FIRE AND (PULSE OR ENERGY))

pheonixstorm:
LOS Range is pg 85

ObviousTech:
F_DIRECT_FIRE appears to have been originally designed for determining if a weapon works with TCs.

F_DIRECT_FIRE currently is also used for determining if a weapon works with LOS Range and Strafing.

- Based on the strafing rules (TW p243) as direct-fire energy weapons, flamers should be able to be used in strafing.  They currently can't in Megamek because they aren't flagged as DIRECT_FIRE.
- MGs and Flamers are not flagged as DIRECT_FIRE, so right now they can fire at LOS Range (but shouldn't be able to according the LOS Range rules).

mkerensky:
AMS has the same issue I think.  Princess fires them at me at LOS range. I've never tried to fire one back.

pheonixstorm:
That would be a different issue entirely. AMS should not be used as a weapon. Ever.

HammerGS:
@pheonixstorm Might want to read TO pg 99-100 Option Rule but still valid.

pheonixstorm:
Didn't think that was actually in MM. Actual option in MM or just a bug that could be turned into that option?

arlith:
Pretty sure there's an option to fire AMS like MGs per the optional rule.

scJazz2:
I think this might have created a bug. Small Lasers are no longer able to fire at Extreme Ranges. Presumably, this effects MGs and Flamers, etc.

arlith:
I cannot reproduce this issue.  If this is in fact a bug, you should create a new ticket for it, and provide steps to reproduce.  A saved game for testing would be useful.

PR #150, opened 2016-04-16
Issue #149: LOS range

ObviousTech:
**DO NOT USE, MGs still bugged**
_Applying the DIRECT_FIRE weapon type flag to MGs will make them work with TCs (which they shouldn't)_

Fix for both parts of issue #149 :
- LOS Range logic bug for non-pulse energy weapons
- Missing DIRECT_FIRE flag for MG weapons

### DIFF
--- a/megamek/src/megamek/common/Compute.java
+++ b/megamek/src/megamek/common/Compute.java
@@ -1033,9 +1033,9 @@ public class Compute {
                 }
             }
             // No Direct Fire Energy or Pulse with range < 7            
-            if (wtype.hasFlag(WeaponType.F_PULSE)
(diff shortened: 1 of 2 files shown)


ObviousTech:
Not sure if MGWeapon should really be F_DIRECT_FIRE as it is already F_BURST_FIRE.  Should look through everything that uses weaponType.F_DIRECT_FIRE and see if MGs would still work correctly if they had this flag...

Issues:
- client\bot\princess\FireControl.java#L898: F_DIRECT_FIRE is used in determining if a weapon can be used with targeting computer.  Flamers and machine guns should be specifically excluded from targeting computer use.  Neither Flamers not MGs are currently F_DIRECT_FIRE.

PR #153, opened 2016-04-16
SPA: Forward Observer

pheonixstorm:
After testing this, this particular SPA should make VTOLs and infantry even more viable as spotters in a game.

No know bugs, ready for review and merge.


megamek/src/megamek/common/weapons/ArtilleryBayWeaponIndirectFireHandler.java
```diff
@@ -181,18 +181,34 @@ public boolean accept(Entity entity) {
             // Out of any valid spotters, pick the best.
             while (spottersAfter.hasNext()) {
                 Entity ent = spottersAfter.next();
-                if ((bestSpotter == null)
-                        || (ent.getCrew().getGunnery() < bestSpotter.getCrew()
-                                .getGunnery())) {
+                if (bestSpotter == null) {
                     bestSpotter = ent;
+                } else {
```
arlith: This addition doesn't make much sense to me.  What is the purpose of the continue line? Is that meant to be a break?  It seems like the the first two conditions should be an if/else if.  This code is repeated again later on in ArtilleryWeaponIndirectFireHandler.

pheonixstorm:
I've got nothing. Best guess is the continue should have been if forward observer was not found continue the loop.

So would pushing the middle choice up and changing the other two into an if/else to use continue or break work?

something like if fo break else continue?

arlith:
I don't see why you need a break of continue at all.  The continue is superfluous because that's what the loop is going to do anyways.  As for breaking when you find a forward observer, the gunnery skill of the forward observer has an effect on the forward observer mod (based on your code), so, just because a crew *has* forward observer doesn't mean it's the best candidate for a spotter (there could be a forward observer with lower gunnery).

pheonixstorm:
Is an if else/if really needed? A few ifs should be fine since there is nothing else to test for and I really don't like leaving out the final else, but.. Your call. I have the code ready for commit. Would the following be fine?

>                 if (bestSpotter == null) {
                    bestSpotter = ent;
                }
                if (ent.getCrew().getOptions().booleanOption("forward_observer")
                        && !bestSpotter.getCrew().getOptions().booleanOption("forward_observer")){
                    bestSpotter = ent;
                }
                if (ent.getCrew().getGunnery() < bestSpotter.getCrew().getGunnery()
                        && !bestSpotter.getCrew().getOptions().booleanOption("forward_observer")) {
                    bestSpotter = ent;
                }
                if (bestSpotter.getCrew().getOptions().booleanOption("forward_observer")
                        && ent.getCrew().getOptions().booleanOption("forward_observer")) {
                    if (ent.getCrew().getGunnery() < bestSpotter.getCrew().getGunnery()) {
                        bestSpotter = ent;
                    }
                }

arlith:
Yes, that seems fine.

arlith:
Could you also update the tooltip for forward observer?  I don't feel like it's terribly descriptive.  Ideally, the tooltip should give you a good idea of what the SPA/Quirk/Option/Whatever does, to save you from having to look it up in the rulebook.

pheonixstorm:
Ok, will change and commit in a few.

pheonixstorm:
If there is nothing else this is now ready.

PR #154, opened 2016-04-16
Replacing the internal representation of mass/tonnage with doubles

Akjosch:
This replaces everything which is related to mass, tonnage, carrying capacity and so on with `double`.

What is left as `float`, due to not needing the precision, are:

* Game options
* Damage
* Random generator weights
* Gravity

This is an API-changing PR. This means I'll need to create PRs for MegaMekLab and MekHQ, and whoever is using MegaMek.jar as a library (like MekWars) will need to be made aware of the change.


Akjosch:
I checked the current MekWars trunk, and it looks like it doesn't interface with these MegaMek classes, so no change necessary there.

PR #155, opened 2016-04-18
Add the possibility to load a part of a file for hex images

Akjosch:
This PR adds a generic method to load images as a replacement/enhancement to `toolkit.getImage(filename)`.

In particular, declaring an image as "image.png(X,Y-W,H)" will load image.png, then return a rectangle of size (W,H) cut at pixel position (X,Y) from the image, enabling relatively easy tile maps and image atlases.


Akjosch:
Also, I made the loading wait for the images to load fully, preventing (in particular) the map editor from having to be started *twice* to see all the hex images ...

Akjosch:
Slight optimisation here: Now every image loaded via `ImageUtil.loadImageFromFile()` will be a BufferedImage ready to be used with hardware acceleration, at least for the default two loaders (custom loaders are free to not care).

PR #169, opened 2016-04-24
Bug#4578.quirks.unit.type

arlith:
Ok, I think this is all set to go.  This fixes [SF bug #4578](https://sourceforge.net/p/megamek/bugs/4578/).

Adding the unitType tag to quirks was the easy bit.  Then I spent a bunch more time writing a utility to help me rewrite the canon quirks XML so that it has the right unitType tags.  I think I've got it now, along with some fixes in the canon quirks list.

Plan to give this a few days for code review, then I'm going to merge it.


pheonixstorm:
Is the unit type tag now required or optional. If required then I, and who knows who else, will need to update the custom quirks files as well, so may want to big announced once merged and released.

arlith:
If it's not present, "mech" is assumed.  I also have a "tool" that could be used to potentially update the quirks file.  Basically, it reads in the canon quirks file (which could be replaced with the custom quirk file) and for each quirk unit Id and unit combination, it looks to see if the ids could match not considering unit type.  If a match is found, it adds an entry to the custom quirks file.

Issue #207, opened 2016-05-19
41.17 : Bot refuses to deploy because of NPE


Question2:
Specifically the clan jade falcon bot won't deploy.

`java.lang.NullPointerException
	at megamek.common.pathfinder.ShortestPathFinder.getLevelDiff(ShortestPathFinder.java:397)
	at megamek.common.pathfinder.ShortestPathFinder$MovePathAStarComparator.compare(ShortestPathFinder.java:212)
	at megamek.common.pathfinder.ShortestPathFinder$MovePathAStarComparator.compare(ShortestPathFinder.java:167)
	at java.util.PriorityQueue.siftUpUsingComparator(Unknown Source)
	at java.util.PriorityQueue.siftUp(Unknown Source)
	at java.util.PriorityQueue.offer(Unknown Source)
	at java.util.PriorityQueue.add(Unknown Source)
	at java.util.AbstractQueue.addAll(Unknown Source)
	at megamek.common.pathfinder.AbstractPathFinder.run(AbstractPathFinder.java:298)
	at megamek.common.pathfinder.AbstractPathFinder.run(AbstractPathFinder.java:325)
	at megamek.client.bot.BotClient.hasPathToCenter(BotClient.java:778)
	at megamek.client.bot.BotClient.getStartingCoordsArray(BotClient.java:724)
	at megamek.client.bot.princess.Princess.calculateDeployment(Princess.java:247)
	at megamek.client.bot.BotClient.calculateMyTurn(BotClient.java:397)
	at megamek.client.bot.BotClient$CalculateBotTurn.run(BotClient.java:99)
	at java.lang.Thread.run(Unknown Source)`

[#battletech.zip](https://github.com/MegaMek/megamek/files/272347/battletech.zip)

pheonixstorm:
Is this the same as issue #33 ?

dericpage:
Related, but not the same no.  In the previous case, the current hex was NULL.  In this case it's the destination.

pheonixstorm:
Ah, ok.

arlith:
With the last referenced commit, the only way I could see this happening has been removed.  I can't verify that this fixes the bug, but we also added logging code that may help work this issue out if it does happen.  I am going to close this ticket for now, and if it crops up again, open a new ticket and we can reference this one.

pheonixstorm:
Will this resolve issue 33 as well or will that require another fix?

PR #209, opened 2016-05-19
Issue #207: Bot refuses to deploy because of NPE

dericpage:
We were already protecting against the current hex being null.  Added protection against the destination hex being null.

### DIFF
--- a/megamek/docs/history.txt
+++ b/megamek/docs/history.txt
@@ -40,6 +40,7 @@ v0.41.18-git
     popup tooltip
 + Issue #101: Princess NPE in PathEnumerator.adjustPathForBidge
 + Feature: Foot Calvary SPA
++ Issue #207: Bot refuses to deploy because of NPE
 
 
 v0.41.17 (2016-04-16 00:30 UTC)
--- a/megamek/src/megamek/client/bot/BotClient.java
+++ b/megamek/src/megamek/client/bot/BotClient.java
@@ -82,7 +82,6 @@ import megamek.common.util.StringUtil;
 public abstract class BotClient extends Client {
 
     private static final HexHasPathToCenterCache hexHasPathToCenterCache = new HexHasPathToCenterCache();
-    private static Coords boardCenter = null;
 
     // a frame, to show stuff in
     public JFrame frame;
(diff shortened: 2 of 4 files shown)


arlith:
I came across this before, but I was reluctant to make this fix because I felt that the destination hex shouldn't be null.  I'm really interested in knowing what cases can lead to a destination hex being null.  I feel like there's probably an underlying problem that needs to be fixed there.

dericpage:
Well, I don't think its valid for curHex to be NULL either.  If a unit is off-board, it's only available move (as I understand it) is to simply return to the board.  I would think there wouldn't be a need to check for a level difference in that case.  Also, Princess shouldn't be moving any off-board units anyway.

How it gets into this state is a good question because the ultimate destination hex is going to be the center of the board (as determined by IBoard.getCenter().  I don't see how that could possibly come back as null.  

However, it would seem that, somehow, ShortestPathFinder is creating paths that might go off-board somewhere along the way.  Unfortunately, I don't understand the path finder well enough to determine why.  Instead of just printing out the coordinates of the bad hex itself, maybe we should print out the entire path to get an idea of what ShortestPathFinder is thinking.

arlith:
Yea, that might  be helpful.  I don't think the ShortestPathFinder ever changes the destination hex; my understanding was that the destination comes from the Princess code.  I was basically hunting this down before and I had the same conclusion; I don't see how you could get a null value being passed in.

One thought I had had is that, `BotClient.boardCenter` is static, and it's only set once (BotClient.java line 759).  If the `BotClient` instance is used for multiple games, that is, `BotClient.hasPathToCenter` is called for two different boards, the second board, if it's far larger than the first board, could have a center that's off the board of the first one, leading to a center `Coords` that could have a null `Hex`.  But, I think when I explored his before I dunno if it could happen in practice.  Regardless, it seems like the null check is probably not totally necessary, and you could always update boardCenter with the current board when `BotClient.hasPathToCenter` is called.

dericpage:
I think I originally coded it as static to avoid doing the getCenter() calculation for dozens of meks (though looking at it, is a pretty simple operation).  I probably wasn't thinking about running multiple games without shutting down in between each.

I'll see about adding the full path to the error log in case of nulls and handling boardCenter better.

arlith:
Sounds like a plan.

PR #222, opened 2016-05-26
SPA: Shaky Stick

pheonixstorm:
See A Time of War Companion p66 for full description.

Makes air units harder to hit by ground units but does not affect targeting by ANY airborne unit


megamek/src/megamek/common/actions/WeaponAttackAction.java
```diff
@@ -1232,6 +1232,10 @@ else if (wtype.getAtClass() == WeaponType.CLASS_LBX_AC) {
 		                || game.getBoard().getHex(te.getPosition()).containsTerrain(Terrains.ROUGH))) {
 		        toHit.addModifier(+1, "urban guerrilla");
 		    }
+		    if (te.getCrew().getOptions().booleanOption("shaky_stick")
```
arlith: I don't think this check is sufficient.  According to this, grounded ASF wouldn't get the shaky_stick penalty.  There's a method somewhere, I think in Compute, that will return whether something is airborne.  This should be an airborne check, and not a unit type check.

pheonixstorm:
Yeah I copied this from the original code. Don't know how I missed that to begin with. I will try to find the airborne check. May need an if/else on if it is airborne for the target modifier. Will see what looks best.

While I am at it though, since this PR now has a conflict (and being it is super simple) would you rather I redo this PR to avoid having to fix the conflict or fix the issue above?

arlith:
Fixing the merge conflict isn't a problem.

megamek/src/megamek/common/actions/WeaponAttackAction.java
pheonixstorm: It is in entity isAirborne()
Attacks by flying units are not affected though so the Aero and VTOL checks should stay. In this case though I think a WiGE would also qualify as an airborne unit. Thoughts before I commit the change?

pheonixstorm:
If the change is satisfactory go ahead and merge.

arlith:
Your conditional seems odd, why isn't it just `&& (isAirborne() || isAirborneVTOLorWIGE())`?  That seems the easiest to read...

pheonixstorm:
The first check is to make sure the target is airborne. The second check is for the attacking entity. We don't want airborne attackers getting the +1 as only ground based attackers are subject to the +1 penalty.

arlith:
Yes, I see it now.  The MM (unwritten) style guide would have you put parens around the last two operands.

pheonixstorm:
Crap, thought I had. Usually do for that type of &&. One last fix then and it should be good to go.

PR #227, opened 2016-05-27
SPA: Golden Goose

pheonixstorm:
This gives us another reason to fear air power. Strike and Bombing get tohit bonuses and bomb scatter is decreased by two hexes.


HammerGS:
Per Arlith, merging this in.

PR #229, opened 2016-05-27
SPA: Oblique Artilleryman

### LINKED ISSUE #228: Artillery maximum distance hard-coded at 19 mapsheets
While I can see the point of such a limit it does not reflect the true distance an artillery piece or cruise missile can fire from.

Additional issue: I do not know (yet) how well the current slider would handle a max distance of 2805 hexes (2550 CM/150 + 10% from Oblique Artilleryman). Will test this and see if they box will expand to accommodate such larger numbers.

I blame 267d8141070aa3c49e478398944b722aa384dc2d
The feature request is so old I can't even pull it up on sf.net

pheonixstorm:
aToWC p 65
Need to resolve #228 to add the second half which is a 10% increase in artillery range. For now its rather moot as I doubt anyone uses artillery beyond 5 mapsheets.

The primary change is the decrease of the scatter distance by two hexes.

This PR is ready for review and merge.


megamek/src/megamek/common/weapons/ArtilleryBayWeaponIndirectFireHandler.java
```diff
@@ -335,7 +335,15 @@ else if (null != bestSpotter) {
             // we do this here to avoid duplicating handle()
             // in the ArtilleryWeaponDirectFireHandler
             Coords origPos = targetPos;
-            targetPos = Compute.scatterDirectArty(targetPos, toHit.getMoS());
+            int moF = -toHit.getMoS();
```
arlith: The original code uses `toHit.getMoS()` as the parameter passed into `scatterDirectArty`whereas the new version uses `-toHit.getMoS()`.  Was this intentional?

pheonixstorm:
Yes. In order to do the scatter bonus I had to change it to "look" correct. I found out on the bomb side that using getMoS returns a negative number. While by itself it doesn't matter, once you try to add any type of modifier you have to turn the negative positive to subtract, or add the modifier. I choose to turn the negative positive. If you would rather it be swapped I can do it though.

toHit.getMoS() + 2 vs -toHit.getMoS() - 2

pheonixstorm:
After looking at the full code I take back what I said above. Partially anyway. In the section you pointed out It doesn't have to be that way but I think it conforms with the rest of the code. But again, it can go either way. Let me know which way you want it, either for that one line of code or for all of it.

arlith:
If it doesn't matter whether it's negated or not, then it's unnecessary to negate it, so the code would be simpler (even if moderately so) if you don't negate it.

pheonixstorm:
So just change the one or both? I prefer passing a positive in both cases, or allow the negative but add a comment to the javadoc that the MoF passed by getMoS is a negative value.

arlith:
I feel that just passing the value from `ToHit.getMoS` is "simpler".  That's the way it was originally, right?  I'm not super picky in either case, really.

pheonixstorm:
Yeah, and it works fine as long you realize that MoS returns a negative for MoF. I made that mistake when I first wrote the golden goose code. I will change it and just comment everything so no one else makes the bone headed mistake I originally did.

PR #252, opened 2016-06-14
HQ Support Vehicle fuel tracking

pheonixstorm:
Awaiting review and possible change from Hammer.

Core code is ready for review. I don't *think* I missed anything.

Waiting to hear back from Hammer about a few changes on large support vees where initially the standard seats were swapped for troop space rather than as written. I changed all references to match the TRO not realizing it might have been intentional.


HammerGS:
I can't provide feedback on the code. But per our offline conversation please reinstate all references to Troopspace.  Going from memory you should be able to leave the in the references to seats in place without affecting the validation.

pheonixstorm:
Instead of using both how about adding the listed standard seating in the history tag. That way you get the troop space for use in MM and the information on the actual passengers with no need to worry about the unit being overweight.

HammerGS:
I'd wouldn't use the history tag but add it with <capabilities> tag

arlith:
Is it not possible to convert between troopspace and seats?  You could have the # of seats be represented by troopspace...

HammerGS:
I think Troopspace is a better way to go since it sort of figures in their equipment. Versus Seats.  I'd prefer to keep it as Troopspace

Akjosch:
I can't comment on the troop space vs. seats debate, but the code changes look fine to me.

In detail though, wouldn't it make sense to put the `getFuelTonnage()` and `setFuelTonnage(...)` methods in `megamek.common.Tank`? They can have ICE engines and then carry fuel with them, after all (10% of their engine's mass per SO p. 35).

pheonixstorm:
For support vehicles there is no reason to check current and max fuel amounts as they do not spend fuel points the way aero do. Also, combat vehicles do not have separate fuel amounts so have no way or reason to track fuel tonnage. It is part of the engines mass rather than a separate entry. For fuel there is no maximum value for support vehicles. The amount of fuel a support vehicle lists is based on the design and not on engine weight. To find the weight of a support vehicles fuel would require loading the unit or adding the value to the mech cache which isn't really needed.

Arlith, based on the code for troopspace.java and standardseatcargobay.java it looks like standard seating was intended to be usable as troop space but was left unfinished.

For now I will switch it back so this can be merged and closed.

Akjosch:
Putting the base method in `Tank` (then overriding it in the others) is simply a way to spare us a lot of `... instanceof ...` checks in MekHQ, where combat vehicle fuel usage, amount and costs matter.

Same for the "maximum" amount of the internal fuel tanks: Currently, MekHQ has to stitch together the full name from the chasis and model name, use that to load the MSC info, and use that to finally load a full brand-new entity just to find out this single value for the `Aero` instances.

pheonixstorm:
Will be fixing this today and creating a new branch so I can fix the seating to be usable as troopspace and show as troopspace. This way all units will match the TRO values and remove this issue entirely (I hope).

Also, will be adding getFuelTonnage() to entity and override elsewhere. There are too many unit types that use fuel such as industrialmechs, tank, aero, warships, etc. plus the fact that fuel is handled differently for the aero and have to be converted in most cases.

Any questions, issues, comments post or PM. Have a kid free day so going to try and get as much work done as possible.

pheonixstorm:
Testing this with my peacetime fuel usage and so far appears to be complete. Give it another once over and let me know if there are any other changes I should make.

PR #285, opened 2016-07-26
+ Bug: +4 Ice PSR penalty applied to Hover/WiGE instead of not Hover/WiGE

wildj79:
Commit to fix logic error in commit
b41db07b925f5ad4b3bf97def4738bc36a0126b6

### DIFF
--- a/megamek/src/megamek/common/Terrain.java
+++ b/megamek/src/megamek/common/Terrain.java
@@ -352,7 +352,7 @@ public class Terrain implements ITerrain, Serializable {
             break;
         case Terrains.ICE:
             if ((moveMode != EntityMovementMode.HOVER)
-                    || (moveMode != EntityMovementMode.WIGE)) {
+                    && (moveMode != EntityMovementMode.WIGE)) {
                 roll.addModifier(4, "Ice");
             }
             break;


arlith:
Curse you De Morgan!

wildj79:
That was fast ;)

arlith:
I was already working on MM, so I saw it, and the fix is obvious.  I should've just paid more attention in the first place.

pheonixstorm:
So why the change from || to &&? Don't get why its a logic error.

wildj79:
With an OR statement, if one of the operands is true, then the whole statement is true. As it was written, that statement would have always evaluated to true. It needs to be AND because we want to ensure that the move mode is not hover AND is not wige

arlith:
De Morgan's law also states that `!(P && Q) <=>  !P || !Q`

pheonixstorm:
So I looked it up and OMG is that whole mess stupid... It seems for this sort of OR test it should be written as | for it to work properly. Why must an OR statement be so stupidly complicated in java.

arlith:
No.  `|` is bitwise or, `||` is logical or.  It's correct in this case.  It's the same in C/C++.

pheonixstorm:
So in C/C++ || would also cause a logic error? So much for professional programmers knowing what they are doing then lol

pheonixstorm:
oh nevermind.. I'm just confusing myself even more digging into this.

arlith:
I couldn't tell if you were being serious or not.

A bitwise and is used for per-forming bit-wise operations (https://en.wikipedia.org/wiki/Bitwise_operation).  So like, `1001 | 0110 = 1111`.  Logical operations are when you're doing Boolean logic, like `false || true = true`.  Logical errors are a separate topic, basically a type of bug that arrives from faulty logic rather than syntax or something else.  Using a (logical) OR when you meant AND is an example of a logical error, but it doesn't necessarily have to involve Boolean logic.

wildj79:
@pheonixstorm You are _way_ overthinking this. If you want to understand why the `||` operator is wrong in this case, just do this simple test.

Take the statement as originally written:
```java
if ((moveMode != EntityMovementMode.HOVER)
    || (moveMode != EntityMovementMode.WIGE)) {
         roll.addModifier(4, "Ice");
}
```
Now, in your head run all of the `EntityMovementMode`'s through the statement and write down if the `if()` statement evaluates to true or fasle. 

```java
public enum EntityMovementMode {

    NONE ("none", "building"), // Future expansion. Turrets?
    BIPED ("biped"),
    TRIPOD ("tripod"),
    QUAD ("quad"),
    TRACKED ("tracked"),
    WHEELED ("wheeled"),
    HOVER ("hover"),
    VTOL ("vtol"),
    NAVAL ("naval"),
    HYDROFOIL ("hydrofoil"),
    SUBMARINE ("submarine"),
    INF_LEG ("inf_leg", "leg"),
    INF_MOTORIZED("inf_motorized", "motorized"),
    INF_JUMP ("inf_jump", "jump"),
    BIPED_SWIM,
    QUAD_SWIM,
    WIGE ("wige", "glider"),
    AERODYNE,
    SPHEROID,
    INF_UMU,
    AIRMECH,
    AEROSPACE, // this might be a synonym for AERODYNE.
    RAIL ("rail"),
    MAGLEV ("maglev");
    
    // rest of enum removed for brevity
}
```

Then, take the statement as it was re-written to:
```java
if ((moveMode != EntityMovementMode.HOVER)
    && (moveMode != EntityMovementMode.WIGE)) {
         roll.addModifier(4, "Ice");
}
```

Do the same as above. What you'll find is that the first statement will always evaluate to true.

Remember to take these two things into account while doing this:
* In a logical OR statement, if the first operand evaluates to `true`, then the second operand will not be evaluated and the whole statement will be evaluated as `true`. This is called short circuiting.
* In a logical AND statement, if the first operand evaluates to `false`, then the second operand will not be evaluated and the whole statement will be evaluated as `false`.

PR #286, opened 2016-07-26
Fixed another logic error in Terrain.java

wildj79:
After updating the fix for the +4 on ice issue, I noticed that
the logic for the Terrains.MUD modifier suffered from the same
logic error.

### DIFF
--- a/megamek/src/megamek/common/Terrain.java
+++ b/megamek/src/megamek/common/Terrain.java
@@ -315,9 +315,9 @@ public class Terrain implements ITerrain, Serializable {
             break;
         case Terrains.MUD:
             if ((moveMode != EntityMovementMode.BIPED)
-                    || (moveMode != EntityMovementMode.QUAD)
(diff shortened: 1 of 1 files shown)


arlith:
Whooo, at least this wasn't my fault.

wildj79:
Actually, I think this is still wrong.  As far as I can tell, it should be:
`if ((moveMode != EntityMovementMode.HOVER) && (moveMode != EntityMovementMode.WIGE)) {
      roll.addModifier(1, "Mud");
  }`

Reference Tactical Operations, p. 34 - 35: Expanded Movement Costs and Planetary Conditions Tables.

wildj79:
Never mind, didn't read p. 50 ;)

arlith:
Sometimes, I hate these rulebooks, you have no idea...

wildj79:
Yep, the table would seem to suggest that there should be a +1 for 'Mechs, but p. 50 explicitly states that they ignore mud.

PR #296, opened 2016-08-01
Removes <omni> tag from Clan second-line ASFs

neoancient:
I noticed that all the Clan ASFs in TRO:3067 (and one from TRO:P) are tagged as omnis, as is the fixed-configuration Eisensturm EST-R3.


HammerGS:
Looks good to me.

PR #307, opened 2016-08-10
Infantry damage for combat within buildings

arlith:
Fixes SF Bug [[#4186]](https://sourceforge.net/p/megamek/bugs/4186/).  For combat within a building, the infantry damage was off.  If a unit is outside of a building and wants to shoot at infantry in a building, they cannot.  Instead, the unit must shoot at the building, and some of the damage to the building is transfered to the infantry.

For combat within a building, the infantry can be targeted directly,  however the building can still absorb damage (although a smaller amount).  However, since the infantry themselves are targeted, the same mechanisms used in the previous case don't work.

### DIFF
--- a/megamek/docs/history.txt
+++ b/megamek/docs/history.txt
@@ -13,6 +13,7 @@ v0.41.22-git
 + Issue #303: Reporting issues with misses on units in buildings
 + Feature: Non-Infantry Weapon Damage Against Infantry changes now reported
 + Bug [#4186]: Combat within building damage for infantry now correct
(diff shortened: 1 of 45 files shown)


arlith:
As a bonus, I also added reports for reporting how much damage is done to the infantry from non-infantry attacks.  This can have significant impacts, such as direct fire weapons only doing 10% of their original damage.

arlith:
I've done some testing, and Hammer has done some testing.  Things seem in order and no one is screaming to stop, so I'm going to merge this in.

PR #318, opened 2016-08-19
run spellcheck on all the help files

ObviousTech:


arlith:
Did you update your repository before making these changes?  I noticed that there was a merge conflict with the readme, and it seems like the conflicting change was made a while ago, so it seemed like you may not have merged in new changes (unless you made these changes a while ago and are just making the PR now).

Anyways, thanks.

ObviousTech:
Ah, that explains that.  I fetched the updates, but did not merge them.  I will remember to do that next time.

PR #331, opened 2016-09-10
Rat generator

neoancient:
Generates random assignment table for any faction appropriate to any year from 2398-3150.

Features:

- In addition to major factions, provides distinct character to a number of very minor factions in the Periphery based on nearby major powers (e.g. the Illyrian Palatinate and the Lothian League have a higher percentage of units that are characteristically Marik and the Tortuga Dominions have both a Davion and a pirate character.
- For major powers (including larger periphery states), the user can select a particular command within that faction's military (such as Sword of Light or Davion Brigade of Guards). Where the fluff indicates that a particular unit is especially popular with or exclusive to a particular command (or rare or absent), it will be reflected in the generated table.
- All unit types are included except space stations and gun turrets (for now).
- The data includes all units from all the TRO series, including XTROs, with the exception of TRO:Vehicle Annex. There are also units from some other sources - mostly for early periods, where the offerings are more sparse. Units that are unique or never entered service (such as most prototypes) are excluded, as are LAMs, trailers, and units not yet in MM.
- Multiple weight classes can be chosen when appropriate for the unit type. The distribution of various types are based on Xotl's work and expanded to include other factions.
- Tables can be generated for a particular type of mission. Depending on the mission type, this can either increase frequency of certain units, decrease frequency, exclude some units from the table, or include units that would not normally appear for a general combat role. There is an adjustment that can vary the amount of change in frequency (up or down).
- Vehicles and conventional infantry can be filtered by motive type. VTOLs can be mixed with tanks to form a traditional all-vehicle table by selecting VTOL as one of the motive types.
- Where appropriate to the unit type, the table can be restricted to a certain type of C3 equipment.
- When the salvage/isorla mechanism selects an omni unit, the configuration is recalculated to reflect the preferences of the faction operating the equipment.


HammerGS:
I've been testing this for the last couple of days and haven't had any issues with it.

arlith:
I haven't had a chance to run this yet, but I glanced over the changes.  Do you have any plans for documentation?  I see there's a bunch of XML files added for different years, each of which has a lot of information in it.  Is there going to be a way to create/edit these files?  Any documentation on what the file format means?

Also, this means we'll be able to remove the neo rats from MM, yea?

neoancient:
Documentation added. Based on discussions with Hammer about maintenance I plan to clean up the graphic utility I use to edit the data and add it to the MM repository. WIth this the Neoancient RATs will be deprecated, but I need to finish adapting HQ to use this system before they can be removed.

PR #335, opened 2016-09-15
SPA: RangeMaster

pheonixstorm:
Cam Ops p80

RangeMaster swaps the given range modifier with the short range modifier.

I have one item concerning this before it can be merged. In the pic below it shows the +0 swapped for extreme range. The issue is should it continue to display as below or should I instead change the text to say +0 (RangeMaster [Extreme]) to be slightly more informative?
rangemaster

The other item is a question I posted about how Sniper and RangeMaster SPAs should work together. I am failry certain the way I have coded it should be correctly so if the answer comes back that I am wrong i will correct the issue then.
http://bg.battletech.com/forums/index.php?topic=54620.0

Last item, I have fixed the sniper SPA to no longer cover LOS range. I don't recall any mention of LOS range being applicable in aToW but it IS covered in Campaign Operations and is excluded from the modified modifiers.


wildj79:
Based on how the abilities are worded in AToW, I would say you have it coded correctly.  Be aware, though, according to AToW p.221, Sniper is a prerequisite to getting Range Master.  Not sure if that is the case in Campaign Operations, though.

 _A Time of War, p. 221_
> *Range Master*
>
> *Prerequisites:* [Abilities] Sniper SPA
> *XP Cost:* 100
>
> The Range Master Ability grants the warrior mastery over
> any range band except Short (Minimum, Medium, Long and so
> forth). Any weapon attacks made in the selected range band
> may swap range modifi ers with the Short range band (if the
> range band selected is Minimum, the Short Range modifi er
> becomes –4 instead).
>
> For example, a Range Master may select the Long range
> band as his area of mastery (a –2 roll modifier, thanks to
> the Sniper SPA), and would receive a +0 modifier at that
> range, but at the cost of a –2 modifier for attack rolls made
> at Short Range.

_Remember to flip the signs when converting AToW modifiers to BT_

pheonixstorm:
No, Campaign Operations follows Alpha Strike for how the SPAs are handled. Since TW does not use skills SPAs in CO are limited by the amount of points used when acquiring SPAs. Regular pilots are limited to 2 point SPAs, etc. All SPAs will move away from aToW prereqs and will be entirely based on CO with the exception of how much XP they cost. That will be determined at 10% aToW costs or AtB costs. I don't see any other way to gain SPAs other than by XP.

With Campaign Operations containing all of the SPAs there is no reason to use aToW since the CO SPAs are based on Total Warfare rules.

pheonixstorm:
as you can see below minimum range is not covered nor is the example from aToW, reason for my forum question. Should probably add minimum range to that question.

> RANGE MASTER
“Too close… Too far… Ah! Just right! Eat nickel, Clanner!”
Unit Type: Any
SPA Cost: 2 points
The Range Master Ability grants the warrior mastery over any
range band except Short (Medium, Long and so forth). Any weapon
attacks made in the selected range band may swap range modifiers
between that range band and the Short range band.
For example, a Range Master may select the Long range band
as their area of mastery (normally a +4 To-Hit Modifier), and would
receive a +0 modifier at that range, but would now suffer a +4 To-
Hit Modifier for attack rolls made at Short Range.

PR #364, opened 2016-10-13
Implementing Black Ice

pheonixstorm:
Added errata change to ice and black ice
Added black ice terrain type

Mostly working though it checks every hex every time you change the movement. Not sure how to stop this behavior.

Also, since this will show up in the map editor how should I handle this?

I think this will need a report entry but I'm falling asleep and half functional. Please check if this is the case or I am just missing something in the code.

There may be other minor errors but I haven't noticed anything.

PR for review and enhancement only. Will be rewritten and reposted after review completed.

### DIFF
--- a/megamek/src/megamek/common/Compute.java
+++ b/megamek/src/megamek/common/Compute.java
@@ -490,6 +490,11 @@ public class Compute {
             return true;
         }
 
+        // check for black ice
+        if (destHex.containsTerrain(Terrains.BLACK_ICE)) {
+            return true;
+        }
+
         // Check for water unless we're a hovercraft or naval or using a bridge
         // or flying.
         if ((movementType != EntityMovementType.MOVE_JUMP)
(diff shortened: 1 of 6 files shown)


pheonixstorm:
Thought of something on how to keep new black ice from forming every time you change move orders during the movement phase.

Line 2570 sets the black ice for the hex. TO says on a 5 or 6 there is black ice, so I figure the if statement could be changed to use a level value for black ice. 0 means no black ice and 1 would be black ice present. This keeps MM from checking the hex again. But is it the proper way to do this??

Also, additionally at the end phase the black ice values for no black ice can be reset so that during the next movement phase the same hex can be checked to see if black ice has formed. 

Is there a proper way to handle this or should I use my idea?

beerockxs:
MoveStep is the wrong place to add terrain, as steps get constructed whenever you plan a new path.
It should be created in Server#processMovement, IMO

pheonixstorm:
when using processMovement would the affects of the black ice be immediate such as skidding?

beerockxs:
Skidding is resolved in processMovement as well, so I think the answer is yes.

pheonixstorm:
Thanks bee

PR #371, opened 2016-10-17
Turret RATs and intro date changes

BLOODWOLF333:
Updated the turret intro dates to the SO common year for the weapon(s).
Created turret RAT files.


BLOODWOLF333:
why did it not put any of my changes in....

PR #396, opened 2016-11-06
TRO 3039 Vehicle source update

pheonixstorm:
All vehicles from the 3039 folder are now updated to show source and era

This should complete TRO 3039


HammerGS:
Thanks.

Issue #401, opened 2016-11-09
Princess deploys multiple units in/on top of one building


kipstafoo:
Thanks for the assist, princess.

As you can see from the screen shot, princess put multiple units in/on one building at deployment and collapsed the building on top of itself.

![2016-11-09](https://cloud.githubusercontent.com/assets/506522/20138652/7c441b78-a64f-11e6-87f8-b3c917603f9c.png)

VampireSeraphin:
I've seen that happen mid game before to.

dericpage:
Change has been merged.  This should be fixed with the next release.

PR #406, opened 2016-11-10
Issue #401: Princess deploys multiple units in/on top of one building

dericpage:
Code to avoid overloading buildings with multiple entities during deployment and movement.


arlith:
Could be nice to split out some of the code formatting changes into their own commits, for the future.  Otherwise, I'm fine with this.

dericpage:
Are you referring to the change of accessibility of the methods?  I'm not seeing much in the way of format changes.

Issue #175, opened 2016-05-02
Foot platoons can move infinite hexes while in woods per turn


BLOODWOLF333:
As long as my foot platoon stays in woods hexes I can move him as many hexes as possible in one turn.

![ingame](https://cloud.githubusercontent.com/assets/16950832/14945685/cf7ac1da-0fdb-11e6-8014-90223d05259c.png)

[FootPlatoonBug.sav.gz](https://github.com/MegaMek/megamek/files/244494/FootPlatoonBug.sav.gz)

pheonixstorm:
Using the latest MM release??

BLOODWOLF333:
Yes this is the 0.3.27 package.

scJazz2:
From convo on MC Forums...
Units ignore the movement penalty from trees if there is pavement as the base layer, the infantry has to be some weird interaction with that. Infantry platoons can also always move at least one hex.

So, I think, what is happening here, is that the foot infantry have an MP of 1, but they get the bonus MP needed to enter a woods hex. But since woods hexes on pavement only cost the underlying amount, in this case, 1, the infantry only ever expends their bonus MP to enter the wooded hexes, causing the cost of entering a woods hex to be effectively 'free' but also never expending their walk MP. Since you have not expended your walk MP, MM thinks you should still be able to move.

BLOODWOLF333:
Player name: Madras as seen in the screenshot.

pheonixstorm:
Should have a fix for this in my next PR. Maybe this weekend.

pheonixstorm:
@arlith 
Still digging around on this one. It seems to have been around for awhile though. I went as far back as 0.35.34 and it was present. I'm shocked this hasn't been reported sooner.

pheonixstorm:
This is being a real bugger to fix properly. Just adding not this or that terrain does nothing but adding not this and not this fixes the issue but makes movement from clear terrain to wooded road or pavement cost 2 mp rather than 1. Will continue to bash my head against it until something breaks.

pheonixstorm:
Fixed.

BLOODWOLF333:
Way to go Phoenix \m/.

arlith:
Are you going to make a PR?

pheonixstorm:
Yes, I had three other bugs to fix before posting.

arlith:
Fixed with PR #183.

VampireSeraphin:
[...]

PR #414, opened 2016-11-15
Refactored hex,added validation checks. Issue #175

Kurios:
Currently working on map validation fix for Issue #175, to prevent maps with terrain like woods:0 from becoming an issue.

### DIFF
--- a/megamek/src/megamek/common/Board.java
+++ b/megamek/src/megamek/common/Board.java
@@ -897,15 +897,31 @@ public class Board implements Serializable, IBoard {
         }
 
         // check data integrity
-        if ((nw > 1) || (nh > 1) || (di == (nw * nh))) {
+        if (isValid(nd) && ((nw > 1) || (nh > 1) || (di == (nw * nh)))) {
             newData(nw, nh, nd);
-        } else {
+        }else{
             System.err.println("board data invalid");
         }
 
     }
 
-    private int indexFor(String hexNum, int width, int row) {
+    public boolean isValid() {
+		//Search for black-listed hexes
+    	for(IHex hex: data) {
+    		if(!hex.isValid()) return false;	
+    	}
+		return true;
+	}
+    
+    private boolean isValid(IHex[] data) {
+		//Search for black-listed hexes
+    	for(IHex hex: data) {
+    		if(!hex.isValid()) return false;	
+    	}
+		return true;
+	}
+
+	private int indexFor(String hexNum, int width, int row) {
         int substringDiff = 2;
         if (row > 99) {
             substringDiff = Integer.toString(width).length();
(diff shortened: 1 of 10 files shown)


Kurios:
Pull request can be merged now. 

WIll validate maps against a small set of rules, and patch the woods:0 problem now. If a random map would violate rules, it picks another map.

Akjosch:
Small inconsistency: Why is woods "fixing" its terrain level, but nothing else does?

While talking implementing limits and auto-fixes, roads should be limited to values of 1 and 2 (with 1 being the default when outside of those values), with a value of 2 a legacy issue and discouraged.

Kurios:
want me to force all roads to 1 then?

And I only changed woods because that's the current known problem.

Akjosch:
Roads should be 1 or 2 for now. Limiting them to just 1 would break too many things at the moment.

What's the current known problem with the woods exactly? Because if it is something in the provided maps or the terrain generator, the solution is to fix *them*.

Kurios:
some of the the current maps have woods:0, which is allowing infantry free movement through them. ( #175 ). Starting a primitive map sanity check was one of the solutions that came out of yesterday's discussion, in order to prevent these breaking maps from hitting. Then this afternoon, it was requested that we don't reject the wood:0, but patch it in place, which was why that was done differently.

Akjosch:
Provided maps are fixed.

beerockxs:
Please format your code to use 4 spaces for indent and to have a space before opening braces.

Kurios:
@beerockxs: Whomever is in charge of formatting should come up with a lint or some sort of formatting tool (like the eclipse code formatter) so the formatting reqs are in code (And if possible, automated)

megamek/src/megamek/common/Board.java
```diff
@@ -897,15 +897,32 @@ public void load(InputStream is) {
...
+	}
+    
+    private boolean isValid(IHex[] data) {
+		//Search for black-listed hexes
+    	for(IHex hex: data) {
+    		if(!hex.isValid()) return false;	
+    	}
+		return true;
+	}
+
```
arlith: It seems like there's unnecessary code duplication here.  Why does `isValid(0)` not just call `isValid(1)`?  Why does `isValid(1)` not check for a null hex but `isValid(0)` does?

arlith:
Another thought that occurred to me... There are some terrain types that require a combination of terrains.  For example, buildings.  There are at least three `Terrain`'s that are required: building, bldg_cf, bldg_elev.  There may be another I'm forgetting.  But, we should validate that in these cases, all of the required `Terrains` are defined.

Kurios:
Arlith: Mind if I put this on back-burner, and start a major formatting only change to make this all cleaner?

arlith:
I think this is pretty close to merging in, and just needs a few small changes and I'll merge it in.  In addition to the comment I made, it would also be nice to be able to get a report of all the boards currently in MM.  Some type of app like in megamek.test (say, MechSetTest).  I could probably write this app myself though.  If you make the one change I commented on, I'll probably merge this in.

Issue #418, opened 2016-11-17
Various visual anomalies with flight path sprites


arlith:
There are a few graphical issues with `FlyOverSprite`.

Firstly, there's a small "spur" on bends in the flightpath.  For example:

![before](https://cloud.githubusercontent.com/assets/13815237/20404455/adf5a516-acca-11e6-8834-1d7bd8f79ca3.png)

Also, with isometric on, the flight path can draw over the top of the entity, and also bends with elevation:

![iso_issues](https://cloud.githubusercontent.com/assets/13815237/20404477/bc4f92fc-acca-11e6-9b01-5213b6c7aae2.png)

arlith:
Another thing I forgot is that, every time `FlyOverSprite.getBounds` is called, it recomputes the flightpath polygon.  This polgyon is the whole flightpath geometric shape.  So this means that it has to trace the whole flightpath and recompute the poly.  `Sprite.getBounds` is called by `BoardView1.drawSprites`, which means that everytime `BoardView1.paintComponent` is called, all of the `FlyOverSprites` completely recompute their shape.  This is incredibly inefficient.

arlith:
Fixed with the branch merge in 05e5ff.

PR #419, opened 2016-11-17
Fix for #418: Graphical anomalis  with FlyOverSprite

arlith:
I changed how FlyOverSprite computes its polygon.  It now has slightly rounded bends without the little spurs.  I also made it so the polygon is only computed once, not every time the bounds are retrieved.  I also made it so the path ignores elevation in isometric mode.  Additionally, the path ends at the edge of the last hex in isometric mode, to prevent issues with the path drawing over the icon.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/boardview/BoardView1.java
+++ b/megamek/src/megamek/client/ui/swing/boardview/BoardView1.java
@@ -3192,7 +3192,7 @@ public class BoardView1 extends JPanel implements IBoardView, Scrollable,
         return p;
     }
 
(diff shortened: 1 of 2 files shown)


arlith:
Here are some before and after images.

Before "spurs":
before

After "rounded corners":
after

Before iso elevation and overlap:
iso_issues

Fixed iso issues:
fixediso

I didn't run any performance benchmarking, but the new code is guaranteed to be faster by a wide margin.

arlith:
No one is screaming at me to stop, so I'm merging this.

Issue #428, opened 2016-11-28
Improper movement in snow?


VampireSeraphin:
I'm not sure if this is working as designed, or a bug.

Tanks in heavy/moderate snow weather that have their MP reduced to 0 by snow are not granted 1st move amnesty. I talked with Hammer about this and it might be a bug.

hammer [2:06 PM]  
unit can always move into the hex directly in
front of it, regardless of the terrain cost, provided such movement
is the only MP expenditure the unit makes in that turn
(i.e. the unit cannot make any facing changes in the same turn
it uses the Minimum Movement rule). The controlling player
can only make this kind of move if the unit has at least 1 MP to
spend (meaning it is mobile) and is not prohibited from entering
that terrain (see Level Change, p. 48); all units reduced
to 1 MP either due to heat or damage can still make use of
the minimum movement rule.

[2:07]  
TW pg 49

To me that sounds like the snow should be increasing the MP cost of the hex, but MM appears to deduct the snow penalty directly from the unit, rendering it unable to move, or even turn. I've attached a save game where I had some mobile long toms reduced to 0mps by heavy snow.

The player name is The Infamous Dust, and this save is before I've moved any units. 

[SnowBlocksMovement.sav.gz](https://github.com/MegaMek/megamek/files/615346/SnowBlocksMovement.sav.gz)

pheonixstorm:
Thin snow adds a cost of +1 per hex for wheeled units while deep snow adds a cost of +1 per hex for all units. Temperatures of below -30 affect the units move. A temp of -40 will subtract 1 from the cruising speed of all vehicles. So a Vedette will go from 5/8 to a 4/6.

In the game you provided there is no snow yet. For heavy snowfall it takes 5 turns to build up to thin snow and an additional 10 turns (15 total) for deep snow.

p60 TO says that heavy snowfall includes the effects of -50 temps which would only remove 2 from the vehicle MP. The code as currently written is calculating the heavy snowfall penalty at -3 which is incorrect on two fronts. First, heavy snowfall itself isn't where the penalty is coming from, rather it should be calculated based on the temperature of -50. The other is reductions due to extreme winds such as storm and tornados.

As for the minimum movement rule... that will need to be asked on the rules forum.

pheonixstorm:
@HammerGS this can be closed

PR #431, opened 2016-12-03
Planetary conditions part 1 fixes sf.net ticket 4124

pheonixstorm:
This fixes several issues people were having with the weather, such as not knowing it had to be set manually. There are several other fixes as well

Visual Spotting: isAero was counting any aero as having the spotting modifiers for aero regardless of where they were. p221 Shows that only aero  at altitude 2-9 should have said modifiers while those below 2 are considered grounded and use the mech column and those at altitude 10 cannot visually spot.

Snow related movement: With the new fix in place this section was removed entirely as it was not being calculated correctly and was causing far higher penalties if the user set the temperature as well as a snow condition. All movement penalties are now calculated correctly based on wind and extreme temperatures.

Blowing Sand/Sleet: Mostly worked before except when shifting winds were in play. As written the affects of both should be lost if wind speed drops below moderate gale. This now works.

Conventional Infantry in Extreme Temps: The now die if not inside a building or transport

Part 2 Will include Hail, Black Ice, and possibly Lightning Storms but will not be done before stable.

This fixes sf.net ticket 4124 ice storm problems


arlith:
Does #428 also get fixed with this PR?

pheonixstorm:
No, needs a ruling from CGL to see if it qualifies for minimum movement since cold weather reduces the amount of movement a vehicle has

> For vehicles, for each 10 degrees C (or fraction thereof) higher than 50 degrees, reduce their Cruising speed by 1 Movement Point. For every 10 degrees C (or fraction thereof) less than –30 degrees, reduce their Cruising speed by 1 Movement Point. Recalculate Flanking speed based on the new Cruising speed.

I want to say that this would be the same as if a vehicle took multiple motive hits that reduced cruise speed to 0 in which case they don't get minimum movement but it really needs a clarification. If VampireSeraphin hasn't posted a question yet I will as I also want to know about infantry regarding this line of text:

> unless they are XCT troops (see p. 351), in which case they ignore such movement restrictions.

pheonixstorm:
I take back what I said. 

After rereading it I think the issue is that VS is thinking the cause of the move penalty is the Snow Terrain in which case there is no bug, just a misunderstanding of how the weather works.

Overall movement penalties work as they are supposed to, and heat bonus/penalty for extreme temperatures also works now. Issue 428 can be closed.

HammerGS:
Per a conversation with Arlith we are going to merge this.  It can receive additional testing this Dev cycle.

Issue #421, opened 2016-11-21
Toughness Text In Pilot Tab Spacing Issue


BLOODWOLF333:
Under the pilot tab in MM the toughness value looks like its smushed up against the last "s" of the word toughness.  So, if it was a value of 1 it appears like the s is written over the 1. Windows 7-8.1

![slack](https://cloud.githubusercontent.com/assets/16950832/20492380/2053414c-afda-11e6-949e-fb1f5b8956a4.png)

BLOODWOLF333:
````        toughBL = createLabel(Messages.getString("PilotMapSet.toughBL"), fm, pilotL.getSize().width + 75, getYCoord()); //$NON-NLS-1$
        content.addArea(toughBL);
        toughBR = createLabel(STAR3, fm, pilotL.getSize().width + 75 + initBL.getSize().width + 25, getYCoord());
        content.addArea(toughBR);````
I dont know how to submit this yet but it works.

BLOODWOLF333:
![image](https://cloud.githubusercontent.com/assets/16950832/21990885/11506f8a-dbd6-11e6-92d9-1a8636ec060e.png)

arlith:
There were some issues with PR #477, particularly the file had its contents doubled.  I was able to figure out what the code changes were and get them applied.

PR #477, opened 2017-01-19
#421 Spacing of toughness value in pilot tab

BLOODWOLF333:
Previously the value of a pilot's toughness (ex 1) would draw in the same space as the : after the toughness label.  The toughness label is wider than the 15 pixels allotted for it causing this issue.  So changed the spacing to 25 pixels after the start of the RH labels as the point when the values will display. #421 

github

I dont know how to remove those remote tracking Commits sorry.


BLOODWOLF333:
Resolved conflict, was missing a } I guess compared to the new master?

arlith:
There are some issues with this PR, and the underlying bug #421  is fixed.

PR #485, opened 2017-02-17
Created isXCT function for checking doomedInExtremeTemp and getMoveme…

pheonixstorm:
…ntMods

isXCT checks for specific temperature related armor kits.
TacOps p350 lists armor kits covered for cold and hot conditions. It defaults to the old method if no armor kits are listed.

### DIFF
--- a/megamek/src/megamek/common/Infantry.java
+++ b/megamek/src/megamek/common/Infantry.java
@@ -1102,6 +1102,16 @@ public class Infantry extends Entity {
 
     @Override
     public boolean doomedInExtremeTemp() {
+        Optional<Mounted> armor = getEquipment().stream()
+                .filter(m -> m.getType().hasFlag(MiscType.F_ARMOR_KIT))
+                .findFirst();
+        if (armor.isPresent()) {
+            if (isXCT() || isMechanized()) {
+                return false;
+            } else {
+                return true;
+            }
+        }
         if (hasSpaceSuit() || isMechanized()) {
             return false;
         }
@@ -1273,6 +1283,37 @@ public class Infantry extends Entity {
                 (getMovementMode() == EntityMovementMode.TRACKED) ||
                 (getMovementMode() == EntityMovementMode.SUBMARINE) ||
                 (getMovementMode() == EntityMovementMode.VTOL);
+    }
+
+    public boolean isXCT() {
+        Optional<Mounted> armor = getEquipment().stream()
+                .filter(m -> m.getType().hasFlag(MiscType.F_ARMOR_KIT))
+                .findFirst();
+        if (armor.isPresent()) {
+            if (armor.get().getType().getName() == "Environment Suit, Hostile"
+                || armor.get().getType().getName() == "Environment Suit, Hostile(Clan)"
+                || armor.get().getType().getName() == "Environment Suit, Marine"
+                || armor.get().getType().getName() == "Environment Suit, Marine(Clan)"
+                || armor.get().getType().getName() == "Spacesuit"
+                || armor.get().getType().getName() == "Spacesuit, Combat") {
+                return true;
+            } else if ((game.getPlanetaryConditions().getTemperature() < -29)
+                    && (armor.get().getType().getName() == "Heat Suit"
+                        || armor.get().getType().getName() == "Heat Suit (Clan)"
+                        || armor.get().getType().getName() == "Snow suit")) {
+                return true;
+            } else if ((game.getPlanetaryConditions().getTemperature() > 49)
+                    && (armor.get().getType().getName() == "MechWarrior Cooling Suit"
+                        || armor.get().getType().getName() == "MechWarrior Cooling Suit (Clan)")) {
+                return true;
+            // TODO Add section here to check for Tainted/Toxic atmosphere equipment once
+            // the conditions are added to PlanetaryConditions.java
+            } else {
+                return false;
+            }
+        } else {
+            return spaceSuit;
+        }
     }
 
     /*
--- a/megamek/src/megamek/common/PlanetaryConditions.java
+++ b/megamek/src/megamek/common/PlanetaryConditions.java
@@ -585,7 +585,7 @@ public class PlanetaryConditions implements Serializable {
         }
 
         // temperature difference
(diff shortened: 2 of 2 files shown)


neoancient:
The issues have have with this PR fall into three areas:
1. Rules:
XCT benefits are the result of a combination of training and equipment. Simply having armor that applies to a certain environment does not automatically provide an XCT benefit. See TacOps, p. 351 and #8 here: http://bg.battletech.com/forums/index.php?topic=286.msg27974#msg27974
Spacesuits confer the ability to survive in a vacuum without special training per the armor kit rules.

2. Implementation:
You check for type of armor by comparing the display name. This is not guaranteed to remain the same. A better approach, and one that will not require updating if other armor kits are added in the future, is to add flags to the `subType` field. See `MiscType`, lines 505-511:
`    // Secondary flags for infantry armor kits
    public static final long S_DEST = 1L << 0;
    public static final long S_SNEAK_CAMO = 1L << 1;
    public static final long S_SNEAK_IR = 1L << 2;
    public static final long S_SNEAK_ECM = 1L << 3;
    public static final long S_ENCUMBERING = 1L << 4;
    public static final long S_SPACE_SUIT = 1L << 5;
`
You could add flags such as `S_HEAT_SUIT`, `S_COLD_SUIT`, etc then add them to the individual suit type initializer.

3. Code quality:
3a. Compare Strings using `equals()` rather than `==`, unless for some reason you need to know that they are stored in the same memory location. This works because of the way the JVM handles String literals, but this approach relies on the internal implementation rather than the public interface of the class.
3b. Instead of copying the code in Infantry::getArmorKit to search the mounted equipment for an armor kit, just call Infantry::getArmorKit. Besides avoiding unnecessary code, it makes the code much more readable.

pheonixstorm:
1. Yes I know that, but MM doesn't deall with the end training and this is only specific to weather conditions in which case the suits listed are valid and why I did not touch doomedInVacuum.

2.  That's fine, I can make that change as this touches so few suits overall

3. I'm a C++ coder so never really used equals() before.

HammerGS:
Couldn't you set the XTC training as a SPA like SCUBA?

pheonixstorm:
SCUBA is a motive type though while XCT isn't so it uses the movement dropdown in MML

XCT as defined by this code is purely an armor kit issue unlike firefighting infantry which get extra equipment or Anti-Mek infantry. XCT while a training specialization does not in itself give out any special equipment like other special infantry do. 

Which is the biggest reason I only focused on XCT related armor kits for hot and cold weather and left the vacuum suits alone since it would be used by say Demo Infantry or any other Combat Engineer infantry.

neoancient:
It just needs a field such as `boolean xctTrained`. The only reason I hadn't already added it was that TacOps isn't really clear whether the different environments amount to separate training. According to the link above, an infantry unit either has XCT training or it doesn't, and the environments it can work in depend on the type of armor kit.

Since Java does not have overloaded operators, comparing any two Objects for equivalent content requires the `equals` method. The == operator compares memory locations.

HammerGS:
Re training. They appear to need training in accordance with the specific enviroments. TO pg 350 (at the bottom)

```
Xenoplanetary Condition-Trained Troops: To create a conventional infantry unit as Xenoplanetary Condition-Trained (XCT) Troops, the unit must select one of the appropriate Infantry Armor types (see Infantry Armor, p. 317) for the environment. XCT Troops may be created for Extreme Temperatures, Hazardous Liquids, Vacuum, Trace Atmospheres, and Toxic or Tainted environments.

Extreme Temperatures (Cold): XCT Troops equipped for extreme cold must use Snowsuit, Heatsuit, Environment Suit (Hostile or Combat), or Spacesuit (any) armor types.

Extreme Temperatures (Hot): XCT Troops equipped for extreme heat must use Cooling Suit (any type), Environment Suit (Hostile or Combat), or Spacesuit (any) armor types.

Hazardous Liquid Environments: XCT Troops equipped for hazardous liquids must use Environment Suit (Any), or Engineering Suit armor types.

Vacuum/Trace Atmospheres: XCT Troops equipped for vacuum or trace atmosphere levels must use Environment Suit (Hostile or Marine), or Spacesuit (any) armor types.

Toxic/Tainted Atmospheres: XCT Troops equipped for toxic or tainted atmosphere conditions must use Environment Suit (Hostile or Marine), or Spacesuit (any) armor types. If the atmosphere is Tainted, but not Toxic, Light Environment Suits may be used instead.

XCT Troops may not be designed as Beast-Mounted or Mechanized (VTOL) infantry if they are to operate in vacuum or trace atmospheres.

Additional construction rules for the required armor types may be found under Infantry Armor (see p. 317).

```

neoancient:
That's what led me to hold off, but see the WoB Mechanized Assault XCT unit (TRO3085, p. 206): "Notes: Xenoplanetary (All Environments) Condition-Trained." That leads me to think that the need to select environment involves making sure that they have the right equipment for that environment, but they are not limited to operating in that environment. The post in the rules clarification thread says as much. There is also only one cost associated with XCT training (I think that might be in the TacOps errata). If each environment is distinct, then units training in multiple environments should cost more.

HammerGS:
For Reference to have in one place.  Errata from XCT Troops.

```
Xenoplanetary Condition-Trained Troops (p. 350)
Under “Construction Rules”
For both “Extreme Temperatures (Cold)” and “Extreme Temperatures (Hot)”, change Environment Suit (Hostile or Combat) to Environment Suit (Hostile or Marine).
```

pheonixstorm:
So far as I can tell XCT troops don't cost extra. Have either of you found a reference to construction costs in TO or IO covering this? I only found a BV modifier but nothing else.

pheonixstorm:
As training is REALLY unimportant to MM it should not be a factor in how XCT troops are created. MM is not A Time of War after all.

If we create a specialization for XCT then the type of XCT troops should be solely based on the type of armor when checking for vacuum, hot, or cold weather conditions and not multiple specialization check boxes. That is just unnecessary clutter imo

pheonixstorm:
Closing this one out for now. Recreating the branch from scratch so will have a newer PR up sometime soon.

PR #493, opened 2017-02-26
Infantry weather update

pheonixstorm:
XCT infantry now have a specialization flag as well as related armor kit flags. BV and Cost multipliers have also been added.

One thing of note. Snowsuit is listed in MiscType as Snow suit which is incorrect based on how it is written in IO and TO. Should I make that change before this is merged or should neo?

### DIFF
--- a/megamek/i18n/megamek/common/messages.properties
+++ b/megamek/i18n/megamek/common/messages.properties
@@ -195,8 +195,8 @@ Infantry.specialization7=Mountain Troops
 Infantry.specialization8=Paramedics
 Infantry.specialization9=Paratroops
 Infantry.specialization10=TAG Troops
-Infantry.specialization11=SCUBA
-Infantry.specialization12=Xenoplanetary Condition-Trained Troops
+Infantry.specialization11=Xenoplanetary Condition-Trained Troops
+Infantry.specialization12=SCUBA
 
 Infantry.specializationTip0=Bridge-building Engineers (Unimplemented)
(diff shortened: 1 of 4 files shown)


HammerGS:
Asked Xotl for clarification on the XCT infantry training.  XCT training is per enviroment.

image

neoancient:
So what about the WoB Mechanized Assault XTC, which is "Xenoplanetary (All Environments) Condition-Trained" per TRO:3085, p. 207?

HammerGS:
I've requested addition clarifications as they related to multiple specializations, cost and BV.

pheonixstorm:
This is now complete. Just need to post the MML PR to add XCT as an option in MML

HammerGS:
This is still in the rules committee. But at this point its leaning this way:

1) XCT infantry can be trained for multiple environments.  BV will add an additional .1 for each environment trained, cost becomes a x5 regardless of the number of environment

2) In MegaMek they can only we equipped to fight in one environment that they are trained in, and that has to be declared prior to the start of the game (i.e. in the lounge)

3) Their will be some rules about what equipment protects against what environment. For example a Spacesuit might protect from all environments.  That is still being debated.

pheonixstorm:
Yeah,  3 was the most specific in the book. X suit protects against X environments. The Combat/Marine Environmental suit was the only suit to protect against all. Not sure on the space suit, would have to look over the rules again to be sure.

HammerGS:
ok, next time you're on Slack PM me about this one.

pheonixstorm:
Updated PR posted, closing this one.

PR #497, opened 2017-03-01
Check length of history, not deployment, when adding history

Alarantalara:


pheonixstorm:
@Alarantalara 
I don't see this getting accepted without any kind of description or reason behind the change

beerockxs:
It's a bugfix for an obvious bug and OK to be merged.

pheonixstorm:
True, but bad habit to start even if it's a no brainer.

PR #506, opened 2017-03-10
TRO 3050 Vehicle source update

pheonixstorm:
Updated source information for 3050 vehicles.
Fixed issue with armor on one VTOL.
Moved several vehicles from 2750 folder to 3050 folder to match the MUL


HammerGS:
Merging in.

PR #514, opened 2017-03-26
Fixes #513 Infantry killed by Inferno ammo not giving kill credit

### LINKED ISSUE #513: Infantry killed by Inferno ammo not giving kill credit
When using SRM Inferno rounds the infantry is not giving kill credit to the attacking unit. It is instead showing as destroyed by pilot error.
`Survivors are:
Javelin JVN-10N (test)
Pilot : Nathan Wentz [1/2]
Kills : 1

Graveyard contains:
Foot Platoon (Rifle) (Princess)
Gunnery Skill : Rita Kanavos [4]
Kills : 0
Destroyed by pilot error.

Foot Platoon (Rifle) #2 (Princess)
Gunnery Skill : Selma Miskarov [4] ( 6 hit(s)   )
Kills : 0
Destroyed by Javelin JVN-10N (test)
`

The second infantry platoon listed was killed by standard SRM rounds. Additionally I tested this with frag rounds and acid warheads, both of which showed as having the correct destroyed by assignment

No errors in the log to report. Attached save game prior to victory screen. Used 0.42.1 stable
[autosave.sav.gz](https://github.com/MegaMek/megamek/files/857251/autosave.sav.gz)

pheonixstorm:
Took longer than expected to find, but I believe it is now fixed.

### DIFF
--- a/megamek/src/megamek/server/Server.java
+++ b/megamek/src/megamek/server/Server.java
@@ -10291,6 +10291,7 @@ public class Server implements Runnable {
                         vPhaseReport.add(r);
                     } else {
                         vPhaseReport.addAll(destroyEntity(te, "damage", false));
+                        creditKill(te, ae);
                         Report.addNewline(vPhaseReport);
                     }
                 } else {


arlith:
This really ignores the fact that there are still a ton of other cases where kills aren't credited correctly, but it fixes this one specific issue.

pheonixstorm:
I really only know about secondary kills such as those from Ammo and Engine explosions. Did those ever get cleared up on who should get the kill? I know you and the other devs had a discussion going for awhile about it.

Is there a list of other kill credits that need work? I'm sure I can find any open tickets for them but not sure if there were some that don't have open bug reports.

neoancient:
I think there's a ticket somewhere about getting a kill credit when a 'Mech is destroyed by a fall caused by an attack.

arlith:
https://sourceforge.net/p/megamek/bugs/4360/

arlith:
Another one would be ejecting.  If a unit takes enough damage that the pilot decides to bail out, I think that the kill should be awarded to a damaging unit from the previous round.

neilfw:
Do you background track damage caused by particular opponents?... 

Not from a rulebook... but a manual method I used to use for attrition destruction was 1pt for an armour hit 2 for internal structure 5 for a successful critical.  So when there is no clear "blew the head off", or clear that a singular unit wore an enemy down, you had an idea of where the damage came from.  Of course many destructions had that near 50-50 split, so you get in to the pain of "half-kills" or "assists"...  Not sure there is a solution that would make everyone happy.  

I do agree that something needs to be done to attribute next-round ejections and death from collapsing buildings or pushed off cliffs (two of my favourite methods).  Even now I play with a notepad open so I track significant damage and I can post-attribute kills correctly.

pheonixstorm:
So track who destroyed a gyro or leg(s)? But in that situation if the pilot is stupid enough to keep trying to stand up and kills his own unit should the other unit still get credit?

Also.. looking over 4360 again it does look like a mess code wise. May not be too difficult but will probably look a bit messy.

arlith:
We do not keep track of who damaged what & when.  Right now, the only "damage tracking" is keeping track of the last unit that dealt damage to another unit, which is used for kill crediting.

You could theoretically keep a more detailed record, but it's going to take memory to store all of that accounting.

pheonixstorm:
I wasn't thinking every little detail, only those few situations of what would put a mech out of action as if it was destroyed. Not sure how possible it really is to check for the crits or leg destroyed, but those are the two biggest causes for ejection. If you have any better ideas I'm all ears.

pheonixstorm:
Leaving this as an idea, but if unit A causes unit B to have an ammo/engine explosion which kills (directly) unit C and D should look for who killed B and give credit get B's attacker. So an additional function may be needed for a secondary unit to find who gets credit. Need to look over how the kill credit code looks first.

Another optional tracking for ejection and abandonment kills would be to track certain crits.

Basically
lastAttacker
lastFatalCrit
getCauseOfExplosion() which should return ??

I really need to get familiar with this particular code again...

PR #517, opened 2017-03-30
Allow adding custom equipments by XML

luiges90:
Revival of https://sourceforge.net/p/megamek/patches/463/ 

I download the patch file and then doing a `git apply`. 

Note: completely untested against newest MM! (it does compile to me tho)

### DIFF
--- a/megamek/docs/custom-ammo.xml
+++ b/megamek/docs/custom-ammo.xml
@@ -1,0 +1,251 @@
+<Weapon>
+	<name>SRM Special Ammo</name>
+	<shortName>SRMSp</shortName> <!-- For RS printing -->
+	<lookupName> <!-- Name(s) of this weapon in MTF/BLK files -->
+		<Name>ISSRMSpecial Ammo</Name>
+		<Name>IS SRM Special Ammo</Name>
+		<Name>IS Ammo SRM-Special</Name>
+	</lookupName>
+	<tonnage>1.0</tonnage>
+	<criticals>1</criticals>
+	<tankSlots>1</tankSlots>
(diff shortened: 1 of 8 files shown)


.gitignore
```diff
@@ -5,3 +5,5 @@
 /.settings/
 /logs/
 .DS_Store
+.idea
+*.iml
```
luiges90: probably I shouldn't do it here but anyway. These are files added by IntelliJ IDEA that are to be ignored

arlith:
Looking through this a bit more, I see that the vast majority of the code added is for serialization/deserialization in XML.  I think that we could probably get the same effect for with far less new code added by using the XStream library that we already use for saved games and such.  It can take a class directly and serialize it to XML.  This would reduce the amount of code added, and is also more maintainable because it means we don't have to keep adding new code every time a new field is added to `Weapon`.

I had debated whether we should use XML or CSV, and I think XML is a better choice in the end (probably compressed).  But, I think some type of UI should be developed to view and possibly edit the values.

I'm not necessarily saying you have to do all or any of this; mostly I'm writing down my thoughts.  This is a feature that I would like to see happen, but I think it will take a lot of work.

Ultimately, I'd like to load all of the weapons/ammo/equipment from files, which is why I think we really need a UI for viewing the equipment and editing it.  Although, we should keep a custom/canon distinction.

Dylan-M:
YAML might be a better choice than XML for this, since the formatting is easier for the human mind...

Then again, if we author an editor it shouldn't matter either.

sixlettervariables:
Without a champion this effort isn't likely to succeed (nor in its current form with the XML).

PR #518, opened 2017-03-30
Quirks update

pheonixstorm:
Added Stabilized Weapon and Barrel Fist quirk

Updated/Fixed Fragile Fuel Tank, Improved Sensors, and Cooling System Flaws

There is still more to add, but I figured this small chunk could be reviewed now while I work on some of the other quirks and update some of the comments/descriptions.


arlith:
To be honest, it'd be a lot nicer if these were submitted individually.  It's less code to review for a single new feature, and if there's an issue with one, it won't hold up all of the rest.

pheonixstorm:
The Full Head Ejection was remove a long time ago as it isn't a quirk and never had any associated code. It somehow managed to slip back in so this is the second time it has been removed. Was removed in adc93ce7dcdb2804d87082c193d61ed6596601d0 then popped back in in 584fd6b1f9370c0b79ee88ee5f977a6945946d78

For the cooling flaws.. I'm not sure on the HQ side, I guess that could be a rules question or just something for all the devs to mull over. Forgot about the reset thanks for catching that, will get a HQ PR for that up sometime tonight.

pheonixstorm:
Side note. MML allows for construction of mechs with full head ejection.

arlith:
Yea, I guess we already support a 'mech having fullhead ejection.  There's a flag in `Mech`, and MML can set that flag and it's handled with printing.  I'll sign off on the review for this once the MHQ PR is updated.

pheonixstorm:
I already posted it awhile ago unless you reviewed it and there is something for me to change

pheonixstorm:
Nevermind, didn't noticed it because Slack doesn't appear to know what the review comments are so it never updates in the github channel. Making the changes now

PR #524, opened 2017-04-02
SPA Human TRO

pheonixstorm:
This SPA can be found on p76 of CamOps.

The code should be correct/finished but I am not sure if anything needs to be done outside `damageEntity()` to make this fully functional.

Before this is accepted though I have a question regarding `areaSatArty` Does this boolean apply to all AOE attacks? I assume yes but want to make sure.


megamek/src/megamek/server/Server.java
```diff
@@ -21141,6 +21141,25 @@ private void resolveShutdownCrashes() {
             critBonus = Math.min((damage_orig - 1) / 5, 4);
         }
 
+        // Find out if Human TRO plays a part it crit bonus
+        Entity ae = game.getEntity(hit.getAttackerId());
+        if (ae != null) {
+            if (!ae.getCrew().getOptions().stringOption(OptionsConstants.MISC_HUMAN_TRO).isEmpty() && !areaSatArty) {
```
arlith: This is purely stylistic, but I would combine the null check and the human TRO check into one conditional.  This reduces the level of indentation and is safe because of short-circuiting.

Also, the check for unit types should be else-if's.  You can't have a Mech-Aero, or Aero-Tank, so only one condition ever applies.

arlith:
areaSatArty will be true for anything that calls `Server.doExplosion` to deal damage.  This is more than just artillery, and could be something like an engine explosion.  You can determine this  by using the `Call Hierarchy` feature.  If you put your cursor on a method definition and right-click, there's an option for `Call Hierarchy`.  This will give you a tree with all of the methods that call the method in question, and you can expand them to go down the hierarchy.  This is super useful for getting an idea of how something gets used.

pheonixstorm:
Didn't know about the call hierarchy usage, that is very useful. Usually I use the reference/declaration search feature. Using both could make finding code sections far easier.

Code changes should be up.

arlith:
I forgot to evaluate before if `damageEntity` is the only place you need to apply the modifier.  This is kind of a thorny question.  The `critBonus` variable ends up being applied to `Server.criticalEntity`.  There are 17 references to this method.  Many of them shouldn't apply, I don't think  (ie, any that come from `Server.processMovement`).

However, some may apply.  Such as `LegAttackHandler.handleEntityDamage`.  Technically, Human TRO is available to all unit types, and CO seems to indicate that the bonus should apply to all rolls on the crit table... There are three other references that arguably should have the bonus.  The full list (which you can get by looking at references to `Server.criticalEntity`) are:

`LegAttackHandler.handleEntityDamage`
`PopUpMineLauncher.handleEntityDamage` -> This is questionable, as I'm not sure about the rules for popup mine launchers
`Server.resolveDfaAttack`
`Server.resolveKickAttack` -> This is questionable, because it's from weakened legs

pheonixstorm:
From what I was able to test any damage applied to itself from a fall would not set off the crit bonus. If it did the poor Urbanmech whose gyro was severely damaged would have been given the bonus.

The LegAttackHandler also calls `damageEntity`, just a different function. The section I added the code to is basically the very bottom of the damageEntity heap 20974 calls 20972 which calls 20998. The call to damageEntity in LegAttack is the function in ln 20972 as is the call from PopUpMine and several others.

This is also why I wanted to know more about the AOE flag. Since this SPA doesn't specify physical attacks I decided to make the assumption that they also apply to physicals, though mines are an odd area but I would think they would not qualify directly.

pheonixstorm:
Also, my original plan was to put each check in the `criticalEntity` sections but they all seemed to point to `damageEntity` which meant I would not have to add a new variable to each function/call

From all the digging I am fairly certain that damageEntity was the best spot. It was also the only spot that seems to deal with using or passing along the critBonus variable.

arlith:
For `LegAttackHandler.handleEntityDamage` specifically, I was referring to the special leg attack critical.  Besides the damage that is dealt, there is a special critical chance because it's a leg attack.  The question is, should human TRO apply to that? I mean, in theory, if an infantry unit has exception knowledge about the weaknesses of 'mechs, they _could_ use that to make more effective leg attacks.

pheonixstorm:
It is open to **any** unit, so infantry could make use of it. 

As such, would this change work?
// Do criticals.
        int critMod = 0;
        if (entityTarget.getArmorType(hit.getLocation()) == EquipmentType.T_ARMOR_HARDENED) {
            critMod -= 2;
        }
        if (ae.getCrew().getOptions().stringOption(OptionsConstants.MISC_HUMAN_TRO).equals(Crew.HUMANTRO_MECH)) {
            critMod += 1;
        }
        vPhaseReport.addAll(server.criticalEntity(entityTarget, hit.getLocation(), hit.isRear(), critMod, damage));

arlith:
Yes, that's basically what I had in mind.

pheonixstorm:
Added.

And why is it when I try to use the code/quote tags mine never looks like that.. always looks crappy or really only handles a single line... grr lol

Issue #539, opened 2017-04-20
Princess GE AI


BLOODWOLF333:
I tested what people were reporting on the forums and it is true that princess does not rotate the turret on gun emplacements.  If you walk into her firing arc she will fire at you like she has done to me but, right now turrets might as well be static emplacements to her.

Requiring the player to deploy her turrets we can live with but, can her AI code for operating a vee's turret be copied for use when firing a gun emplacement?  Obviously she still will not do anything with arty turrets and will require a human player for that case but, getting her to operate the turrets right now would be great and I wont feel like I wasted so much time making all those new turret icons lol.

NickAragua:
This code for this is in pull request #578.

BLOODWOLF333:
Thank you we were needing that.

PR #578, opened 2017-07-30
#539 - Enable Princess to rotate turrets and use "extended torso twist" quirk

NickAragua:
Introduced the possibility for "torso twists" larger than one hex side. Princess will also take that kind of possibility into account when planning her moves.


megamek/src/megamek/client/bot/princess/FireControl.java
```diff
@@ -1948,30 +1948,27 @@ FiringPlan getBestFiringPlanWithTwists(Entity shooter, Targetable target,
...
-        rightTwistPlan.setTwist(1);
-
-        // Turn to the left.
-        shooter.setSecondaryFacing(correctFacing(originalFacing - 1));
-        FiringPlan leftTwistPlan = getBestFiringPlan(shooter, target, game, ammoConservation);
-        leftTwistPlan.setTwist(-1);
-
-        // todo extended torso twist.
-
+        ArrayList<Integer> validFacingChanges = getValidFacingChanges(shooter);
```
dericpage: Typically, it's better to use the interface (List in this case) for variable declarations as it reduces overhead and makes it easier to convert to other List subtypes if needed.

megamek/src/megamek/client/bot/princess/FireControl.java
```diff
@@ -1948,30 +1948,27 @@ FiringPlan getBestFiringPlanWithTwists(Entity shooter, Targetable target,
...
+        
+        // Now, we loop through all possible facings. If one facing produces a better plan 
+        // than what we currently have as the best plan then use that. Start with "no twist" as default.
+        FiringPlan bestFiringPlan = noTwistPlan;
+        for(int index = 0; index < validFacingChanges.size(); index++)
+        {
```
dericpage: You should replace this with a For...Each loop.

megamek/src/megamek/client/bot/princess/FireControl.java
```diff
@@ -1948,30 +1948,27 @@ FiringPlan getBestFiringPlanWithTwists(Entity shooter, Targetable target,
...
+        	int currentTwist = validFacingChanges.get(index);
+        	
+        	shooter.setSecondaryFacing(correctFacing(originalFacing + currentTwist));
+        	FiringPlan twistPlan = getBestFiringPlan(shooter, target, game, ammoConservation);
+        	twistPlan.setTwist(currentTwist);
+        	
+        	if(twistPlan.getUtility() > bestFiringPlan.getUtility())
+        		bestFiringPlan = twistPlan;
```
dericpage: Put brackets around your clauses.  It makes it easier to extend the clause later if need be.

megamek/src/megamek/client/bot/princess/FireControl.java
```diff
@@ -2001,32 +1998,27 @@ FiringPlan guessBestFiringPlanUnderHeatWithTwists(Entity shooter,
...
-        rightTwistPlan.setTwist(1);
-
-        // Turn to the left.
-        shooter.setSecondaryFacing(correctFacing(originalFacing - 1));
-        FiringPlan leftTwistPlan = guessBestFiringPlanUnderHeat(shooter,
-                shooterState, target, targetState, maxHeat, game);
-        leftTwistPlan.setTwist(-1);
-
-        // todo extended torso twist.
+        ArrayList<Integer> validFacingChanges = getValidFacingChanges(shooter);
```
dericpage: Again, List vs ArrayList.

megamek/src/megamek/client/bot/princess/FireControl.java
```diff
@@ -2001,32 +1998,27 @@ FiringPlan guessBestFiringPlanUnderHeatWithTwists(Entity shooter,
...
+        
+        // Now, we loop through all possible facings. If one facing produces a better plan 
+        // than what we currently have as the best plan then use that. Start with "no twist" as default.
+        FiringPlan bestFiringPlan = noTwistPlan;
+        for(int index = 0; index < validFacingChanges.size(); index++)
+        {
```
dericpage: Another for...each candidate.

megamek/src/megamek/client/bot/princess/FireControl.java
```diff
@@ -2001,32 +1998,27 @@ FiringPlan guessBestFiringPlanUnderHeatWithTwists(Entity shooter,
...
+        	int currentTwist = validFacingChanges.get(index);
+        	
+        	shooter.setSecondaryFacing(correctFacing(originalFacing + currentTwist));
+        	FiringPlan twistPlan = guessBestFiringPlanUnderHeat(shooter, shooterState, target, targetState, maxHeat, game);
+        	twistPlan.setTwist(currentTwist);
+        	
+        	if(twistPlan.getUtility() > bestFiringPlan.getUtility())
+        		bestFiringPlan = twistPlan;
```
dericpage: Brackets

megamek/src/megamek/client/bot/princess/FireControl.java
```diff
@@ -2048,32 +2040,27 @@ FiringPlan guessBestFiringPlanWithTwists(Entity shooter,
...
-        rightTwistPlan.setTwist(1);
-
-        // Turn to the left.
-        shooter.setSecondaryFacing(correctFacing(originalFacing - 1));
-        FiringPlan leftTwistPlan = guessBestFiringPlan(shooter, shooterState,
-                target, targetState, game);
-        leftTwistPlan.setTwist(-1);
-
-        // todo extended torso twist.
+        ArrayList<Integer> validFacingChanges = getValidFacingChanges(shooter);
```
dericpage: List vs ArrayList

megamek/src/megamek/client/bot/princess/FireControl.java
```diff
@@ -2048,32 +2040,27 @@ FiringPlan guessBestFiringPlanWithTwists(Entity shooter,
...
+        
+        // Now, we loop through all possible facings. If one facing produces a better plan 
+        // than what we currently have as the best plan then use that. Start with "no twist" as default.
+        FiringPlan bestFiringPlan = noTwistPlan;
+        for(int index = 0; index < validFacingChanges.size(); index++)
```
dericpage: For...each

megamek/src/megamek/client/bot/princess/FireControl.java
```diff
@@ -2048,32 +2040,27 @@ FiringPlan guessBestFiringPlanWithTwists(Entity shooter,
...
+        {
+        	int currentTwist = validFacingChanges.get(index);
+        	
+        	shooter.setSecondaryFacing(correctFacing(originalFacing + currentTwist));
+        	FiringPlan twistPlan = guessBestFiringPlan(shooter, shooterState, target, targetState, game);
+        	twistPlan.setTwist(currentTwist);
+        	
+        	if(twistPlan.getUtility() > bestFiringPlan.getUtility())
+        		bestFiringPlan = twistPlan;
```
dericpage: brackets

megamek/src/megamek/client/bot/princess/FireControl.java
```diff
@@ -2836,4 +2823,38 @@ protected Mounted getAntiAirAmmo(List<Mounted> ammoList, WeaponType weaponType,
 
         return returnAmmo;
     }
+
+    // Helper method that figures out the valid facing changes for the given shooter
+    private ArrayList<Integer> getValidFacingChanges(Entity shooter)
```
dericpage: Definitely prefer using interfaces in method returns.

megamek/src/megamek/client/bot/princess/FireControl.java
```diff
@@ -2836,4 +2823,38 @@ protected Mounted getAntiAirAmmo(List<Mounted> ammoList, WeaponType weaponType,
...
+    {
+    	// figure out all valid twists or turret turns
+        // mechs can turn:
+        //		one left, one right unless he has "no torso twist" quirk
+        //		two left, two right if he has "extended torso twist" quirk
+        // vehicles and turrets can turn any direction unless he has no turret
+        ArrayList<Integer> validFacingChanges = new ArrayList<Integer>();
```
dericpage: List interface vs ArrayList

megamek/src/megamek/client/bot/princess/FireControl.java
```diff
@@ -2836,4 +2823,38 @@ protected Mounted getAntiAirAmmo(List<Mounted> ammoList, WeaponType weaponType,
...
+        {
+        	validFacingChanges.add(1);
+        	validFacingChanges.add(-1);
+        	validFacingChanges.add(2);
+        	validFacingChanges.add(-2);
+        	validFacingChanges.add(3);
+        }
+        
+        return validFacingChanges;
+    }
```
dericpage: Overall, seems like there's a lot of duplicate code here (which is true of the MM project in general) that could be moved into helper methods.

megamek/src/megamek/client/bot/princess/FireControl.java
```diff
@@ -2836,4 +2752,39 @@ protected Mounted getAntiAirAmmo(List<Mounted> ammoList, WeaponType weaponType,
...
+
+    // Helper method that figures out the valid facing changes for the given shooter
+    private List<Integer> getValidFacingChanges(Entity shooter)
+    {
+    	// figure out all valid twists or turret turns
+        // mechs can turn:
+        //		one left, one right unless he has "no torso twist" quirk or is on the ground
+        //		two left, two right if he has "extended torso twist" quirk
+        // vehicles and turrets can turn any direction unless he has no turret
+        ArrayList<Integer> validFacingChanges = new ArrayList<Integer>();
```
dericpage: Should be List rather than ArrayList for the declaration, and actually, this should probably be a Set/HashSet since the order doesn't matter and each facing should be unique.

megamek/src/megamek/client/bot/princess/FireControl.java
```diff
@@ -2836,4 +2752,39 @@ protected Mounted getAntiAirAmmo(List<Mounted> ammoList, WeaponType weaponType,
...
+        if(shooter instanceof Mech &&
```
dericpage: You should replace these instanceof comparasons with equals comparisons to Entity.ETYPE_*.  It's a faster operation and the reason the ETYPEs exist.  So, you'd have (Entity.ETYPE_MECH == shooter.getEntityType()).  The only potential gotcha would be shooter being null, but you should have a null check that throws an NPE with a specific message at the start of the method as a best practice anyway.

megamek/src/megamek/client/bot/princess/FiringPlanCalculationParameters.java
```diff
@@ -0,0 +1,77 @@
...
+     * @return the 'best' firing plan under a certain heat, includes the option of twisting.
+     */
+    public FiringPlanCalculationParameters(Entity shooter, Targetable target, Map<Mounted, Double> ammoConservation)
+    {
+    	this.shooter = shooter;
+    	this.target = target;
+    	this.ammoConservation = ammoConservation;
+    	calculationType = FiringPlanCalculationType.Get;
+    }
+}
```
dericpage: This should probably be an immutable class and you might want to look at using a Builder pattern rather than overriding the constructor.

megamek/src/megamek/client/bot/princess/FiringPlanCalculationParameters.java
```diff
@@ -0,0 +1,77 @@
...
+public class FiringPlanCalculationParameters
+{
+	//The type of firing plan calculation to carry out
+	public enum FiringPlanCalculationType
+	{
+		// We're guessing the firing plan based on our estimate of enemy movement
+		Guess,
+		// We're getting a firing plan based on exact known enemy movement results
+		Get
+	}
```
dericpage: FYI, enum values should be all caps by the Java coding standard as they're constants.

megamek/src/megamek/client/bot/princess/FiringPlanCalculationParameters.java
NickAragua: Did a little reading on the builder pattern, and this situation seems tailor-made for it.

NickAragua: Not sure I can make it completely immutable (without quite excessive code changes to other code), since the various object properties (Entity, EntityState) contain mutable objects.

dericpage: Immutable classes containing mutable objects is pretty normal.  In that case, the standard practice is to use 'defensive copies' of the mutable object to prevent anything outside the immutable class changing it.
http://www.javapractices.com/topic/TopicAction.do?Id=15

NickAragua: My point is that I can make a defensive copy of (for example) an Entity, but that Entity has a number of mutable properties itself, so we're not really getting the full benefits of immutability.

That aside, I'm concerned about the performance impact/memory footprint of repeatedly creating copies of an Entity. This "guess"/"get" firing plan method is called very frequently (many times per path, and the moving unit also tries to predict the location of unmoved enemy units and evaluate *their* best firing plans) by the AI routine and Entity is not a light-weight object. Garbage collection or not, I have trouble imagining that it won't have a performance impact.

I'll see if I can gather up some statistics tonight re: usage of this method, maybe my concerns are misplaced. I'm pretty sure we're talking about #legal paths for current unit * (sum over all other units of # legal paths for that unit), though.

arlith: I don't think it's a great idea to create a defensive copy of `Entity` (I'm kind of jumping in the middle of this, so I may have missed important details).  `Entity` is already a pretty heavy object, and has some inefficient uses of memory.

dericpage: @NickAragua Well, in those situations what you do is create copies in both your setters and getters, so the copy in the immutable class is completely isolated.  However, you're right that the performance hit is likely not worth it.  Looks like you've got it about as immutable as can be reasonably expected.

Ideally, we'd store only those parts of Entity that Princess actually uses in EntityState and make that immutable but that's not much smaller a change than trying to Entity immutable.

dericpage:
If you resolve the conflicts (looks like line-wrapping changes are the source of the conflict) and update the code to match the project coding style (https://github.com/MegaMek/megamek/wiki/Code-Formatting), we can get this merged.

NickAragua:
Ok, I'll take a look at it tonight.

NickAragua:
Ok, resolved the conflicts and made the style changes.

Issue #479, opened 2017-01-25
Feature Request: Tooltip BV


ThomasMarik:
When playing games that have BV victory conidtions, such as ATB games,  I am finding myself constantly wanting to check the current BV of both enemy and my own units.  I would like to request that this information be put on the tooltip of the unit for easy access.

SJuliez:
BV victory conditions check the team BV, don't they? Wouldn't the team BV then be more interesting than the individual unit BV?

ThomasMarik:
checkbv will give me the total team BV.  I am wanting individual BVs so I can easily assess priority targets on the map.

PR #586, opened 2017-08-05
Added feature request for Issue 479

wildj79:
Adds a feature requested in issue #479. This allows a user to quickly see the current battle value of a unit in game, without having to view the unit. The listing shows the current bv, the initial bv, and the percentage of bv remaining. If playing with double blind and the "Hide Battle Value report" option is selected, the bv listings will not be visible on enemy units, but will be visible on your units.

### DIFF
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -217,6 +217,9 @@ public abstract class Entity extends TurnOrdered implements Transporter,
     protected boolean useManualBV = false;
     protected int manualBV = -1;
 
+    protected int initialBV = -1;
+    protected boolean initialBVSet = false;
+
     protected String displayName = null;
     protected String shortName = null;
     public int duplicateMarker = 1;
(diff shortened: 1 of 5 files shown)


megamek/src/megamek/common/Entity.java
```diff
@@ -217,6 +217,9 @@
     protected boolean useManualBV = false;
     protected int manualBV = -1;
 
+    protected int initialBV = -1;
+    protected boolean initialBVSet = false;
```
neoancient: I don't see any reason for the `initialBVSet` field, when you can test whether it has been set with `initialBV < 0`.

Since `initialBV` is only used in the client and does not change, it could be marked `transient` to keep from sending the value back and forth between the client and the server each time the `Entity` is updated.

wildj79: I think I would agree with you.  My initial thought was that we would want to make sure this value doesn't get updated more than once.  But, after submitting the pull request, I came to the conclusion that it was superfluous, so I'll remove it.

PR #601, opened 2017-08-22
Here's my fix for the null checks required in Server around line 894

mkerensky:


megamek/src/megamek/server/Server.java
```diff
@@ -887,6 +887,16 @@ private void receivePlayerVersion(Packet packet, int connId) {
             System.out.println("ERROR: Client/Server Version Mismatch -- Client: "+version+" Server: "+MegaMek.VERSION);
             needs = true;
         }
+        else if (clientChecksum == null) {    // print a message indicating client doesn't have jar file
```
NickAragua: I'm not sure that this should be part of the previous if block, as it's relevant to the checksum rather than the version.

arlith:
This is good save for two things. I will add in-line comments.

megamek/src/megamek/server/Server.java
arlith: This line should be on the same line as the closing brace in line 889, to match coding conventions.

megamek/src/megamek/server/Server.java
```diff
@@ -887,6 +887,16 @@ private void receivePlayerVersion(Packet packet, int connId) {
...
+            buf.append("Client Checksum is null. Client may not have a jar file");
+            System.out.println("ERROR: Client does not have a jar file");
+            needs = true;      	
+        }
+        else if (serverChecksum == null) {    // print message indicating server doesn't have jar file
```
arlith: Same as line 890 -- this should go up a line to be on the same line as the closing bracket on line 894.

megamek/src/megamek/server/Server.java
```diff
@@ -887,6 +887,16 @@ private void receivePlayerVersion(Packet packet, int connId) {
...
+            buf.append("Server Checksum is null. Server may not have a jar file");
+            System.out.println("ERROR: Server does not have a jar file");
+            needs = true; 
+        }
         if (!clientChecksum.equals(serverChecksum)) {
```
arlith: This needs to be an else-if with the previous conditional block, otherwise you won't actually prevent the null pointer exception.

megamek/src/megamek/server/Server.java
arlith: Yes, I overlooked that before.  You're right, the version and checksum checks should be separate.  Additionally, there should be a new-line check, like when the checksums don't match.

megamek/src/megamek/server/Server.java
mkerensky: Hah, missed that one.  I had it as an else-if in my dropship branch and copied it over wrong.  Anything else I should know about the coding conventions?  I am trying to use 4 spaces instead of tab to indent, based on a convention I remember from my 2010 work.  I'm trying to put braces inline with their statements, though I know I missed a bunch of those in my dropship branch before I got the word.

Dylan-M: You can configure your IDE (for Eclipse and Intellij at least) to do 4 spaces for a tab keypress on a per project basis. You should do so.

mkerensky:
Latest update should take care of the requested changes.

megamek/src/megamek/server/Server.java
mkerensky: I haven't understood whether I need to make this change or not, or exactly what the change is.   Do I need the                 buf.append(System.lineSeparator());?   Should the null checks go after the checksum mismatch block?

arlith:
With the latest changes, it looks like there's one big if/else.  There should be two: there should be a conditional that checks whether there's a version mismatch, and then there should be a conditional that checks for a checksum mismatch.  The conditions checking for nulls in the checksums, should be one if/else block that is not part of the version check.

These should be separate, as the version check is different from the checksum check, and they can have independent outcomes.

megamek/src/megamek/server/Server.java
arlith: You need to add code like this:
```
if (!version.equals(MegaMek.VERSION)) {
                 buf.append(System.lineSeparator());
}
```
What this is doing is adding a new line.  Otherwise, if you have a version mismatch **and** checksum mismatch, the two error messages will run together.  The new line will keep them on a separate line.

Dylan-M: I would add TWO line separators, so that there is also a blank line between them.

mkerensky:
Round 2. Is this more what you had in mind?

mkerensky:
@arlith  Any thoughts on the latest revision?  I think the biggest issue I'm having with this PR is that I don't know where to look for the changes I'm adding to see if I'm producing the desired results. I thought it was the log, but then I saw some of the output appear in the chat window in the lobby.

mkerensky:
Please forgive the extra commits. I switched git clients and the new one seems to have an update delay. 'Null Checksum, round 3' contains the changes I meant to post.

arlith:
You're going to have to walk back that last commit.  f229978 has diffs in *all* of Server.java.  Usually, this is a whitespace issue (tabs going to spaces, or spaces to tabs or something like that).  The diff is so large it won't load in GitHub.  I could look at it later, but I don't want to merge in anything that changes all of server.  I'm going to be gone without internet the rest of the weekend.

mkerensky:
I see what you're talking about, but have no idea how that happened and I'm not having much luck fixing it. I'm going to close this PR, delete the branch and start over with a new, clean copy of master.

PR #613, opened 2017-08-30
Aero partial repairs

mkerensky:
+ Feature: Adds all partial repairs from SO that apply to spacecraft. 
+ Fix: Removes from Aero options a few newer quirks that shouldn't apply to them.


arlith:
Have you tested these changes thoroughly?

mkerensky:
I believe that I have, yes.

PR #633, opened 2017-09-08
Princess aerotech handling

NickAragua:
A brief summary of the changes:

Aero units on atmospheric ground maps use an entirely new pathfinding algorithm. Behavior includes:

- attempting to fly over enemies
- striking
- dive bombing
- staying at altitude 5
- flying off the board
- mostly avoiding PSRs

Behavior does not include:
- altitude bombing
- air to air combat untested

There are a few improvements that I'd like to make, most notably:
- Add evasive action at the end of selected move path if it does not fly over an enemy
- Adjust path generation algorithm to respect board edges a little better

But I wanted to get this out and get people's feedback on it before making further changes.

Also, as a bonus, slight general improvement to Princess performance by caching results of "getMovePathSuccessProbability" method, as it was being invoked multiple times for the same path.


NickAragua:
Well, I fubared this one up pretty good.

PR #635, opened 2017-09-08
SPA Terrain Master

pheonixstorm:
This is 4 of the 6 Terrain Master SPAs. These are all independent so these 4 can be merged after review.

Question: In Terrain.java the movementCost function is starting to get a bit messy. Should I rewrite the function to make it easier to deal with in the future now or create a separate PR for it later?


HammerGS:
This should be merged in before we do the IO merge, as IO some quirk fixes in optionsconstants.

megamek/src/megamek/common/Entity.java
```diff
@@ -6871,6 +6874,9 @@ public PilotingRollData checkBogDown(MoveStep step,
             if ((this instanceof Mech) && ((Mech) this).isSuperHeavy()) {
                 roll.addModifier(1, "superheavy mech avoiding bogging down");
             }
+            if (getCrew().getOptions().booleanOption(OptionsConstants.PILOT_TM_SWAMP_BEAST)) {
+                roll.addModifier(-1, "swamp beast");
```
arlith: Should fix capitalization to match other entries.

megamek/src/megamek/common/Entity.java
```diff
@@ -10268,6 +10278,10 @@ public void addPilotingModifierForTerrain(PilotingRollData roll, Coords c,
         }
         IHex hex = game.getBoard().getHex(c);
         hex.terrainPilotingModifier(getMovementMode(), roll, enteringRubble);
+
+        if (hex.containsTerrain(Terrains.JUNGLE) && getCrew().getOptions().booleanOption(OptionsConstants.PILOT_TM_FOREST_RANGER)) {
+            roll.addModifier(-1, "Forest Ranger");
```
arlith: The tooltip for forest ranger didn't really elude to this, bit the tooltip for Mountaineer did.  Should mountaineer also have a check for rubble here?  Perhaps I don't understand what this modifier is for.

megamek/src/megamek/common/Terrain.java
```diff
@@ -370,6 +370,22 @@ public int movementCost(Entity e) {
...
+                    }
+                    return 1;
+                }
+                if (((e instanceof Mech) && ((Mech)e).isSuperHeavy())
+                    || (e.getCrew().getOptions().booleanOption(OptionsConstants.INFANTRY_FOOT_CAV)
+                            && (moveMode == EntityMovementMode.INF_LEG))) {
+                    return 0;
+                }
+                return 0;
+            }
```
arlith: I think it would be easier to write something like
```
int mountaineerModifier = 0;
if (mountaineer) {
    mountaineerModifier = 1;
}
```
And then always subtract the mountaineer Modifier.  This requires 4 new lines, and then modifier old lines, instead of adding a big block of mostly duplicated code.

megamek/src/megamek/common/Terrain.java
```diff
@@ -389,6 +405,12 @@ public int movementCost(Entity e) {
                     && (moveMode == EntityMovementMode.INF_LEG) && (level > 1)) {
                 return level - 1;
             }
+            if (e.getCrew().getOptions().booleanOption(OptionsConstants.PILOT_TM_FOREST_RANGER)) {
+                if ((e instanceof Mech) && ((Mech)e).isSuperHeavy()) {
+                    return level - 2;
```
arlith: Should this have a `Math.max(0, level -2)` instead?  I think this could go negative (although I didn't verify it too much).

megamek/src/megamek/common/actions/WeaponAttackAction.java
```diff
@@ -1141,6 +1141,18 @@ else if (wtype.getAtClass() == WeaponType.CLASS_LBX_AC) {
                     && (!ae.isAirborne() || !ae.isAirborneVTOLorWIGE())) {
                 toHit.addModifier(+1, OptionsConstants.PILOT_SHAKY_STICK);
             }
+            if (te.getCrew().getOptions().booleanOption(OptionsConstants.PILOT_TM_FOREST_RANGER)
+                    && (game.getBoard().getHex(te.getPosition()).containsTerrain(Terrains.WOODS)
+                       || game.getBoard().getHex(te.getPosition()).containsTerrain(Terrains.JUNGLE))
+                    && te.moved == EntityMovementType.MOVE_WALK) {
+                toHit.addModifier(+1, "forest ranger");
```
arlith: Capitalization

megamek/src/megamek/common/actions/WeaponAttackAction.java
```diff
@@ -1141,6 +1141,18 @@ else if (wtype.getAtClass() == WeaponType.CLASS_LBX_AC) {
...
+            }
+            if (te.getCrew().getOptions().booleanOption(OptionsConstants.PILOT_TM_SWAMP_BEAST)
+                    && (game.getBoard().getHex(te.getPosition()).containsTerrain(Terrains.MUD)
+                        || game.getBoard().getHex(te.getPosition()).containsTerrain(Terrains.SWAMP))
+                    && te.moved == EntityMovementType.MOVE_RUN) {
+                toHit.addModifier(+1, "swamp beast");
```
arlith: Capitalization

megamek/src/megamek/common/Terrain.java
pheonixstorm: One of the reasons I asked if movecost should get a rewrite. It would be easier to make all of rely on a modifier. At the top of the switch add say int mp = 0; then in each case add the TO per hex cost. so for case WOODS use mp += level then subtract each possible modifier from it and return the final value.

arlith: I don't generally get the full context looking at diffs.  Yea, I can expand them, but you have to expand a lot.  I think, overall, the structure is probably okay.  It's not really dense code, but it just has a lot of white space.  In this particular case, you duplicated a big block of code, and I think it could be written without any real duplication.

megamek/src/megamek/common/Entity.java
pheonixstorm: Jungle has a PSR modifier unlike woods. Forest Ranger gives a bonus to that roll. If its not in the tooltip I will add it. Pretty sure I had it there at one point though...

megamek/src/megamek/common/Terrain.java
pheonixstorm: Went ahead and pushed a fix I had on the moveCost test branch I linked you in clack. This should work better than testing in each case.

pheonixstorm:
I am going to go ahead and push the test branch changes over. After starting to work on the Animial Mimicry SPA it would appear that it would be better to have movementCost rewritten anyway as Animal Mimicry adds new move modifiers for jungle and woods. Adding cross-country may also add new material as well and the old way would then become really cluttered.

megamek/src/megamek/common/Entity.java
pheonixstorm: This was corrected with 01c807d

pheonixstorm:
All reviews should be covered now, let me know if there is any other changes that need to be made.

arlith:
I would recommend going through and replacing all `return mp;` lines with `return Math.max(0, mp)`.  There are some cases where you could technically return a -1 (rubble is the first example), however I would apply the check to all cases as a preventative measure.  It's really minuscule overhead, and if someone copies/pastes some code, the check will be there.

arlith:
After that last change, I'll merge the PSR.  If you commit the history.txt, I can merge through the web.

pheonixstorm:
That was dealt with in f681b67 by adding
`for (ITerrain terrain : terrains.values()) {
     rv += terrain.movementCost(entity);
     rv += Math.max(0, terrain.movementCost(entity));;
         }`
to Hex.java

arlith:
I don't think that's a very suitable fix.  Certainly, it's the most efficient (less overhead), but I think efficiency is of little concern in this case.  I feel like this fix is error-prone.  I'm not sure what all calls `Terrain.movementCost` right now, but this could be counter-intuitive to future uses and lead to bugs.  I think that it's best that `Terrain.movementCost` not return negative values.

megamek/docs/history.txt
```diff
@@ -1,6 +1,7 @@
 VERSION HISTORY:
 ----------------
 v0.43.4-git
++ PR #635: Terrain Master SPA (All but Drag Racer and Nightwalker)
```
arlith: We generally append new entries onto the bottom.  The top most entry is the first thing that got fixed/added in a release, and subsequent entries happen after that.

Also, I think it's better to enumerate the SPAs added rather than list the ones that weren't (it's more informative to people who don't know all of the terrain master SPAs.  You could just say "Terrain Master SPAs: blah, blah 2, and blah 3".

pheonixstorm: So for future reference or go ahead and change it?

Also, the only spot it is called from is where I changed it in Hex.java Can have it changed in 5-10 if you still think it may be an issue in the future 👍

arlith: This isn't really written down anywhere yet, but I think the policy we should generally have is that the person merging in the PR should add the history.txt entries.  The reason is, any subsequent changes to history.txt are all but guaranteed to cause a merge conflict.  This means we'll have to sort out this conflict for every PR.  In this case, I asked you to add them because then I can merge this PR before I get home tonight, and I don't expect anyone to update history.txt before then.

Yes, I think it's best to ensure that `Terrain.movementCost` never returns negative.  I think it's a good preventative measure.

pheonixstorm:
Pushed final changes and fixed the history.txt

PR #660, opened 2017-09-18
Run unit tests during packaging.

dericpage:
Modified the packaging scripts to run the unit tests to ensure that unit tests are run any time the project is built (unless the 'skip.tests' environment variable is set to "true").  This should prevent new code that breaks the tests being added to the project.

### DIFF
--- a/megamek/build.xml
+++ b/megamek/build.xml
@@ -161,8 +161,6 @@
 
     <target depends="clean, compile, unit.test, jar" name="all" description="Clean, compile and build a jar"/>
 
-    <target depends="clean, compile, jar" name="all-no-test" description="Clean, compile and build a jar"/>
-
     <target name="compileTests" depends="compile" unless="skip.tests"  description="Compile unit tests">
         <delete dir="${dir.build.test}"/>
         <mkdir dir="${dir.build.test}"/>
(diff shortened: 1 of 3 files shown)


Dylan-M:
So... did you make whitespace changes or something? Both actual packaging scripts are a bitch to compare because it's a huge number of lines.

dericpage:
Yeah,sorry.  I corrected the tab characters to be 4 spaces.  Should have checked that in separately. The functional changes are near the bottom, with the new test-... targets.

arlith:
Are you going to fix the white-space diffs?  Should we update the task descriptions to indicate you can use a no test env variable?  What's the format of the environment variable supposed to be?  skip.tests?  UNIX conventions would have it be all upper-case (I feel like I missed something in the diffs that defines this).

dericpage:
You mean unfix the whitespace diffs? They were tab characters instead of 4 spaces before I changed them.

Updating the task description to explain the new variable is a good idea.  

I guess calling it an environment variable is a bit misleading.  It can actually be a property, environment variable or a command-line argument and the normal ant convention is for them to be all lowercase.  So, you could access it from the command line using 
`ant -build.file packaging.xml -Dskip.tests=true jar pkgdir-build`

Also, your IDE may allow you to define it.  In IDEA there's a Build File Properties dialog that allows you to define these.

There's more on how the `unless` attribute works here: https://ant.apache.org/manual/properties.html#if+unless

Dylan-M:
Essentially yes, I prefer whitespace fixes like that to be separate from real work because it makes the real work impossible to review in cases like this. I can't see what actually changed, because diff thinks everything did.

arlith:
Ah, I misunderstood the issue.  I thought it was a mistaken switch, not an actual fix.

arlith:
In general, I think this is a good idea.  I haven't reviewed the changes too closely because of the white space issues, but I'm in support of this change in general.

dericpage:
To make the review easier w/out the whitespaces, I'll abandon this branch and resubmit a new one and do two separate check-ins.

Issue #417, opened 2016-11-17
Kill Count Victory Condition Should Count Ejections as Kills


VampireSeraphin:
This is an enhancement I've been meaning to ask for a for a while.

The kill count victory condition should count enemy ejected units as kills. Right now it only counts hard kills, and you can eliminate the entire enemy team and never reach your kill count it there are several ejected units.

arlith:
This is kind of a duplicate of this bug on SF: https://sourceforge.net/p/megamek/bugs/4360/.  The name doesn't make this clear, but to me that issue represents the a problem with how kills are reported in MM.  The kill tracking system was kind of tacked on, and hence isn't as robust as it needs to be.  Basically, kills are indicated when weapon damage is resolved.  Since weapons haven't dealt damage when a unit ejects, the proper information isn't updated.

Bonepart:
Updating kill tracking would require a lot of work, but for now PR #661 should let destruction by ejection and pilot error count. This will only work in the case of 2 teams playing. More than that and you would have to actually credit the kill accurately.

pheonixstorm:
Yes, but read up on the notes from PR 661.

HammerGS:
Related - https://github.com/MegaMek/megamek/issues/416

PR #661, opened 2017-09-19
Update so non-direct kills count for kill victory

Bonepart:
Fixing https://github.com/MegaMek/megamek/issues/417

Before it would not count the kill if the unit did not have a 'killer' assigned. I changed it to award credit towards the kill count victory to the opposite team. Since this wouldn't work very well for more than 2 teams in that event the credit is skipped.

### DIFF
--- a/megamek/src/megamek/server/victory/KillCountVictory.java
+++ b/megamek/src/megamek/server/victory/KillCountVictory.java
@@ -17,6 +17,7 @@ import java.io.Serializable;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.Hashtable;
+import java.util.List;
 
 import megamek.common.Entity;
 import megamek.common.IGame;
@@ -101,9 +102,8 @@ public class KillCountVictory implements Victory, Serializable {
             Entity killer = game.getEntityFromAllSources(wreck.getKillerId());
             
             if (killer == null){
-                if (game.getNoOfTeams() == 2) {
-                    for (Enumeration<Team> e = game.getTeams(); e.hasMoreElements();) {
-                        Team team = e.nextElement();
+                if (game.getNoOfTeams() == 2 && wreck.getOwner().getTeam() != Player.TEAM_NONE) {
+                    for (Team team : game.getTeamsVector()) {
                         if (team.getId() != wreck.getOwner().getTeam()) {
                             Integer kills = teamKills.get(team.getId());
                             if (kills == null){
--- a/megamek/src/megamek/server/victory/KillCountVictory.java
+++ b/megamek/src/megamek/server/victory/KillCountVictory.java
@@ -23,6 +23,7 @@ import megamek.common.IGame;
 import megamek.common.IPlayer;
 import megamek.common.Player;
 import megamek.common.Report;
+import megamek.common.Team;
 
 /**
  * Implements a kill count victory condition.  Victory is achieved if a team (or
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/server/victory/KillCountVictory.java
```diff
@@ -100,6 +102,19 @@ private void updateKillTables(IGame game,
             Entity killer = game.getEntityFromAllSources(wreck.getKillerId());
             
             if (killer == null){
+                if (game.getNoOfTeams() == 2 && wreck.getOwner().getTeam() != Player.TEAM_NONE) {
```
arlith: Each of the comparisons should be surrounded with parents, aka `(game.getNoOfTeams() == 2) && (wreck.getOwner().getTeam() != Player.TEAM_NONE)`

Dylan-M:
I tend to agree with all of Arlith's comments. This is a very hacky solution and isn't something I'd want to try and deal with bugs on.

neilfw:
If the kill tracking code needs refactoring, would it not make sense to capture a log per unit of the sources of damage, then if ther eis a clear KILL it is assigned to that unit, if there is an ejection or other event that disables the unit then the log of the damage is reviewed and the kill assigned to the source of the most significant damage supplier.  That way it would work across any number of opponents, plus the code is only used in the otherwise ambiguous cases.

Dylan-M:
@neilfw That's the direction I would take, but would also need to account for chain kills. If I blow up a fuel tank and it kills 3 mechs, I should get the kill credit for those 3 mechs.

neilfw:
Agreed, Fuel tanks, engine explosions, field artillery and other factors that can cause multiple kills would need to be considered, tracked and catered for.  Also extensive fires overheating meachs and causing ammo explosions... There are an awful lot of edge cases.

Bonepart:
I don't think you can always count on assigning the killer. In the case of explosions if I cause an explosion that kills an enemy unit, but also one of mine I would be the killer, but that unit being gone should still count for the other teams victory. 

ie 4 vs 4, if Kill Victory is set for 3 and I kill 2 of my own by accident that would make victory for the other team impossible. Granted that will probably be fairly rare, but should still be accounted for.

I think improving kill assignment definitely needs done as well, but that will be no small task, and this will make kill victories work in the meantime.

dericpage:
I'm also in the camp of not wanting to introduce a temporary fix that will have to be reverted out when we get around to actually fixing the kill tracking code.  Better to break the needed kill tracking fixes into smaller tasks and tackle those one at a time.  Off the top of my head we'd need:
- A way of tracking the last unit to damage the target (this should account for AoE attacks like artillery).
- A way of tracking the unit that did the most damage to the target (ditto).
- A way of tracking 'indirect' kills (this will be the tricky part as we'd have to track who cause what 'environmental' effect: fire/stackpole/fuel explosion/etc).

kipstafoo:
I'll pile on the 'no hacks, please' camp. I agree on the comments related to tracking. For example, in my TT campaign we track damage done and award XP based upon BV differences (allows an OTT pilot to actually get some XP every now and then). My point being, If you get the killing blow, great, you get credit. If 5 people pile on damage and then the enemy decides to eject, give the kill to who every did the most damage. If we also propagated that damage log back to MHQ that would allow for different ways to handling XP awards.

arlith:
I think @Bonepart brought up a fair point though.  You can't always count on who scored the kill, as you could potentially end up killing your own units.

Perhaps Kill Count isn't really the right victory condition.  I feel like the desired result is really "enemy has lost X units".  I guess this runs into the same problem with multi-team games though (if one team loses the victory condition, every other team technically wins, rather than having one winning team).  What does BV % destroyed handle this?

Bonepart:
I've never tried BV % destroyed with more than 2 teams before. Granted I've never played with more than 2 teams period.

It occurs to me that Battletech seems to be built around the assumption of 2 teams opposing each other. Does anyone play with more than 2 teams, or know a case where someone does?

arlith:
I commonly play multi-team free-for-all games, but I get the impression very few people play with more than 2 teams.

Issue #644, opened 2017-09-11
NPE at the start of firing phase


Question2:
Couldnt hit done because there was a NPE : 

```
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
	at megamek.common.Game.getNextEntityNum(Game.java:1843)
	at megamek.common.Game.getNextEntity(Game.java:1813)
	at megamek.client.ui.swing.FiringDisplay.endMyTurn(FiringDisplay.java:919)
	at megamek.client.ui.swing.FiringDisplay.ready(FiringDisplay.java:1292)
	at megamek.client.ui.swing.AbstractPhaseDisplay$1.actionPerformed(AbstractPhaseDisplay.java:118)
	at javax.swing.AbstractButton.fireActionPerformed(Unknown Source)
	at javax.swing.AbstractButton$Handler.actionPerformed(Unknown Source)
	at javax.swing.DefaultButtonModel.fireActionPerformed(Unknown Source)
	at javax.swing.DefaultButtonModel.setPressed(Unknown Source)
	at javax.swing.plaf.basic.BasicButtonListener.focusLost(Unknown Source)
	at java.awt.Component.processFocusEvent(Unknown Source)
	at java.awt.Component.processEvent(Unknown Source)
	at java.awt.Container.processEvent(Unknown Source)
	at java.awt.Component.dispatchEventImpl(Unknown Source)
	at java.awt.Container.dispatchEventImpl(Unknown Source)
	at java.awt.Component.dispatchEvent(Unknown Source)
	at java.awt.KeyboardFocusManager.redispatchEvent(Unknown Source)
	at java.awt.DefaultKeyboardFocusManager.typeAheadAssertions(Unknown Source)
	at java.awt.DefaultKeyboardFocusManager.dispatchEvent(Unknown Source)
	at java.awt.Component.dispatchEventImpl(Unknown Source)
	at java.awt.Container.dispatchEventImpl(Unknown Source)
	at java.awt.Component.dispatchEvent(Unknown Source)
	at java.awt.EventQueue.dispatchEventImpl(Unknown Source)
	at java.awt.EventQueue.access$500(Unknown Source)
	at java.awt.EventQueue$3.run(Unknown Source)
	at java.awt.EventQueue$3.run(Unknown Source)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(Unknown Source)
[...]

PR #681, opened 2017-09-25
Fix for #644 and another NPE

NickAragua:
Just what it says.

For some reason, the quirks collection includes a null. Meanwhile, addEquipment() invokes addComponent() on a null compositeTechLevel.

### DIFF
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -3395,9 +3395,12 @@ public abstract class Entity extends TurnOrdered implements Transporter,
             throws LocationFullException {
         mounted.setLocation(loc, rearMounted);
         equipmentList.add(mounted);
-        compositeTechLevel.addComponent(mounted.getType());
-        if (mounted.isArmored()) {
-            compositeTechLevel.addComponent(TA_ARMORED_COMPONENT);
+        
+        if(null != compositeTechLevel) {
+            compositeTechLevel.addComponent(mounted.getType());
+            if (mounted.isArmored()) {
+                compositeTechLevel.addComponent(TA_ARMORED_COMPONENT);
+            }
         }
 
         // add it to the proper sub-list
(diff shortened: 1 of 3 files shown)


NickAragua:
Looks like there's going to be about 20 more of these scattered throughout the code. Maybe a better approach would be to ensure that calls to IOptionGroup.getOptions() don't contain null elements.

megamek/src/megamek/common/Entity.java
```diff
@@ -3395,9 +3395,12 @@ public void addEquipment(Mounted mounted, int loc, boolean rearMounted)
             throws LocationFullException {
         mounted.setLocation(loc, rearMounted);
         equipmentList.add(mounted);
-        compositeTechLevel.addComponent(mounted.getType());
-        if (mounted.isArmored()) {
-            compositeTechLevel.addComponent(TA_ARMORED_COMPONENT);
+
```
neoancient: 1. These issues are unrelated and should not be in the same PR.
2. The problem here is that `compositeTechLevel` should never be null. Skipping this section of code if it has an illegal value does not fix the problem. It hides it.

NickAragua: I'll implement the "right" fix for these issues in another PR then. 

For my education:

At what point is the quirks list initialized? Why does it get continuously reloaded in the middle of a tactical game?

At what point during a tactical game is the compositeTechLevel information of an on/offboard unit used?

pheonixstorm: Entity ln822 `quirks.initialize();`
I didn't find any other areas that call initialize in Quirks.java outside of Entity and AbstractOptions

arlith:
I think that the quirk fix is in the same class as the `compositeTechLevel` fix; you're not actually fixing the problem but just hiding it.  This is a crash I've seen often (and never really done anything about) that is generally caused by loading a saved game across versions, where the new version that has changes related to quirks (additions, removals, name changes).

In this case, ignoring the null may be a viable option, but you should clearly understand the implications of this, and not just pushing the problem down to a crash later.

dericpage:
I concur with Arlith.  At the very least, if there's a null element in the quirks list, a log entry should be made (probably at the Warning level) listing the name of the group that contained the null.

NickAragua:
Closing for now, will resubmit with different set of fixes.

PR #688, opened 2017-10-01
Princess disabled unit humping fix

NickAragua:
I've noticed this fairly often and it kind of bugged me - Princess routinely clusters its units around disabled hostile Princess units (such as immobilized vehicles, ejected mechwarriors, etc). Culprit was the code that looks for "closest enemy". Now, we eliminate such units from consideration as "closest enemy", which should increase the bot's effectiveness quite a bit.


NickAragua:
Commit history looks a little weird because I switch branches from other megamek code I was working on.

elementx54:
I tested this out and the changes still make the game playable as I did 3 games against the bot with advance rules. I don't play against the bot much but, I think she did ignore my ejected pilot. However, I'd like the AI creator to take a quick look, at least at the code lines.

Also, if you don't like a commit history, 9 commits for a single changed file, you might be consider squashing. Just an idea.

PR #691, opened 2017-10-04
Allow conventional fighters to drop bombs too

NickAragua:
Found this while doing some princess aero handling playtesting. Conventional fighters don't track heat so they always default to the alpha strike without consideration for bomb drops. This change fixes that behavior.

### DIFF
--- a/megamek/src/megamek/client/bot/princess/FireControl.java
+++ b/megamek/src/megamek/client/bot/princess/FireControl.java
@@ -2010,8 +2010,11 @@ public class FireControl {
         // between firing different weapons, because swarm/leg attacks are
         // mutually exclusive with normal firing, so we treat them similarly to
         // heat-tracking units.
+        
+        // conventional fighters can drop bombs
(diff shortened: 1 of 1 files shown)


arlith:
Does this have an effect on Tanks?

NickAragua:
I don't think so? It resolves to "Shooter does not track heat" and "Shooter is not infantry" and "Shooter is not conventional aircraft".

If someone wanted to enable the AI to have VTOLs drop bombs then this is one of the places they'd need to change.

arlith:
Alright, I looked at the greater context, and it does make sense.  This is exactly for tanks.  It's basically saying: if you don't don't track heat, always alpha-strike.  However, infantry (and now conventional fighters) may have other things to do, so just blindly alpha-striking doesn't make sense.   Tanks, on the other hand, don't have much reason for not alpha-striking (ammo conservation is the only thing that really comes to mind, but I don't think that Princess worried too much about that).

Really, this section of code is just trying to short-circuit things to reduce computation, but I'm not so certain it's a great way to do things anyways.

PR #700, opened 2017-10-12
Fix for soft lock when firing missiles at buildings

NickAragua:
Nothing complicated here, just a couple of null checks. Fixes the behavior where, if you fire missiles of any kind at a building, the game locks up while resolving weapons fire. Buildings aren't entities, so entityTarget was being calculated as null. I suspect the same issue would be happening if you were to shoot missiles at the dirt.


HammerGS:
Merged in.

PR #706, opened 2017-10-20
Quirk NPEs fix

NickAragua:
This fixes several NPEs that result in game hang/UI display failures. Mostly, the situations where I encountered this were when loading a saved game from a previous version of Megamek (I upgraded because I needed a fix for the "mech can't punch" bug). The units involved had quirks that must have been replaced, and so the quirks were loading as null. As I finished that game, I uncovered more of these issues.

Then I did a text search for what kept causing the NPEs and bulletproofed those as well.

Then I noticed that a lot of this code is identical so I pulled it up to separate methods.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/widget/GeneralInfoMapSet.java
+++ b/megamek/src/megamek/client/ui/swing/widget/GeneralInfoMapSet.java
@@ -40,15 +40,15 @@ import megamek.common.Mech;
 import megamek.common.QuadVee;
 import megamek.common.Tank;
 import megamek.common.Warship;
-import megamek.common.options.AbstractOptions;
 import megamek.common.options.IOption;
 import megamek.common.options.IOptionGroup;
+import megamek.common.options.IOptions;
 import megamek.common.options.OptionsConstants;
 import megamek.common.options.PilotOptions;
 import megamek.common.util.MegaMekFile;
 
 /**
- * Set of elements to reperesent general unit information in MechDisplay
+ * Set of elements to represent general unit information in MechDisplay
  */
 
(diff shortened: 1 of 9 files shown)


megamek/src/megamek/client/ui/swing/widget/GeneralInfoMapSet.java
```diff
@@ -356,15 +356,18 @@ public void setEntity(Entity en) {
...
         if ((null != en.getGame())
                 && en.getGame().getOptions().booleanOption(OptionsConstants.ADVANCED_STRATOPS_QUIRKS)) {
-            addOptionsToList(i, en.getQuirks(), quirksR);
+            addOptionsToList(en.getQuirks(), quirksR);
         }
         
+        // Note that, due to the fact that the quirks display is hardcoded to 40 labels in height,
+        // unless you're running at 1024x12000 resolution, you're never going to see partial repairs on the unit info card.
+        // That fix will require changing the data types of partRepsR and quirksR to a dynamically allocated list,
+        // which is not a very simple change, so I'm leaving it for another time.
```
dericpage: Seems like a case for a couple JScrollPanes so that the full lists can be displayed.

As for switching from arrays to lists, that shouldn't be too hard.  From what I looked at, it'd only be about 7 or 8 lines of code that would actually have to be changed.

NickAragua: JScrollPane didn't quite work out, but I was finally able to figure out a solution to dynamically display a variable number of rows. So now we'll be seeing quirks and partial repairs on that unit card.

PR #714, opened 2017-10-23
Large space craft ammo

mkerensky:
+Enhancement: AR10 magazines have a base tonnage from the BLK file and can be loaded with alternate missile types in the lobby accordingly
+Enhancement: Added Peacemaker and Santa Anna nuclear munitions (S2S only) per IO rules. Can be loaded into appropriate capital missile launchers just like other alternate muntions, instead of the 'replace x killer whale warheads with nukes' option. 
+Enhancement: Ability to set number of shots in each aero ammo bin, even if you can't change the munition
+Enhancement: added functionality for Alamo (S2S) and ASEW bombs

### DIFF
--- a/megamek/data/mechfiles/dropships/Arondight Pocket Warship (RAF Mk II).blk
+++ b/megamek/data/mechfiles/dropships/Arondight Pocket Warship (RAF Mk II).blk
@@ -1,0 +1,207 @@
+#building block data file
+<BlockVersion>
+1
+</BlockVersion>
+
+#Write the version number just in case...
+<Version>
+MAM0
+</Version>
+
+<UnitType>
+Dropship
+</UnitType>
+
+<Name>
+Arondight Pocket Warship
+</Name>
+
+<Model>
+(RAF Mk II)
+</Model>
+
+<Tonnage>
+12000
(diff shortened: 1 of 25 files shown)


HammerGS:
Can suggest splitting this into a different PR's

- The Ammo one 
- The ASEW one (including bombs costs here)
- The Peacemaker/Alamo one

neoancient:
Munitions flags are used to create variant ammo types from a single AmmoType entry in bulk, using all the same values except name, shots/ton, rules reference, and tech intro dates. It should not be used for AR10 ammos because they cannot be created by modifying those values. The flags field is sufficient to determine which type of missile it is, which is why it was set up that way. Furthermore, the `long` only has 64 bits, so trying to lshift a 1L more than 63 places leaves you with zero.

neoancient:
Configurable ammo for large craft is implemented in the Large_craft branch, including capital missiles with nuclear warheads.

mkerensky:
I'm keeping this open just in case I find I need something in it that I forgot about when I separated ASEW missiles and Nukes out. If they work when you merge large craft, this can be closed.

PR #717, opened 2017-10-24
Fix for bug # 716

mkerensky:
Changes calcCounterAV() in homing missile handler to check for a non-entity target instead of a null target.

### DIFF
--- a/megamek/src/megamek/common/weapons/ArtilleryWeaponIndirectHomingHandler.java
+++ b/megamek/src/megamek/common/weapons/ArtilleryWeaponIndirectHomingHandler.java
@@ -24,7 +24,6 @@ import megamek.common.Building;
 import megamek.common.Compute;
 import megamek.common.Coords;
 import megamek.common.Entity;
-import megamek.common.GunEmplacement;
 import megamek.common.HitData;
 import megamek.common.IGame;
 import megamek.common.Infantry;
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/weapons/ArtilleryWeaponIndirectHomingHandler.java
```diff
@@ -589,7 +589,7 @@ protected int getAMSHitsMod(Vector<Report> vPhaseReport) {
     //If you're firing missiles at a capital ship...
     @Override
     protected int calcCounterAV () {
-        if ((target == null) || !advancedPD) {
+        if (!(target instanceof Entity) || !advancedPD) {
```
NickAragua: Good start. Can we investigate the possibility of firing one of these at a building that has a gun emplacement on it, and the gun emplacement has AMS? I bet it'll throw an NPE. I ask because I ran into a similar situation recently.

NickAragua: Thinking about it some more, it won't throw an NPE, but it might not fire the AMS.

mkerensky:
Sure.  I thought about the AMS turret while I was working on this, but wasn't sure how buildings were handled as targets. This specific change is in the calcCounterAV section, which only deals with AMS fire of dropships and other bay-using targets.  There's a different call for single AMS that did not throw an NPE when targeting the ground, so I didn't work on that section. I'll test against the AMS turret and see what it does.

mkerensky:
Yes, I've confirmed that the existing AMS code can't handle a targetable building, which affects missiles and thunderbolts too - not just Arrow IVs.  I'll work on that.

megamek/src/megamek/common/weapons/ArtilleryWeaponIndirectHomingHandler.java
```diff
@@ -466,15 +467,27 @@ protected boolean handleSpecialMiss(Entity entityTarget,
...
+        Entity entityTarget;
+        if (target.getTargetType() == Targetable.TYPE_BLDG_TAG) {
+            ArrayList<Entity> turrets = new ArrayList<Entity>();
+            for (Entity turret : game.getEntitiesVector(target.getPosition())) {
+                if (turret instanceof GunEmplacement) {
+                    turrets.add(turret); 
+                }
+            }
+        } else {
+
```
NickAragua: Empty else block is unnecessary, let's get rid of it.

mkerensky:
Yeah, that was work in progress on the gun turret thing, which I determined was going to need a new PR

mkerensky:
Done. This PR now just fixes the NPE

PR #723, opened 2017-10-27
Rebuild of nuclear weapon overhaul

mkerensky:
Rebuild of PR714. This contains only Alamo/santa anna/peacemaker code - prices, damage values, ammotypes (per IO rules) and capitalmissilehandler code for Alamos.
You still can't load or fire any of these without additional work.

### DIFF
--- a/megamek/src/megamek/common/AmmoType.java
+++ b/megamek/src/megamek/common/AmmoType.java
@@ -1199,10 +1199,8 @@ public class AmmoType extends EquipmentType {
         EquipmentType.addType(AmmoType.createKrakenAmmo());
         EquipmentType.addType(AmmoType.createKillerWhaleAmmo());
         EquipmentType.addType(AmmoType.createPeacemakerAmmo());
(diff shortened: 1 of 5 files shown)


megamek/src/megamek/common/AmmoType.java
```diff
@@ -11609,97 +11604,6 @@ private static AmmoType createAR10WhiteSharkAmmo() {
                     .setReintroductionFactions(F_FS,F_LC);
                 return ammo;
             }
-
-            private static AmmoType createAR10BarracudaTAmmo() {
```
NickAragua: Where did these deleted methods go? And why?

mkerensky: Per the TM tables, the AR10 is not supposed to be able to use telemissiles, only standard versions.

mkerensky: Per IO, it can fire peacemaker and santa anna nukes, which is why those are in.

PR #724, opened 2017-10-27
Asew missiles

mkerensky:
+Feature: Adds ASEW missile functionality
+Fix: Adds cost for a number of bombs that had 0.
Enables Alamo bombs for selection (functionality is in PR#723)

### DIFF
--- a/megamek/src/megamek/common/Warship.java
+++ b/megamek/src/megamek/common/Warship.java
@@ -47,13 +47,25 @@ public class Warship extends Jumpship {
     }
     
     //ASEW Missile Effects, per location
+    //Values correspond to Locations, as seen above: NOS,FLS,FRS,AFT,ALS,ARS,LBS,RBS
     private int asewAffectedTurns[] = { 0, 0, 0, 0, 0, 0, 0, 0};
     
+    /*
+     * Sets the number of rounds a specified firing arc is affected by an ASEW missile
+     * @param arc - integer representing the desired firing arc
+     * @param turns - integer specifying the number of end phases that the effects last through
+     * Technically, about 1.5 turns elapse per the rules for ASEW missiles in TO
+     * Because Warships have 8 arcs instead of 6, this overrides the method in Jumpship
+     */
     @Override
     public void setASEWAffected(int arc, int turns) {
         asewAffectedTurns[arc] = turns;
     }
     
+    /*
+     * Returns the number of rounds a specified firing arc is affected by an ASEW missile
+     * @param arc - integer representing the desired firing arc
+     */
     @Override
     public int getASEWAffected(int arc) {
         return asewAffectedTurns[arc];
--- a/megamek/src/megamek/common/Warship.java
+++ b/megamek/src/megamek/common/Warship.java
@@ -46,6 +46,20 @@ public class Warship extends Jumpship {
         damThresh = new int[] { 0, 0, 0, 0, 0, 0, 0, 0 };
     }
     
+    //ASEW Missile Effects, per location
+    private int asewAffectedTurns[] = { 0, 0, 0, 0, 0, 0, 0, 0};
+    
+    @Override
+    public void setASEWAffected(int arc, int turns) {
+        asewAffectedTurns[arc] = turns;
+    }
+    
+    @Override
+    public int getASEWAffected(int arc) {
+        return asewAffectedTurns[arc];
+    }
+ 
+    
     @Override
     public double getStrategicFuelUse() {
     	double tonsperday = 0;
--- a/megamek/src/megamek/common/actions/WeaponAttackAction.java
+++ b/megamek/src/megamek/common/actions/WeaponAttackAction.java
@@ -311,6 +311,12 @@ public class WeaponAttackAction extends AbstractAttackAction implements Serializ
             return (new ToHitData(TargetRoll.AUTOMATIC_FAIL, "Cannot launch bomb in mech mode."));
         }
         
+        // ASEW Missiles cannot be launched in an atmosphere
+        if ((wtype.getAmmoType() == AmmoType.T_ASEW_MISSILE)
+                && !game.getBoard().inSpace()) {
+            return (new ToHitData(TargetRoll.AUTOMATIC_FAIL, "Cannot launch ASEW missile in an atmosphere."));
+        }
+        
         Targetable swarmSecondaryTarget = target;
         Targetable swarmPrimaryTarget = oldTarget;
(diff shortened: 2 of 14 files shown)


megamek/src/megamek/common/Warship.java
```diff
@@ -46,6 +46,20 @@ public Warship() {
         damThresh = new int[] { 0, 0, 0, 0, 0, 0, 0, 0 };
     }
     
+    //ASEW Missile Effects, per location
+    private int asewAffectedTurns[] = { 0, 0, 0, 0, 0, 0, 0, 0};
+    
+    @Override
+    public void setASEWAffected(int arc, int turns) {
```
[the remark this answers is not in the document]
NickAragua: The comment should probably include a list of what number maps to what arc, here and for the other "set/getASEWAffected" methods.

megamek/src/megamek/common/actions/WeaponAttackAction.java
```diff
@@ -760,6 +766,31 @@ else if ((te != null) && (ae.getSwarmTargetId() == te.getId())) {
...
+            int loc = weapon.getLocation();
+            if (j.getASEWAffected(loc) > 0) {
+                toHit.addModifier(4, "weapon arc affected by ASEW missile");
+            } 
+        } else {
+            if (ae.getASEWAffected() > 0) {
+                toHit.addModifier(4, "attacker affected by ASEW missile");
+            }
+        }
+
```
NickAragua: This method is huge and could stand to be broken up a little, but that's probably beyond the scope of this PR.

megamek/src/megamek/common/weapons/ASEWMissileWeaponHandler.java
```diff
@@ -0,0 +1,140 @@
...
+     */
+    private static final long serialVersionUID = 6359291710822171023L;
+
+    /**
+     * @param t
+     * @param w
+     * @param g
+     * @param s
+     */
+    public ASEWMissileWeaponHandler(ToHitData t, WeaponAttackAction w, IGame g,
```
NickAragua: Let's actually fill out the comment.

megamek/src/megamek/server/Server.java
```diff
@@ -1881,6 +1881,37 @@ private void resetEntityPhase(IGame.Phase phase) {
...
+    
+    /*
+     *  Called during the end phase. Checks each entity for ASEW effects counters and decrements them by 1 if > 0
+     */
+    
+    public void decrementASEWTurns() {
+        for (Iterator<Entity> e = game.getEntities(); e.hasNext(); ) {
+            final Entity entity = e.next();
+            // Decrement ASEW effects
+            if (entity instanceof Dropship) {
```
NickAragua: The Dropship and Jumpship code seems identical. Why not coalesce it?

Also, you can check entity.getEntityType() & ETYPE_JUMPSHIP instead of using reflection to improve performance a bit.

megamek/src/megamek/common/weapons/ASEWMissileWeaponHandler.java
```diff
@@ -0,0 +1,140 @@
...
+        hit.setGeneralDamageType(generalDamageType);
+        hit.setCapital(wtype.isCapital());
+        hit.setBoxCars(roll == 12);
+        hit.setCapMisCritMod(getCapMisMod());
+        hit.setFirstHit(firstHit);
+        hit.setAttackerId(getAttackerId());
+        if (weapon.isWeaponGroup()) {
+            hit.setSingleAV(attackValue);
+        }
+        Report r = new Report(3405);
```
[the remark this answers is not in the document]
neoancient: These magic numbers are ids for report messages. The text for the messages is found is i18n/megamek/common/report-messages.properties. This is how report messages are generated throughout the code. The report class is well-documented and provides the explanation through the javadoc comment of this constructor. Providing named constants for the thousand or so possible server messages would be an added complexity with no real benefit.

megamek/src/megamek/common/weapons/ASEWMissileWeaponHandler.java
mkerensky: So umm, I inherited this from the parent handler, and I went back through all of its parents and found that the comment is never defined more than this. I don't know how it should be filled out...

megamek/src/megamek/common/actions/WeaponAttackAction.java
mkerensky: Yeah. I'm still relatively green, so I try not to change the structure of the code unless I'm working on something specific enough that I know exactly what the consequences will be.

megamek/src/megamek/server/Server.java
mkerensky: I separate dropship and jumpship Because the closest common ancestor that Dropship and Jumpship have in common is Aero, and Aero (or Small Craft) are not supposed to have the get/set ASEWAffected(loc) methods.    I thought that if I did something like "if ((entity instanceof Dropship) || (entity instanceof Jumpship)) {
Dropship d = (Dropship) entity;
Jumpship j = (Jumpship) entity;     that I would get a class cast exception

mkerensky: I do not know how to use the entity.getEntityType() & ETYPE_JUMPSHIP  in Server- it gives me errors

mkerensky: Also, if I did that, will the ETYPE still apply to the Warship and Space Station children of Jumpship?

megamek/src/megamek/common/weapons/ASEWMissileWeaponHandler.java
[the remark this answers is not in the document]
NickAragua: Fair enough, let's leave it then.

megamek/src/megamek/server/Server.java
NickAragua: In that case, I think if you want to check for jumpship and all its child types, you probably want entity.getEntityType() & ETYPE_JUMPSHIP != 0, rather than entity.getEntityType() & ETYPE_JUMPSHIP == ETYPE_JUMPSHIP. I'm pretty sure ETYPE_WARSHIP is something like 0010001 (not the exact value, but something like it), while ETYPE_JUMPSHIP is just 0010000. Basically, with >= 0 you're seeing if a particular group of bits is set ignoring any other bits. With == ETYPE_WHATEVER, you're seeing if only that group of bits is set and nothing else.

mkerensky:
I think that I have fixed all requested changes at this time.

PR #730, opened 2017-10-30
Updates the symlink in README.md

wildj79:
The symlink for `README.md` was broken when the `docs/` folder was reorganized.  This fixes the symlink.


HammerGS:
Thanks for the fix.

Issue #205, opened 2016-05-19
 Artillery on dropships needs "Ground Dropships Use Individual Weapons" option (Airborne dropship unable to fire arrow IV aritllery)


Question2:
Megamek says : Airborne aerospace units can't fire non-Arrow-IV artillery

But the unit is an airborne dropship attempting to fire arrow IV artillery in an artillery bay...

[#battletech.zip](https://github.com/MegaMek/megamek/files/271882/battletech.zip)

pheonixstorm:
And what rule says it can still fire? Source and page.

Question2:
That's not the point, the error message implies it can since it specifies non-arrow IV artillery.

Arrow IV rules are in tac ops btw.

BLOODWOLF333:
p180 TacOps:  ....A unit mounting an Arrow IV (no other artillery weapons may be used to make an airborne attack) may make an artillery attack against ground hexes while airborne; direct-fire airborne artillery attacks cannot be made. When making such an attack, follow all the standard rules for a unit on the ground making an artillery attack, with the following additions. 

.....attacks from Altitude 10 or higher cannot be made.

mkerensky:
I was under the impression that airborne units could only use Arrow IV 'bombs' under that rule (though I don't remember why I thought this). Maybe it's the artillery bay triggering the error, not the Arrow IVs it contains...

djmbm:
I don't recall the exact rules, but there are definitely rules about airborne artillery. However, that aside, there is also a known issue with the ArtilleryBayWeaponHandler being broken. This is why we say you must use individual dropship weapons when landed option.

arlith:
I loaded up the saved game, and this is coming down to ArtilleryBays.  Ultimately, this is a duplicate of [SF [#3766]](https://sourceforge.net/p/megamek/bugs/3766/).

HammerGS:
Reopening this is include some information recently receiving from a CGL source.

When asked about how rules would handle this in TT the reply was:

If 4 Long Toms combine to form a bay then damage for the bay would be calculated they these for 20 point damage weapons. So 8 Cap Scale Damage. In situations where a bay attack is used that would be the value.

But if used on a Ground Map treat each weapon as an individual weapon purposes for attack.  So the Bay becomes 4 Long Toms each making its own attack roll.

For reference this bug is the same as here:
https://sourceforge.net/p/megamek/bugs/3766/

I've changed the title here to reference the SF bug and will be closing the SF bug.

[...]

PR #734, opened 2017-11-01
Airborne Aero Artillery Fire (Bug #205)

mkerensky:
This fixes bug #205 , correcting a number of issues within the Artillery Bay Indirect Handler that threw NPEs and some that prevented artillery weapons from being fired by airborne units.

Note that per errata, Artillery weapons should not be grouped in bays. This fix works properly when separate artillery bays are created for each artillery weapon and its ammunition. It will waste ammo and only resolve a single shot impact if used if used with a bay of multiple weapons. Pending work from Hammer and Neoancient will change the unit files to resolve this.

### DIFF
--- a/megamek/src/megamek/common/weapons/ArtilleryBayWeaponIndirectFireHandler.java
+++ b/megamek/src/megamek/common/weapons/ArtilleryBayWeaponIndirectFireHandler.java
@@ -149,10 +149,6 @@ public class ArtilleryBayWeaponIndirectFireHandler extends AmmoBayWeaponHandler 
             return true;
         }
         Mounted ammoUsed = ae.getEquipment(aaa.getAmmoId());
-        //TODO: Not sure why atype is defined the way it is here. 
-        //I'm removing most of the references to it in order to handle bay ammo.
-        final AmmoType atype = ammoUsed == null ? null : (AmmoType) ammoUsed
-                .getType();
         // Are there any valid spotters?
         if ((null != spottersBefore) && !isFlak) {
             // fetch possible spotters now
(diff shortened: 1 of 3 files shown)


megamek/src/megamek/common/weapons/ArtilleryBayWeaponIndirectFireHandler.java
```diff
@@ -86,6 +86,33 @@ public boolean cares(IGame.Phase phase) {
...
+    @Override
+    protected void useAmmo() {
+        for (int wId : weapon.getBayWeapons()) {
+            Mounted bayW = ae.getEquipment(wId);
+            // check the currently loaded ammo
+            Mounted bayWAmmo = bayW.getLinked();
+
+            if (bayWAmmo == null) {// Can't happen. w/o legal ammo, the weapon
+                // *shouldn't* fire.
+                System.out.println("Handler can't find any ammo!  Oh no!");
```
NickAragua: You will want to log this instead of using System.out.println

mkerensky: This is a spot of code I found and repurposed where logging hasn't been changed yet. What's the new format supposed to be?  Is it the Try {   Catch ?  I've seen a few of those and I think I could change to use that.

NickAragua: There should be a member variable called "game", which probably has a logger instance. You don't need a try-catch block or anything like that. Wildj79 provided some good examples in the contributors channel just a few minutes ago: https://megamek-public.slack.com/files/U2FKHG8EL/F7UG5SZD0/log4j_example.java

NickAragua:
WeaponAttackAction really needs to have its text tokenized for translation purposes, but that can be handled at another time.

PR #749, opened 2017-11-13
Added client GUI option to display a mini armor status visual in the boardview mouse-over tooltip.

Qwertronix:
A small QoL hack that adds "gradient status bars", which I call here collectively the mini armor visualization, to the mouse-over tooltip on units in the board view. It is entirely optional and defaults to off (like weapons listing in tooltips) and the color and special characters used are configurable in the advanced client settings. 

Since I'm a beginner with Java, this is the best I could make in a reasonable amount of time that gives a quick glance at which units are damaged and how badly. 

wip_at-a-glance_armor-tooltip
wip_at-a-glance_armor-tooltip_2
wip_at-a-glance_armor-tooltip_3

### DIFF
--- a/megamek/src/megamek/client/ui/swing/boardview/EntitySprite.java
+++ b/megamek/src/megamek/client/ui/swing/boardview/EntitySprite.java
@@ -25,6 +25,7 @@ import megamek.common.Entity;
 import megamek.common.EntityMovementType;
 import megamek.common.GunEmplacement;
 import megamek.common.IAero;
+import megamek.common.IArmorState;
 import megamek.common.IBoard;
 import megamek.common.IGame;
 import megamek.common.IGame.Phase;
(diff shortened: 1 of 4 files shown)


BLOODWOLF333:
Well, is it possible to make it smaller?  Like the text and the bars?  I could see this being like Mechwarrior 2 or 3 where they are solid green bars kinda like what we have over the unit sprites for each location that then turns yellow/red/black with damage or just fills up with red like in MW3.

image

image

If my idea is a retarded amount of hard work then never mind.

arlith:
I think it would look a lot better if you could align the start of each bar.  There are probably tons of different ways to do this, but the one that springs to mind for me would be to make another HTML table.  Define a table that has two columns, and put the hit location abbreviation in the first column, and then the armor bar in the second column.

wildj79:
@arlith recommendation should work.  It is tabular data ;)

Qwertronix:
Yeah it should be possible to adjust the size, for sure, and the color is already adjustable easily. The styling seems done through a basic HTML renderer. OTOH I'm not sure how fancy you can get with the tooltip in terms of rendering, as my first idea was to use some fancy-dandy SVG tags to draw minified paperdoll like in the Armor tab, and put color in there. Either I was doing it really wrong every time, or the tooltip renderer doesn't support advanced functions like that (which would make sense). One of the regular devs would probably have to take a look and weigh in on that front. 

Just in terms of size, though, that's definitely something that could be tweaked.

Qwertronix:
Now an attempt with bright green/yellow/red color scheme and size -2: 

wip_at-a-glance_armor-tooltip_5

BLOODWOLF333:
What is the white text-stuff on the bars?  But, that is so much better and more professional looking.
Do you plan to display the rear torso armor as well?  Hard to tell on github but bars would prob have to be twice as thin to fit the rear in.
slack

Qwertronix:
Basically the current bars display everything in blocks of 10 units. The strange-looking stuff (▓) is a special "dark shade" unicode character that here represents Internal Structure. Rear armor is placed on the left of the IS if there's any. So when you have:

LT: █▓▓▓███

Then that's a 'Mech that has 1-10 LTR armor, 21-30 LT internal structure, and 21-30 LT front armor.

megamek/src/megamek/client/ui/swing/boardview/EntitySprite.java
```diff
@@ -780,6 +781,91 @@ else if (numAdv > 1)
         addToTT("ArmorInternals", BR, entity.getTotalArmor(),
                 entity.getTotalInternal());
 
+        // Build a "status bar" visual representation of each
+        // component of the unit using block element characters.
+        if (GUIPreferences.getInstance().getBoolean(GUIPreferences.SHOW_ARMOR_MINIVIS_TT)) {
```
NickAragua: Code-wise, this mostly looks fine. The only change I would recommend is splitting this functionality out to a separate method.

arlith:
With the red/yellow/green colors, it's super hard to tell what's IS in your screenshot.

Qwertronix:
Yeah it's really hard to tell. I still left the default color scheme to what is shown in the first screenshots, but will keep the table and -2 size as default. 

I'll just need to add a BR/NOBR condition for the next tooltip item (since that's what's causing the extra line, as tables automatically create a newline when closed in HTML4 from what I've found), then move out most of the code into its own method. I'll update the PR with the changes once done!

MGM175:
This idea looks very good to me, especially in colored variant, but I think it should use an much darker colors due to light background in tooltip balloon.

And yes, I've used HTAL system in MW4, so I'm quite familiar with it :)

Qwertronix:
@MGM175 The good news is that you'll be able to set whichever colors you want on them in the View menu -> Client Settings -> Advanced tab. Standard RGB format scaled from 0 to 255 like everything else in those advanced options.

@arlith @NickAragua @BLOODWOLF333  Updated to include your suggestions to format it as a table, split up the Rear armor into its own bar when present for clarity, and make the whole thing smaller by default (configurable ofc). 

Image of how it looks as a table the current defaults for reference: 

image

Qwertronix:
Welp, I still need to push another commit with a small tweak to remove that extra line after the table, and I also messed up the code for completely destroyed part (it ends up generating a new row with the part name and "destroyed" symbol for each 10 pts worth of initial armor right now... oops!)

Qwertronix:
That should be it, I think. Tested this and it works as intended on my end. Due to variability in the font used depending on the user's selected MekHQ Theme (and/or possibly the MM skin?) the exact visual end result can vary. 

For instance with MekHQ's `Metal` theme and MM defaultSkin (not BW's version) I get these horizontally wider block characters: 
image

To be honest, I personally find it charming that the look and feel of the visualization changes when you, erm, change the overall look and feel of the application, so I'm fine with this.

PR #754, opened 2017-11-17
Implements Caseless Ammo

HammerGS:
Implements Caseless Ammo from Tactical Operations.  The handler is basically copied from the Rapid Fire handler and modified.  

There has been additional code help from @NickAragua


HammerGS:
No such thing as a Caseless Autocannon.  Basically once a AC is set to fire Caseless at the start of the game it can only fire Caseless Ammo in game. That is coded.

MGM175:
Ok, then why those nonexistent weapon is in Sarna.net?
http://www.sarna.net/wiki/Caseless_AC/20

Qwertronix:
I believe TacOps also confirms that ACs have to be declared as Caseless once purchased/installed and cannot ever be reloaded with Cased ammo.

HammerGS:
The Caseless AC was a weapon developed for Tactical Handbook, then replaced as Caseless Ammunition when Tactical Operations was written.

From Sarna - _"As this weapon was not referenced in later publications they should not be considered tournament legal and border on being apocryphal.[1]"_

From pg 352 TO- 
_Finally, because the feed mechanisms for caseless ammo replace those of standard autocannons, a weapon designated as firing caseless ammo cannot draw from any other alternative ammunition types, even if they would otherwise be compatible with the autocannon._

The rules in TO are Game specific not Campaign specific. So the wording "cannot draw" is saying in game they can't switch ammo types.

Qwertronix:
> The rules in TO are Game specific not Campaign specific. So the wording "cannot draw" is saying in game they can't switch ammo types.

Exactly. So the current implementation matches RAW perfectly.

HammerGS:
Got some retooling to do.

PR #755, opened 2017-11-18
RISC Hyperlaser

HammerGS:
Implements the RISC Hyperlaser from Int Ops. pg 93


megamek/src/megamek/common/verifier/TestMech.java
```diff
@@ -1449,6 +1452,22 @@ public boolean hasIllegalEquipmentCombinations(StringBuffer buff) {
             }
         }
         
+        if (mech.hasWorkingWeapon(WeaponType.F_HYPER)) {
+            switch (mech.hasEngine() ? mech.getEngine().getEngineType() : Engine.NONE) {
```
NickAragua: Since FUSION is the only legitimate engine type, I would recommend just checking for that instead, rather than all of the other engine types (case Engine.FUSION: break, default: illegal)

megamek/src/megamek/common/verifier/TestMech.java
```diff
@@ -1438,6 +1438,9 @@ public boolean hasIllegalEquipmentCombinations(StringBuffer buff) {
         if (mech.hasWorkingWeapon(WeaponType.F_TASER)) {
             switch (mech.hasEngine() ? mech.getEngine().getEngineType() : Engine.NONE) {
                 case Engine.FISSION:
+                case Engine.BATTERY:
```
NickAragua: I'm pretty sure my comment on line 1456 applies here as well.

megamek/src/megamek/common/verifier/TestMech.java
neoancient: Checking for a fusion engine is common enough that there's a convenience method for it.
`if (mech.hasWorkingWeapon(WeaponType.F_HYPER) && !(mech.hasEngine() && mech.getEngine().isFusion()) {
  //yada yada
}

NickAragua:
One last change (substitute that engine check in the taser with the same one you did for the hyperlaser), and we'll be good to go.

PR #757, opened 2017-11-19
Case less ammunition

HammerGS:
Implements Caseless Ammunition from TO pg 352.


NickAragua:
Looks fine to me, but I'm not sure how much use there is in me reviewing my own code.

HammerGS:
Reviews complete and merging in.

PR #760, opened 2017-11-20
XCT infantry

pheonixstorm:
XCT Infantry are trained to fight in hostile environments. As such they do not have the same penalties normal infantry with the same equipment would.

In MM terms that means no movement penalty in extreme weather. The other environments that would have a significant benefit are Tainted/Toxic Atmosphere as well as Hazardous Liquid Pools. None of these are currently coded in MM

For XCT troops to be effective you will need to make sure they have the proper armor kits. These are detailed on p350 of TacOps.

### DIFF
--- a/megamek/src/megamek/common/MiscType.java
+++ b/megamek/src/megamek/common/MiscType.java
@@ -3653,7 +3653,7 @@ public class MiscType extends EquipmentType {
         misc.setInternalName(misc.name);
         misc.addLookupName("SpacesuitCombat");
         misc.damageDivisor = 1.0;
-        misc.subType = S_ENCUMBERING | S_SPACE_SUIT | S_XCT_VACUUM |S_COLD_WEATHER;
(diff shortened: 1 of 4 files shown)


megamek/src/megamek/common/MiscType.java
```diff
@@ -3644,7 +3653,7 @@ public static MiscType createSpacesuitCombatInfArmor() {
         misc.setInternalName(misc.name);
         misc.addLookupName("SpacesuitCombat");
         misc.damageDivisor = 1.0;
-        misc.subType = S_ENCUMBERING | S_SPACE_SUIT;
+        misc.subType = S_ENCUMBERING | S_SPACE_SUIT | S_XCT_VACUUM |S_COLD_WEATHER;
```
NickAragua: Hate to nitpick, but |S_COLD_WEATHER should have a space between the | and the S.

pheonixstorm: That's a tiny nit to pick lol

Qwertronix: "It was itchy okay?!"

PR #762, opened 2017-11-21
Fix for some issues with simultaneous phases

arlith:
There's been a long standing issue with Princess and simultaneous phases where too many `CalculateBotTurn` threads will get spawned.  These threads get spawned based on receiving a `GameTurnChangeEvent`, which happens every time the server processes a turn.  In sequential play, the client is able to check the current turn and not act when it's another players turn.  However, simultaneous play makes it so every turn can be acted upon.

What this leads to is Princess spawning a `CalculateBotTurn` thread every time any player ends a turn.  The most frequent result of this is a bunch of invalid attack/physical/deploy packets, as Princess tries to tell the server to move a unit that's already been used.

This situation can also lead to issues with a human players Client as well.  There can be situations where players client could select an already done unit.  Say Client 1 sends a done packet for a unit, while the server is already processing the turn for another unit.  The server caches the client's packet and will issue a turn change when processing the first packet.  Client 1 will then receive this done packet, before the server has responded with the result of the units movement, so Client 1's game has not yet marked that unit as done, nor updated the current turn.  This can lead to the client selecting the same unit again, which when moved will result in an invalid turn packet.

Both of these cases can be fixed by having the client only respond to `GameTurnChangeEvent`'s  that are created by the client's player.  The server has been updated to pass this information along, and I updated both `Client` and `BotClient` to utilize this information.


arlith:
This is related to issue #644, and I will consider that bug fixed once this is merged in.

megamek/src/megamek/client/bot/BotClient.java
```diff
@@ -112,24 +119,40 @@ public void gamePlayerChat(GamePlayerChatEvent e) {
...
                     sendUnloadStranded(getStrandedEntities());
                 }
             }
 
+            @Override
+            public void gamePhaseChange(GamePhaseChangeEvent e) {
+                calculatedTurnThisPhase = false;
+                if (e.getOldPhase().isPhaseSimultaneous(game)) {
+                    int numOwnedEntities = game.getEntitiesOwnedBy(getLocalPlayer());
+                    System.out.println("BotClient calculated turns, " + getName() + " phase " + e.getOldPhase()
```
NickAragua: Do we want to use that fancy pants new logging mechanism here?

arlith: Yes, but `BotClient` doesn't use it from what I could tell.  I wanted to keep this PR tight on one specific issue.

As an aside, in looking at the logs with the new logging, it looks like it adds a lot of whitespace.   It seems like lines logged from server went from being 1 line to 3.  It looks like there's a linebreak between the "status" block of the log message, the actual message, and then a blank line after that.  I'm not a fan of this, but it seems like it's part of the logging library.  I think maybe there's a logging properties setting somewhere to change this?  I didn't have much time to look into this.

arlith: I also think we need to adjust how we're using the logging.  I think `MMLogger` should have some static helper functions, like `logDebug` and `logWarn`, which get used with a passed-in logger.  Right now, classes that use the logging define these methods, but they're defined locally so they will end up getting repeated everywhere.

NickAragua: Ok, let's leave it for now and get this merged, the rest looks fine.

wildj79: @arlith you can set the message format in the `log4j.xml` config file. There is a laundry list of options you can use, and you can pretty well format the message any way you like.

PR #772, opened 2017-12-02
Make tooltips disappear when mouse is over the StatusDisplay / buttons

Qwertronix:
Fixed one of my most frequent irritants with the MM interface. That's all.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/StatusBarPhaseDisplay.java
+++ b/megamek/src/megamek/client/ui/swing/StatusBarPhaseDisplay.java
@@ -23,6 +23,8 @@ import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.awt.event.KeyEvent;
 import java.awt.event.KeyListener;
+import java.awt.event.MouseListener;
+import java.awt.event.MouseEvent;
 import java.util.ArrayList;
 import java.util.Comparator;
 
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/client/ui/swing/StatusBarPhaseDisplay.java
```diff
@@ -52,14 +54,49 @@
...
+
+    }
+
+    @Override
+    public void mouseExited(MouseEvent e) {
+
+    }
+
+    @Override
+    public void removeAllListeners() {
```
NickAragua: Is this particular one necessary?

Qwertronix: It's part of the MouseListener interface and gives me compile errors if it's not implemented. Is there a better alternate implementation in Java?

NickAragua: Just that you don't override it in the other class and it seems like it might have some other unintended side effects. Maybe call super.removeAllListeners() inside?

Qwertronix: Huh. I might have things mixed up then. I'll try removing it and the overrides again and give it another go.

PR #776, opened 2017-12-05
Nc3 fcs tons fix

mkerensky:
Per the Naval C3 construction rules on TO p332, any FCS tonnage for having more than (12) weapons on a given facing is doubled if a Naval C3 is installed. This is in addition to the 1% of the ship base weight of the system.  This adds the mm side to TestSmallCraft.java

### DIFF
--- a/megamek/src/megamek/common/verifier/TestSmallCraft.java
+++ b/megamek/src/megamek/common/verifier/TestSmallCraft.java
@@ -376,8 +376,8 @@ public class TestSmallCraft extends TestAero {
     }
     
     //Determines if we have a Naval C3 installed. Needed for FCS tonnage calculations.
-    protected boolean hasNC3() {
-        for (Mounted m : smallCraft.getEquipment()) {
+    public static boolean hasNC3(SmallCraft sc) {
+        for (Mounted m : sc.getEquipment()) {
             if (m.getType().hasFlag(MiscType.F_NAVAL_C3)) {
                 return true;
             } 
@@ -446,7 +446,7 @@ public class TestSmallCraft extends TestAero {
     @Override
     public double getWeightMisc() {
         double weight = 0.0;
-        boolean NC3 = hasNC3();
+        boolean NC3 = hasNC3(smallCraft);
         // Add in extra fire control system weight for exceeding base slot limit
         for (double extra : extraSlotCost(smallCraft)) {
             //Naval C3 doubles the weight of FCS, if present
--- a/megamek/src/megamek/common/verifier/TestSmallCraft.java
+++ b/megamek/src/megamek/common/verifier/TestSmallCraft.java
@@ -374,7 +374,17 @@ public class TestSmallCraft extends TestAero {
     public boolean isSmallCraft() {
         return true;
     }
-
+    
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/verifier/TestSmallCraft.java
```diff
@@ -436,9 +446,15 @@ public double getWeightAmmo() {
     @Override
     public double getWeightMisc() {
         double weight = 0.0;
+        boolean NC3 = hasNC3(smallCraft);
         // Add in extra fire control system weight for exceeding base slot limit
         for (double extra : extraSlotCost(smallCraft)) {
-            weight += extra;
+            //Naval C3 doubles the weight of FCS, if present
+            if (NC3 == true) {
```
neoancient: This is already a boolean value and there is no reason to add an extra comparison operation. Simply `if (NC3)`

megamek/src/megamek/common/verifier/TestSmallCraft.java
```diff
@@ -374,7 +374,17 @@ public boolean isAero() {
     public boolean isSmallCraft() {
         return true;
     }
-
+    
+    //Determines if we have a Naval C3 installed. Needed for FCS tonnage calculations.
+    public static boolean hasNC3(SmallCraft sc) {
+        for (Mounted m : sc.getEquipment()) {
```
neoancient: Rather than iterating through weapons and ammo along with MiscType, you can just look through the MiscType list with `Entity.getMisc()`. You can also replace this whole construction with `sc.hasWorkingMisc(MiscType.F_NAVAL_C3)`. It adds the extra step of checking whether the equipment is operational, but this is not a place where the very slight performance hit will matter and validation is only done on newly loaded units. There are other places in the validation code where the same assumption is made.

megamek/src/megamek/common/verifier/TestSmallCraft.java
neoancient: Rather than doubling it during final weight calculations it should be doubled when the extra tonnage per arc is calculated. This way requires it to be doubled everywhere it is used which makes for more fragile code.

mkerensky:
Changes made and tested.

megamek/src/megamek/common/verifier/TestSmallCraft.java
```diff
@@ -374,7 +378,7 @@ public boolean isAero() {
     public boolean isSmallCraft() {
         return true;
     }
-
+
```
neoancient: Remove extraneous spaces

megamek/src/megamek/common/verifier/TestSmallCraft.java
```diff
@@ -438,7 +442,7 @@ public double getWeightMisc() {
         double weight = 0.0;
         // Add in extra fire control system weight for exceeding base slot limit
         for (double extra : extraSlotCost(smallCraft)) {
-            weight += extra;
+            weight += extra;
```
neoancient: Remove trailing spaces.

mkerensky: Done.

PR #778, opened 2017-12-07
Ams bay heat fix

mkerensky:
My original implementation of Advanced Point Defense for large craft wasn't able to account for the heat generated by weapon attack declarations when checking to see if point defenses could fire without overheating the vessel.  This corrects that issue.


megamek/src/megamek/common/weapons/ArtilleryBayWeaponIndirectHomingHandler.java
```diff
@@ -474,6 +475,8 @@ protected int calcCounterAV () {
         Entity entityTarget = (Entity) target;
         // any AMS bay attacks by the target?
         ArrayList<Mounted> lCounters = waa.getCounterEquipment();
+        //We need to know how much heat has been assigned to offensive weapons fire by the defender this round
+        int weaponHeat = getLargeCraftHeat(entityTarget) + entityTarget.heatBuildup;
```
NickAragua: It might just be me, but it looks like the contents of this calcCounterAV() method across all 8 source files in this code review are just about identical. Is there a way to consolidate this logic? And maybe not implement getLargeCraftHeat() 8 different times?

mkerensky: I didn't find any when I wrote calcCounterAV() originally - I think there are 4 versions that are different in subtle ways, either for reporting or null checks, and since bay weapons and normal weapons follow a different tree, there have to be 8 copies.  I tried to consolidate getLargeCraftHeat at a higher level, but found it was returning 0 when it shouldn't when I did that

NickAragua: I just did a diff on ArtilleryBayWeaponIndirectHomingHandler and ArtilleryWeaponIndirectHomingHandler's calcCounterAV and getLargeCraftHeat methods and only found one diff, which is the entity null check at the top of calcCounterAV. Diffs with other methods do show some more substantial differences, so consolidating the whole method may not be feasible. However, you're still duplicating a lot of logic (how many times do you invoke the code that checks if a weapon has 'heat as dice'?)

In this situation, I would suggest creating several smaller-scope helper methods to consolidate the functionality that is identical. We want to avoid getting into a situation where someone needs to make changes to this code later on and winds up either a) having to change 8 different files or b) missing some of these 8 different files and creating bugs.

megamek/src/megamek/common/weapons/ArtilleryBayWeaponIndirectHomingHandler.java
```diff
@@ -607,6 +622,33 @@ protected int getCounterAV() {
         return CounterAV;
     }
     
+    /*
+     * Returns the heat generated by a large craft's weapons fire declarations during the round
+     * Used to determine whether point defenses can engage.
+     * @Param ae - the entity you wish to get heat data from
+     */
+    @Override
+    protected int getLargeCraftHeat(Entity e) {
```
NickAragua: Splitting the getLargeCraftHeat() discussion from the other one, it looks like you have 7 identical implementations of this method and one different one for the WeaponHandler class. So, you definitely need to consolidate these implementations.

Also, I recently discovered that there's a convenience method in Entity called "hasETypeFlag" or something similar, that takes an int (which should be one of the ETYPE_ flags) so you don't have to fool around with manual bit comparisons.

PR #791, opened 2017-12-12
Add missing load NC3 from MUL function

mkerensky:
Oops.  One more piece that was supposed to be part of NC3 and missed. I really think this is it at this point...


HammerGS:
Merged

Issue #362, opened 2016-10-12
Board Tooltip following cursor


BLOODWOLF333:
Tooltip behavior is unwanted/odd.  @Ralgith and @arlith are aware, this is just a reminder.  If you hover your cursor over a hex long enough to get the tooltip window to pop up and then move your mouse down to click fire or some other button the tooltip window follows your mouse down to the button panel and stays open for whatever the players delay time is set at and just covers up the map and gets in the way.  Other programs you would expect the tooltip to close on mouse movement or at least once you move the mouse away from the hex once your done with it.
```ralgith [1:34 PM]  
I’ve had it follow my mouse __off__ the MM window
[1:34]  
Which… is very wrong.

bloodwolf [1:35 PM]  
i thought so
[1:35]  
my thinking says it should close the tooltip as soon as my mouse moves away from the hex, im done with it so close

ralgith [1:35 PM]  
It should close as soon as there is any mouse movement… like… oh… every other program?```

Dylan-M:
Please add the details anyway, we forget things.

BLOODWOLF333:
So did you fix this @Qwertronix ?

Qwertronix:
This one I did: https://github.com/MegaMek/megamek/pull/772 

So that should be a fix for the part where it stays stuck open when going down to the buttons, but I didn't touch the code that keeps moving and refreshing the tooltip when moving across the board.

arlith:
Fix for this in PR #795.

NickAragua:
I believe we can call this fixed.

PR #795, opened 2017-12-16
Fix for Issue #362

arlith:
I fixed the issue by keeping track of the last mouse position when the tooltip text was updated.  If the distance between the current and last position is large enough, the tooltip is instead disabled.  This makes it so stopping the mouse on the board view will display the tooltip and small motions will move the tooltip but large motions will reset/hide it.


megamek/i18n/megamek/client/messages.properties
```diff
@@ -747,6 +747,8 @@ CommonSettingsDialog.title=Client Settings
 CommonSettingsDialog.tooltipDelay=Tooltip popup delay:
 CommonSettingsDialog.tooltipDismissDelay=Tooltip dismiss delay:
 CommonSettingsDialog.tooltipDismissDelayTooltip:Sets the delay before a tooltip is dismised.  -1 uses the default for your look and feel.
+CommonSettingsDialog.tooltipDistSuppressionTooltip=<html>To prevent tooltips from following the cursor on the board, the tooltip will be supressed if the mouse moves more than this distance between updates. <br>The default value is the diagonal of a single</html>
```
NickAragua: the diagonal of a single what?

NickAragua:
Looks good to me, just missing some text at the end of the long tooltip token.

megamek/i18n/megamek/client/messages.properties
arlith: The answer is "hex", although that's not strictly accurate.  It's really the diagonal of the hex's bounding box, which is longer, but I don't know that the distinction is super important.  Although the diagonal of a hex is a curious concept.

PR #796, opened 2017-12-17
Board editor terrain list display name

arlith:
This makes it so the terrain `JListBox` uses "displayable" names instead of the terrainType:lvl:exit format.  This makes things more human readable, but has the effect of hiding some of the gory details (although right now the old format is displayed as a tooltip).

I'm not yet convinced this is a good approach without further updates to the map editor.  It may hide information that's still kind of useful, even if it's a pain to have to understand the internal mechanisms.

It would be nice if the "level" text box could be changed to an editable combo-box, so things with actual levels like woods you could just select from the drop down but still edit it for things like building CF.  However, we don't have this information encoded yet to my knowledge.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/BoardEditor.java
+++ b/megamek/src/megamek/client/ui/swing/BoardEditor.java
@@ -42,7 +42,10 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 
 import javax.imageio.ImageIO;
 import javax.swing.Box;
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/client/ui/swing/BoardEditor.java
```diff
@@ -137,12 +199,19 @@ public Component getListCellRendererComponent(JList list, Object value, int inde
...
+                list.setToolTipText(terrainTypes.get(index).getTooltip());
+            }
             return comp;
         }
 
         public void setTerrains(TerrainHelper[] terrains) {
             this.terrains = terrains;
         }
+        
+        public void setTerrainTypess(List<TerrainTypeHelper> terrainTypes) {
```
NickAragua: I hope we haven't been infiltrated by snake people.

Dylan-M:
Since this has been pulled into the much larger PR by @SJuliez, I'm going to close this one.

PR #798, opened 2017-12-18
Added ability to total bay personnel

pheonixstorm:
Needed for CamOps reputation rating.

This is part of the overall fix/completion of the CamOps unit rating (p33)


neoancient:
This method already exists in `Entity`. There is no reason to override it in every aerospace unit class to do the exact same thing.

pheonixstorm:
... Wasn't expecting it there. Ok, no need for this then.

PR #800, opened 2017-12-18
Broke out flee conditions by unit type and allowed aerospace to escap…

DaveLely:
…e the low altitude map

### DIFF
--- a/megamek/src/megamek/common/Aero.java
+++ b/megamek/src/megamek/common/Aero.java
@@ -208,7 +208,7 @@ public class Aero extends Entity implements IAero, IBomber {
 
     @Override
     public boolean canFlee() {
-    	return (super.canFlee() && getAltitude() == 10 && hasMP());
+    	return (underControl() && getAltitude() == 10 && hasMP());
     }
     
     public Aero() {
--- a/megamek/src/megamek/common/Aero.java
+++ b/megamek/src/megamek/common/Aero.java
@@ -206,6 +206,11 @@ public class Aero extends Entity implements IAero, IBomber {
 
     private int eccmRoll = 0;
 
+    @Override
+    public boolean canFlee() {
+    	return (super.canFlee() && getAltitude() == 10 && hasMP());
+    }
+    
     public Aero() {
         super();
         // need to set altitude to something different than entity
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -9061,17 +9061,23 @@ public abstract class Entity extends TurnOrdered implements Transporter,
     	return (getWalkMP() > 0);
     }
     
+    public boolean underControl() {
+    	Coords pos = getPosition();
+        return (pos != null)
+           && !isProne()
+           && !isStuck()
+           && !isShutDown()
+           && !getCrew().isUnconscious()
+           ;
+    }
+    
+    
     /**
      * Returns whether an entity can flee.
      */
+    
     public boolean canFlee() {
-    	Coords pos = getPosition();
-            return (pos != null)
-               && !isProne()
-               && !isStuck()
-               && !isShutDown()
-               && !getCrew().isUnconscious()
-               ;
+    	return underControl() && atEdge() && hasMP();
     }
 
     public void setEverSeenByEnemy(boolean b) {
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -9042,21 +9042,36 @@ public abstract class Entity extends TurnOrdered implements Transporter,
         return unitNumber;
     }
 
+
     /**
-     * Returns whether an entity can flee from its current position. Currently
-     * returns true if the entity is on the edge of the board.
+     * Returns whether an entity is on the edge of the board.
+     */
+    public boolean atEdge() {
+    	Coords pos = getPosition();
+        return ((pos.getX() == 0) || (pos.getX() == (game.getBoard().getWidth() - 1))
+                || (pos.getY() == 0) || (pos.getY() == (game.getBoard()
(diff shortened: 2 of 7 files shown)


neoancient:
1. Since nearly all unit types have the same criteria for fleeing, it makes sense to have that in `Entity` as the default behavior. I see no reason to have the same code repeated in multiple subclasses when there is only one unit type that behaves differently.
2. I don't think fleeing is the right mechanism to use to implement this. First, it requires two thrust points to increase altitude, rather than simply having some MP remaining. Second, the rules state that the unit moves to the first atmosphere row of the high altitude map. This is not currently implemented, but likely will be in the next few months. If the game is being played without the high altitude map (always true currently; later by choice) this is the same as fleeing, but with the high altitude map it will trigger moving the unit from one board to another.

megamek/src/megamek/common/Entity.java
```diff
@@ -9042,21 +9042,36 @@ public short getUnitNumber() {
...
     /**
-     * Returns whether an entity can flee from its current position. Currently
-     * returns true if the entity is on the edge of the board.
+     * Returns whether an entity can flee.
      */
     public boolean canFlee() {
-        Coords pos = getPosition();
-        return (pos != null)
-               && ((getWalkMP() > 0) || (this instanceof Infantry))
+    	Coords pos = getPosition();
```
NickAragua: As per neoancient's comment, the default base class implementation should be canFlee() && hasMP() && atEdge(). That way, you don't have to override it for mechs, tanks, etc, just aeros and infantry.

neoancient:
My entire point #2 should be read as my personal opnion with the exception that the unit needs to have two thrust points available.

megamek/src/megamek/common/Aero.java
```diff
@@ -206,6 +206,11 @@
 
     private int eccmRoll = 0;
 
+    @Override
+    public boolean canFlee() {
+    	return (super.canFlee() && getAltitude() == 10 && hasMP());
```
NickAragua: As per neoancient's comment, let's make sure that the Aero has at least 2 MP remaining.

arlith:
One thing to keep in mind is that if an Aero flees, its flight path should still be present.  Unlike ground units, Aeros that flee should still be able to be shot at.  I think this is currently something that "fly-off" doesn't do.

Also, I believe that Aeros currently can flee, it's just really difficult to trigger the conditions, as they need to be on the map edge at the start of movement.

DaveLely:
there are two units that behave differently, and they're both subtractive
cases:  so how do we subtract out the conditions from the base
class.canFlee in the child class.canFlee?  If the base canFlee includes
atEdge and hasMP but the aero canFlee doesn't require atEdge I would have
to do all the checks in the current base canFlee in the Aero canFlee, which
is even more code duplication than we have here.  Ditto for infantry that
doesn't need hasMP.

Also hasMP() already requires 2 MP remaining in the case of aero.  if
walkingMP is 1 then running is 2.  Only if walking is zero can they not
make it.

megamek/src/megamek/common/Entity.java
DaveLely: Both Infantry and Aero behave differently, and they're both subtractive
cases.  Originally I just added the Aero check to the base class evaluation by putting an or clause on the position that said if it was an aero at altitude 10 ignore the edge requirement.  People on slack thought it was better not to clutter Entity and instead put that stuff into the type specific classes, so in order to avoid duplicating the base code in aero and infantry I needed to break it out into components.  I don't see another way to take out the edge and MP requirements for those two cases..

megamek/src/megamek/common/Aero.java
DaveLely: hasMP() already requires 2 MP remaining in the case of aero.  if
walkingMP is 1 then running is 2.  Only if walking is zero can they not
make it.

megamek/src/megamek/common/Entity.java
NickAragua: You can selectively invoke the helper functions you wrote in the base class from the inheriting classes, since you're overriding the base class implementation of canFlee(). The aero logic is purely additive, while, with the infantry, you can omit the call to "hasMP".

megamek/src/megamek/common/Aero.java
NickAragua: Ok, fair enough. Flee generally only works at the beginning of the movement phase and cannot be combined with other orders, so the mere presence of MP should be fine for now.

neoancient:
I didn't even consider the fact that fleeing onky happens at the beginning of the turn. Per TW p. 78 the unit stays on the map until the end of the turn. (N.B.. @arlith's comment above). This implements the rule incorrectly and while it would partially fulfill the concept of the rule I'm wary of presenting the option to the player one way then somewhere down the road changing the way it works rather than doing it correctly the first time.

megamek/src/megamek/common/Entity.java
DaveLely: ah so I add a new helper function that returns the current base class implementation of canFlee, called, say 'underControl' and then make the base class implementation of canFlee underControl and atEdge and hasMP, then override that function in aero as underControl, hasMP and altitude ==10, and the infantry implementation as underControl and atEdge?

megamek/src/megamek/common/Aero.java
DaveLely: I _am_ noodling on how to force a control check if MP = 1 since they have to overthrust, but I need to look up the consequences of a failed control roll in that circumstance.

NickAragua:
We've had some internal discussions about this, and I've sent you the summary over Slack.

sixlettervariables:
Closing as stale, feel free to reopen when up to date with the latest on master.

Issue #799, opened 2017-12-18
0.43.5 MiniMap Double Blind error


SnyderPhilip:
The mini map shows unit Icon indicators for all units detected, even those that have just been spotted by sensors and not IDed visually.  The game map shows correctly the unidentified sensor contact icon.  This error did NOT exist in the 43.5BetaNov11 version.

NickAragua:
Probably an issue with changes I made for not displaying hidden units on the minimap.

NickAragua:
Fixed with PR#801

PR #801, opened 2017-12-19
Fix for issue #799

NickAragua:
Consolidated logic used to determine if a unit should be a sensor blip, fully rendered or not rendered at all.


NickAragua:
Ok, did some more testing, and I'm satisfied that sensor pings now show up on the minimap and board consistently.

PR #809, opened 2017-12-20
Additional maps - various authors.

pheonixstorm:
This is the start of the new additions to the unofficial maps. Considering the number of new additions I have broken these down into small commits for anyone who wants to review or test them.

All of these maps have been validated previously but may get a second pass once I round up all of the old maps that had the issue with incorrect terrain numbering. I will continue adding new sets as I finish renaming them and moving them over. Not sure how many are left, there are still a few duplicates among them as well that need to be sorted out. Readme files includes for those sets that had them.

If anyone wants to do some quality control on these be my guest. I don't have the time for QC at the moment. I will post another comment when this can be merged or if further information is needed.


Dylan-M:
I'm confused about the purpose of this PR... all of my maps and @jayof9s' maps should have been in. I'd like to review and approve it before stable if I can get an understanding. Though @HammerGS is assigned to it and I'll defer to him.

PR #814, opened 2017-12-21
Change guessed firing plan sorting algorithm

NickAragua:
Previously, for whatever reason, the "guess firing plan" method (called when evaluating the utility of a particular move path) was sorting a path's potential firing plans from lowest to highest then *returning the lowest utility firing plan*. 

So, for example, a unit of field gunners would consider not moving, but since some of those firing plans had utility 0, considered the "firing utility" of the zero-move move path to be 0 and opt to move instead, even though they could stand still and fire the field guns.

### DIFF
--- a/megamek/src/megamek/client/bot/princess/BasicPathRanker.java
+++ b/megamek/src/megamek/client/bot/princess/BasicPathRanker.java
@@ -51,8 +51,6 @@ import megamek.common.TargetRoll;
 import megamek.common.Targetable;
 import megamek.common.Terrains;
 import megamek.common.TripodMech;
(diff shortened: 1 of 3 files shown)


dericpage:
Looks good to me.  

I was going to suggest that we make FiringPlan and RankedPath both implement Comparable, but I see that they already do.  Of course, it looks like the compareTo methods could probably stand to be rewritten a bit, but that's not a big deal.  

Still, it might be quicker to put the heatPlans into an ArrayList and sort that rather than iterating through an array, but that would also require rewriting calcFiringPlansUnderHeat and that's probably not worth it this close to the freeze.

PR #847, opened 2018-01-04
bulletproofing turn order switching

NickAragua:
Fixes a bug where the client freezes if there are stranded entities to unload.

### DIFF
--- a/megamek/src/megamek/server/Server.java
+++ b/megamek/src/megamek/server/Server.java
@@ -2268,7 +2268,7 @@ public class Server implements Runnable {
         boolean turnsChanged = false;
         boolean outOfOrder = false;
         GameTurn turn = game.getTurn();
-        if (game.isPhaseSimultaneous() && (entityUsed != null)
+        if (game.isPhaseSimultaneous()
             && (entityUsed != null)
             && !turn.isValid(entityUsed.getOwnerId(), game)) {
             // turn played out of order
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/server/Server.java
```diff
@@ -2268,7 +2268,7 @@ private void endCurrentTurn(Entity entityUsed) {
         boolean turnsChanged = false;
         boolean outOfOrder = false;
         GameTurn turn = game.getTurn();
-        if (game.isPhaseSimultaneous() && (entityUsed != null)
+        if (game.isPhaseSimultaneous()
```
NickAragua: This might look like I removed a null check, but since it was being done twice in a row...

megamek/src/megamek/server/Server.java
```diff
@@ -2452,9 +2452,9 @@ else if (((infMoved && infMoveMulti)
 
         // move along
         if (outOfOrder) {
-            send(createTurnIndexPacket(entityUsed.getOwnerId()));
+            send(createTurnIndexPacket(playerId));
```
NickAragua: playerId is earlier set to entityUsed.getOwnerId(), except it checks for entityUsed == null and sets playerId to -1 if it is.

megamek/src/megamek/server/Server.java
```diff
@@ -3444,7 +3444,7 @@ private void changeToNextTurn(int prevPlayerId) {
         if (prevPlayerId != -1) {
             send(createTurnIndexPacket(prevPlayerId));
         } else {
-            send(createTurnIndexPacket(player.getId()));
+            send(createTurnIndexPacket(player != null ? player.getId() : IPlayer.PLAYER_NONE));
```
NickAragua: "unload stranded entities" turn doesn't happen for a specific player, it happens for every player, so the turn's player ID is -1.

PR #851, opened 2018-01-08
Change Default KeyBinds

BLOODWOLF333:
Several times we have seen posts on the forums of people accidentally pressing Y and turning the unit labels off which tells them what units have moved already.  So, I am proposing a new default keybinds.xml that is close to the keybinds from mechwarrior 2 and 3.  Toggling ISO view, unit labels, field of fire, anything like that is done while holding ctrl + first letter of what it is called like F for field of fire overlay so they are less likely to accidentally happen.  Fire a weapon is Space, next weapon is Enter, etc.  I think a lot of players have played the video games so this should be more intuitive for them in my opinion.

### DIFF
--- a/megamek/mmconf/defaultKeyBinds.xml
+++ b/megamek/mmconf/defaultKeyBinds.xml
@@ -1,312 +1,297 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- 
-  This is the default key binding for MegaMek.  
-
-   Key code numeric values are defined in awt.event.KeyEvent.
-   See: http://docs.oracle.com/javase/7/docs/api/constant-values.html#java.awt.event.KeyEvent.CHAR_UNDEFINED
-
-   The modifier numeric value is a mask that determines what modifier keys are pressed (ie; ctrl, shift, etc). 
-   These are defined at the same webpage, however they should be combined using bitwise ors.
-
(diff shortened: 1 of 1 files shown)


scJazz2:
right now most of the commands are under the left hand which is very convenient for right handers using the mouse.

HammerGS:
I think their needs to a conversation before changing the default key binds.  We have players just getting used to the existing key binds.  Changing them could create confusion.

Not against it, just saying needs some conversation.

arlith:
Personally, I think that the keybinds should be designed around WASD.  That's what most people are going to be familiar with.  WASD should scroll, and then operations should flow out from there.  All of the prev/next operations are implemented around that: Q/E and Z/C along with with ctrl/shift modifiers are used for things like prev/next target, weapon, unit, etc.  It's where some of the ancillary options start to get hairy (like turn on/off unit labels or isometric).

Ultimately, it ends up being an issue of visibility.  You could argue that, if people don't know these keybinds exist, and they don't accidentally discover them, they may never know they exist.  Maybe it's not clear as it could be that keybinds can be edited?  It seems somewhat natural to me to think that someone presses some keys and something happens and now end up in bad state, then they look at the list of key bindings and figure out what key they pressed to get in the bad state and fix it.

BLOODWOLF333:
Changed back to the default keys but added CTRL to only the binds that toggled overlays and unit labels.

Issue #864, opened 2018-01-14
SRMs Deal Zero Damage To Aero


BLOODWOLF333:
### Environment
MegaMek 0.43.7-betaJan2 Standalone
Java version 1.8.0_144
Platform Windows 8.1 6.3 (amd64)

### Description
Not sure if this applies to LRMs and other missile types but, definitely does to SRMs.  Enemy ASF is at altitude 4, unit display states it's range is 9, needs an 8+ TH for the Gargoyles SRMs, both hit and both did zero damage which left me asking why?

![image](https://user-images.githubusercontent.com/16950832/34913070-b0207876-f8b8-11e7-9218-de15fa6e5108.png)

![image](https://user-images.githubusercontent.com/16950832/34913073-b7a59d2e-f8b8-11e7-8083-481376e8d152.png)

### Files
Player Name:  BLOODWOLF
[GargoylevsAero.sav.gz](https://github.com/MegaMek/megamek/files/1629069/GargoylevsAero.sav.gz)
[megameklog.txt](https://github.com/MegaMek/megamek/files/1629070/megameklog.txt)

neoancient:
What's going on here is that MM is using ground combat rules to create the attack action but resolving the damage against any airborne unit as aerospace combat. Under aerospace rules the SRM has a maximum range of short, which is 0-6 hexes. At a greater distance the SRM has an attack value of zero, which is why it's doing no damage. In looking at the code history this issue appears to have been present a long time, possibly since the inclusion of aerospace combat in MM. Fixing it without breaking aero use of AMS may be tricky; I haven't looked deeply enough at this point to know.

mkerensky:
You can assign this one to me. I don't know that I can fix it in time for the RC, but I will try.

mkerensky:
Fixed in PR #868 .  It wasn't hard and didn't have any negative effect on aero AMS.

neoancient:
The problem is not that the range needs to be fudged. MM incorrectly uses the aerospace damage rules based on whether the target is airborne when it should be the attacker. Damage from a ground-to-air attack should be resolved the same way a ground-to-ground attack is. See the example on TW, p. 247 in which a UAC uses the cluster table and streak SRMs are allocated by missile instead of a total damage value in 5-point clusters.

Dylan-M:
@mkerensky Are you still going to fix this, including @neoancient's feedback?

mkerensky:
[...]

PR #868, opened 2018-01-17
Fix for Bug #864

mkerensky:
CalcAttackValue in WeaponHandler returns the shortAV if the attacker is not airborne.  I had forgotten to include this in the CalcAttackValue override in MissileWeaponHandler.

There are many other overrides for other missile types, but they all have at least a mediumAV (range of 12) and don't have this problem when shot at strafing aeros.

This fix does not break AMS in anyway.


neoancient:
This just fudges something that is implemented incorrectly in the first place. A ground-to-air attack should not be using AV in the first place, but should resolve an SRM attack by rolling on the cluster hits table and apply the damage in 2-point groupings. The problem is that MM uses the target to determine the damage method when it should use the attacker.

PR #893, opened 2018-01-31
Add check to ignore SI damage for squadron status display

mkerensky:
Fixes something that's been bugging me for a while - squadrons show as crippled/heavy damage/light damage, etc  based on how many fighters are in the squadron because the number of fighters is counting as the squadron's SI.

Squadrons now ignore the SI calculation for their damage status display, but damage to individual fighters is still tracked and applied properly on disband/unloading from the squadron.


Dylan-M:
Okay, so, looking at this again, I was mistaken in my understanding. Though, this does render the getInternalRemainingPercent() method in FighterSquadron unused, does it not?

PR #908, opened 2018-02-18
Weapons bay ammo display/MUL loading fix

NickAragua:
This is a change that prevents the following behavior:

- Experimental tech level and year 3150
- User adds an Overlord A3 dropship
- User goes to configure -> equipment, looks at an AR-10 launcher, drops the two default warheads lower and adds some tele-operated ones in addition to the regular ones. clicks ok
- User goes back to configure -> equipment, now there are two ammo rows for the AR-10 launcher, one with just the regular warheads, one with the tele-operated ones.

This still preserves the behavior where clan weapons can't select IS ammo by setting the row's "tech base" to that of the weapon being displayed.

The changes to MULParser and EntityFileList prevent the following behavior:

- Unit adds Overlord A3
- User configures equipment, changes AR-10 launcher to have non-default ammo load out with at least three ammo types
- Save unit list to MUL file
- Load unit list from MUL file, the AR-10 launcher only has the first two ammo types loaded.

### DIFF
--- a/megamek/src/megamek/common/EntityListFile.java
+++ b/megamek/src/megamek/common/EntityListFile.java
@@ -312,13 +312,10 @@ public class EntityListFile {
                         haveSlot = true;
                     }
 
-                    
                     // Record ammunition slots in undestroyed locations.
                     // N.B. the slot CAN\"T be damaged at this point.
-                    // If the ammo is mounted in a bay, it will be handled differently
                     else if (!isDestroyed && (mount != null)
-                            && (mount.getType() instanceof AmmoType) 
-                            && (mount.getEntity().getBayByAmmo(mount) == null)) {
+                            && (mount.getType() instanceof AmmoType)) {
                         thisLoc.append(indentStr(indentLvl+1) + "<slot index=\"");
                         thisLoc.append(String.valueOf(loop + 1));
                         thisLoc.append("\" type=\"");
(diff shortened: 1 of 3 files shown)


megamek/src/megamek/common/EntityListFile.java
```diff
@@ -268,6 +274,13 @@ public static String getLocString(Entity entity, int indentLvl) {
                     if (CriticalSlot.TYPE_EQUIPMENT == slot.getType()) {
                         mount = slot.getMount();
                     }
+                    
+                    // if the "equipment" is a weapons bay, 
+                    // then let's make a note of it
+                    if(entity.usesWeaponBays() && 
+                            slot.getMount().getBayAmmo() != null) {
```
neoancient: `Mounted` initializes `bayAmmo` to an empty vector. `getBayAmmo()` is never going to return `null`, so this map is going to contain every `Mounted` slot up to the current one. I can't think of a way to test for a weapon bay other than `slot.getMount() instanceof BayWeapon`.

This also relies on the bay critical slot always coming before the ammo. This may be a valid assumption, but I'm not certain of that.

NickAragua: Probably checking getBayAmmo().length ought to do it then. I'll look at the possibility of a bay coming after the ammo tonight. Seems unlikely from looking at some random blk files, but worth double checking.

NickAragua: I added a weapon before a bay in a dropship .blk file and it actually auto-generates a bay in the previous slot if there isn't one already.

PR #910, opened 2018-02-24
Resolve Keybind Conflicts

BLOODWOLF333:
Changed a few xml keybinds that conflicted with the hardcoded menu hotkeys like the round report for example.


BLOODWOLF333:
why is this gone now in the current version?

PR #916, opened 2018-03-06
Make MUL files work for tracking ammo on GunEmplacements

Dylan-M:
This makes MUL files work for GunEmplacements in a manner similar to how Protomechs were handled.


Dylan-M:
This is to fix the bug in MHQ where GunEmplacements don't reload their changed ammo on loading a campaign save.

PR #918, opened 2018-03-08
BoardEditor GUI

### LINKED ISSUE #920: Map Editor - Selecting Terrain Features
### Environment
0.43.8-RC2
Windows7
Most recent java8

### Description
When you have terrain types in the Terrain window, you can no longer select them (single or double click) and have the dropdown below update so you can easily change.

I.e. I copy a building from the map and want to change the elevation, CF, and building type. I now have to dig into the drop down for each to make updates. In the past I used to just be able to click on Building and the drop down would automatically change to 'Building' and fill in the current value.

Currently if I double click, it will carry the current value down but leave whatever the previous selection is for the dropdown.

See the example screenshot. I clicked 'Building' in the window and it added the value to "Arms".

why

SJuliez:
Tried to improve the BoardEditor's GUI. 
Fixes #920 

Note: arlith's pull request concerning BoardEditor display names is included here.

Also, I added restrictions - valid are:
- snow:1 and snow:2 (thin and deep snow)
- mud:1 and mud:2
- pavement:1 or higher
- tundra:1 

I changed boards to reflect this.


seem8:
Data Thu, 08 Mar 2018 08:19:11 -0800
SJuliez <notifications@github.com> napisał(a):
>   * Undo/Redo functions for the Map Editor

No kidding! Thank You very much for this.

-- 
Łukasz Posadowski

megamek/src/megamek/client/ui/swing/BoardEditor.java
```diff
@@ -136,69 +226,111 @@ public Component getListCellRendererComponent(JList list, Object value, int inde
...
+                list.setToolTipText(terrainTypes.get(index).getTooltip());
+            }
             return comp;
         }
 
         public void setTerrains(TerrainHelper[] terrains) {
             this.terrains = terrains;
         }
+        
+        public void setTerrainTypess(List<TerrainTypeHelper> terrainTypes) {
```
Dylan-M: This needs fixed, there's too many 's' here... `setTerrainTypess` should be `setTerrainTypes`

SJuliez: Is corrected.

SJuliez:
Not sure if I should be doing something; I corrected that method name.

Dylan-M:
@SJuliez I need to re-review it, and I didn't have time this morning before pushing out the latest RC due to how large this change is. Though, I guess the actual code changes aren't overwhelming :)

arlith:
I haven't really reviewed the code, but I tried this out.  This isn't a definitive list by any means, but a couple of things I noticed...

It seems like the control panel isn't in a scroll box.  It seems larger than my screen, and there's no way to see the UI widgets that are off the edge (all of the map control buttons are there).

Alt-click is also a bad option.  I know we used it before, but in Linux (at least for the Gnome-based DEs I've used), alt-click is "drag window", and this will consume the event before it gets to any running application.

I feel like the UI layout for the new control bar could use some work.  It feels very cluttered.  The preset buttons are nice, but they are huge and take up a ton of space and probably aren't really all that relevant most of the time.  A lot of the buttons seem really huge, larger than they need to be.

The exits button has a broken tooltip.  Also, it might be good to have a hex image in the center of the exit dialog.

The undo/redo feature is great, and something that should have been added ages ago.

arlith:
One idea would be to fit the preset buttons into a slightly shorter scroll pane, so the real estate taken up by the preset buttons isn't as larger.  Then you could also make it so users could specify their own presets.  Really fancy would be to allow you to click and drag the buttons to reorder the presets.

HammerGS:
What is the status on this guy?  Could we pull it into the next Dev release?

NickAragua:
I'll take a look tonight. I believe the requested changes were very minor, so I don't expect to see a problem with it.

Windchild292:
Will be picking this up to finish the review fixes and get it merged in

Issue #859, opened 2018-01-12
Game Options Tooltip Exception


BLOODWOLF333:
### Environment
MegaMek 0.43.7-betaJan2 Standalone
Java version 1.8.0_144
Platform Windows 8.1 6.3 (amd64)

### Description
I dont notice anything broken or wrong but the logfile fills up with exceptions that seem to correspond to each time I hover over a game rule so that the tooltip pops up and I can read what it does.

> Exception in thread "AWT-EventQueue-0" java.lang.ClassCastException: megamek.client.ui.swing.ClientGUI$3 cannot be cast to javax.swing.JComponent
	at javax.swing.ToolTipManager$AccessibilityKeyListener.keyPressed(Unknown Source)

### Files
Occurs in megamek lobby
[megameklog.txt](https://github.com/MegaMek/megamek/files/1624912/megameklog.txt)

BLOODWOLF333:
Confirmed during a game tonight that other players also receive this error in their logs and seems to occur anytime any tooltip is created/or does not get created.  Seemed like on my side at least while hovering over units that it did not want to give me the tooltip pop-up but, thought it was just lag.

PR #921, opened 2018-03-09
Fixes boardview tooltip NPEs

SJuliez:
`null` tooltips were generating NPEs.


BLOODWOLF333:
Were you getting the same errors that I was in #859 still?

SJuliez:
I think these are different ones. They're from uuuuh.... javax.TooltipSomething.InsideTimerEvent and came in through the tooltip fix that stopped the board tooltip from following the mouse around all the time. (And in #859 it seems they were ClassCastExceptions, here it's NPEs).

BLOODWOLF333:
You should be on our slack http://megamek.org:3000/

PR #932, opened 2018-03-12
Add Handle override to CapitalMissileBayHandler

mkerensky:
Add separate Handle method to CapitalMissileBayHandler to try and fix the Glancing/Direct blow reporting issue when AMS fire increases the ToHit.  

Works in space without obviously breaking things.  Needs extensive testing on ground maps and with grounded dropships to ensure damage to buildings/targets in buildings isn't broken.


Dylan-M:
I don't know that Capital Missiles can be fired on ground maps, so I'm okay with that for now ;)

I mean, Orbital Bombardment isn't in yet, and I wouldn't wast CapMis shots for it anyway! Naval Lasers, ACs, and Gauss FTW.

Issue #936, opened 2018-03-15
0.43.9-RC3 Princess crashes


mhjacks:
Fill out the fields below with as much pertinent information as possible to help the developers solve the issue.
Search for your issue inside open and closed issues first before reporting so as not to make duplicates.

### Environment
What version of MegaMek does your issue apply to?  Note whether it is the stand alone MegaMek or HQ's MegaMek.
mekhq's megamek from 0.43.9-RC3
What operating system are you using?  Windows, Mac, or Linux? Linux (Fedora 27)
What java version are you using?  If unsure it will state your java version at the top of the megameklog.txt file. 1.8.0_161

### Description
Describe the issue and provide screenshots ingame if able.
Game becomes unresponsive after last fire declared for a turn.  NPE reported in logfile.

Provide the rulebook, page number, and text if you feel like this is a rules issue.
And provide the steps you have taken to reproduce the issue in the saved game if applicable.
Played two different games with different forces that included artillery, tag and c3
[autosave.sav.gz](https://github.com/MegaMek/megamek/files/1813777/autosave.sav.gz)
[megameklog.txt](https://github.com/MegaMek/megamek/files/1813778/megameklog.txt)

### Files
Provide a saved game of when the issue was found.  If possible try to save the game during the phase before the issue occurs so the 
developers can step through the code and find exactly what it is that is breaking in the code.
Provide the megameklog.txt file for any NPE's or errors that MegaMek might be reporting.
Do not paste the entire logfile text into this ticket. Just attach the actual megameklog.txt file
And provide your playername so that they can load the saved game.

HammerGS:
Various NPE's from the Logs.
First Up. (Almost too Numerous to count)
```
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
	at javax.swing.ToolTipManager$insideTimerAction.actionPerformed(ToolTipManager.java:664)
	at javax.swing.Timer.fireActionPerformed(Timer.java:313)
	at javax.swing.Timer$DoPostEvent.run(Timer.java:245)
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:756)
	at java.awt.EventQueue.access$500(EventQueue.java:97)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.awt.EventQueue$3.run(EventQueue.java:703)
	at java.security.AccessController.doPrivileged(Native Method)
[...]

PR #938, opened 2018-03-15
Fix for tooltip NPEs, #936

SJuliez:
Suggestion for a fix for #936 (the tooltip NPEs part).

This uses the dismissal delay to get rid of the tooltip. The TT does not always vanish at once; when the mouse is moved really fast the tooltip stays visible for a few frames but it mostly works. No NPEs.

I tried the other few settings of the ToolTipManager but none of them seemed useful.

I also checked the suppression distances and it's really difficult to move the mouse without getting 1, 2, 4 as the first distances (in that order). Therefore I'd recommend less than 5 as the default suppression distance. For me 0 works best.


Dylan-M:
Fixes #936

PR #947, opened 2018-03-22
Fix for #936 - check if an entity has been moved before trying to move it

NickAragua:
Just what the title says.

### DIFF
--- a/megamek/src/megamek/client/bot/princess/Princess.java
+++ b/megamek/src/megamek/client/bot/princess/Princess.java
@@ -712,6 +712,12 @@ public class Princess extends BotClient {
         final StringBuilder msg = new StringBuilder("Deciding who to move next.");
         for (final Entity entity : myEntities) {
             msg.append("\n\tUnit ").append(entity.getDisplayName());
+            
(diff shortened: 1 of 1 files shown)


arlith:
Did you determine what the added `&& !getGame().isPhaseSimultaneous()` was for?  Why not just remove  that?

NickAragua:
The test case where I was able to reproduce the bot sending illegal orders for an already-moved unit was when a mech was destroyed and produced an ejected mechwarrior. Before I added the "isDone" check, the bot was repeatedly attempting to move that mechwarrior.

Attempting to remove the line got me "entity is null" errors in the bot thread. It's probably due to the bot attempting to move a destroyed unit. Which it shouldn't do, since the "isValidEntity()" method includes a "isSelectableThisTurn", which includes a check for the entity being destroyed and null? Now that I think about it, it's also kind of a mystery why the "isDone()" check earlier prevents the illegal orders, since it's also part of "isSelectableThisTurn()". 

Not really sure I like how mysterious this whole thing is, but, for a bandaid, it works fine.

PR #950, opened 2018-03-23
Fixes #764: Mount List Too Long

### LINKED ISSUE #764: Mount List Too Long
My 'mount' list in MM lobby is too long, but there's no scroll bar to scroll down.

image

Dylan-M:
This was a simple matter of moving the existing utility class for handling this from MHQ over to MM and making use of it. I'm sure there are more places it could be made use, but not as critically since they're far less likely to have enormous lists.


NickAragua:
Would the file change list contain the moved mekhq file?

Dylan-M:
I may have accidentally pushed that direct to master

PR #952, opened 2018-03-24
Ground to air fix (fix for #864)

mkerensky:
Changes behavior of ground-to-air fire so that attacks use the cluster hits table and ground ranges rather than converting to AV and inflicting 0 damage if range > short. 

Also allows aero AMS to engage these attacks and cleans up the getAMSHitsMod method to the same standards as CalcCounterAV. 

In further testing, I found and resolved a number of bugs affecting reporting/AMS/Damage resolution and its interaction with the optional rules "Grounded Dropships use individual weapons"  and "Single fighters not capital"  I have resolved these and I'm satisfied that this doesn't break any previously working functionality around aero combat in space/atmo. 

Probably needs a new review, though.

### DIFF
--- a/megamek/src/megamek/common/weapons/MissileWeaponHandler.java
+++ b/megamek/src/megamek/common/weapons/MissileWeaponHandler.java
@@ -593,8 +593,7 @@ public class MissileWeaponHandler extends AmmoWeaponHandler {
                         apdsEngaged = true;
                     }
                 }
-                
-                // Determine modifier and report
+                //Determine APDS mod
                 if (apdsEngaged) {
                     int dist = target.getPosition().distance(
                             pdEnt.getPosition());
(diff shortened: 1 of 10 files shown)


megamek/src/megamek/common/weapons/MissileWeaponHandler.java
```diff
@@ -485,55 +486,121 @@ protected int getAMSHitsMod(Vector<Report> vPhaseReport) {
...
+                                counter.getLocation(), false))
+                        // shutdown means no AMS
+                        || pdEnt.isShutDown()) {
+                    continue;
+                }
+                
+                //If we're an AMSBay, heat and ammo must be calculated differently
+                if (isAMSBay) {
+                    for (int wId : counter.getBayWeapons()) {
+                        Mounted bayW = entityTarget.getEquipment(wId);
```
NickAragua: I seem to recall some kind of function you wrote for a previous PR that handles AMS activity. Is there a way to reuse any of that code?

mkerensky: No, though I did use it as a reference for streamlining this method.  I had to add AMSBays as a possible counter for standard cluster table missile attacks, and the code here breaks the bay down and assigns 1 AMS in it to each attack rather than combining the entire bay - which is incompatible with the damage and heat code for bays in calcCounterAV().

mkerensky:
Don't merge this quite yet. I found some weirdness that impacts grounded/atmospheric dropships that I'm trying to fix.

PR #957, opened 2018-04-01
Atlas images

SJuliez:
Bundles around 1000 hex tileset images into a few Atlas images. Also added reference images in data/images/hexes/_Reference_Images/

Should not introduce changes that are visible in MM except for yellow lines that I added to the runway fluff (I think these are guidance lines, they are present on all the big airports I googlemapped)


arlith:
I had initially decided against committing image atlases to the repository.  My reasoning was that I thought it made the images easier to work with.  For maps, this probably isn't the case.  If you're editing an image file, it's likely just as easy to work with the atlas.  However, for units there is the question of creating the mechset file, which is more tedious with the atlas (much nicer to let a program do that).

One thing to note, how does this interact with the image atlasing program?  It should ignore and not atlas images that aren't the proper size, but it will spit out a warning.  Might be good to add some code to ignore atlas files, to prevent a list of "known" warnings.

SJuliez:
First I have to correct myself: I did add the roads with cars as road:3 and SirMegaV's more saturated roads as road:4 (except for the bw_atmospheric tileset where the road:x are numbered differently)

As to the atlas images: I guess it's even easier to work with them than with single images; and GIMP or the like allow to restrict operations to a selected area anyway. I would add the layered versions of the files but I don't know a good format for that (the native format of Greenfish Icon Editor Pro is probably not a good choice...).

I was not sure about the atlassing program (it seems ready to use but it hasn't been used?). In any case, it seemed necessary to sort the directories first because images/hexes really is a big jumble. And even sorted it seems the images should be put into atlasses in a tidied-up way, not just all-in, to make it possible to have practical reference images (in plain words, we want to be able to interact with the stuff in the BoardEditor). The same isnt true for the unit images. Therefore my suggestion would be to use the atlassing program for images/units.

arlith:
The atlassing program is definitely used; it's part of the packaging scripts.  It gets called to atlas up all of the images that go into the releases.  The releases only ship with image atlases.  The code repo doesn't have atlases, but as part of the release process all images get atlassed automatically.

SJuliez:
Ah ok, I didn't know that. Now I think I also understand your earlier comment better. Then it would be interesting if the atlasses and atlassing interfere somehow. And you'll have to decide if atlasses are wanted at all in the code repo :-)

Dylan-M:
I personally prefer no atlases in repo since it prevents people who don't understand them from working on the images and we have a utility that will automatically build them anyway.

Arguably, images shouldn't be in the repo at all, since binary files in git are considered bad joo-joo, but meh... that was screwed up way back in the CVS days of this repo (or maybe SVN if it started as SVN, I seem to recall it started as CVS though).

arlith:
My initial preference was for not atlases in the repo, but I actually think it may work out better for hex images.  For instance, it allows you to see all of the road variants in one image.  For things where we have a bunch of rotated variants, it can be helpful to see everything at once, imo.  Plus, it's not like editing texture templates is anything new.

I don't have strong preferences either way though, as I'm generally not the one who works with these files.

HammerGS:
For the record, I don't want to have to deal with Atlases with the unit files or camos, since those can change a lot and I need a easy way to switch them around.

arlith:
I'm in full agreement with that.

SirMegaV:
Actually I was told by Arlith not to use road:3 and road:4 etc because it
was meant for road type. Road:2 is wrong and will be removed eventually.
But that was years ago.

sixlettervariables:
Closing as stale, feel free to reopen when updated to master.

PR #962, opened 2018-04-05
Refactoring HexTileSet

SJuliez:
and a bit of modernization...


arlith:
I skimmed through this and it looks fine.  Nothing really changes functionally, just formatting and removing `Iterator`.

PR #963, opened 2018-04-05
Iso changes

SJuliez:
The draw order of hexes in Iso mode was a bit strange. Of each row of hexes, the slightly higher (= farther back) half should be drawn first, then the slightly lower ones. The elevation doesnt matter.
In this way, the hex gfx can extend upwards out of the hex shape, e.g. for trees or iso buildings (although actually having such gfx drawn at the correct position needs more adjustment).

Also corrects the calculation of hexes to draw given the clipping rectangle "view". Didnt work very well when zooming in or out on big maps.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/boardview/BoardView1.java
+++ b/megamek/src/megamek/client/ui/swing/boardview/BoardView1.java
@@ -2342,12 +2342,15 @@ public class BoardView1 extends JPanel implements IBoardView, Scrollable,
      */
     private void drawHexes(Graphics g, Rectangle view, boolean saveBoardImage) {
         // only update visible hexes
-        int drawX = (int) (view.x / (HEX_WC * scale)) - 1;
-        int drawY = (int) (view.y / (HEX_H * scale)) - 1;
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/client/ui/swing/boardview/BoardView1.java
```diff
@@ -2342,25 +2342,28 @@ private void drawHexes(Graphics g, Rectangle view) {
      */
     private void drawHexes(Graphics g, Rectangle view, boolean saveBoardImage) {
         // only update visible hexes
-        int drawX = (int) (view.x / (HEX_WC * scale)) - 1;
-        int drawY = (int) (view.y / (HEX_H * scale)) - 1;
-
-        int drawWidth = (int) (view.width / (HEX_WC * scale)) + 3;
-        int drawHeight = (int) (view.height / (HEX_H * scale)) + 3;
+        double xs = (int) (HEX_WC*scale);
```
wildj79: if `xs` and `ys` are declared as a `double`, why are you casting them to an `int`?

Issue #387, opened 2016-10-31
ASF Visibility


jayof9s:
In instances with limited visibility (darkness, rain, etc.) ASF visibility is incorrect. They currently cannot target enemies they flew over during the round if they end up too far away from them.

VampireSeraphin:
Unless they have a searchlight.

I've had this problem as well in night missions. Can't target a non illuminated unit.

mkerensky:
I also run into this a lot. I've previously suggested that Aero units get assigned a sensor suite. They all obviously have one, and given the requirements of space combat it's probably much superior to what most ground units have (though perhaps NOT as good at clutter suppression for things like trees and buildings), but yet aeros got left off the sensor tables in TacOps.

jayof9s:
The problem is it doesn't matter where an ASF ends up. They aren't shooting from their final position, they're shooting from where they moved during the round. So if they fly directly over a target, the target is in LOS regardless of conditions. It doesn't matter that they end up 60 hexes away.

It's currently just not functioning correctly.

arlith:
Yea, it's probably a simple bug to fix.  Basically, when computing visibilities it's using the ASF's current hex, and not the hex it would attack from.  Probably, we just need to add an exception for ASF's in the code that's computing visibility.  I think there a lot of minor bugs like this around air-to-ground attacks, because the ASF isn't attacking from `Entity.getPosition`, but virtually all of the code uses `Entity.getPosition` unless specific exceptions have been considered and added...

VampireSeraphin:
I'd believe that. I find new and interesting problems with fighters almost every time.

arlith:
It's not clear to me how this should work in the rules.  It doesn't seem like the visual spotting rules in TO cover how ASF on ground maps work.  I've reached out to Xotl on this issue.  Until I get a response, I'm going to mark this as pending.

pheonixstorm:
has anyone tested this in reverse? How do ground units act when trying to attack asf during these conditions?

VampireSeraphin:
IIRC, from the last time I played this kind of mission, you need to hit the
fighters with your spot light to attack them.

pheonixstorm:
How do the rules say to handle attacks against ASF in the dark, illuminate to attack or are they visible if their flight path is within visual range.

arlith:
The rules don't say, from what I can see.  That's the problem.  There's nothing that indicates to me how this should be handled.

mkerensky:
[...]

PR #964, opened 2018-04-05
Aero Sensor Suites and Visibility in Atmosphere/Low Vis Conditions

mkerensky:
This fix addresses bug #387 

It accounts for the December 2017 errata surrounding aero spotting and sensors, and adds a basic sensor package to all ASFs and CFs.

I still have a bit more testing to do around range reporting vs actual effect, but I wanted to go ahead and put it out here for review in case there's anything major I need to change and retest. 

This testing was completed and it seems to work.  I will admit that I didn't think about the low atmosphere map when writing this, just space and ground (mostly ground).  We found major issues with low atmo during testing last night anyway, so once that's fixed I'll make another PR to integrate this with it.

### DIFF
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -12700,12 +12700,19 @@ public abstract class Entity extends TurnOrdered implements Transporter,
         }
         int bracket = Compute.getSensorBracket(getSensorCheck());
         int range = getActiveSensor().getRangeByBracket();
+        int groundRange = 0;
+        if (getActiveSensor().isBAP()) {
+            groundRange = 2;
+        } else {
+            groundRange = 1;
+        }
         int maxSensorRange = bracket * range;
         int minSensorRange = Math.max((bracket - 1) * range, 0);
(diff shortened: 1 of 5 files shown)


megamek/src/megamek/common/Entity.java
```diff
@@ -12700,10 +12700,24 @@ public String getSensorDesc() {
...
         int minSensorRange = Math.max((bracket - 1) * range, 0);
+        int maxGroundSensorRange = bracket * groundRange;
+        int minGroundSensorRange = Math.max((maxGroundSensorRange - 1), 0);
         if (game.getOptions().booleanOption(OptionsConstants.ADVANCED_INCLUSIVE_SENSOR_RANGE)) {
             minSensorRange = 0;
+            minGroundSensorRange = 0;
+        }
+        if (isAirborne() && game.getBoard().onGround()) {
+            return getActiveSensor().getDisplayName() + " (" + minSensorRange + "-"
+                    + maxSensorRange + ")" + " {Ground Range" + " (" + minGroundSensorRange + "-"
```
NickAragua: The "ground range" text should probably be tokenized.

mkerensky:
Changes made.

PR #967, opened 2018-04-16
Ant bulid.xml fix

dericpage:
Fixed issue with ant build not seeing the localized messages.properties files for certain tasks, resulting in a "NoClassDefFoundError: megamek/common/messages" during the build.


dericpage:
I have not tested the packaging scripts, so I have no idea if the same issue is happening there.

PR #997, opened 2018-06-18
Manual Tele-Operated Missiles v2

mkerensky:
Restores functionality of manually guided telemissiles (removed in 0.43.5), updating them to work with new space combat features and fixing bugs in damage, ammo use, etc.

With the exception of missiles still processing in the turn order as a physical ram attack (not sure how to change this without breaking Server), they now follow the rules outlined on TW p251


mkerensky:
Neither of those things should be possible.   I'm not exactly sure where the fly off the board/return to the board code is (if you know, a pointer would be helpful...).  Telemissiles should not be eligible.

I guess I need to revisit the safe thrust calculation code and have it "getFuel" or something at the start of each round.

mkerensky:
Fixed both.

megamek/src/megamek/client/ui/SharedUtility.java
```diff
@@ -743,7 +744,12 @@ public static MovePath moveAero(MovePath md, Client client) {
                 if (!game.getBoard().contains(md.getLastStep().getPosition())) {
                     md.removeLastStep();
                     if(game.getOptions().booleanOption(OptionsConstants.ADVAERORULES_RETURN_FLYOVER)) {
-                        md.addStep(MoveStepType.RETURN);
+                        //Telemissiles shouldn't get a return option
+                        if (entity instanceof TeleMissile) {
```
NickAragua: This does not appear to solve the problem, as I'm still able to fly missiles off the board and have them return.

I would suggest that you have the missile get removed from the game when it goes off board instead.

NickAragua: Never mind, I needed to update the workspace.

PR #1024, opened 2018-07-06
Adds debug, error, fatal, info, trace, and warning conveince methods …

wildj79:
…to the `MMLogger` interface and `DefaultMmLogger` class.

This is a simple PR that adds some convience methods to the logger class. Also deprecated one of the `log` methods in favor of using an equivalent `error` method.


arlith:
This PR forgot `unittests.megamek.common.logging.FakeLogger`.  There are compile issues with the unit tests now.

arlith:
I fixed this in e106ee, and I thought the commit message would automatically reference this PR, but I guess it didn't.

PR #1025, opened 2018-07-07
Enable Princess to fly spacecraft using the "advanced" flight model

NickAragua:
Features include:
- Fixing a long-standing bug with bot geometry where she was incorrectly evaluating the "closest coordinates" of a movable area.
- Fixing a bot cheat where an out-of-control aircraft could still issue orders and evade
- Oh yeah, and allowing the bot to fly spacecraft on space maps, using the stratops advanced newtonian spaceflight rules. This includes all spacecraft without restriction.

### DIFF
--- a/megamek/src/megamek/client/bot/princess/BasicPathRanker.java
+++ b/megamek/src/megamek/client/bot/princess/BasicPathRanker.java
@@ -96,6 +96,10 @@ public class BasicPathRanker extends PathRanker implements IPathRanker {
     void setPathEnumerator(PathEnumerator pathEnumerator) {
         this.pathEnumerator = pathEnumerator;
     }
+    
+    PathEnumerator getPathEnumerator() {
+        return pathEnumerator;
+    }
 
     Map<Integer, Double> getBestDamageByEnemies() {
         return bestDamageByEnemies;
@@ -375,7 +379,7 @@ public class BasicPathRanker extends PathRanker implements IPathRanker {
 
         EntityEvaluationResponse returnResponse = new EntityEvaluationResponse();
 
-        int distance = calculateDistance(path, enemy.getPosition());                
+        int distance = enemy.getPosition().distance(path.getFinalCoords());                
         
         // How much damage can they do to me?
         double theirDamagePotential = calculateDamagePotential(enemy,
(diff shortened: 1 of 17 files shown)


megamek/src/megamek/client/bot/princess/BasicPathRanker.java
```diff
@@ -647,6 +650,15 @@ protected RankedPath rankPath(MovePath path, IGame game, int maxRange,
 
     }
 
+    /**
+     * Worker function that determines if a given enemy entity should be evaluated as if it has moved.
+     */
+    protected boolean evaluateAsMoved(Entity enemy) {
+        // Aerospace units on ground maps can go pretty much anywhere they want, so it's
+        // somewhat pointless to try to predict their movement.
+        return !enemy.isSelectableThisTurn() || enemy.isImmobile() || enemy.isAero();
```
neoancient: Do you want to do a check for airborne here? Since aerodyne on the ground move like tanks, and spheroids are even more predictable. (This is not a request for a change, just a question.)

megamek/src/megamek/client/bot/princess/Princess.java
```diff
@@ -172,6 +174,8 @@ public LogLevel getVerbosity() {
     IPathRanker getPathRanker(Entity entity) {
         if(entity.hasETypeFlag(Entity.ETYPE_INFANTRY)) {
             return pathRankers.get(PathRankerType.Infantry);
+        } else if (entity.hasETypeFlag(Entity.ETYPE_AERO) && game.useVectorMove()) {
```
neoancient: This will exclude LAMs in fighter mode. Entity.isAero() includes all aeros and potentially LAMs.

megamek/src/megamek/client/bot/princess/Princess.java
```diff
@@ -1724,18 +1733,31 @@ int getHighestEnemyInitiativeId() {
...
     
     /**
      * Helper function that insinuates an "evade" step for aircraft that will not be shooting.
      * @param path The path to process
      */
     private void evadeIfNotFiring(final RankedPath path) {
+        Entity pathEntity = path.getPath().getEntity();
+        
+        // we cannot evade if we are out of control
+        if(pathEntity.isAirborne() && ((Aero) pathEntity).isOutControlTotal()) {
```
neoancient: Tracing back through the code I could not find anything that limits this line to aerospace units. `isAirborne()` returns true for non-aero units executing a combat drop, which will throw a ClassCastException here. In addition to a check for aero, it should also be cast to IAero so it can apply to LAMs in fighter mode.

neoancient:
Threw an NPE when Princess tried to calculate a path close to the edge of the board. It might have had something to do with the fact that my vector would have taken me off the board.
image
[megameklog.txt](https://github.com/MegaMek/megamek/files/2177194/megameklog.txt)
[autosave.sav.gz](https://github.com/MegaMek/megamek/files/2177195/autosave.sav.gz)

megamek/src/megamek/client/bot/princess/BasicPathRanker.java
NickAragua: Yes, it looks like I missed a spot here.

NickAragua:
Ok, I got the NPE, and fixed an issue where the bot couldn't generate paths for a unit if the initial vector took it over the map edge. Also got the places where flying LAMs would cause errors or be ignored.

As a bonus, the bot won't keep re-deploying crippled units that have gone off map if forced withdrawal is on.

Issue #1028, opened 2018-07-09
Infantry TAG has 0 range


ndrwstn:
### Environment
Megamek 0.44.0
MacOS 10.13.5
Java 8 Update 171?

### Description
Using TAG Spotter Infantry Stalwart Support - Spotter Infantry the TAG has 0/0/0 range.

### Files
Save attached (probably unnecessary, but I had it from another bug.)
[targetoutofarc.sav.gz](https://github.com/MegaMek/megamek/files/2177197/targetoutofarc.sav.gz)

PR #1033, opened 2018-07-09
Fix for #1028: Infantry TAG has 0 range

neoancient:
InfantryWeapon overrides getLongRange and getExtremeRange to provide values calculated from the infantry weapon range. This does not work for infantry TAG, which uses standard unit ranges.


neoancient:
My first approach was to set the infantryRange to 3 on the TAG, but that only works because the range bands are all the same size. This approach allows for other weapon types with uneven range bands.

Issue #1027, opened 2018-07-09
Off-board Infantry Field Artillery "Target Out of Arc"


ndrwstn:
### Environment
Megamek 0.44.0
MacOS 10.13.5
Java 8 Update 171?

### Description
Using Infantry Field Artillery (Arrow IV) with Homing ammo. Infantry deployed off-board  East 17 hexes.  Trying to target the south/southwest quadrant of a 32x34 generated map (hex 0119 for example), I can't fire ("Target Out of Arc").

### Files
Save attached.
[targetoutofarc.sav.gz](https://github.com/MegaMek/megamek/files/2177130/targetoutofarc.sav.gz)

PR #1036, opened 2018-07-10
Fix for #1027: Off-board Infantry Field Artillery "Target Out of Arc"

neoancient:
When deploying an entity offboard its position is set as the designated number of hexes from the middle of the indicated edge. When placed east or west, the facing is not directly toward the middle of the map due to the hex constraints. If placed close enough, this can cause part of the play area to be out of the firing arc. The player has no control over the facing, or the ability to rotate turret-mounted weapons during play (which includes field artillery). There is also no mention in the off-board rules in TacOps about checking firing arcs.

Added a line that ignores out of arc computation for offboard attacks.


neoancient:
I think that line exempts air-to-ground fire from the firing arc check unless it's indirect artillery.

PR #1041, opened 2018-07-11
Force generator

neoancient:
Generates campaign to regiment sized forces (larger for some factions) using faction-specific organization rules.

This is the product of work I have been doing on and off for about four years. It was originally written as an independent program during the earlier part of the development period, with the code moved into MegaMek over a year ago, with development continuing there. The RAT generator in MM and the Bloodname generator in MekHQ were both developed to serve as pieces of this project.

It is primarily additive, though some modifications were made to the RAT Generator and some bugs fixed along the way. There were also some changes to RandomArmyDialog to add the tag and alter the layout as needed.

The generated force includes values that are not used by MegaMek, but can be used by MekHQ to assign ranks, bloodnames, and force icons. I have also written some test code that is not included in this branch that can check whether it is being run from a MekHQ installation and if so it will dynamically load the necessary classes to show ranks, Bloodnames, and force icons directly in MM without a circular dependency.


megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -0,0 +1,1696 @@
...
+		attached = new ArrayList<ForceDescriptor>();
+		flags = new HashSet<String>();
+		topLevel = false;
+		element = false;
+		positionIndex = -1;
+		nameIndex = -1;
+		fluffName = null;
+	}
+	
+	public boolean matches(ChassisRecord cRec) {
```
NickAragua: Let's add comments for these "matches" methods, as I'm not entirely sure what they do.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -0,0 +1,1696 @@
...
+            } else {
+                baSubs.get(i).getRoles().add(MissionRole.MECHANIZED_BA);
+            }
+        }
+        generateLance(baSubs);
+        
+        return baseUnitList;
+	}
+	
+	public void generateLance(List<ForceDescriptor> subs) {
```
NickAragua: Self-explanatory, but it's a large method that does a lot, so we may want some documentation on it.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -0,0 +1,1696 @@
...
+					}
+				}
+			}
+			if (ut == null || ut == UnitType.MEK) {
+				baseModel = null;
+			}
+		}
+	}
+	
+	public void setUnit(ModelRecord unit) {
```
NickAragua: This method could probably use some comments. Also, the indentation seems off near the bottom.

PR #1062, opened 2018-07-24
Aerospace unit fly-off/return issue fixes

NickAragua:
This PR does three things:

1) Fix a long-standing bug where, if a unit flew off board but was initially facing in a direction different from the edge it flew off of, it could "warp" to a different edge. (e.g. aircraft starts facing north, flies off south edge, would come back from the north edge for some reason)

2) Implement a CGL ruling stating that "more than 2x velocity" modifiers don't apply to control rolls in space

3) Add logic for the bot to avoid deploying in such a way as to instantly fly off the board.


megamek/src/megamek/server/Server.java
```diff
@@ -6785,8 +6785,9 @@ else if (game.getBoard().onGround()) {
      *                   if it can't return)
      * @return
      */
-    private Vector<Report> processLeaveMap(Entity entity, int facing,
+    private Vector<Report> processLeaveMap(MovePath movePath,
```
neoancient: The documentation for this method still reflects the former signature.

megamek/src/megamek/server/Server.java
```diff
@@ -6785,8 +6785,9 @@ else if (game.getBoard().onGround()) {
...
                                            boolean flewOff, int returnable) {
+        Entity entity = movePath.getEntity();
```
neoancient: I tested fleeing with a ground unit and got this:
Exception in thread "Packet Pump" java.lang.NullPointerException
        at megamek.server.Server.processLeaveMap(Server.java:6846)
        at megamek.server.Server.processMovement(Server.java:6938)
        at megamek.server.Server.receiveMovement(Server.java:5208)
        at megamek.server.Server.handle(Server.java:32004)
        at megamek.server.Server$PacketPump.run(Server.java:368)

It appears that when sending the flee command the movepath is generated without setting the Entity. I didn't look further than this, but it seems to me that the client should probably do that.

NickAragua: The "processMovement" function already receives an entity from the client, so I just have it set the path's entity.

neoancient:
It seems to me that we could probably mark the entity field of MovePath as transient if the Entity always gets sent as part of the movement packet anyway and gets set on the server side. I'm not proposing doing that now because I haven't done a thorough check to make sure there's not a gotcha somewhere.

Issue #1073, opened 2018-08-01
Princess able to manuever out of control atmospheric fighter


ndrwstn:
### Environment
Megamek 0.44.0
MacOS 10.13.5
Java 8 Update 171

### Description
Princess is able to manuever/evade (at least) atmospheric fighters despite being "out of control". Attached game has Heavy Strike Fighter Meteor out of control for the last two turns, but she has been turning and evading.

### Files
[princessnotoutofcontrol.sav.gz](https://github.com/MegaMek/megamek/files/2250957/princessnotoutofcontrol.sav.gz)

NickAragua:
In the future, when providing saved games, please specify the names of the players to save time.

PR #1074, opened 2018-08-02
Fix issues with out-of-control aeros

NickAragua:
Fixes issue #1073, and also forces out-of-control aeros under bot control to actually do the out-of-control move.

Also a minor performance improvement, as the bot will no longer attempt to generate or evaluate potential move paths for out of control aeros on ground/low atmospheric maps.


NickAragua:
Also, got the bot to use out-of-control aeros as init sinks.

PR #1092, opened 2018-08-10
Forcegen fixes

neoancient:
Various fixes for the force generator, particularly in applying unit ratings and experience levels correctly. Also involves some adjustments to skill generation, correcting the logic to matching rules using the ifName predicate, and corrections the Wolf and Cloud Cobra rule sets.


neoancient:
Given how old the original force generator code is, I suspect I had Eclipse on the default settings using tabbed indents at the time, and I can't say whether I ever did a cleanup.

PR #1100, opened 2018-08-15
Fix for wrong bot chase behavior

NickAragua:
This fixes a bot behavior that's been annoying me for quite a while. Normally, bot units under "forced withdrawal" don't fire until fired upon, but, when the bot is under "immediate withdrawal" rules, that goes out the window.

Best way to test this is to give the bot a tank of some kind, yourself a unit with a lot of low damage/clustering weapons (SRMs, small lasers, MGs). Set 'forced withdrawal' to on and give the bot a 'destination edge'. 

The expected behavior is that it'll shoot at you normally. When you cripple it (armor breach, immobilized, etc), it should not shoot at you the following round.

If you shoot or kick it subsequently, it should return fire on the following round.

### DIFF
--- a/megamek/src/megamek/client/bot/princess/Princess.java
+++ b/megamek/src/megamek/client/bot/princess/Princess.java
@@ -1585,6 +1585,11 @@ public class Princess extends BotClient {
      * whichever is the more recent.
      */
     public void refreshCrippledUnits() {
+        // if we're not following 'forced withdrawal' rules, there's no need for this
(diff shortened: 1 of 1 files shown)


NickAragua:
Let's hold off on this until I make the unit tests pass.

NickAragua:
Ok, unit tests are working now, but I'm going to wait until the gradle branch gets merged in to make that PR.

Issue #1103, opened 2018-08-19
Implement/verify support for small basements


giorgiga:
The [BasementType](https://github.com/MegaMek/megamek/blob/7a147e30247d87ba7e80f8d781b1c1a062672ff2/megamek/src/megamek/common/Building.java#L227) enumeration contains a value `ONE_DEEP_NORMALINFONLY` that is intended to represent the "small" basements one gets by rolling  a 9 on the basement table (see TW, p179).

The [Building. rollBasement()](https://github.com/MegaMek/megamek/blob/7a147e30247d87ba7e80f8d781b1c1a062672ff2/megamek/src/megamek/common/Building.java#L531) method responsible for actually rolling basement types does _never_ return `BasementType.ONE_DEEP_NORMALINFONLY` (the value 9 is missing from the if/else chain).

Investigation is needed to verify editing `rollBasement()` is enough to introduce small basements and/or if that causes problems somewhere else.

PR #1105, opened 2018-08-19
Building refactoring

giorgiga:
The goals of this PR are:

1. making the `Building` code (and the closely related) intelligible
2. breaking the circular dependency between `Building` and `Board`

The changes in this PR are split across multiple commits, which hopefully will allow code reviewers to "follow along" as the various changes are applied.

Personally, I would however recommend to just skim trough the overall diff and then carefully read all code in `megamek.common.building` (it shouldn't be too difficult to trace it to parts of `megamek.common.Building`).

# Summary of changes

The functionality in `megamek.common.Building` has been split into multiple files/classes and all have been placed into the new package `megamek.common.building`.

Outside this new package, changes are widespread but miminal in scope (mostly, it's updated imports to point into the new package).

Changes break compatibility with old save files.

# The classes in megamek.common.building

### Building
- Besides deprecated code, `Building` no longer depends on `IBoard` (thus breaking a circular dependency between the two).
- The two server-related methods have been moved to a helper class.
- Hex-specific information is no longer spread across different `Map`s.

### BuildingSection
New class to collect all building info about a specific hex in one place.

### Buildings
New class that collects the code to instantiate a `Building` from a `Board`.

_I don't particularly like the class name, but `BuildingFactory` didn't sound good and `BuildingBuilder` was outright ridiculous._

### BuildingServerHelper
The two server-related methods from `Building` (ie: `collapseBasement` and `rollBasement`) have been moved to this class as a temporary step before the corresponding parts of `Server` are refactored.

### BuildingClass
New enum that replaces constants and methods with hardcoded values originally in `Building`.

### ConstructionType
New enum that replaces constants and methods with hardcoded values originally in `Building`.

### BasementType
Originally an inner class (enum) of `Building`, `BasementType` has been moved to its own file and refactored to make its API resemble that of the other enums in.

### DemolitionCharge
Originally a (for some reason, non-static) inner class in `Building`, it's been moved to its own file but left otherwise untouched.

### BuildingTest and BuildingSectionTest (in /unittests)
Contain (simple) unit test for the corresponding classes. Admittedly, the tests are quite basic at the moment.

# Files changed elsewhere
Besides the changes to imports because of moved classes and those due to the refactoring of BasementType (which, in a lapse of thoughtfulness-less, I refactored via the IDE instead of deprecating the old methods), changes outside `megamek.common.building` are limited to the following.

### build.xml
The compileTests target has been updated to copy all resources from unittest/ into the output folder (it only copied xml files).

### megamek.common.Board
Has been impacted by the removal of `FuelTank` and the switch to factory methods over constructors for `Building` instances.

### megamek.common.BuildingTarget
Has been impacted by the removal of the `UNKNOWN` constant from `Building`.

### megamek.common.IHex
Two new utility (default) methods have been added as specialized versions of the ubiquitous `terrainLevel(int)` (which is most commonly called with a constant as the parameter).

### megamek.common.util.BuildingTemplate
_This class is actually unchanged, as I restored it to its initial state after initially moving it to megamek.common.building._

### megamek.server.Server

Has been impacted by the removal of `FuelTank`.

### megamek.test.TestUtilities (in unittests/)

I needed to share one utility method between the two new test cases... I didn't find anywhere obvious to put it so I made a new class.

# Additional work

To be completely satisfied, I would like to also do the following (either appending to this PR or in a new one later):

1. demolition charges beg to be moved from `Building` to `BuildingSection` (they are deployed in a specific hex, not in a building).
1. `Building` should be further simplified as information like construction type (light, medium, ...) and building class (hangar, fortress, ...) really is per-hex.
1. `BasementType.UNKNOWN` should be removed and `null`/`Optional` used instead .

I guess there's no problem about them but I did reach the main goals of this PR and I want to hold off until I see how this PR is received before putting in more work.

# Possible/suggested follow-up changes

These are additional/related changes that I do recommend, but I am not personally comfortable doing unless there appear to be some consensus about.

- All deprecated stuff in `Building` will need to be removed (callers will have to be updated - this will have somewhat of a wide impact!).
- Code for collapsing basements needs to be consolidated in one place (currently, it's split mostly between `Server.collapseBasement()` and `BuildingServerHelper`).
- `IHex` should be merged with its only implementor `Hex`, `IBoard` with `Board`, `ITerrain` with `Terrain` (at the very least, constants should not be split between `ITerrain` and `Terrain`).
- Additional utility methods in the same spirit of the two I've added in should be written for `IHex`/`Hex`, with the long-term goal of getting rid of most of the references to the constants in `Terrains`.

# Impact on other projects

MegaMekLab and MekHQ are not affected by these changes.

Mekwars needs 1 import to be updated: `import megamek.common.Building;` needs to be replaced with `import megamek.common.building.Building;` in `/MekWars/src/common/campaign/clientutils/GameHost.java`. I can provide a patch against the current SVN head if required, but I guess this modification is more easily done than patched in)


megamek/src/megamek/common/building/BuildingServerHelper.java
```diff
@@ -0,0 +1,116 @@
...
+
+            String msg = String.format("Basement at %s already collapsed (basement typet: %s)", hex.getCoords().getBoardNum(), bs.getBasementType()); //$NON-NLS-1$
+            DefaultMmLogger.getInstance().warning(BuildingServerHelper.class, "collapseBasement", msg); //$NON-NLS-1$
+
+        } else {
+
+            String msg = String.format("Basement at %s is collapsing (basement type: %s)", hex.getCoords().getBoardNum(), bs.getBasementType()); //$NON-NLS-1$
+            DefaultMmLogger.getInstance().info(BuildingServerHelper.class, "collapseBasement", msg); //$NON-NLS-1$
+
+            disseminator.accept((newHexCollapsedReport(b, bs)));
```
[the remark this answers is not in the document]
giorgiga: good catch!

megamek/src/megamek/common/building/Buildings.java
```diff
@@ -0,0 +1,179 @@
...
+            throw new IllegalArgumentException(msg);
+        }
+
+        Map<Coords,IHex> receptacle = new HashMap<>();
+        getSpannedHexesRecurse(hex, board, structureType, receptacle);
+        return receptacle;
+
+    }
+
+    private static void getSpannedHexesRecurse(IHex hex , IBoard board, int structureType, Map<Coords,IHex> receptacle) {
```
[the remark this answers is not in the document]
giorgiga: It's only called when initialising a `Board` (ie: in `Board.initializeAll()`) – so I guess performance is not really critical here.

giorgiga:
I was testing small basements (#1103) and found out these changed actually caused an IllegalArgumentException to be thrown when using infantry and/or protomechs (I had both - not sure which caused the issue). Must be that I only tested with mechs and tanks until now :-(

giorgiga:
With this I think I'm finished: there's nothing outstanding on my TODO list for this PR :-)

sixlettervariables:
I've played around with this and haven't had any issues, that being said, as I'm less familiar with MM I think somebody more seasoned should double check that I've not missed anything.

megamek/src/megamek/common/building/Building.java
```diff
@@ -0,0 +1,405 @@
...
+    @Deprecated public boolean isIn(Coords coords) {
+        return sectionAt(coords).isPresent();
+    }
+
+    /** @deprecated use {@code sectionAt(coords).isPresent() } instead */
+    @Deprecated public boolean hasCFIn(Coords coords) {
+        return sectionAt(coords).isPresent();
+    }
+
+    /** @deprecated use {@link #iterateCoords()} instead */
```
neoancient: `iterateCoords()` does not exist.

megamek/src/megamek/common/building/Building.java
```diff
@@ -0,0 +1,405 @@
...
+
+    /**
+     * @return the building class of this building, if any
+     */
+    public Optional<BuildingClass> getBuildingClass() {
+        return Optional.ofNullable(buildingClass);
+    }
+
+    /**
+     * @returns the magnitude of the explosion this building causes when destroyed, if any
```
neoancient: @return

megamek/src/megamek/common/building/ConstructionType.java
```diff
@@ -0,0 +1,158 @@
...
+     */
+    public int getDefaultCF() {
+        return defaultCF;
+    }
+
+    /**
+     * Per page 175 of Total Warfare, this is how much damage the building
+     * absorbs when combat occurs inside the building.
+     *
+     * @returns the fraction of damage absorbed by the building
```
neoancient: @return

megamek/src/megamek/common/building/Building.java
```diff
@@ -0,0 +1,405 @@
...
+            case MEDIUM:   buf.append("Medium ");   break;
+            case HEAVY:    buf.append("Heavy ");    break;
+            case HARDENED: buf.append("Hardened "); break;
+            case WALL:     // fall-through
+            default:       // do nothing
+        }
+
+        getBuildingClass().ifPresent(bc -> {
+            switch (bc) {
+            case HANGAR:          buf.append("Hangar "); break;
```
neoancient: indentation

megamek/src/megamek/common/building/BuildingSection.java
```diff
@@ -0,0 +1,294 @@
...
+     * Sets whether this building section is burning
+     */
+    public void setBurning(boolean burning) {
+        this.burning = burning;
+    }
+
+    /**
+     * @return a stream of the demolition charges planted in this building section
+     */
+    public Stream<DemolitionCharge> streamDemolitionCharges() {
```
neoancient: Provide a way to iterate.

giorgiga: I don't get the point here, sorry.
What do you mean?

giorgiga:
@neoancient 2141869 should address all the points in your review, **except** the last one, where I need some clarification

megamek/src/megamek/common/building/BuildingSection.java
neoancient: Sorry, you're right. That's a half-completed thought that doesn't make any sense. What I meant to say is provide a way to iterate using a loop. Previously it returned the list directly, which is not usually the best approach, but it allowed processing by loops or streams. You have restricted access to streams only, which makes it harder to step through code. And some people don't like streams and would much rather use loops.

neoancient: That goes for streamSections as well.

giorgiga: Oh, I see your point there.

The reason why I'm more at ease with exposing stuff as streams rather than collections/iterables/iterators/arrays is that streams are read-only.

With collections, people often wrap via `Collections.unmodifiableXXX` before returning, and arrays are (should be) treated as copy-on-read entities – streams don't require any of this craziness and one can just `.collect(toList())` or `.toArray()` if he wants to use a different API.

Personally, I advocate for using streams whenever possible, but... (or course) it is your call when it comes to MM! :-) Just let me know if you want me to use collections instead of streams (I guess providing both as alternatives wouldn't make much sense) – no big deal.

neoancient: I thought that was probably the reason. I tend to use streams frequently within the body of a method, but when it comes to public accessors I usually wrap a collection in an unmodifiable, which is not a perfect solution but it's a lightweight wrapper that prevents modification of the collection itself and easily lends itself to iteration by loop or stream without the overhead of converting a `Stream` to an `Iterable` or an `Iterator` to a `Stream`.

giorgiga:
@HammerGS there is no pending change request on this PR as far as I know. Can you clarify?

pheonixstorm:
Whatever happened to this?

HammerGS:
The person closed this as such we never proceeded with it.

PR #1111, opened 2018-08-23
Gradle

neoancient:
Uses Gradle for build automation, replacing ant and removing binary libraries from the repository.
The distribution tasks are designed to produce release packages that match what we are currently producing with ant. Using Gradle will result in some changes to workflow, which are described here: https://github.com/MegaMek/megamek/wiki/Working-With-Gradle
Some minor changes:
1. Libraries have been upgraded to current versions.
2. Some changes were necessary in unit tests because as of Mockito v. 2.1+, the `anyX()` matchers will no longer match a `null` value. A few matchers had to be replaced by the newer `nullable(Class)`, which will match a `null`.
3. Gradle expects source code and source resources to be in separate directories, so the *.properties files were moved from the src and unittests directories into their own.
4. The suffix for the current development version has been changed from "-git" to "-SNAPSHOT", which has a special meaning in Maven repositories. Any artifacts published as a snapshot version have the suffix converted to a time stamp to aid Gradle in using the latest version.

I did not switch to the standard Gradle directory layout (inherited from Maven) in order to avoid a diff that includes every source file. Though as it currently stands at 283 changed files, that might not have made much difference. The vast majority of those were in the packaging_utils directory.


megamek/build.gradle
```diff
@@ -0,0 +1,422 @@
...
+	implementation 'log4j:log4j:1.2.17'
+
+	implementation 'javax.xml.bind:jaxb-api:2.3.0'
+	runtimeOnly 'org.glassfish.jaxb:jaxb-runtime:2.3.0'
+	runtimeOnly 'org.glassfish.jaxb:jaxb-core:2.3.0'
+	runtimeOnly 'com.sun.activation:javax.activation:1.2.0'
+	
+	jarbundler 'com.ultramixer.jarbundler:jarbundler-core:3.3.0'
+
+    testImplementation 'junit:junit:4.12'
```
sixlettervariables: mixed tabs/spaces

PR #1126, opened 2018-09-08
Adopt Log4j2

giorgiga:
This PR switches the logging system from the legacy log4j 1.2.x to log4j2.

## Impact on functionalities

1. The log file is now overwritten every time the game is started.  
   _The previous behaviour was to append to the file and keep the last 60MB worth of logs in 6 rolling files)._
2. The default log file is now `./logs/mm.log`, the same for MM, MML and MHQ (see the wiki writeup, below, for how to log to adifferent file).
   _The default log file was ./logs/megameklog.txt for MM, ./logs/megameklablog.txt for MML and ./logs/mekhqlog.txt
   for MHQ_
3. MM will cleanup any legacy logs on startup (this is to avoid confusion when people upgrade).
3. The default log configurations now live in the classpath (no more `mmconf/log4j.xml`). It is still possible to edit the logging configuration without opening the .jar file (see the wiki writeup, below).
4. All logging output is routed through the logging system, including sysout/syserr.

## Summary of changes

- The `MMLogger` hierarchy (`MMLogger` implemented by `DefaultMMLogger` and `FakeLogger`) have been flattened out and
  now only `MMLogger` (now a class) and the static method `DefaultMMLogger.getInstance()` remain.
- All methods in `MMLogger` have been made static and are now implemented directly in terms of log4j calls, with no indirection through other `MMLogger` methods (ie: they are ready to inline). Methods that were unused in MM, MML and MHQ have been made package-private. `MMLogger` has _not_ been deprecated, as that would generate a bazillion deprecation warnings and the inlining is (IMHO) best done in one go across the whole project.
- Test classes that used `FakeLogger` don't use it anymore (it was basically an elaborate workaround to shut down logging during unit tests, which is better controlled in log4j2 test configuration).
- Classes in `megamek.common.logging` have been widely modified or removed (preserving source compatibility in all three of MM, MML and MHQ).
- The `MegaMek` main class is obviously impacted for what concerns setting up the logging subsystem; since it was being modified anyway, I also went on and replaced any reference to the "old" logging system with the equivalent log4j2 calls.

## Impact on MML and MHQ and related PRs

This PR introduces incompatible dependency changes and breaking changes to the MM API: both MML and MHQ will have to be updated shortly after merging this PR.

See the related PRs on MML and MHQ:  
- https://github.com/MegaMek/megameklab/pull/197
- https://github.com/MegaMek/mekhq/pull/922

## Known Issues

The log configuration doesn't observe `PreferenceManager.getClientPreferences().getLogDirectory()`.  
Since there is no way to set that preference (except maybe manually editing the preferences file?), this doesn't seem a very problem (especially considering that "logs" is already hardcoded, eg. in `MekHQ.java` ).

## Additional work

All calls to `MMLogger` should be inlined or otherwise replaced with their Log4j equivalents (in all three projects); `MMLogger` and other classes in `megamek.common.logging` should then be removed.

Usage of `IClientPreferences.getLogDirectory()` should be analysed and the method either removed or renamed (for example, `getLogDirectory()` is used in `ClientGUI.gameListener` to determine where `salvage.mul` should be saved, which doesn't seem quite right).

Usage of `MegaMek.TIMESTAMP` should be reviewed with the aim of, if possible, removing the field and the file `logs/timestamp` it relies on. The rationale is that if the field is used only with informational value, it's not a very interesting information while if it's used for "serious" matters (eg. checking server/client compatibility), then it is to easy to tinker with (am I the only one who periodically runs `rm logs/*`?) .


megamek/build.gradle
```diff
@@ -394,7 +402,7 @@ task javadocJar(type: Jar) {
 publishing {
     publications {
         publishMMLibrary(MavenPublication) {
-            artifactId = "megamek${mmBranchTag}"
+            artifactId = "megamek${mmBranchTag.replaceAll('[^A-Za-z0-9_\\-.]','-')}"
```
giorgiga: Some characters (notably `/`) are allowed for branch names but not for maven artifact names (I changed this in all three projects - no need to check).

megamek/src/megamek/common/logging/Technicalities.java
```diff
@@ -0,0 +1,85 @@
...
+     *         (eg: from {@code public static void main(String[] args)})
+     */
+    public static StackTraceElement getCaller() throws NoSuchElementException {
+        StackTraceElement[] stackTrace = Thread.currentThread().getStackTrace();
+        // 0 is getStackTrace(), 1 is us, 2 is our caller
+        if (stackTrace.length < 4) {
+            throw new NoSuchElementException();
+        }
+        return stackTrace[3];
+    }
```
giorgiga: This `getCaller()` and the following `getFirstCallerOutsideClass()` are two methods I sneaked in because I plan to use them in MHQ.

giorgiga:
I started preparing a wiki page on how logging is configured and how devs can customise it locally - I'll finish writing it if this PR looks any good (let me know)

megamek/resources/log4j2-release.xml
```diff
@@ -0,0 +1,25 @@
...
+    <appenders>
+        <file name="file" fileName="${logfile.filename}" append="false" createondemand="true">
+            <patternlayout pattern="${logfile.pattern}" disableAnsi="true" />
+        </file>
+    </appenders>
+
+    <loggers>
+        <logger name="megamek" level="info" />
+        <logger name="System"  level="info" />
+        <root level="warning">
```
neoancient: Console message when running from a Jar:
2018-09-08 10:31:06,947 main WARN Error while converting string [warning] to type [class org.apache.logging.log4j.Level]. Using default value [null]. java.lang.IllegalArgumentException: Unknown level constant [WARNING].

megamek/resources/log4j2-release.xml
```diff
@@ -0,0 +1,25 @@
...
+            <appenderref ref="console" />
```
neoancient: 2018-09-08 10:31:06,955 main ERROR Unable to locate appender "console" for logger config "root"

giorgiga: Great catch! (both of these) - I forgot to actually try and run MM from the jar.

(in case you want to fix it locally to go on with the testing, `<root level="warning">` should be `<root level="warn">` and the `<appenderref ref="console" />` should be removed – I'll test and update the PR anyways).

PhoenyxCullen:
I understand the need to keep session logs seperate, but can we timestamp the logs instead of overwritting it? Therefore we keep the last x logs in the log directory.

giorgiga:
3 instances in parallel? I thought you were talking about successive sessions...

I'm not sure logging from different JVMs to the same files is even supported in log4j2 (or log4j 1.x for what matters) – how did that work with log4j 1.x? (I guess it all went to one file, and that output from the different processes was occasionally mixed?)

PhoenyxCullen:
That's why I'm suggesting a time-stamped log file.

giorgiga:
Including the startup time in the log filename is certainly doable... the tricky part is deciding what files should be deleted (age alone is not very reliable: I, for example, often hibernate the laptop with MM running, which results in sessions longer than 24hrs).

IIRC Log4j supports locking the log files, which would certainly help... I've never used that feature though – I'll need to experiment a bit to understand how it works.

BTW: if we just want to keep a certain number of sessions worth of logs, a rolling file appender seems to do nicely:

```xml
<rollingfile name="file" fileName="${logfile.filename}" filePattern="${logfile.filename}.%i.gz">
    <patternlayout pattern="${logfile.pattern}" disableAnsi="true" />
    <policies><onstartuptriggeringpolicy /></policies>
    <defaultrolloverstrategy fileIndex="min" min="1" max="3" compressionLevel="9" />
</rollingfile>
```

giorgiga:
I thought log4j2 would acquire locks when a file is opened and kept them until the appender was shut down, but this is not the case:

> When set to true, I/O operations will occur only while the file lock is held allowing FileAppenders in multiple JVMs and potentially multiple hosts to write to the same file simultaneously. This will significantly impact performance so should be used carefully. Furthermore, on many systems the file lock is "advisory" meaning that other applications can perform operations on the file without acquiring a lock. The default value is false.

The bad news is, we can't use log4j's locks as a marker for files that shouldn't be deleted - the good news is that if we enable locking, multiple MM instances will safely log to the same file (ok, log lines will be all mixed up).

At this point all I can thing about is implementing some file-based locking system on our side... but I'm not 100% sure how much I want to work on it.

I'll see if I can come up with something (I'm busy this weekend - so it will most probably be next week and).

BTW @PhoenyxCullen for now, you can specify a `-log` option on the command line to keep your logs separate (I've not tested it, but it should work with the current MM version too)

giorgiga:
@PhoenyxCullen 

Your remarks appear to be an enhancement request on the logging behaviour in general and not about log4j in particular. Can you open an issue about the ER?

The point is that, while your comments are certainly related to the issue under discussion here, this is a PR and it can very well be discarded (together with your ER if it only lives here) if – for example – devs decide to go with logback instead of log4j2.

(Also, I realised I really don't want to put effort into implementing your ER, at least until after the fate of this PR is decided: any work I do on top of this PR risks vanishing down the drain in case this PR is not merged).

giorgiga:
@PhoenyxCullen

I messed around with file locks a bit. It may very well be that I am just not getting how to use them, but they seem to be too fiddly/unreliable to be practical (I've managed to happily lock the same file twice from two separate JVM instances).

I can't think of any other alternative to tell log files that are in use apart from those that can be deleted, so... I guess you'll have to stick with the `-log` command-line parameter until someone comes up with an idea for this :-)

giorgiga:
The file locks thing was way too exciting :-) so I forgot about the log4j2 writeup.

Below is a proposed wiki page on the subject: please do edit it to adapt it to the project communication style and also because... english is not my first language an the page could surely have been written better than I did :-)

----
# Logging in MegaMek

As of 0.45.2, MegaMek (and related projects) have switched from apache log4j 1.2 to [apache log4j2](https://logging.apache.org/log4j/2.x/).

This page will get you started with the new logging API and explain how it is configured in MegaMek.

## Logging API

You should refer to the [log4j2 doc](https://logging.apache.org/log4j/2.x/manual/api.html) for details on the log4j2 API, but -if you are in a hurry- the example below should cover some 95% of the use cases:

```java
import java.time.*;
import java.util.*;
import org.apache.logging.log4j.LogManager;
import org.apache.logging.log4j.Logger;

public class SomeClass {

    // obtaining a Logger
    private static Logger log = LogManager.getLogger(SomeClass.class);

    public static void main(String[] args) {

        // log, log, log...
        log.info("simple message");
        log.info("message with a parameter of value {}", UUID.randomUUID());
        log.info("message with {} {}", 2, "parameters");

        // logging exceptions
        try {
            Optional.of(null);
        } catch (Exception e) {
            log.error("It's {} and exceptions can always be specified after parameters", LocalTime.now(), e);
        }

        // conditional logging
        if (log.isDebugEnabled()) {
            String msg = "generating this log message requires {}, "
                       + "so we don't want to do it unless it's "
                       + " going to actually be logged";
            log.debug(msg, lengthyOperation());
        }

        // functional-style conditional logging
        log.debug("this also spares {}, in case !log.isDebugEnabled()", SomeClass::lengthyOperation);
    }

    private static Object lengthyOperation() { return "a ton of computations"; }
}
```

### Caveat coder!

When you start using log4j2 in your code, you'll quickly notice that `Logger` and `LogManager` are quite popular as class names and that you have several classes with those name in the classpath, so importing the wrong class can become a nuisance.

If you use eclipse, an editor template will help importing the right classes (and spare you the tedium of typing the `Logger` initialisation).

You can copy the following to an `.xml` file and import it into your IDE (_Preferences -> Java -> Editor -> Templates_ and _Import..._ - of course you can also hit _New..._ and fill the form instead).

```xml
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<templates>
<template autoinsert="true" context="java" deleted="false" description="" enabled="true" name="Log4j2 logger">${:import(
org.apache.logging.log4j.LogManager,
org.apache.logging.log4j.Logger
)}private static final Logger log = LogManager.getLogger(${enclosing_type}.class);${cursor}</template>
</templates>
```

## Logging configuration

Megamek includes three separate log4j2 configurations:

Megamek includes three different log4j2 configuraiton files:
- `/megamek/resources/log4j2.xml`
- `/megamek/resources/log4j2-release.xml`
- `/megamek/testresources/log4j2-test.xml`

`resources/log4j2.xml` and `testresources/log4j2-test.xml` are handled by the log4j2 auto-configuraion mechanism (you should read about it in the [log4j2 documentation](https://logging.apache.org/log4j/2.x/manual/configuration.html#AutomaticConfiguration), but basically what happens is that log4j2 tries to load the `-test` configuration and, failing that, falls back to the "regular" `log4j2.xml`).

Additionally, when starting the applicaiton through `megamek.MegaMek.main()`, MegaMek will:

1. detect if it is being run from withing a `.jar` file and switch to `log4j2-release.xml` if appropriate, and
2. check if a file named `log4j2.xml` exists in the current working directory and, if found, use it to override the log4j configuration (note this will happen _regardless_ of whether MegaMek is run from a `.jar`, an IDE or `gradle run`)

**All in all, in order of priority, the configuration used will be**:

- `$PWD/log4j2.xml`, if starting via `megamek.MegaMek.main()`;
- `resources/log4j2-release.xml`, if starting via `megamek.MegaMek.main()` from a `.jar` file;
- `testresources/log4j2-test.xml`, if the test classpath is available (ideally, this only happens when running unit tests, but please check your IDE's configuration if you start code from there); and eventually
- `resources/log4j2.xml` if everything else fails.

As a developer, you can exploit this overcomplicated mechanism in two ways:
1. you can `cp resources/log4j2.xml .` and enable `TRACE` on the classes you are working on, with no fear of flooding the other developer's logs (`/megamek/log4j2.xml` is in `.gitignore`), and
2. you can prepare a `log4j.xml` and instruct end users to put it in the working directory if you are trying to debug an issue you can't replicate locally.

## MegaMekLab and MekHQ

MegaMekLab and MekHQ piggyback on MegaMek logging, so all that was said about MegaMek above, also apply to its sister applications.

Please note that if you launch MegaMek from inside MekHQ, log originating from either one will go to the same file (per default, `logs/mm.log`).

## Integration with other logging libraries

`build.gradle` contains lines that add the various adapters log4j2 comes with to the classpath:

```groovy
runtimeOnly    "org.apache.logging.log4j:log4j-jul:${log4jVersion}"        // routes java.util.logging to log4j2
runtimeOnly    "org.apache.logging.log4j:log4j-1.2-api:${log4jVersion}"    // routes log4j 1.2.x to log4j2
runtimeOnly    "org.apache.logging.log4j:log4j-jcl:${log4jVersion}"        // routes commons-logging to log4j2
runtimeOnly    "org.apache.logging.log4j:log4j-slf4j-impl:${log4jVersion}" // routes slf4j to log4j2
```

At the moment no adapter is needed and so these lines are commented out.

Shall a library that depends on the log4j 1.2.x, or commons-logging, or slf4j or that uses `java.util.logging` be added in the future, uncommenting the relevant line (and excluding the transitive dependencies on other logging libraries) will allow to route its logs to log4j2.

giorgiga:
In case you can't copy the source markdown from the previous comment (I have no idea), here's the text as a text file: [logging-wiki-page.md.gz](https://github.com/MegaMek/megamek/files/2389527/logging-wiki-page.md.gz)

giorgiga:
Can you add it and push to this PR (or just apply the change while merging, if it's otherwise ok to do so)?

I don't mind the chore, but it will take a while before I can do it.

giorgiga:
@HammerGS there is no pending change request on this PR as far as I know. Can you clarify?

neoancient:
If there is a way to add additional configuration information from another file, I think that would be the simplest solution. After processing either the built-in or the override, it could look for a supplemental file and process the directives in that one as well. The additional file could be required to have a particular name, or you could search the mmconf directory and include any files that end in .l4j.xml. Otherwise I would think there would need to be a copy of the log file that can be used in place of the one in the jar by renaming it, but that seems somewhat more awkward. The only real requirement is that the option for enabling debugging is with the hands of the average user (or at least the average user who would take on a task like customizing force generation files). Here is the section from the documentation:
```
IV. Customizing the Formation Rules

This section documents the xml-based language used to define the force generation rules for each faction. This
is provided for anyone who wants to create custom formation building rules, but is made with no promise of
support if your file produces unexpected results. When troubleshooting a problem with generation, it can
be helpful to edit the logging configuration file (mmconf/log4j.xml) and uncomment the two sections pertaining
to force generator logging by removing the <!-- at the beginning of the line and the --> at the end. This
will enable debugging output and log messages from force generation to a separate file.
```

giorgiga:
I guess our comments above overlapped - but I can't be sure so be patient if I repeat the same idea (I'll explain it in more detail in order to not waste your time completely if you already read the above) :-)

If those people work with a released version of megamek rather than  a git clone, they will have no access to the default `log4j.xml` in `resources/` (that will be inside the jar). Also, rather than instructing them to edit a configuration file, it may be easier to have a separate cfg file for them and instruct them to just place it in the root directory.

We could add a separate cfg in `docs/RAT Stuff/force-generator-log4j2.xml` and tell users that:

> When troubleshooting a problem with generation, it can be helpful to insulate the log messages originating from the force generator. To do that,  you can copy `docs/RAT Stuff/force-generator-log4j2.xml` to the root MegaMek directory (the same directory where `docs/` is) and rename it to `log4j2.xml`. This
will enable debugging output and log messages from force generation to a separate file.

What do you think?
(If you approve the idea) can you write that file and update the docs or should I do it? (in case confirm that the only doc to edit is  `docs/RAT Stuff/force-generator.txt`)

neoancient:
That will work. Why the root directory instead of mmconf where it is currently?

giorgiga:
No specific reason... the root directory seemed like the right place for a local configuration override file that is not under version control and nor part of the distribution (similar to the gradle local properties file): I've written the mechanism for this log4j2.xml in the root directory basically because I'm tired of  dealing with modifying `mmconf/log4j.xml`, avoiding committing it and occasionally having it reset by a `git checkout -- .`.

Can you write that file and update the docs or should I do it? (in case confirm that the only doc to edit is docs/RAT Stuff/force-generator.txt)

megamek/src/megamek/common/logging/Log4j.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+package megamek.common.logging;
+
+import java.io.File;
+import java.io.IOException;
+import java.net.URI;
+
+import org.apache.logging.log4j.Level;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.core.LoggerContext;
```
neoancient: While I was looking at ways to make this easier to use I noticed that you're using internal code that is not part of the public API.

giorgiga: AFAIK `log4j-api` does not provide any mean for programmatic configuration (and certainly does not provide the `InputStream`/`OutputStream` implementations used in `Log4j.snatchSytemStreams()`).

As far as my understanding goes, `log4j-core` is the logging backend and `log4j-api` is a facade that one could use with different backends (doesn't make much sense for an app, but it allows a library that internally uses `log4j` to be integrated in an app that uses `logback` as a logging backend).

The `log4j-core` API will certainly be less stable than `log4j-api`, but it is documented and –IMHO– far from deserving the label of "internal code" :-) Specifically, the use of `org.apache.logging.log4j.core.LoggerContext` (complete with the horrible cast from `(LoggerContext) LogManager.getContext(false)`) is documented at https://logging.apache.org/log4j/2.x/manual/customconfig.html#AddingToCurrent

giorgiga: BTW the dependency on `org.apache.logging.log4j.core` is basically the reason why I made this `Log4j` utility class: the idea was to concentrate all log4j config functionality in one place.

Outside `MegaMek` this stuff will only be used by `Princess` to set its own logging level.

neoancient:
Closing this because it adds too much unnecessary complexity and you seem to expect me to fix it.

giorgiga:
Now, this is uncalled for.
Do you want to reconsider, @neoancient ?

PR #1132, opened 2018-09-12
Zweihander

AaronGullickson:
Implementation of the Zweihander SPA from CamoOps. When a unit with this SPA and two functioning hand actuators attempts to punch or hit with a melee weapon, a pop-up will come up after the usual melee pop-up asking if they want to ZWEIHANDER, much like the retractable blades approach. 

The rules for punching with Zweihander are a bit vague. It doesn't say that the unit is limited to one punch. If they were limited to one punch, then Zweihander would be totally worthless for punching as your single punch would do basically the same amount of damage as two separate non-Zweihander punches, plus the risk you face of damage/falling. So, I assume they meant that you could still make two punch attacks, but it does seem a little strange and hard to visualize.


AaronGullickson:
CamOps, not CamoOps.

AaronGullickson:
Ok, I am going to hold off on this for a bit, because after reflection and discussion, I am pretty sure that you should be allowed only one punch attack.

AaronGullickson:
I also missed the "only front arc" business.

AaronGullickson:
Ok, this should now be ready for review again. Punching is only done once and front arc limitations are added. To test, I recommend an Axman-2N and a stealth 1D. Enable friendly fire and give them both zweihander and then let them smack each other around. You can give them some upper and lower arm damage as well to start with if you want to see how that affects attacks.

AaronGullickson:
I have changed all the tabs to white spaces and added all of the requested javadocs. I have also added a condition to BipedMech.canZweihander to ensure that neither arm is missing as I am not sure if this will be picked up by the hand actuator check.

PR #1134, opened 2018-09-12
pilot portrait option for tooltip

AaronGullickson:
Added a client setting option (false by default) to add pilot portraits to tooltip. Here is an example:

screen shot 2018-09-12 at 12 12 46 pm

If players are using the default image or the image is not found, nothing will show here. So, to see another player's portraits, you will need to have their portrait set installed.


megamek/src/megamek/client/ui/swing/boardview/EntitySprite.java
```diff
@@ -860,6 +865,20 @@ else if (numAdv > 1)
                 addToTT("InfSpec", BR, Infantry.getSpecializationName(spec));
             }
         }
+        
+        //add portrait?
+        String category = entity.getCrew().getPortraitCategory(0);
```
sixlettervariables: Is `getPortraitXXX(0)` valid when you hover over ejected units? And are all units crewed?

AaronGullickson: Yes, I just checked and the portrait goes with the ejected pilot. I have added a check for null crews to avoid any NPEs.

AaronGullickson:
I cleaned up whitespace and added a check for null crews. It should be good to go now.

megamek/src/megamek/client/ui/swing/boardview/EntitySprite.java
sixlettervariables: I don't see the check for null crews.

AaronGullickson:
@sixlettervariables  Whoops, pushed to the wrong branch.  Its there now.

PR #1144, opened 2018-09-16
Fix for high-heat aeros refusing to fire on ground targets

NickAragua:
Fix an issue where aeros that have more guns than heat capacity would refuse to fire on ground targets.

Basically, the "alpha strike" would come back with a utility of -4000 or whatever because aeros don't like overheating, and then the getFullFiringPlan method would return the no-action bombing run as the alpha strike instead. Whoops!

Now we check whether the bombing plan actually involves doing something before considering it as the alpha strike.


NickAragua:
Incidentally, this may fix some of the issues the bot has with flying off board using advanced aerospace movement.

PR #1163, opened 2018-09-29
Deprecate BehaviourSettings.getCopy()

giorgiga:
This deprecates
```
public BehaviourSettings getCopy() throws PrincessException
```
in favour of a new
```
public BehaviourSettings copy()
```
which doesn't declare exceptions that can't actually be thrown.

The declared `PrincessException` is only thrown (via `setDescription()`) if the copied settings don't have a valid description - which is a class invariant.

Note that I don't necessarily endorse that `BehaviourSettings.setDescription()` throw a checked exception instead of `IllegalArgumentException` or that `getCopy()` be implemented in terms of the public getters and setters: I'm proposing this PR just for the sake of MekHQ, where the handlers for the impossible `PrincessException` could be removed.

### DIFF
--- a/megamek/src/megamek/client/bot/princess/BehaviorSettings.java
+++ b/megamek/src/megamek/client/bot/princess/BehaviorSettings.java
@@ -132,13 +132,20 @@ public class BehaviorSettings {
         this.logger = logger;
     }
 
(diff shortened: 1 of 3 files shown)


giorgiga:
Canelling, as I realized it has a but

PR #1186, opened 2018-10-25
Protomech equipment

neoancient:
This implements four remaining pieces of Protomech equipment, which should complete the implementation of the Protomech unit. I had intended to split these into three separate branches, but they ended up depending on some common code that made it difficult to split. Sorry.

Protomech Melee Weapon (TacOps, p. 337) and Quad Melee System (IOps, p. 67)
Very simple implementation that adds damage to the melee attack if the equipment is present and operational.

Fusillade Launcher (IOps, p. 65)
Effectively a poor man's iATM which uses the Artemis IV cluster hit bonus instead of the streak effect. Since the iATM handler already had code to ignore the streak effect when fired indirectly I just expanded that to check for the protomech weapon flag as well. The tricky part was that it is effectively a one-shot launcher with another reload. By itself that wouldn't be too bad but each of the two shots per launcher can be a different one of the five available munition types. I set it up with a second ammo `Mounted` that is linked by the first, so the weapon and the two shots form a double-linked list with the weapon at the head and the ability to bring any ammo in the chain to the second position. I wrote it so that if needed in the future it could support an arbitrary list size.

Magnetic Clamp System (IOps, p. 66)
Competes with the fusillade for the most annoying. The protomech system occupies the same space as the BA handles/magclamp transports but the mech can't use both at the same time. I thought about expanding the BA handles class to be able to mount either BA or protomechs, but the fact that each trooper in a BA squad occupies a particular position but the protomech can be mounted back or front at the player's discretion made it enough different to go another route. I ended up giving each mech two protomech transports, one front and one rear, which can be loaded specifically using the bay number field that is used for aerospace units. Since the transports don't know anything about the unit or each other, it is the responsiblity of the code handling the loading to make sure it doesn't load both BA and protos, or load an ultraheavy in the rear, or an ultraheavy and a second proto. Not as clean as I would like it, but to get it as clean as I would like it would require a significant rewrite of all the transport code.


NickAragua:
The Fusillade and mounting/unmounting stuff seems to work fine, as far as I checked. Quad melee system seems to work fine as well, although I'm not sure why there's a separate "quad melee system" button (eventually figured out I should use "protomech frenzy" instead).

neoancient:
I never noticed the quad melee button. If I had I would have removed it.

neoancient:
Turns out it's an unintended by-product of using the F_CLUB flag for protomech melee weapon systems. It was enabling the club attack button and changing the name to the equipment with the club flag. Giving them a unique flag fixed that.

PR #1187, opened 2018-10-28
Factor networking code in Client and Server

nbtdev:
This PR repackages (but not rewrites) the code that deals with networking aspects in `Client` and `Server`. Events in the lifetime of the network connections are handled via overridable methods in the new `SocketServer` and `SocketClient` classes, which `Server` and `Client` (respectively) extend. All of the unit tests that should pass, do pass (there is one that also fails in upstream master)

The reason for this is that I need to be able to replace the use of raw Socket and ServerSocket with [WebSocket](https://tools.ietf.org/html/rfc6455) (looking to use https://github.com/TooTallNate/Java-WebSocket), and the details of managing and polling the connections and messages to and from the connected endpoints are hidden behind the implementations in the WebSocket library. I would rather not maintain a downstream fork of the MegaMek code and have to integrate, for the rest of eternity, any changes that come from upstream, so the best middle ground was this solution -- make it simpler to replace the networking implementation with something else. 

The reason I need to do this is probably outside the scope of this MR, but in general, I need to be able to kick off game servers on demand, accessed through an nginx reverse proxy. This isn't really possible (at least not easily) with standard nginx stream proxying but it's relatively trivial if WebSocket is used instead (since WebSocket begins with an HTTP transaction and nginx speaks native HTTP).


.gitignore
```diff
@@ -12,3 +12,4 @@ Thumbs.db
 /.project
 /build/
 /out/
+/.idea/
```
nbtdev: I use IntelliJ and it seems to think that putting more files in the source tree is a good idea...

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.9-bin.zip
+#Fri Oct 26 21:08:02 PDT 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-zipStorePath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
+zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.9-all.zip
```
nbtdev: More IntelliJ -- I believe it made this change when I imported the project. Can be removed if there is a reason we only wanted to use bin.

megamek/build.gradle
```diff
@@ -291,6 +291,7 @@ ${project.ext.jvmOptions.join('\n')}
 
 windowsDistZip {
     description = 'Creates Windows distribution packaged as a zip archive'
+    dependsOn stageFiles
```
nbtdev: This is missing if you just want to make a Windows build -- I suspect it wasn't a problem as typically when zips are made, they are all made and some other step invokes the `stageFiles` step.

megamek/src/megamek/common/net/DataStreamConnection.java
```diff
@@ -27,7 +27,7 @@
  * <code>DataInputStream</code> and <code>DataOutputStream</code> to
  * send/receive data.
  */
-class DataStreamConnection extends AbstractConnection {
+class DataStreamConnection extends AbstractSocketConnection {
```
nbtdev: I think this may have been a holdover from when I was considering inserting an abstraction at this level  -- can be removed if it's an issue

megamek/src/megamek/common/net/IConnection.java
```diff
@@ -16,33 +16,33 @@
...
+ * original Connection class which was moved to AbstractSocketConnection
  */
 public interface IConnection {
     /**
      * Opens the connection
      * 
      * @return <code>true</code> on success, <code>false</code> otherwise
      */
-    public boolean open();
+    boolean open();
```
nbtdev: IntelliJ (correctly) pointed out that for interfaces, there is no need to mark the methods as public (they are inherently public)

megamek/startup.sh
```diff
@@ -154,7 +154,7 @@ RUNPATH=`ls -1 $MEGAMEK_CLASSPATH/*.jar | awk 'BEGIN {ORS=":"} {print $1}' | sed
 # We'll just test both versions to see if they are greater than 8
 JAVA_VERSION1=`java -version 2>&1 | grep "version" | awk '{print $3}' | tr -d \" | awk '{split($0, array, ".")} END{print array[1]}'`
 JAVA_VERSION2=`java -version 2>&1 | grep "version" | awk '{print $3}' | tr -d \" | awk '{split($0, array, ".")} END{print array[2]}'`
-if [ $JAVA_VERSION1 -gt 8 -o $JAVA_VERSION2 -gt 8 ]; then
+if [ "0$JAVA_VERSION1" -gt 8 -o "0$JAVA_VERSION2" -gt 8 ]; then
```
nbtdev: Without this, I get an error message in the console when starting up on *nix; apparently one or both of these come back empty on my system. This simply makes sure that the numbers aren't empty (putting a zero in front of anything will not change the numeric comparison)

sixlettervariables:
I'm going to try and review this sometime tomorrow.

nbtdev:
> I'm going to try and review this sometime tomorrow.

Any thoughts?

HammerGS:
This pull request **fixes 1 alert** when merging 3ef7bd586a7f97d25a69c9e2ebdcff6363dc3e04 into e9fac32d93d93a6510efd7cf76d0c549f1192eff - [view on LGTM.com](https://lgtm.com/projects/g/MegaMek/megamek/rev/pr-7efb7f4f8a0ef872ad3e3d38f3b4c07255de4080)

**fixed alerts:**

* 1 for Deserialization of user\-controlled data

---

megamek/src/megamek/common/net/SocketServer.java
```diff
@@ -0,0 +1,159 @@
...
+                connection.addConnectionListener(connectionListener);
+
+                ConnectionHandler ch = new ConnectionHandler(connection);
+                Thread newConnThread = new Thread(ch, "Connection " + connection.getId());
+                newConnThread.start();
+                connectionHandlers.put(connection.getId(), ch);
+            } catch (InterruptedIOException iioe) {
+                // ignore , just SOTimeout blowing..
+            } catch (IOException ex) {
+
```
sixlettervariables: should these too be ignored or logged?

sixlettervariables:
I've begun reviewing this in detail, but overall it looks good.

megamek/src/megamek/common/net/SocketServer.java
nbtdev: I believe those were in the original code -- I just lifted the existing code out of the `Server` class and packaged it into this one

sixlettervariables:
(closed/reopened to kickstart build checks)

gradle/wrapper/gradle-wrapper.properties
sixlettervariables: @neoancient thoughts on this?

neoancient: The runtime is sufficient to the task. It won't hurt anuthing, but I see no need to download and save sample code and documentation.

nbtdev: Updated to use the runtime (bin) only

SJuliez:
Closing this as this doesn't look like it's going to get merged.

PR #1192, opened 2018-11-12
Reset crew state

neoancient:
Replaces the method that resets the fatigue counter with one that resets all game state. For Megamek/mekhq#978

### DIFF
--- a/megamek/src/megamek/common/Crew.java
+++ b/megamek/src/megamek/common/Crew.java
@@ -738,7 +738,7 @@ public class Crew implements Serializable {
         s = s.trim();
         int index = s.indexOf(" ");
         if (index == -1) {
-            return new Boolean(true);
+            return Boolean.TRUE;
         }
         String t = s.substring(index + 1, s.length());
         Object result;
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/Crew.java
```diff
@@ -1097,10 +1097,17 @@ public void incrementFatigueCount() {
...
+     * Sets crew state fields back to defaults. Used by MekHQ to clear game state.
      */
-    public void resetFatigue() {
+    public void resetGameState() {
         fatigueCount = 0;
+        doomed = false;
+        ejected = false;
+        for (int i = 0; i < crewType.getCrewSlots(); i++) {
+            unconscious[i] = false;
+            dead[i] = false;
```
NickAragua: Do we really want to reset "dead" and "missing"? Or is the death of a particular crew member handled elsewhere?

neoancient: It's handled elsewhere. During scenario resolution this method gets called after crew casualties are applied to the Person, but before the crew is recalculated. That way we start with a blank slate, then any changes to the Crew are applied based on the availability and status of the MekHQ Person.

NickAragua: Good to go, then.

PR #1193, opened 2018-11-13
TRO style summary

neoancient:
Uses Apache's FreeMarker template library to format unit stats in the official TRO format. The primary motivation for this was to make a more complete unit summary available for exporting unit designs from MML in plain text.

Each unit type has both an HTML and a plain text template. The HTML template is used to add an alternate unit display in the unit selector dialog:
image

I have also factored the fluff fields out of `Entity` and expanded them to include manufacturing data that can be included in the fluff header of the TRO display. A companion branch in MML adds a tab for editing fluff. It still has a couple things left to do, but the branch is available for looking at now. Here is an example of what the display looks like with the fluff filled in:
image

Other than finishing export feature in MML, future plans include adding support for assigning manufacturers and models for individual pieces of mounted equipment and an option to integrate it with factory data in MekHQ. With a template library in place a rework of the BV and cost summaries is also in order.


NickAragua:
Is there some set of steps I need to take to set up the freemarker library?

neoancient:
If you're using Eclipse it doesn't update automatically when the build file changes so you'll need to right click on the project and have Gradle refresh it. Then it will grab the library and reconfigure the class path to include it

megamek/src/megamek/client/ui/swing/MechViewPanel.java
```diff
@@ -130,9 +131,30 @@ public void setMech(Entity entity, MechView mechView){
...
+            }            
+            icon = new ImageIcon(image);
+            lblMek.setIcon(icon);
+        }  
+    }
+    
     public void setMech(Entity entity) {
         MechView mechView = new MechView(entity, false);
         setMech(entity,mechView);
+//    	TROView troView = TROView.createView(entity, true);
```
sixlettervariables: This intended as a TODO?

megamek/src/megamek/common/EntityFluff.java
```diff
@@ -0,0 +1,290 @@
+/*
```
sixlettervariables: Whitespace is a bit off in this file.

megamek/src/megamek/client/ui/swing/MechViewPanel.java
neoancient: Obsolete test code that I've now removed.

megamek/src/megamek/common/EntityFluff.java
neoancient: I forgot to change the editor settings when I set up a new workspace post-Gradle conversion. I missed this file when I fixed all the others with tabbed indents.

PR #1212, opened 2018-12-24
New Edge options for Aerospace units

mkerensky:
Adds a new set of edge reroll options for aeros.

### DIFF
--- a/megamek/src/megamek/server/Server.java
+++ b/megamek/src/megamek/server/Server.java
@@ -26906,10 +26906,7 @@ public class Server implements Runnable {
                                     ammoRoll = Compute.d6(2);
                                     // To explode, or not to explode
                                     if (ammoRoll >= boomTarget) {
-                                        r = new Report(9121);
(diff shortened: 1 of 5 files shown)


megamek/src/megamek/server/Server.java
```diff
@@ -26667,7 +26685,30 @@ else if (secondaryEffects) {
                 }
                 // check for possible explosion
                 int fuelroll = Compute.d6(2);
+                r = new Report(9120);
+                r.subject = aero.getId();
                 if (fuelroll > boomTarget) {
```
sixlettervariables: I realize this is existing code, but why `>` and not `>=` like elsewhere?

mkerensky: Kind of wondered why the previous author did it that way too, to be honest - I'd rather have the actual numbers from SO show up in the code instead of the offset-by-1 needed to use >.  Fixed.

Issue #1242, opened 2019-02-02
Static analysis "big ticket" items (lgtm)


sixlettervariables:
These are the items on lgtm for MM that deserve attention (i.e. likely bugs):

- [X] #1243 Array index check https://lgtm.com/projects/g/MegaMek/megamek/snapshot/0ec4a4fb9d5d774373d49892fb5ab705c5c26d33/files/megamek/src/megamek/common/TechAdvancement.java?sort=name&dir=ASC&mode=heatmap#x9029b2ad7948272e:1
- [X] #1243 Array index check https://lgtm.com/projects/g/MegaMek/megamek/snapshot/0ec4a4fb9d5d774373d49892fb5ab705c5c26d33/files/megamek/src/megamek/common/ITechnology.java?sort=name&dir=ASC&mode=heatmap#xec6172734ac7aa11:1
- [X] #1243 Array index check https://lgtm.com/projects/g/MegaMek/megamek/snapshot/0ec4a4fb9d5d774373d49892fb5ab705c5c26d33/files/megamek/src/megamek/common/CompositeTechLevel.java?sort=name&dir=ASC&mode=heatmap#xc1cc8453fb343711:1
- [X] #1243 Array index check https://lgtm.com/projects/g/MegaMek/megamek/snapshot/0ec4a4fb9d5d774373d49892fb5ab705c5c26d33/files/megamek/src/megamek/client/ui/swing/util/PlayerColors.java?sort=name&dir=ASC&mode=heatmap#x8a702797bf5724c3:1
- [X] #1244 Wrong type used https://lgtm.com/projects/g/MegaMek/megamek/snapshot/0ec4a4fb9d5d774373d49892fb5ab705c5c26d33/files/megamek/src/megamek/common/ArtilleryTracker.java?sort=name&dir=ASC&mode=heatmap#x4af049aa508d658b:1
- [X] #1245 Wrong choice of sync object https://lgtm.com/projects/g/MegaMek/megamek/snapshot/0ec4a4fb9d5d774373d49892fb5ab705c5c26d33/files/megamek/src/megamek/client/ui/swing/ForceGeneratorOptionsView.java?sort=name&dir=ASC&mode=heatmap#xac5fcf88b46e7c1e:1
[...]

PR #1244, opened 2019-02-02
Fix incorrect usage of Hashtable::contains in ArtilleryTracker (lgtm)

sixlettervariables:
Fixes ArtilleryTracker::removeWeapon. Part of #1242.


NickAragua:
Out of curiosity, what kind of problem would this cause?

PR #1261, opened 2019-02-18
Quads and Extended Torso Twist quirk

### LINKED ISSUE #1259: [MegaMek 0.44.0] Extended Torso Twist Quirk not applying to Quads
### Environment
MegaMek 0.44.0
Windows

### Description
A quad with the Extended Torso Twist quirk should be able to twist like a Bipedal mech normally would but that is not the case.

blah2355:
Fixes #1259: Extended Torso Twist Quirk not applying to Quads. Borrowed code from megamek/common/Mech.java

### DIFF
--- a/megamek/src/megamek/common/QuadMech.java
+++ b/megamek/src/megamek/common/QuadMech.java
@@ -209,25 +209,22 @@ public class QuadMech extends Mech {
         return getWalkMP(gravity, ignoreheat);
     }
     
-    /**
+    /*
      * Normally Quads can't torso twist. Extended torso twist allows regular bipedal mech twisting
      */
     @Override
     public boolean canChangeSecondaryFacing() {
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/QuadMech.java
```diff
@@ -208,10 +208,37 @@ public int getRunMPwithoutMASC(boolean gravity, boolean ignoreheat, boolean igno
...
+     */
     @Override
     public boolean canChangeSecondaryFacing() {
-        return false;
+    	return hasQuirk(OptionsConstants.QUIRK_POS_EXT_TWIST) && !isProne();
+    }
+    
+    @Override
+    public boolean isValidSecondaryFacing(int dir) {
+        int rotate = dir - getFacing();
```
neoancient: You left out the call to Math.abs, which will result in it not working properly when rotating anticlockwise. Since this has been sitting here idle while I've been away I went ahead and fixed this so we can get it merged.

PR #1280, opened 2019-03-03
Add proper labels for host dialog controls.

BlindGuyNW:


megamek/src/megamek/client/ui/swing/HostDialog.java
```diff
@@ -25,8 +25,8 @@
  */
 public class HostDialog extends JDialog implements ActionListener {
     /**
-     *
-     */
+     *	
+     */
```
sixlettervariables: I'm not certain about the accessibility features of GitHub's reviews, but there are some trailing spaces in this file introduced by your edit.

BlindGuyNW: It's possible those were an inadvertent result of my fussing with a few editors last night. I'm trying to find a decent and accessible IDE. :)

sixlettervariables: Don't worry about it then, we need to setup editorconfig support for the repo so that this sort of thing is automagically fixed by the IDE.

Issue #1290, opened 2019-03-06
Add accessibility support to DialogOptionComponent


sixlettervariables:
Improve the accessibility of [DialogOptionComponent](https://github.com/MegaMek/megamek/blob/master/megamek/src/megamek/client/ui/swing/DialogOptionComponent.java) by ensuring labels point to their components. Also ensure setAccessibleDescription is done anywhere a tooltip contains HTML.

sixlettervariables:
Implemented by #1292.

PR #1292, opened 2019-03-06
Options fix

BlindGuyNW:
Add labels to all DialogOptionComponents. #1290.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ Thumbs.db
 /.project
 /build/
 /out/
+/.vscode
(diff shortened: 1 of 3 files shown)


BlindGuyNW:
Accidental close due to user error. Please carry on :)

sixlettervariables:
No worries! So the code looks good, but I'm trying to find a way I can verify this stuff works. I installed NVDA and cannot really figure out how to get it to "see" inside MM (which is perhaps the problem).

Are there tools you use that I should be using to review this work?

BlindGuyNW:
Hey,

Okay, yeah. I forgot that you need to enable the Java ACcess Bridge in WINdows. If you go into the Java bin directory there’s a program called jabswitch.exe. I think the command line syntax is “jabswitch /enable.” If you enable the JAB NVDA should suddenly see inside MM and all the other related programs, possibly with a restart.

Hope it works :)
Zack.

sixlettervariables:
I have it working now, thanks!

Issue #1095, opened 2018-08-12
[0.44.0; Errata Fix] TAG now designates for Indirect LRMs


ProjectDarkFox:
Per the latest round of Errata on Total Warfare as posted here https://bg.battletech.com/wp-content/uploads/2018/06/Total-Warfare-2018-06-06-v5.01-Fourth-Printing.pdf?x64300 on page 13:

**TAG (Target Acquisition Gear) (p. 142)**
_Replace the second and third paragraphs with the following:_

To  use  TAG  equipment  for  target  designation,  calculate  the  to-hit  number  as  for  a  standard  weapon  attack.  As  using  TAG  occurs  before  the  Weapon  Attack  Phase,  it  does  not  count  towards  the  number  of  targets  a  unit  engages  in  the Weapon Attack Phase.
If the to-hit roll fails (meaning the TAG spotter fails to designate the target), TAG has no further effect. If the to-hit roll is  successful,  the system designates the target for that turn’s Weapon Attack Phase; the target unit is designated for any number of attacks from any number of units using TAG-guided ammunition, such as semi-guided missiles (p. 142).
**Successful TAG designation also spots the target for Indirect LRM Fire (see p. 111),** but as it occurs before the Weapon Attack Phase, no spotting to-hit modifier is incurred if the unit also fires weapons in the Weapon Attack Phase. This counts as the one target a unit can spot each turn.

PR #1296, opened 2019-03-08
Enable spotting with TAG

NickAragua:
Implements #1095 - getting TAGged now serves as a slightly better spot for indirect-fire LRMs.

### DIFF
--- a/megamek/src/megamek/common/Compute.java
+++ b/megamek/src/megamek/common/Compute.java
@@ -861,9 +861,14 @@ public class Compute {
                 los.setTargetCover(LosEffects.COVER_NONE);
                 mods.append(Compute.getAttackerMovementModifier(game,
                                                                 other.getId()));
-                if (other.isAttackingThisTurn() && !other.getCrew().hasActiveCommandConsole()) {
+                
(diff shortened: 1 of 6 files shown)


megamek/src/megamek/common/Compute.java
```diff
@@ -861,9 +861,14 @@ public static Entity findSpotter(IGame game, Entity attacker,
                 los.setTargetCover(LosEffects.COVER_NONE);
                 mods.append(Compute.getAttackerMovementModifier(game,
                                                                 other.getId()));
-                if (other.isAttackingThisTurn() && !other.getCrew().hasActiveCommandConsole()) {
+                
+                // a spotter suffers a penalty if it's also making an attack this round
+                // unless it has a command console or has TAGged the target
+                if (other.isAttackingThisTurn() && !other.getCrew().hasActiveCommandConsole() && 
+                        (!isTargetTagged(null, target, game) || (taggedBy != -1))) {
```
NickAragua: This logic isn't quite right, as the target has to be tagged by the specific spotter for the penalty to not apply.

Issue #1315, opened 2019-04-01
Add map annotations and description support to .board files


sixlettervariables:
In order to support visually impaired players, it would be nice to provide support for descriptions and annotations to boards.

1. Annotations of hexes, used to point out landmarks.
2. A general description of the board, explaining where features are (hopefully with annotations to help direct the user to those locations).

Based on the code in Board.java, this seems easy to add and would likely be backwards compatible.

sixlettervariables:
The two new items would be:
```
note <HEX> "Annotation about the hex"
```
and
```
description "A general description of the map"
description "each of these being their own paragraph"
```

PR #1319, opened 2019-04-05
Map annotations and description

sixlettervariables:
This extends the board files to support an overall description and hex annotations per #1315 . This feature serves two purposes:
1. Allows text descriptions of the map and certain hexes to be available to visually impaired players.
2. Allows fluff to be added to scenario maps.

The board files now support:
```
description "This is the first paragraph of the description."
description "This is the second paragraph."
note 0716 "This is a note about hex (7, 16)"
note 0716 "This is another note about hex (7, 16)"
```

This is not currently wired up into the GUI in any way.


sixlettervariables:
Yes there is, but the first go will be to enable chatbox commands to get the information.

PR #1329, opened 2019-04-17
Multi target shooting

NickAragua:
This set of changes allows the bot to consider shooting at multiple targets, as long as the unit doing the shooting isn't subject to secondary target penalties (incorporating secondary target penalties is a lot more difficult).

This affects large spacecraft, multi-crewed vehicles (when the appropriate tacops rule is turned on), units with the MultiTrac quirk and units with the MultiTasker SPA, greatly increasing their effectiveness on the field.

A side effect is that the bot will also incorporate the "combat computer" quirk into its firing plan for those units.

### DIFF
--- a/megamek/src/megamek/client/bot/princess/MultiTargetFireControl.java
+++ b/megamek/src/megamek/client/bot/princess/MultiTargetFireControl.java
@@ -1,3 +1,17 @@
+/*
+ * MegaMek - Copyright (C) 2019 Megamek Team
+ *
+ *  This program is free software; you can redistribute it and/or modify it
+ *  under the terms of the GNU General Public License as published by the Free
+ *  Software Foundation; either version 2 of the License, or (at your option)
+ *  any later version.
+ *
+ *  This program is distributed in the hope that it will be useful, but
+ *  WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ *  or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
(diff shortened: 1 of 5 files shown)


megamek/src/megamek/client/bot/princess/MultiTargetFireControl.java
```diff
@@ -0,0 +1,389 @@
...
+     * @param weapon Weapon to fire.
+     * @return The weapon fire info with the most expected damage. Null if no such thing.
+     */
+    WeaponFireInfo getBestShot(Mounted weapon) {
+        WeaponFireInfo bestShot = null;
+        
+        for(Targetable target : getTargetableEnemyEntities(weapon.getEntity(), owner.getGame(), owner.getFireControlState())) {
+            final int ownerID = (target instanceof Entity) ? ((Entity) target).getOwnerId() : -1;
+            if(owner.getHonorUtil().isEnemyBroken(target.getTargetId(), ownerID, owner.getBehaviorSettings().isForcedWithdrawal())) {
+                owner.log(getClass(), "MultiTargetFireControl.getBestShot", LogLevel.INFO, target.getDisplayName() + " is broken - ignoring");
```
sixlettervariables: Should this be WARN or DEBUG instead of Info?

NickAragua: It's "INFO" in the other places where isEnemyBroken is invoked, so I followed that pattern. I could be convinced to call it DEBUG as well.

PR #1330, opened 2019-04-18
Accessibility WIndow Integration

BlindGuyNW:
The Accessibility work is nowhere near complete, but I wondered if it wasn't a good place to integrate the window into the main code. I've got a bunch of new stuff for the client commands and some planned documentation I want to add later, but for now this is the raw window I've been using to play.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/AccessibilityWindow.java
+++ b/megamek/src/megamek/client/ui/swing/AccessibilityWindow.java
@@ -28,8 +28,8 @@ public class AccessibilityWindow extends JDialog implements KeyListener {
     private Coords selectedTarget;
     private JTextField inputField;
     private String[] args;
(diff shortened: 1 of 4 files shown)


megamek/src/megamek/client/ui/swing/AccessibilityWindow.java
```diff
@@ -0,0 +1,190 @@
...
+    public static final String ACCESSIBLE_GUI_SHORTCUT = ".";
+
+    Client client;
+    ClientGUI gui;
+    JTextArea chatArea;
+
+    private Coords selectedTarget;
+    private JTextField inputField;
+    private String[] args;
+    public LinkedList<String> history;
```
NickAragua: It looks like neither history, nor historyBookmark need to be public, so let's make them private.

PR #1336, opened 2019-04-22
Fix for issue 1331, Ejected Pilot Allowed Excess Movement

### LINKED ISSUE #1331: Ejected Pilot Allowed Excess Movement
### Environment
MM 0.45.4 with MHQ 0.45.4
Java (build 1.8.0_211-b12)
Win10

### Description
An ejected mech pilot was able to move through light woods for free.  I do not believe that I've seen infantry units doing this.  The game did complete the move.

sprint
[sprint.sav.gz](https://github.com/MegaMek/megamek/files/3099695/sprint.sav.gz)

[Steel Recon30261027.zip](https://github.com/MegaMek/megamek/files/3099700/Steel.Recon30261027.zip)
[customs.zip](https://github.com/MegaMek/megamek/files/3099701/customs.zip)

wildj79:
Fixes #1331.

In my testing, I found that infantry, which includes ejected mechwarriors, that had the Terrain Master (Forest Ranger) SPA were able to move through light woods at no apparent cost. The best fix that I could find for now was adding an `if` statement in `MoveStep::calcMovementCostFor(IGame,MoveStep)` that checks if `mp <= 0` and if it is sets `mp = 1`. I included this `if` statement inside of the code block starting on line 3086, which is specific to checking if an infantry is moving through woods or not. I was going to place it at the very end of the method, but wasn't sure if that could potential foul up calculations for other units.

### DIFF
--- a/megamek/src/megamek/common/MoveStep.java
+++ b/megamek/src/megamek/common/MoveStep.java
@@ -29,6 +29,8 @@ import java.util.TreeMap;
 import java.util.Vector;
 
 import megamek.common.MovePath.MoveStepType;
+import megamek.common.logging.DefaultMmLogger;
+import megamek.common.logging.MMLogger;
 import megamek.common.options.OptionsConstants;
 
 /**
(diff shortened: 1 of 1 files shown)


wildj79:
That was my initial thought and how I had originally wrote it. But, the movement code is... let's just say it's complex. I wasn't sure that keeping it at the end of the method wasn't going to have a negative affect on other unit's or not. So, I just put it in with the infantry specific code, since they seem to be the only unit's affected... at the moment.

PR #1358, opened 2019-05-31
fixes for a logic error and NPE

NickAragua:
This fixes a logic error that prevented aerospace fighters on ground maps from considering non-redundant paths (there are plenty of redundant paths that can be generated normally, but when you slow an ASF down to 1/2, that's not the case).

Also, fixes an NPE the bot encounters when attempting to drop thunder bombs on a target.


megamek/src/megamek/client/bot/princess/Princess.java
```diff
@@ -610,7 +610,8 @@ protected void calculateFiringTurn() {
                     // damageMap for the target enemy.
                     for(WeaponFireInfo shot : plan) {
                         int targetId = shot.getTarget().getTargetId();
-                        double newDamage = damageMap.get(targetId) + shot.getExpectedDamage();
+                        double existingTargetDamage = damageMap.containsKey(targetId) ? damageMap.get(targetId) : 0;
```
sixlettervariables: If `damageMap` implements `Map<K,V>` you could use [`getOrDefault`](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#getOrDefault-java.lang.Object-V-):
```suggestion
                        double existingTargetDamage = damageMap.getOrDefault(targetId, 0.0);
```

NickAragua: It required an extra line of code change, but done.

PR #1362, opened 2019-06-05
Reset LAM/QV conversion mode after getting movement stats.

neoancient:
When formatting the TRO view, LAMs and QuadVees are cycled through their modes to get the movement stats, but they are not returned to the original mode. This causes the MP calculations to show incorrectly in MML.

Fixes MegaMek/megameklab#301: LAM reports Walk MP same as its Jump MP


neoancient:
The MechView preview uses the same approach, but does reset the conversion mode.

PR #1365, opened 2019-06-06
Fix lam deployment npe

NickAragua:
This change fixes an NPE when the bot considers deploying a LAM in fighter mode within weapons range of a hostile unit. The number of locations changes between fighter and mech/airmech modes, so we need to translate the location appropriately.

Got lazy and based it off my previous PR, which is why you see the extra file.

### DIFF
--- a/megamek/src/megamek/common/LandAirMech.java
+++ b/megamek/src/megamek/common/LandAirMech.java
@@ -2108,18 +2108,4 @@ public class LandAirMech extends BipedMech implements IAero, IBomber {
             }
         }
     }
-    
-    /**
-     * Override of Entity's getBadCriticals which translates the given location
-     * to its aero equivalent if the LAM is in fighter mode.
-     */
-    @Override
-    public int getBadCriticals(int type, int index, int loc) {
-        if(getConversionMode() == CONV_MODE_FIGHTER) {
-            return(super.getBadCriticals(type, index, getAeroLocation(loc)));
-        } else {
-            return(super.getBadCriticals(type,  index, loc));
-        }
-    }
-
 }
--- a/megamek/src/megamek/common/LandAirMech.java
+++ b/megamek/src/megamek/common/LandAirMech.java
@@ -2108,5 +2108,18 @@ public class LandAirMech extends BipedMech implements IAero, IBomber {
             }
         }
     }
+    
+    /**
+     * Override of Entity's getBadCriticals which translates the given location
+     * to its aero equivalent if the LAM is in fighter mode.
+     */
+    @Override
+    public int getBadCriticals(int type, int index, int loc) {
+        if(getConversionMode() == CONV_MODE_FIGHTER) {
+            return(super.getBadCriticals(type, index, getAeroLocation(loc)));
+        } else {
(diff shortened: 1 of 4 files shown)


megamek/src/megamek/common/LandAirMech.java
```diff
@@ -2108,5 +2108,18 @@ public void updateSensorOptions() {
             }
         }
     }
+    
+    /**
+     * Override of Entity's getBadCriticals which translates the given location
+     * to its aero equivalent if the LAM is in fighter mode.
+     */
+    @Override
+    public int getBadCriticals(int type, int index, int loc) {
```
neoancient: LandAirMech#getAeroLocation is intended to find the equivalent location for weapon firing arcs (and the JavaDoc should probably say so). It does not work for checking critical locations. This change breaks a lot of things. For instance, trying to check for damaged gyro while in fighter mode (which prevents conversion and adds a penalty to control rolls) will check locations in the head instead and never find them.

NickAragua: Is there something similar that maps critical locations between mech and fighter mode?

neoancient: I don't understand what you're trying to do. Changing mode doesn't change critical locations.

NickAragua: Combat this exception, which the bot gets when calculating deployment for a fighter-mode LAM.

java.lang.ArrayIndexOutOfBoundsException: 10
	at megamek.common.Entity.getCritical(Entity.java:4313)
	at megamek.common.Entity.getBadCriticals(Entity.java:4506)
	at megamek.common.LandAirMech.getBadCriticals(LandAirMech.java:2121)
	at megamek.common.Compute.getDamageWeaponMods(Compute.java:2085)
	at megamek.common.actions.WeaponAttackAction.toHit(WeaponAttackAction.java:1797)
	at megamek.common.actions.WeaponAttackAction.toHit(WeaponAttackAction.java:273)
	at megamek.client.bot.BotClient.getDeployDamage(BotClient.java:934)
	at megamek.client.bot.BotClient.getStartingCoordsArray(BotClient.java:707)
	at megamek.client.bot.princess.Princess.calculateDeployment(Princess.java:385)
	at megamek.client.bot.BotClient.calculateMyTurn(BotClient.java:498)
	at megamek.client.bot.BotClient.access$0(BotClient.java:466)
	at megamek.client.bot.BotClient$CalculateBotTurn.run(BotClient.java:104)

Basically, the botClient class is checking potential attacks to the potential deployment hex, and the toHit method is checking for shoulder actuator damage in location 10 of the potential firing unit, which is the wings, but the entity only has 7 crits (it's a Stinger LAM STG-A10).

NickAragua: Ok, changed the approach. Now, the getDamageWeaponMods only checks for arm actuator damage in mechs, and only if the weapon in question is mounted on an arm. As you pointed out, this will also result in a slight performance boost as we are scanning through a lot fewer critical slots.

Also fixed a cast error that occurred when processing weapon hits by LAMs in fighter mode + part of a squadron.

Issue #1352, opened 2019-05-26
Changing Planetary Conditions prior to starting a new game causes a -40C without input


CrimsonThunderSB:
### Environment
Megamek 0.45.4

### Description
In the initial megamek menu, when I attempted change any parameters with the planetary conditions (temperature change, weather conditions, wind speeds, etc.) It will automatically set the temperature to -40C in which I cannot change back unless I start a new game again, and I am trying to create a custom scenario where Infantry is involved, and the bugged temperature to that will make the infantry die instantly on deployment due to extreme temperature.

### Files
N/A as this can be reproduced without a saved game file.

wildj79:
Believe it or not, this is actually working rules as written, see pages 60 and 62 in Tactical Operations. That being said, I think it is reasonable to allow for adjusting the temperature up or down a few degrees.

PR #1369, opened 2019-06-06
Update for PlanetaryConditions.

wildj79:
This updates `PlanetaryConditions.java` to fix a couple pieces of errata, and to allow a user to edit the temperature field after setting a weather condition.

Here's a list of changes to `PlanetaryConditions.java`:

* The logic in the `PlanetaryConditions::alterConditions(PlanetaryConditions)` method that was using `runOnce` seemed backwards to me. I flipped the logic so it seems more "correct"
* The `setRunOnce(boolean)` method takes a `boolean` paramter, but was setting the `runOnce` field explicitly to `true`. I corrected the method to set `runOnce` to whatever the value of the parameter is.
* The `clone()` method wasn't coping the `runOnce` field, which caused some issues.

Changes made to `PlanetaryConditionsDialog.java`:

* `setConditions()` was setting `conditions.setRunOnce(true)` on every call, so the `PlanetaryConditions::alterConditions(PlanetaryConditions)` method was always calling the `PlanetaryConditions::setTempFromWeather()`, `PlanetaryConditions::setWindFromWeather()`, and `PlanetaryConditions::setSandStorm()` methods.
* Added a `currentWeather` field to the class, to keep track of the currently selected weather condition is.
* Added an `ItemListener` to the `choWeather` combo box.
* Added a check to make sure that the inputted temperature isn't above 30 degrees Celsius when using snow or hail.

This fixes issue #1352

### DIFF
--- a/megamek/src/megamek/client/ui/swing/PlanetaryConditionsDialog.java
+++ b/megamek/src/megamek/client/ui/swing/PlanetaryConditionsDialog.java
@@ -57,6 +57,7 @@ public class PlanetaryConditionsDialog extends JDialog implements
     private ClientGUI client;
     private JFrame frame;
     private PlanetaryConditions conditions;
+    private int currentWeather;
     public PlanetaryConditions getConditions() {
         return conditions;
     }
@@ -145,6 +146,23 @@ public class PlanetaryConditionsDialog extends JDialog implements
             @Override
             public void windowClosing(WindowEvent e) {
                 setVisible(false);
+            }
+        });
+
+        choWeather.addItemListener(e -> {
+            int index = choWeather.getSelectedIndex();
+            if (currentWeather != index && 
+                    (index == PlanetaryConditions.WE_LIGHT_HAIL ||
+                     index == PlanetaryConditions.WE_HEAVY_HAIL ||
+                     index == PlanetaryConditions.WE_LIGHT_SNOW || 
+                     index == PlanetaryConditions.WE_SLEET ||
+                     index == PlanetaryConditions.WE_SNOW_FLURRIES ||
+                     index == PlanetaryConditions.WE_HEAVY_SNOW ||
+                     index == PlanetaryConditions.WE_ICE_STORM || 
+                     index == PlanetaryConditions.WE_BLIZZARD ||
+                     index == PlanetaryConditions.WE_MOD_SNOW)) {
+                currentWeather = index;
+                conditions.setRunOnce(false);
             }
         });
 
@@ -301,6 +319,7 @@ public class PlanetaryConditionsDialog extends JDialog implements
                     .addItem(PlanetaryConditions.getWeatherDisplayableName(i));
         }
         choWeather.setSelectedIndex(conditions.getWeather());
+        currentWeather = conditions.getWeather();
 
         choWind.removeAllItems();
         choMinWind.removeAllItems();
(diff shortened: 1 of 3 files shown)


megamek/src/megamek/client/ui/swing/PlanetaryConditionsDialog.java
```diff
@@ -148,6 +149,23 @@ public void windowClosing(WindowEvent e) {
             }
         });
 
+        choWeather.addItemListener(e -> {
```
neoancient: In Swing an `ItemListener` fires twice when the selection changes: once for deselecting the previous item and again for selecting the new item. An `ActionListener` only fires once when the selection changes, and is preferred except in the rare cases when you want to know which item was previously selected.

wildj79:
I noticed that as well. I'll play around with making the field update dynamically, instead of only when the dialog is closed.

megamek/src/megamek/client/ui/swing/PlanetaryConditionsDialog.java
```diff
@@ -148,6 +149,24 @@ public void windowClosing(WindowEvent e) {
...
+                    (weather == PlanetaryConditions.WE_LIGHT_HAIL ||
+                     weather == PlanetaryConditions.WE_HEAVY_HAIL ||
+                     weather == PlanetaryConditions.WE_LIGHT_SNOW || 
+                     weather == PlanetaryConditions.WE_SLEET ||
+                     weather == PlanetaryConditions.WE_SNOW_FLURRIES ||
+                     weather == PlanetaryConditions.WE_HEAVY_SNOW ||
+                     weather == PlanetaryConditions.WE_ICE_STORM || 
+                     weather == PlanetaryConditions.WE_BLIZZARD ||
+                     weather == PlanetaryConditions.WE_MOD_SNOW)) {
+                currentWeather = weather;
```
neoancient: `currentWeather` only gets updated if set to winter conditions. If you change from snow to non-snow, `currentWeather` remains set to the snow conditions. So if I go to light snow, then adjust the temp from -40 to 5, then clear, I get different behavior if I set it back to light snow vs. a different snow condition. While seems likely to confuse and frustrate users. in addition to the weird snow temperature rules that we didn't make up to begin with.

megamek/src/megamek/client/ui/swing/PlanetaryConditionsDialog.java
```diff
@@ -348,7 +368,10 @@ private void refreshConditions() {
     }
 
     private void setConditions() {
+            setConditions(true);
+    }
 
+    private void setConditions(boolean shouldSendToClient) {
```
neoancient: This has no effect on the way it works, but this doesn't actually send it to the client. It instructs the client to send it to the server.

wildj79: Changed it the parameter name to `shouldSendToServer`.

megamek/src/megamek/client/ui/swing/PlanetaryConditionsDialog.java
wildj79: So, moving `currentWeather = weather;` outside of the `if` statement produces an unwanted side effect. I think it produces the dynamic behavior we're looking for, but the temperature doesn't stick. I think it's getting set correctly when the dialog box is closed, it's just that the field get's reset when you open the dialog back up, so it looks like it isn't working (need to test that assumption further). That being said, I think the question to ask is, what is the desired behavior? I think refactoring the dialog would be easier if we have a clear cut idea of how it's supposed to function based on the funky weather rules.

neoancient: In that case I think this is probably enough at this point. Looking at this makes me want to rewrite all the planetary conditions using enums but I'm not going to do that right now.

PR #1371, opened 2019-06-07
Fix for flamer heat damage.

### LINKED ISSUE #1353: Battle Armor Heavy Flamer only raise 2 heat per hit
### Environment
Ver 0.45.4
Windows 7

### Description
As title says. Heavy Flamer says it raises heats as much as it can normally damages, but Battle Armor Heavy Flamer only raise 2 heat per hit when I choose to raise the heat.

neoancient:
The flamer weapon handler always applies two points of heat damage. The
BA heavy flamer should get four points, and the ER flamers should do
half damage when used in heat mode (TacOps, 312). The ammo-based flamers
(vehicle, heavy) use a different handler which applies the correct
damage.

Fixes #1353: Battle Armor Heavy Flamer only raise 2 heat per hit

### DIFF
--- a/megamek/src/megamek/common/weapons/FlamerHeatHandler.java
+++ b/megamek/src/megamek/common/weapons/FlamerHeatHandler.java
@@ -30,6 +30,7 @@ import megamek.common.Mech;
 import megamek.common.Report;
 import megamek.common.TargetRoll;
 import megamek.common.ToHitData;
+import megamek.common.WeaponType;
 import megamek.common.actions.WeaponAttackAction;
 import megamek.common.options.OptionsConstants;
 import megamek.server.Server;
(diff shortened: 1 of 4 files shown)


megamek/src/megamek/common/weapons/FlamerHeatHandler.java
```diff
@@ -82,21 +83,27 @@ protected void handleEntityDamage(Entity entityTarget,
             r = new Report(3400);
             r.subject = subjectId;
             r.indent(2);
-            if (entityTarget.getArmor(hit) > 0 && 
+            int heatDamage = wtype.getDamage();
+            if (wtype.hasFlag(WeaponType.F_ER_FLAMER)) {
+                heatDamage = Math.max(1,  heatDamage / 2);
+            }
+           if (entityTarget.getArmor(hit) > 0 &&
```
sixlettervariables: whitespace got a little off at this point

PR #1382, opened 2019-06-14
Fix LAM bomb bay intro dates.

neoancient:
I'm not sure where the dates we have for LAM bomb bay tech progression came from, but an intro date of 3071 is clearly wrong. I couldn't find a listing for it in the master tech advancement table in IO, but since it is a LAM-only piece of equipment I combined the dates for bimodal and standard LAMs, using the earliest intro date and latest extinction date.

Fixes MegaMek/megameklab#313


neoancient:
Actually I do know where it came from. Before IO, 3071 was our reference date for things like availability and tech level. This one never got updated because it doesn't have an entry in the master table.

Issue #1388, opened 2019-06-16
Successful TAG designation cause the unit to suffer the penalty of spot for LRM indirect fire on the same turn


binaryspica:
### Environment
Ver 0.45.4
Windows 7

### Description
Honestly, I have no idea why it does. It seems that whatever the artillery is onboard or offboard it have nothing to do with this. And both homing and nonhoming rounds are affected by it. Although the save files are only uses Arrow IV, I have seen the same problem on tube artillery as well before that(and that's why I did the test run with this).

On the game, friendly fire option is checked and all units are on your control. I doubt that it would be the reason for this, though.

And... by the result of savegame002 and 003, I suspect that the unit that already fires TAG on the same round are suffered by it. Units that not fired TAG are unaffected by it, if I check it correctly.

### Files
[savegame001.sav.gz](https://github.com/MegaMek/megamek/files/3294332/savegame001.sav.gz)
[savegame002.sav.gz](https://github.com/MegaMek/megamek/files/3294345/savegame002.sav.gz)

[savegame003.sav.gz](https://github.com/MegaMek/megamek/files/3294349/savegame003.sav.gz)
(savegame003 have few turns more after 002) 

![p_](https://user-images.githubusercontent.com/47907439/59566988-57951300-90a2-11e9-8bb3-4e9934861a59.png)

![p_02](https://user-images.githubusercontent.com/47907439/59567071-ad1def80-90a3-11e9-918d-dc242b94485d.png)
![p_03](https://user-images.githubusercontent.com/47907439/59567072-b018e000-90a3-11e9-9823-8ef840fdc0c3.png)

binaryspica:
![p_15](https://user-images.githubusercontent.com/47907439/59845098-7bed3a00-9397-11e9-9c2b-b40d45f6ed61.png)
![p_16](https://user-images.githubusercontent.com/47907439/59845101-7ee82a80-9397-11e9-9607-73795f4359b2.png)



It seems that I was totally get the wrong point. The problem is **TAG itself**, not the artillery or anything else.

Per errata of Total Warfare, **TAG (Target Acquisition Gear) (p. 142)** section says;

> .. Successful TAG designation also spots the target for Indirect LRM Fire (see p.111), but as it occurs before the Weapon Attack Phase, no spotting to-hit modifier is incurred if the unit also fires weapons in the Weapon Attack Phase. This counts as the one target a unit can spot each turn.

[...]

PR #1396, opened 2019-06-24
Remove spot for LRM indirect fire penalty for the unit that was successfully designated TAG on the same turn.

binaryspica:
Fix for issue #1388 .

I was reviewed the pull request #1296 before making a fix for the problem, because I think that the problem is related with this. 

Actually, pull request #1296 was intended to fix for it. But, it seems that it wasn't fully cover the problem. Anyway my solution is just adding something to that.

It is simple - add a flag on Entity that notes the entity spots without penalty, and see if the unit will not suffer the penalty before put it.

But, because I am not familiar to MegaMek's code, and I am not fully understand the structure, so I think that it may cause a bug. So please check it with care. Also, I fear that I might cause a serious problem because I have touched Entity class, which seems one of the most used class in the application. But, after search through Entity.Java it seems that a new flag was needed to solve the problem.

### DIFF
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -354,6 +354,8 @@ public abstract class Entity extends TurnOrdered implements Transporter,
     private boolean evading = false;
 
     public boolean spotting;
+    public boolean spottingWithoutPenalty; // the unit has spotted for LRM indirect fire but suffers no penalty for it
+
     private boolean clearingMinefield = false;
     protected int killerId = Entity.NONE;
     private int offBoardDistance = 0;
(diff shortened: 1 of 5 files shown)


binaryspica:
So, if it is possible to simply check if the entity is on the list of unit that was tagged something, I don't need to touch Entity. Thanks, but I think that it needs some time to delve into it....

NickAragua:
That's fine, we're not in a hurry. :)

NickAragua:
Yep, looking at the code, you're looking for game.getTagInfoForTurn. Maybe write a method that takes an entity ID and returns true if that tag info collection contains the entity ID you're looking for.

binaryspica:
Thanks for the info. But I think that I should close it and make a new pull request, because I can't modify the list of it....

PR #1398, opened 2019-06-27
Remove spot for LRM indirect fire penalty for the unit that was successfully designated TAG on the same turn.  - take 2

binaryspica:
It is from #1396 . The reason for the change is same as it - fix for issue #1388 .

By the advice, I did making it with only touch Physical/WeaponAttackAction. Only these two files are have the phrase "attacker is spotting for indirect LRM fire" in MegaMek so it would be enough.

And, while check [TagInfo.java](https://github.com/MegaMek/megamek/blob/master/megamek/src/megamek/common/TagInfo.java) I found that the class have a boolean variable named 'missed'. It seems that the code need to check it as well but is the variable have any use on MegaMek? I have searched the code but it seems that nothing uses it, so it would cause a bug if I add it to the checker but nothing else uses or changes it.


megamek/src/megamek/common/actions/PhysicalAttackAction.java
```diff
@@ -211,8 +211,23 @@ protected static void setCommonModifiers(ToHitData toHit, IGame game,
...
 
         // if we're spotting for indirect fire, add +1
-        if (ae.isSpotting() && !ae.getCrew().hasActiveCommandConsole()) {
-            toHit.addModifier(+1, "attacker is spotting for indirect LRM fire");
+        if (ae.isSpotting() && !ae.getCrew().hasActiveCommandConsole()) {           
+            //but if it spots by TAG it doesn't.
+            boolean isSpotByTag = false;//temporal boolean flag
+            
+            //search through tagInfoForTurn and see if the entity has successfuly designated TAG on something.
+            Vector<TagInfo> tmpTagInfoForTurn = game.getTagInfo();
```
neoancient: This can be simplified a lot by using the foreach type of for loop:
```
for (TagInfo tmpInfo : game.getTagInfo()) {
    if (tmpInfo.attackerId == attackerId) {
        isSpotByTag = true;
        break;
    }
}

File: megamek/src/megamek/common/actions/PhysicalAttackAction.java [hunk 2]
```diff
@@ -211,8 +211,23 @@ protected static void setCommonModifiers(ToHitData toHit, IGame game,
         }
 
         // if we're spotting for indirect fire, add +1
-        if (ae.isSpotting() && !ae.getCrew().hasActiveCommandConsole()) {
-            toHit.addModifier(+1, "attacker is spotting for indirect LRM fire");
+        if (ae.isSpotting() && !ae.getCrew().hasActiveCommandConsole()) {           
+            //but if it spots by TAG it doesn't.
+            boolean isSpotByTag = false;//temporal boolean flag
```
neoancient: Comments that explain how the code works are usually good. Comments like this that explain how the language works are superfluous and add unnecessary noise.

megamek/src/megamek/common/actions/PhysicalAttackAction.java
```diff
@@ -211,8 +211,23 @@ protected static void setCommonModifiers(ToHitData toHit, IGame game,
...
+            
+            //search through tagInfoForTurn and see if the entity has successfuly designated TAG on something.
+            Vector<TagInfo> tmpTagInfoForTurn = game.getTagInfo();
+            for (int i = 0; i < tmpTagInfoForTurn.size(); i++){
+                TagInfo tmpInfo = tmpTagInfoForTurn.elementAt(i);
+                if (tmpInfo.attackerId == attackerID){
```
neoancient: Java is case sensitive. The local variable you are trying to match is `attackerId`, not `attackerID`.

megamek/src/megamek/common/actions/WeaponAttackAction.java
```diff
@@ -789,7 +789,22 @@ else if ((te != null) && (ae.getSwarmTargetId() == te.getId())) {
...
         if (ae.isSpotting() && !ae.getCrew().hasActiveCommandConsole()) {
-            toHit.addModifier(+1, "attacker is spotting for indirect LRM fire");
+	    //but if it spots by TAG it doesn't.
+            boolean isSpotByTag = false;//temporal boolean flag
+            
+            //search through tagInfoForTurn and see if the entity has successfuly designated TAG on something.
+            Vector<TagInfo> tmpTagInfoForTurn = game.getTagInfo();
+            for (int i = 0; i < tmpTagInfoForTurn.size(); i++){
+                TagInfo tmpInfo = tmpTagInfoForTurn.elementAt(i);
+                if (tmpInfo.attackerId == attackerID){
```
neoancient: Java is case sensitive. The local variable you are trying to match is `attackerId`, not `attackerID`.

binaryspica: Ha, what a silly mistake! Thanks for review that.

neoancient:
FYI we don't get a notification when you push more changes to the open branch. I saw that you made the requested changes because I was looking over the open PRs.

PR #1403, opened 2019-07-03
fix nonsensical infantry behavior

NickAragua:
Now that the bot regularly gets field gunners, I noticed that they love running out into the open to meet my Firestarters in glorious battle, instead of camping out in the woods and firing at long range.

Once I made this change, a platoon of AC/5 field gunners changed their behavior from running out into the open and opening fire with their auto-rifles to camping out in the woods and getting in two engine crits. This will probably also benefit their non-field-gunner counterparts.

Technical notes:
- Max damage evaluation wasn't taking into account the fact that field guns cannot be fired if a unit has moved.
- "Unit in open" check was returning results based on the unit's current hex rather than unit's destination hex.
- "aggression" modifier was causing infantry to want to close in anyway, even though it's not to their benefit. Changed it to only apply aggression modifier if the path being considered will result in no damage being inflicted by the unit.

### DIFF
--- a/megamek/src/megamek/client/bot/princess/InfantryFireControl.java
+++ b/megamek/src/megamek/client/bot/princess/InfantryFireControl.java
@@ -36,20 +36,25 @@ public class InfantryFireControl extends FireControl {
      *            Is the extreme range optional rule in effect?
      * @return The most damage done at that range.
      */
-    public double getMaxDamageAtRange(final Entity shooter, final MovePath path, final int range,
+    public double getMaxDamageAtRange(final MovePath shooterPath, final MovePath targetPath, final int range,
             final boolean useExtremeRange, final boolean useLOSRange) {
         double maxFGDamage = 0;
         double maxInfantryWeaponDamage = 0;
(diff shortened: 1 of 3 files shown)


HammerGS:
This pull request **introduces 1 alert** when merging c53cff1985b8e04a2d7809397b90e7490e442de4 into 3f0e37d8ee5b06fe8e161d9f879c1a652fda8663 - [view on LGTM.com](https://lgtm.com/projects/g/MegaMek/megamek/rev/pr-976cd6a13d066e0846119a887762ec497b7186bd)

**new alerts:**

* 1 for Spurious Javadoc @param tags

---

This project has automated code review enabled, but doesn't use the <a href="https://github.com/apps/lgtm-com">LGTM GitHub App</a>. Migrate over by <a href="https://github.com/apps/lgtm-com/installations/new">installing the app</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

PR #1406, opened 2019-07-06
first pass at bot spotting functionality

NickAragua:
Allow the bot to spot for indirect fire. In this implementation it will do so as an afterthought, if it can't shoot at anything or if doing so won't cause any "revisions" to a particular unit's firing plan. It also won't be particularly smart about what it tries to spot, opting for the nearest target. Future revisions will make an attempt to spot the target with the lowest to-hit modifier instead. But still, this ought to give those stationary MG turrets a little more bite when paired with LRM-equipped units stuck in a cul-de-sac half a map away.

To support this, I needed to implement a few things:

- units that can do indirect fire now go after units that cannot
- this evaluation is cached for the turn to avoid repeated re-calculation


megamek/src/megamek/client/bot/princess/FireControlState.java
```diff
@@ -32,4 +40,31 @@ public FireControlState() {
...
+    
+    /**
+     * Accessor for the data structure containing a mapping between entities and whether or not
+     * they have indirect fire capability as in LRMs.
+     */
+    public Map<Integer, Boolean> getEntityIDFStates() {
+    	return entityIDFStates;
+    }
+
+	public LinkedList<Entity> getOrderedFiringEntities() {
```
sixlettervariables: Whitespace got off at this point.

megamek/src/megamek/client/bot/princess/Princess.java
```diff
@@ -854,6 +868,41 @@ protected void calculateTargetingOffBoardTurn() {
         }
     }
 
+    /**
+     * Gets an entity eligible to fire from a list contained in the fire control state.
+     */
+    Entity getEntityToFire(FireControlState fireControlState) {
```
sixlettervariables:
```suggestion
    private Entity getEntityToFire(FireControlState fireControlState) {
```

megamek/src/megamek/client/bot/princess/Princess.java
```diff
@@ -552,7 +553,7 @@ protected void calculateFiringTurn() {
         try {
             // get the first entity that can act this turn make sure weapons 
             // are loaded
-            final Entity shooter = game.getFirstEntity(getMyTurn());
+            final Entity shooter = getEntityToFire(fireControlState);//ame.getFirstEntity(getMyTurn());
```
sixlettervariables:
```suggestion
            final Entity shooter = getEntityToFire(fireControlState);
```

megamek/src/megamek/client/bot/princess/Princess.java
```diff
@@ -624,7 +624,13 @@ protected void calculateFiringTurn() {
                     }
     
                     // tell the game I want to fire
-                    sendAttackData(shooter.getId(), plan.getEntityActionVector());
+                    Vector<EntityAction> actions = plan.getEntityActionVector();
+                    EntityAction spotAction = getFireControl(shooter).getSpotAction(plan, shooter, fireControlState);
```
sixlettervariables: `getFireControl` is always non-null?

NickAragua: Yeah, they get initialized whenever a Princess first loads up. The default is the standard FireControl.

Issue #1407, opened 2019-07-07
[0.45.4] Artillery attack on APC with 2 infantry units stops phase from ending


Da-Nuker:
Megamek 0.45.4 as part of MekHQ
Win Java 1.8.0_211 64-bit

Launching an artillery weapon (tested Arrow IV artillery missile) onto an APC carrying two units of Battle Armor (tested Badger (C) Prime, and two Elemental Points) causes Megamek to freeze at end of firing phase.

The game does not freeze if it was there was only a single unit of BA carried.

This error message log was found in megameklog:
Exception in thread "Packet Pump" java.lang.ArrayIndexOutOfBoundsException: 6
	at megamek.common.Entity.getInternalForReal(Entity.java:3146)
	at megamek.common.Entity.getInternal(Entity.java:3139)
	at megamek.common.Infantry.getInternal(Infantry.java:732)
	at megamek.common.BattleArmor.getInternal(BattleArmor.java:783)
	at megamek.common.BattleArmor.getTrooperAtLocation(BattleArmor.java:705)
	at megamek.server.Server.damageExternalPassenger(Server.java:24894)
	at megamek.server.Server.damageEntity(Server.java:23675)
	at megamek.server.Server.damageEntity(Server.java:23018)
	at megamek.server.Server.artilleryDamageHex(Server.java:36732)
	at megamek.server.Server.artilleryDamageArea(Server.java:36843)
	at megamek.server.Server.artilleryDamageArea(Server.java:36804)
	at megamek.common.weapons.ArtilleryWeaponIndirectFireHandler.handle(ArtilleryWeaponIndirectFireHandler.java:452)
	at megamek.server.Server.handleAttacks(Server.java:37078)
	at megamek.server.Server.handleAttacks(Server.java:37007)
	at megamek.server.Server.endCurrentPhase(Server.java:3136)
	at megamek.server.Server.changeToNextTurn(Server.java:3473)
	at megamek.server.Server.endCurrentTurn(Server.java:2463)
	at megamek.server.Server.receiveAttack(Server.java:14058)
	at megamek.server.Server.handle(Server.java:32801)
	at megamek.server.Server$PacketPump.run(Server.java:369)

### Files
Load the save game as player "Naga" and with target player "BA".
As Naga, Launch Arrow IV in direct fire mode at the surviving Badger (C) hex.

[autosave.sav.gz](https://github.com/MegaMek/megamek/files/3365354/autosave.sav.gz)

[...]

PR #1410, opened 2019-07-08
Fix for #1407

mkerensky:
+ Bugfix #1407 - Ensure array of BA troopers contains a trooper 5 or 6 before checking for damage

### DIFF
--- a/megamek/src/megamek/common/BattleArmor.java
+++ b/megamek/src/megamek/common/BattleArmor.java
@@ -702,17 +702,20 @@ public class BattleArmor extends Infantry {
                 }
                 break;
             case Tank.LOC_REAR:
-                //These troopers only exist when you have Clan and CS/WoB units, so we need
-                //to ensure the array is large enough before checking whether they're dead.
-                if (((locations() >= 6) && (getInternal(5) > 0)) && ((locations() >= 7) && (getInternal(6) > 0))) {
+                //Troopers 5 and 6 only exist when you have Clan and CS/WoB units, so we need
+                //to ensure the array is large enough before checking their status.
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/BattleArmor.java
```diff
@@ -702,9 +702,11 @@ public HitData getTrooperAtLocation(HitData hit, Entity transport) {
...
                 break;
             case Tank.LOC_REAR:
-                if ((getInternal(5) > 0) && (getInternal(6) > 0)) {
+                //These troopers only exist when you have Clan and CS/WoB units, so we need
+                //to ensure the array is large enough before checking whether they're dead.
+                if (((locations() >= 6) && (getInternal(5) > 0)) && ((locations() >= 7) && (getInternal(6) > 0))) {
                     loc = Compute.randomInt(2) + 5;
-                } else if (getInternal(5) > 0) {
+                } else if ((locations() >= 6) && (getInternal(5) > 0)) {
                     loc = 5;
```
NickAragua: Seems like you'd want loc to be 6 here and 5 in the else block?

mkerensky: The arrays have 1 extra location for squad

NickAragua: I think I'm having trouble understanding the logic then. 

If we've got 7 troopers then we either pick trooper 6 or 7
If we've got 6 troopers then we pick trooper 6
Otherwise we pick trooper 7?

mkerensky: I made a couple of adjustments to the test logic that I think will take care of the issues here.

PR #1412, opened 2019-07-14
Shrink save game sizes by using XStream's ID_REFERENCES

sixlettervariables:
Currently MegaMek uses XStream's default means of serializing object graphs, which uses relative XPath references. During testing of larger aerospace games this resulted in a fair amount of save game bloat, and many object references were duplicated rather than being coalesced. This switches to using `ID_REFERENCES` which results in a cleaner, smaller XML file for save games. It also makes diffing save games much cleaner.


sixlettervariables:
As a note to those reading this PR in the future, this breaks savegame compatibility; however, we do not guarantee that between releases.

PR #1422, opened 2019-07-26
Game.java has a number of sync methods which confer no benefit (LGTM)

sixlettervariables:
- Multiple setters had synchronized modifiers without the remainder of the class properly synchronizing the data
- Playtesting showed no downside to removing them, however this is probably due some playtesting beyond just my 13 round slugfest with Princess.

### DIFF
--- a/megamek/src/megamek/common/Game.java
+++ b/megamek/src/megamek/common/Game.java
@@ -930,7 +930,7 @@ public class Game implements Serializable, IGame {
         return Collections.unmodifiableList(entities);
     }
 
-    public synchronized void setEntitiesVector(List<Entity> entities) {
+    public void setEntitiesVector(List<Entity> entities) {
         //checkPositionCacheConsistency();
         this.entities.clear();
         this.entities.addAll(entities);
@@ -1273,7 +1273,7 @@ public class Game implements Serializable, IGame {
         addEntity(entity, true);
     }
 
-    public synchronized void addEntity(Entity entity, boolean genEvent) {
+    public void addEntity(Entity entity, boolean genEvent) {
         entity.setGame(this);
         if (entity instanceof Mech) {
             ((Mech) entity).setBAGrabBars();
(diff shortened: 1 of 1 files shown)


HammerGS:
This pull request **fixes 3 alerts** when merging a4570367533c1b26481c53e094091f4a64ca3059 into 1449d302a7693904434df071e42b9c97a69ab239 - [view on LGTM.com](https://lgtm.com/projects/g/MegaMek/megamek/rev/pr-c212fd2b46ee75cc04ef230f82dd177c07c7ff6a)

**fixed alerts:**

* 3 for Inconsistent synchronization of getter and setter

---

This project has automated code review enabled, but doesn't use the <a href="https://github.com/apps/lgtm-com">LGTM GitHub App</a>. Migrate over by <a href="https://github.com/apps/lgtm-com/installations/new">installing the app</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

neoancient:
It's taken me a while to get a good look at this because concurrency is not something I can work on with any distractions.

LGTM has correctly identified that synchronizing these methods confers no benefit, but I think that rather than just removing the illusion of thread safety they should actually be made thread-safe. Most of that could be accomplished by replacing the collections with thread-safe implementations: ConcurrentHashMap, CopyOnWriteArrayList, and CopyOnWriteArraySet. I would also change the lastEntityId field to AtomicInteger, and generate new ids with getAndIncrement().

That would leave at least one outstanding concurrency issue, which is coordination between the entity list and the id lookup map. I don't know that there's a good reason to keep two collections, since Map#values will return an iterable collection of entities. If they had to be a list for some reason it would probably be cheaper overall to instantiate new ArrayList<>(entityMap.values()) than to maintain and coordinate the two collections. If they need to be ordered for some reason, then ConcurrentSkipListMap is the thread-safe alternative to TreeMap.

PR #1424, opened 2019-07-26
Work on making Vector's into List's

sixlettervariables:
This is a chunky PR that should be done one commit at a time. Basically, the aim here is to get rid of outdated usages of Vector / Enumeration and replace them with their more contemporary forms. I chose areas of IGame that don't appear to have active work being done in/around them.

This is not an essential PR, nor one I would have undertaken without working nights and having a lot of free time. If this level of updating isn't worth it, I'm happy to close the PR or cherry pick commits that were more useful than others.

### DIFF
--- a/megamek/src/megamek/client/Client.java
+++ b/megamek/src/megamek/client/Client.java
@@ -28,6 +28,7 @@ import java.util.Collection;
 import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.Hashtable;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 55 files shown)


neoancient:
`Game` should be thread-safe. It's deficient in that area, but I would be wary of replacing the thread-safe `Vector` with non-thread-safe `ArrayList`s. Without a more thorough examination I don't know which lists require a thread-safe `List` implementation, but I think replacing `Vector` with the more efficient (but still thread-safe) `CopyOnWriteArrayList` instead would be prudent.

PR #1430, opened 2019-08-07
Bot aero anti dropship behavior fix

NickAragua:
Four sets of changes:

1) Fixes a bot behavior where aircraft would love to present their tail directly to the front battery of hostile dropships.
2) Minor bot optimization: there are a lot of calls to "entity.getMaxWeaponRange()", repeated multiple times for each entity during each phase (probably # paths x # enemies on board). Caches these calls, since enemy max weapon range isn't affected by the path the unit is considering.
3) Fix bot estimation for weapon ranges of hostile air/spaceborne units
4) Clean-up user-visible range calculation adjustment for air/spaceborne units

### DIFF
--- a/megamek/src/megamek/client/bot/princess/Princess.java
+++ b/megamek/src/megamek/client/bot/princess/Princess.java
@@ -297,12 +297,8 @@ public class Princess extends BotClient {
      * @return
      */
     public int getMaxWeaponRange(Entity entity, boolean airborneTarget) {
-        if(!getFireControlState().getWeaponRanges(airborneTarget).containsKey(entity.getId())) {
-            getFireControlState().getWeaponRanges(airborneTarget).put(entity.getId(), 
-                    entity.getMaxWeaponRange(airborneTarget));
-        }
-        
-        return getFireControlState().getWeaponRanges(airborneTarget).get(entity.getId());
+        return getFireControlState().getWeaponRanges(airborneTarget).
+            computeIfAbsent(entity.getId(), ent -> entity.getMaxWeaponRange(airborneTarget));
     }
 
     public void setFallBack(final boolean fallBack,
--- a/megamek/src/megamek/client/bot/princess/Princess.java
+++ b/megamek/src/megamek/client/bot/princess/Princess.java
@@ -283,6 +283,27 @@ public class Princess extends BotClient {
     Precognition getPrecognition() {
         return precognition;
     }
+    
+    public int getMaxWeaponRange(Entity entity) {
+        return getMaxWeaponRange(entity, false);
+    }
+    
+    /**
+     * Retrieves maximum weapon range for the given entity.
+     * Cached version of entity.getMaxWeaponRange() 
+     * @param entity Entity we're checking
+     * @param airborneTarget Whether the potential target is in the air, only relevant for
+     *          aircraft shooting at other aircraft on ground maps.
+     * @return
+     */
+    public int getMaxWeaponRange(Entity entity, boolean airborneTarget) {
+        if(!getFireControlState().getWeaponRanges(airborneTarget).containsKey(entity.getId())) {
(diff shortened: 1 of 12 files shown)


megamek/src/megamek/client/ui/swing/unitDisplay/WeaponPanel.java
```diff
@@ -2053,8 +2053,8 @@ else if (atype.getMunitionType() == AmmoType.M_IATM_IMP)
         }
 
         // Aero
-        if (entity.isAirborne() 
-                || entity.usesWeaponBays()) {
+        if (entity.isAirborne()){
+                //|| entity.usesWeaponBays()) {
```
sixlettervariables: should this be removed?

megamek/src/megamek/client/bot/princess/Princess.java
```diff
@@ -283,6 +283,27 @@ public PathRankerState getPathRankerState() {
...
+    /**
+     * Retrieves maximum weapon range for the given entity.
+     * Cached version of entity.getMaxWeaponRange() 
+     * @param entity Entity we're checking
+     * @param airborneTarget Whether the potential target is in the air, only relevant for
+     *          aircraft shooting at other aircraft on ground maps.
+     * @return
+     */
+    public int getMaxWeaponRange(Entity entity, boolean airborneTarget) {
+        if(!getFireControlState().getWeaponRanges(airborneTarget).containsKey(entity.getId())) {
```
sixlettervariables: You could consider something like:
```java
return getFireControlState().getWeaponRanges(airborneTarget)
    .computeIfAbsent(entity.getId(), _ -> entity.getMaxWeaponRange(airborneTarget));
```

megamek/src/megamek/client/ui/swing/unitDisplay/WeaponPanel.java
mkerensky: What happens when you have a grounded dropper that still uses weapon bays?  I think that's what the || statement was there for.

megamek/src/megamek/client/bot/princess/Princess.java
NickAragua: Did not know that was a thing. I'll take a look at it.

megamek/src/megamek/client/ui/swing/unitDisplay/WeaponPanel.java
NickAragua: Do we know where the rules for grounded dropships weapon ranges actually are?

megamek/src/megamek/client/bot/princess/Princess.java
sixlettervariables: BTW: Java complains about underscore as an identifier, so I guess make that `id` instead.

megamek/src/megamek/client/ui/swing/unitDisplay/WeaponPanel.java
mkerensky: Errata actually changes them.  It's p249 of TW:

Attacks by DropShips are carried out with the exceptions noted above concerning firing arcs for grounded spheroid DropShips. Resolve each individual weapon as a separate attack, using ranges and damage values as found on pp. 303-304.

So no more bay fire by grounded droppers.

mkerensky: Fun.  Guess I know what my next task is...   At least we have the framework for it already in place - the trick is how to handle bay ammo so everything is linked properly once the dropper lifts off

NickAragua: On the plus side, the ranges from grounded dropship weapon bays never displayed correctly anyway, from what I can tell.

NickAragua:
Fixed the mentioned issues, and fixed one of the many unit test errors. Those I'll handle in a separate PR.

PR #1432, opened 2019-08-08
Equipment type key

neoancient:
This was a bit tedious, but designed to make me feel better about some of the code I'm writing. There are some places in the code where a specific equipment instance must be retrieved using `EquipmentType#get`. This replaces the use of String literals as parameters with String constants. Only a small portion of the EquipmentType objects have named constants because most of them do not require direct specific lookup.

Besides guarding against null EquipmentTypes by typo, it provides assurance that the equipment will exist because the constant is used to initialize the internal name. A unit test provides an extra measure of security against changing the possibility of lookup failure due to the lookup name being changed.

### DIFF
--- a/megamek/src/megamek/common/Infantry.java
+++ b/megamek/src/megamek/common/Infantry.java
@@ -1993,7 +1993,7 @@ public class Infantry extends Entity {
         if ((spec & TRENCH_ENGINEERS) > 0 && (infSpecs & TRENCH_ENGINEERS) == 0) {
             // Need to add vibro shovels
             try {
-                EquipmentType shovels = EquipmentType.get("Vibro-Shovel");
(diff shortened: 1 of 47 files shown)


megamek/src/megamek/common/Infantry.java
```diff
@@ -1993,7 +1993,7 @@ public void setSpecializations(int spec) {
         if ((spec & TRENCH_ENGINEERS) > 0 && (infSpecs & TRENCH_ENGINEERS) == 0) {
             // Need to add vibro shovels
             try {
-                EquipmentType shovels = EquipmentType.get("Vibro-Shovel");
+                EquipmentType shovels = EquipmentType.get(EquipmentTypeLookup.VIBRO_SHOVEL);
```
NickAragua: lol vibro-shovel

PR #1434, opened 2019-08-11
Remove unnecessary syncs around entities in Game.java

sixlettervariables:
...and replace them with a single concurrent data structure: a ConcurrentSkipListMap.

New patterns are introduced in IGame to:
1. Reduce unnecessary allocations.
2. Ensure correct iteration (w.r.t. concurrency).
3. Improve code readability (e.g. findEntitiesAt or hasEntitiesAt).
4. Speed up algorithms.

This is an omnibus PR which should be taken cautiously. I'm only a few hours into play testing this.

### DIFF
--- a/megamek/src/megamek/client/Client.java
+++ b/megamek/src/megamek/client/Client.java
@@ -26,6 +26,7 @@ import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Enumeration;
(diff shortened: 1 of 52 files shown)


sixlettervariables:
Okay, so this does not crash under heavy play testing...but it also looks like Physical attacks are no longer an option. Trying to track down exactly why.

PR #1451, opened 2019-08-21
enable bot usage of searchlights; minor optimizations

NickAragua:
Three changes:

- the bot calls the getEnemyEntities() and getFriendlyEntities() calculation a lot, so I changed it to a lazy-loaded method.
- get rid of the completely unused "PathSearcher" class so it quits clogging up my search results
- oh yeah, and give the bot the ability to use searchlights to light things up at night, preferring against lighting up friendly units.


NickAragua:
Some updates to searchlight logic to prevent the bot from piling multiple searchlights onto a single unit.

PR #1452, opened 2019-08-21
Support vee construction data

neoancient:
These are the MM changes needed for support vehicle construction in MML. It does not yet include support vehicle validation. I want to do that in a separate branch (since this is already such a large set of changes). There's a good chance the unit validation will reveal some things that need to be fixed so I'm posting this as a draft PR at this point.


megamek/src/megamek/common/verifier/TestSupportVehicle.java
```diff
@@ -14,34 +14,602 @@
...
-         {.000, .000, .000, .057, .051, .047},
-         {.000, .000, .000, .063, .056, .052},};
+public class TestSupportVehicle extends TestEntity {
+
+    /**
+     * Support vehicle categories for construction purposes. Most of these match with a particular movement
+     * mode, but the construction rules treat naval and rail units as single types.
+     */
+    public enum SVType implements ITechnologyDelegator {
+        AIRSHIP (300, EntityMovementMode.AIRSHIP,
```
NickAragua: I'm incapable of figuring out what the values in these data structures mean, is it possible to put an explanatory comment?

PR #1457, opened 2019-08-25
WeaponAttackAction Rework

mkerensky:
[MegaMek.zip](https://github.com/MegaMek/megamek/files/3551099/MegaMek.zip)
Cleans up WeaponAttackAction, removing some duplicate calculations and organizing the various to-hit modifiers and impossible shot strings.  Also improves on comments, fixes a couple of issues where rules were incorrectly implemented and adds a number of null guards.

This still needs testing and a couple of other tweaks before it's merged, but I wanted to open it up for comments and let LGTM chew on it.

[MegaMek.zip](https://github.com/MegaMek/megamek/files/3542637/MegaMek.zip)

I have attached a Jar of the latest build for this - I could really use some help testing it out.


megamek/src/megamek/common/actions/WeaponAttackAction.java
```diff
@@ -662,3829 +629,4254 @@ private static ToHitData toHit(IGame game, int attackerId, Targetable target, in
...
+        // hull down vees can't fire front weapons
+        if ((ae instanceof Tank) && ae.isHullDown() && weapon != null && (weapon.getLocation() == Tank.LOC_FRONT)) {
+            return Messages.getString("WeaponAttackAction.FrontBlockedByTerrain");
         }
 
-        // Is the pilot a weapon specialist?
-        if (ae.hasAbility(OptionsConstants.GUNNERY_WEAPON_SPECIALIST, wtype.getName())) {
-            toHit.addModifier(-2, Messages.getString("WeaponAttackAction.WeaponSpec"));
-        } else if (ae.hasAbility(OptionsConstants.GUNNERY_SPECIALIST)) {
-			// aToW sty
```
NickAragua: Is it even possible for a ground unit to get under a grounded dropship (without being squashed)?

NickAragua: Samuel L. Jackson style plasma?

NickAragua: This might be a little out of scope, but I'm seeing a lot of these things where we loop through the game actions looking for a specific one. What if we were to introduce some kind of mechanism for setting a flag for these specific actions (perhaps a key-value pair with the key being an entity ID and the value being a set of flags indicating one of these special conditions such as "bomb action scheduled")? The only wrinkle would be that removing a particular action would have to unset that action's flag as well, but that should be doable as well.

NickAragua: Not sure how expensive "isAirToGround" is, but I've seen it before, so maybe its result should be stored.

NickAragua: Would the attacker having sensor ghosts really cause the attack to be harder to hit?

megamek/src/megamek/common/actions/WeaponAttackAction.java
```diff
@@ -662,3829 +629,4254 @@ private static ToHitData toHit(IGame game, int attackerId, Targetable target, in
...
-            if (toHit.getValue() == TargetRoll.IMPOSSIBLE) {
-                return toHit;
-            }
-
-            if (te instanceof Tank) {
-                toHit.addModifier(-2, Messages.getString("WeaponAttackAction.TeVehicle"));
+        
+        // determine some more variables
+        int aElev = ae.getElevation();
+        int tElev = target.getElevation();
```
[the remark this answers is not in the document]
mkerensky: I calculate the target's relative height, which can be different from elevation.

megamek/src/megamek/common/actions/WeaponAttackAction.java
```diff
@@ -662,3829 +629,4254 @@ private static ToHitData toHit(IGame game, int attackerId, Targetable target, in
...
-                toHit.addModifier(-1, Messages.getString("WeaponAttackAction.Bracket80"));
-            }
-            if (wtype.hasModes() && weapon.curMode().equals("Bracket 60%")) {
-                toHit.addModifier(-2, Messages.getString("WeaponAttackAction.Bracket60"));
-            }
-            if (wtype.hasModes() && weapon.curMode().equals("Bracket 40%")) {
-                toHit.addModifier(-3, Messages.getString("WeaponAttackAction.Bracket40"));
-            }
+        //if LOS is blocked, block the shot except in the case of indirect artillery fire
+        if ((losMods.getValue() == TargetRoll.IMPOSSIBLE) && !isArtilleryIndirect) {
```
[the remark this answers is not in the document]
mkerensky: still works.  losMods gets set from the spotter in the case of indirect fire, so this only returns true if the spotter can't see the target either. Start at line 567.

megamek/src/megamek/common/actions/WeaponAttackAction.java
mkerensky: Grounded, I don't think so, but this is also looking at airborne dropships. Spheroids aren't restricted to shooting at targets along their flight path like aerodyne craft

mkerensky: lol yeah I see that now.  I didn't write that one in, but I think it got stashed next to the snakes

mkerensky: Yeah that's out of scope for this, but now things are organized so such improvements are more easily found and made.

mkerensky: That's a thought.  Maybe @sixlettervariables can run a cost analysis on it?

mkerensky: Yeah. This represents faulty radar on the attacker giving ghost returns.  Not the same thing as generating ghost targets defensively with ECM

NickAragua: I was always under the impression spheroids followed the same rules in that regard, don't suppose you can point me to the relevant rulebook?

megamek/src/megamek/common/actions/WeaponAttackAction.java
```diff
@@ -168,6 +168,20 @@
      * Can be checked to allow casting of attack handlers to the proper homing handler.
      */
     protected boolean isHomingShot = false;
+    
+    /**
+     * Boolean flag that determines if this shot was fired using a weapon with special to-hit handling.
+     * Allows this waa to bypass all the standard to-hit modifier checks
+     */
+    protected static boolean specialResolution = false;
```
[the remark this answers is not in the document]
mkerensky: I was unable to call it from within another static method if it wasn't...

megamek/docs/history.txt
```diff
@@ -1,61 +1,45 @@
 VERSION HISTORY:
```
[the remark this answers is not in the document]
mkerensky: Done. I was trying to avoid introducing errors to it with an out of date copy

megamek/src/megamek/common/actions/WeaponAttackAction.java
NickAragua: Let's leave it for the time being.

megamek/src/megamek/common/actions/WeaponAttackAction.java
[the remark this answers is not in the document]
NickAragua: Looks better now.

PR #1460, opened 2019-08-29
Save engine tech base when it differs from unit tech base.

### LINKED ISSUE #1459: Using XXL(Clan) fusion engine on combat vehicle by mixed IS tech is not loaded properly
### Environment
Ver 0.46.0-rc.2

### Description
As title says. On Mixed Inner Sphere tech, you may pick XXL(Clan) engine for your vehicle(both land and VTOL), and it passes the validation(on MegaMekLab), but if you load the file on either MegaMekLab or MegaMek, both programs are read XXL(Clan) engine as XXL Inner Sphere engine and it throws 'invalid design' if the end result exceeds the numbers of slots. 

For now, if I modify the tech to 'Mixed Clans' I can solve the problem and nothing was changed significally on MegaMek, though.

neoancient:
Mixed tech units are incorrectly having the engine tech base set to the unit tech base.
Support vehicle engines do not distinguish between IS and Clan, so there is no need to set the flag.

Fixes #1459: Using XXL(Clan) fusion engine on combat vehicle by mixed IS tech is not loaded properly

### DIFF
--- a/megamek/src/megamek/common/loaders/BLKSupportVTOLFile.java
+++ b/megamek/src/megamek/common/loaders/BLKSupportVTOLFile.java
@@ -1,5 +1,6 @@
-/**
+/*
  * MegaMek - Copyright (C) 2004 Ben Mazur (bmazur@sev.org)
+ * Copyright (C) 2019 The MegaMek Team
  *
  * This program is free software; you can redistribute it and/or modify it under
  * the terms of the GNU General Public License as published by the Free Software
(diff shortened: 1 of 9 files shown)


megamek/src/megamek/common/loaders/BLKSupportVTOLFile.java
```diff
@@ -97,9 +97,6 @@ public Entity getEntity() throws EntityLoadingException {
             t.setFuelTonnage(dataFile.getDataAsDouble("fuel")[0]);
         }
         int engineFlags = Engine.TANK_ENGINE | Engine.SUPPORT_VEE_ENGINE;
-        if (t.isClan()) {
```
NickAragua: How come this block is removed here but added to in the other files?

neoancient: Support vehicle engines don't distinguish between IS and Clan, so this block serves no purpose.

PR #1473, opened 2019-09-05
fix an NPE when deploying grounded dropship

NickAragua:
The logic for checking prohibited hexes forgot about buildings, which lead to an NPE further down the line. Also added IMPASSABLE while I was at it.

### DIFF
--- a/megamek/src/megamek/common/Dropship.java
+++ b/megamek/src/megamek/common/Dropship.java
@@ -175,11 +175,7 @@ public class Dropship extends SmallCraft {
         // Check prohibited terrain
         // treat grounded Dropships like wheeled tanks,
         // plus buildings are prohibited
-        boolean isProhibited = hex.containsTerrain(Terrains.WOODS) || hex.containsTerrain(Terrains.ROUGH)
-                || ((hex.terrainLevel(Terrains.WATER) > 0) && !hex.containsTerrain(Terrains.ICE))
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/Dropship.java
```diff
@@ -270,6 +259,20 @@ public boolean isLocationProhibited(Coords c, int currElevation) {
...
+     * Worker function that checks if a given hex contains terrain onto which a grounded dropship
+     * cannot deploy. 
+     */
+    private boolean hexContainsProhibitedTerrain(IHex hex) {
+        return hex.containsTerrain(Terrains.WOODS) || hex.containsTerrain(Terrains.ROUGH)
+                || ((hex.terrainLevel(Terrains.WATER) > 0) && !hex.containsTerrain(Terrains.ICE))
+                || hex.containsTerrain(Terrains.RUBBLE) || hex.containsTerrain(Terrains.MAGMA)
+                || hex.containsTerrain(Terrains.JUNGLE) || (hex.terrainLevel(Terrains.SNOW) > 1)
+                || (hex.terrainLevel(Terrains.GEYSER) == 2) 
+                || hex.containsTerrain(Terrains.BUILDING) || hex.containsTerrain(Terrains.IMPASSABLE);
```
neoancient: Does this also catch bridges? I can't remember.

NickAragua: Water implicitly catches bridges over water, but bridges can apparently go over land too, so I better add it in.

PR #1478, opened 2019-09-06
Grounded Dropships and Individual Weapons

mkerensky:
+ Errata: Grounded dropships always use individual weapons
+ Enhancement: Improves matching of dropship weapon and ammo slots with weapon bays


megamek/src/megamek/common/Mounted.java
```diff
@@ -145,6 +145,9 @@
     public static final int MINE_INFERNO = 3;
     public static final int MINE_EMP = 4;
     public static final int MINE_COMMAND_DETONATED = 5;
+    
+    // This mounted is not associated with a bay...
+    public static final int NO_BAY = -1;
```
neoancient: This isn't used anywhere. It looks like a leftover from something you ended up doing a different way.

mkerensky: Ah, yeah it sure is.  Thanks.

PR #1484, opened 2019-09-10
JS/SS max armor calculation

neoancient:
While working on support vehicle validation I noticed that most of the jumpships and space stations were getting flagged for exceeding the maximum allowed armor. I tracked it down to all advanced aero using the warship formula for SI weight instead of the correct one for the unit type.


neoancient:
I think it's mostly that they have a higher SI, since JS and SS have SI of 1.

Issue #1480, opened 2019-09-08
Mech Mortar Smoke round is unavailable at all.


binaryspica:
### Environment
Ver 0.47.0
Windows 7

### Description
I have tried to use them but none of them are available. On both MegaMek and MegaMekLab, if you put Smoke mortar round it is saved as Shaped Charge mortar round instead. The funny thing is it is not able to save them as having smoke round.

In MegaMekLab, I have tried but when I open it to notepad and I found that they are saved as Shaped Charge.
In MegaMek, you can change the rounds on the lobby but even if you choose smoke round and saved it is turned to shaped charge instead.

All Mech Mortar 1 to 8, and both Inner Sphere and Clans have the same bug.

### Files

[Test1 Mortar Tanks.zip](https://github.com/MegaMek/megamek/files/3587926/Test1.Mortar.Tanks.zip)

binaryspica:
I have a question: During make a pull request #1493, I found that the code seems weird.

On AmmoType, 'Mech Mortar Smoke rounds are defined as M_SMOKE. That's the same definition as the artillery rounds, not LRMs(M_SMOKE_WARHEAD). I don't think that it seems important' though.

But... on [MekMortarWeapon](https://github.com/MegaMek/megamek/blob/4de5dc6fbafda1eb8747d431acd09563324c0adc/megamek/src/megamek/common/weapons/mortars/MekMortarWeapon.java) and [MekMortarSmokeHandler](https://github.com/MegaMek/megamek/blob/4de5dc6fbafda1eb8747d431acd09563324c0adc/megamek/src/megamek/common/weapons/MekMortarSmokeHandler.java) they calls M_SMOKE_WARHEAD, not M_SMOKE.

Is it intended or is a mistake? And if it is a mistake, then is it related with the problem?

PR #1494, opened 2019-09-17
Attempt to fix 'Mech Mortar Smoke rounds

binaryspica:
For Issue #1480 

Actually, I am not sure that it will fix the issue, but at least it would make it consistent.

Although I can't fully understand the whole mechanism, but I suspect that conflict between defines Mech Mortar Smoke round as M_SMOKE but calls for M_SMOKE_WARHEAD when the other parts want to use Smoke round would be the root of the problem that can't save Smoke round. 

I suspect that, when MegaMek or MegaMekLab try to save it, it loads its Ammo Type and check if the weapon uses such type of ammo, and if it is failed to get the correct type it saves the default ammo type instead. And, even if it doesn't, it is better to make it consistent for it will be better to find out what is the problem.

So, please help me if I missed something.

Also, what is correct, M_SMOKE, or M_SMOKE_WARHEAD, for 'Mech Mortar's intended type of definition in AmmoType? Because it was defined as M_SMOKE in AmmoType, I changed all the M_SMOKE_WARHEAD that refers mech mortars to M_SMOKE. But if you intended to make it M_SMOKE_WARHEAD to be type of Mech Mortar Smoke round, I have to change it as well. M_SMOKE is the definition of artilliery smoke round, and M_SMOKE_WARHEAD is the definition of LRM round. And, I suspect that M_SMOKE_WARHEAD seems correct for Mech Mortars are similar to LRMs, but I am not sure.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/FiringDisplay.java
+++ b/megamek/src/megamek/client/ui/swing/FiringDisplay.java
@@ -2537,6 +2537,7 @@ public class FiringDisplay extends StatusBarPhaseDisplay implements
                     || (aType.getAmmoType() == AmmoType.T_LRM)
                     || (aType.getAmmoType() == AmmoType.T_LRM_IMP))
                     && ((munitionType == AmmoType.M_AIRBURST) 
+                            || (munitionType == AmmoType.M_SMOKE)
                             || (munitionType == AmmoType.M_SMOKE_WARHEAD))) {
                 return new HexTarget(pos, game.getBoard(),
                         Targetable.TYPE_HEX_CLEAR);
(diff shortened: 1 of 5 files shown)


neoancient:
What keeps the ammo from saving correctly is that the munitions are created as M_SMOKE (AmmoType, lines 1389 and 1456) but when creating the name that will be used for saving, it checks for M_SMOKE_WARHEAD (AmmoType, line 15535). I don't think it makes a lot of difference which is used (and I'm not sure why we need two different constants) but it needs to be consistent throughout the code. It looks like M_SMOKE_WARHEAD is more common, and the places that use M_SMOKE should be changed to that.

binaryspica:
> What keeps the ammo from saving correctly is that the munitions are created as M_SMOKE (AmmoType, lines 1389 and 1456) but when creating the name that will be used for saving, it checks for M_SMOKE_WARHEAD (AmmoType, line 15535). I don't think it makes a lot of difference which is used (and I'm not sure why we need two different constants) but it needs to be consistent throughout the code. It looks like M_SMOKE_WARHEAD is more common, and the places that use M_SMOKE should be changed to that.

Thanks. And I think that it will solve the issue. In AmmoType, Mech Mortar Smoke ammunition was defined as M_SMOKE_WARHEAD in createMunitionType with its abbreviation(SM), so it seems that M_SMOKE_WARHEAD was right.

Issue #1489, opened 2019-09-15
VTOL Jet Booster's tonnage is not correctly applied


binaryspica:
### Environment
Ver 0.47.0

### Description
Although the issue was found on MegaMekLab, but because it seems the actual data is on MegaMek I post it on here.

As title says, VTOL Jet Booster's weight is weird. On pg. 350. TO;

> **VTOL Jet Booster**: .. The booster weights 10 percent of the VTOL's engine weight(rounded up to the nearest half-ton), an takes up one item slot. .. 

But, in MegaMekLab its weight is exactly 10 percent of the engine weight, but is not rounded up to the nearest half-ton. For example, a VTOL with engine weight of 1.0 ton has a VTOL Jet Booster with 0.1 ton(per the rules it must be 0.5), and if the engine's weight is increased to 5.5 its weight is also increased to 0.55 ton(it must be 1.0).
![r001](https://user-images.githubusercontent.com/47907439/64924316-1c7bba80-d81e-11e9-8758-549a1790a2d2.png)
![r002](https://user-images.githubusercontent.com/47907439/64924318-1ede1480-d81e-11e9-87aa-e2f7c5ef5127.png)
![r003](https://user-images.githubusercontent.com/47907439/64924319-20a7d800-d81e-11e9-80a8-b6fd8b419fc9.png)
![r004](https://user-images.githubusercontent.com/47907439/64924321-243b5f00-d81e-11e9-9856-059ac412fefc.png)
On these screenshots, I have removed all the equipment and only put VTOL Jet Booster, so tonnage shown on the equipment section of Summary is equal to the weight of VTOL Jet Booster on the unit.

sixlettervariables:
The code appears correct:
```java
        } else if (hasFlag(F_JET_BOOSTER)) {
            Engine e = entity.getEngine();
            return Math.ceil((e.getWeightEngine(entity) *.1 ) * 2.0f) / 2.0;
```

sixlettervariables:
Figured it out. VTOL Jet Boosters count as MASC:
https://github.com/MegaMek/megamek/blob/113fb9a1d0cdfced317fcb79065e7e41d3f8f92c/megamek/src/megamek/common/MiscType.java#L2281-L2282

Which gets its tonnage calculated earlier than `F_JET_BOOSTER`:
[...]

PR #1498, opened 2019-09-21
Fix jet booster and supercharger weights #1489

sixlettervariables:
Fixes VTOL jet booster weights (#1489) and supercharger weights (found during code inspection).

### DIFF
--- a/megamek/src/megamek/common/MiscType.java
+++ b/megamek/src/megamek/common/MiscType.java
@@ -480,6 +480,13 @@ public class MiscType extends EquipmentType {
             return Math.ceil(entity.getWeight() / 10.0);
         } else if (hasFlag(F_CLUB) && hasSubType(S_RETRACTABLE_BLADE)) {
             return 0.5 + Math.ceil(entity.getWeight() / 10.0f) / 2.0;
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/MiscType.java
```diff
@@ -489,15 +496,10 @@ public double getTonnage(Entity entity, int location) {
...
-                }
                 if (hasSubType(S_SUPERCHARGER)) {
-                    Engine e = entity.getEngine();
-                    if (e == null) {
-                        return 0.0f;
-                    }
-                    return Math.ceil((e.getWeightEngine(entity) / 10.0f) * 2.0f) / 2.0;
+                    // pg 411, TO
+                    // 10% of engine weight
+                    return entity.hasEngine() ? entity.getEngine().getWeightEngine(entity) / 10.0f : 0.0f;
```
neoancient: Per TacOps, p. 344 the weight of the supercharger is rounded up to the nearest half ton for units over 5t and to the nearest kg for units under 5t. I suspect a rules inquiry would say the same is true of the VTOL jet booster but that's not what the rules currently say.

PR #1519, opened 2019-10-17
Sv validation

neoancient:
Part 2 of the support vehicle construction project. This branch adds the ability to check support vehicles against the construction rules, and includes adjustments to weight and slot calculations. I removed the unit file fixes from this branch (cf. sv_unit_fixes, which is identical to the branch but with the unit file fixes included). Git still converts the line endings on commit, but hiding whitespace changes should hide those ~60 files.


megamek/src/megamek/common/verifier/TestAero.java
```diff
@@ -1018,11 +1014,80 @@ public boolean correctEntity(StringBuffer buff, int ammoTechLvl) {
         correct &= correctControlSystems(buff);
         correct &= !hasIllegalTechLevels(buff, ammoTechLvl);
         correct &= !hasIllegalEquipmentCombinations(buff);
+        correct &= !hasMismatchedLateralWeapons(buff);
         correct &= correctHeatSinks(buff);
         
         return correct;
     }
 
+    /**
```
neoancient: This check is present in the larger aerospace vessels. I added it for use by fixed wing support vehicles, but it also fixes the lack in aerospace and conventional fighters.

megamek/src/megamek/common/MiscType.java
```diff
@@ -745,10 +745,20 @@ public double getTonnage(Entity entity, int location) {
             return TestEntity.ceil(weight, roundWeight);
 
         } else if (hasFlag(F_BASIC_FIRECONTROL)) {
+            // Omni support vees have a fixed weight for the chassis, which may be
```
NickAragua: This appears to be identical to the logic for advanced fire control, except the weight multiplier near the bottom. Let's put it into a function?

megamek/src/megamek/common/verifier/TestBattleArmor.java
```diff
@@ -1177,8 +1177,8 @@ public StringBuffer printEntity() {
         buff.append(printSource());
         buff.append(printShortMovement());
         if (correctWeight(buff, true, true)) {
-            buff.append("Weight: ").append(getWeight()).append(" (")
-                    .append(calculateWeight()).append(")\n");
+            buff.append("Weight: ").append(getWeight() * 1000).append(" kg (")
```
sixlettervariables: Does this need the `%3.1f` format like below?

megamek/src/megamek/common/verifier/TestProtomech.java
```diff
@@ -587,8 +587,8 @@ public StringBuffer printEntity() {
         buff.append(printSource());
         buff.append(printShortMovement());
         if (correctWeight(buff, true, true)) {
-            buff.append("Weight: ").append(getWeight()).append(" (")
-                    .append(calculateWeight()).append(")\n");
+            buff.append("Weight: ").append(getWeight() * 1000).append(" kg (")
```
sixlettervariables: Same as before regarding format string.

megamek/src/megamek/common/MiscType.java
neoancient: I consolidated them in a single block. I also spotted and fixed an error in the logic to exclude certain weapons used by small support vehicles. It's supposed to skip light weapons (e.g. any infantry weapons that aren't flagged as support) but it was excluding all weapons with either the infantry or infantry support flags.

megamek/src/megamek/common/verifier/TestBattleArmor.java
neoancient: I don't see the format string you're referring to.

PR #1521, opened 2019-10-20
Pd bay errata

mkerensky:
+ Errata Treat Arrow IV and similar 'large missiles' as capital missiles for AMS/PD bay resolution
+ Rules Correction  Grounded units cannot fire capital weapons directly at ground targets
+ Feature Ground units can fire capital missiles surface to surface as artillery
+ Bugfix Corrects some handler extension issues
+ Bugfix Corrects handling of capital fighter squadron Alamo nuclear attacks
+ Bugfix Corrects max offboard deployment distance for artillery armed units
+ Bugfix Corrects some attack value calculations for AR10 weapons


megamek/src/megamek/common/weapons/CapitalMissileHandler.java
```diff
@@ -305,39 +338,54 @@ public boolean handle(IGame.Phase phase, Vector<Report> vPhaseReport) {
...
-        // aero range brackets
-        if (!ae.isAirborne() || game.getOptions().booleanOption(OptionsConstants.ADVCOMBAT_UAC_TWOROLLS)) {
-            if (usesClusterTable()) {
-                // for cluster weapons just use the short range AV
-                av = wtype.getRoundShortAV();
+        //AR10 munitions
+        if (atype != null) {
+            if (atype.getAmmoType() == AmmoType.T_AR10) {
+                if (atype.hasFlag(AmmoType.F_AR10_KILLER_WHALE)) {
+                    av = 4;
```
NickAragua: I'm curious, is this something that can be put into the ammo type definition instead?

mkerensky: I think this is as it is due to handling nuclear warheads.  I'm inclined to leave it alone at the moment, but since there's more to do with different missile warheads and WMDs I can revisit the topic when I get to those.

megamek/src/megamek/common/weapons/CapitalMissileHandler.java
```diff
@@ -137,8 +137,10 @@ public boolean handle(IGame.Phase phase, Vector<Report> vPhaseReport) {
             CounterAV = bayHandler.getCounterAV();
             nDamPerHit = calcDamagePerHit();
         } else {
-            // Should only be used when using a grounded dropship with individual weapons
-            // Otherwise we're using CapitalMissileBayHandler
+            // Used when using a grounded dropship with individual weapons
+            // or a fighter squadron loaded with ASM or Alamo bombs.
+            nDamPerHit = calcDamagePerHit();
```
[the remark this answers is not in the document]
mkerensky: Done. Good catch.

megamek/src/megamek/common/weapons/MissileWeaponHandler.java
```diff
@@ -396,9 +395,14 @@ protected int calcAttackValue() {
...
             av = av * 2;
         }
+        //Set the Capital Fighter AV here. We'll apply counterAV to this later
+        originalAV = av;
                 
         //Point Defenses engage the missiles still aimed at us
-        counterAV = calcCounterAV();
+        if (ae.usesWeaponBays() || ae.isCapitalFighter()) {
+            // Grounded dropships should fire with individual AMS, not bays
+            counterAV = calcCounterAV();
```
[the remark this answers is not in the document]
mkerensky: I see how the comment would be unclear.  usesWeaponBays  accounts for the dropship being grounded or not. 
This does work as intended.  The interaction of grounded dropships, airborne dropships, aero sanity and fighter squadrons with this handler and the MissileBayWeaponHandler are...complicated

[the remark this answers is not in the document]
mkerensky: Either way, fixed the extra internal variable and cleaned up the comment a bit.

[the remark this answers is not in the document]
neoancient: I didn't even think about usesWeaponBays being dependent on being airborne, but now that you say it I remember you changing that recently.

megamek/src/megamek/common/weapons/MissileWeaponHandler.java
```diff
@@ -400,10 +399,9 @@ protected int calcAttackValue() {
                 
         //Point Defenses engage the missiles still aimed at us
         if (ae.usesWeaponBays() || ae.isCapitalFighter()) {
-            // Grounded dropships should fire with individual AMS, not bays
-            counterAV = calcCounterAV();
+            CounterAV = calcCounterAV();
         }
-        av = av - counterAV;
+        av = av - CounterAV;
```
neoancient: It wasn't just about adding a local variable. It was that you had three lines of code where only one is needed. Now it's two lines. If you're only changing the value of av if it meets the previous condition, then put the arithmetic operation inside the if block. As written this is effectively identical to
```
if (ae.usesWeaponBays() || ae.isCapitalFighter()) {
    av = av - calcCounterAV();
} else {
    av = av - 0;
}
```
Instead just use `av = av - calcCounterAV();` as the body of the if block.

mkerensky: Let me look at it some more.  Part of the reason this is written as it is is that I don't want calcCounterAV() being called unless we meet the usesWeaponBays or isCapitalFighter conditions.  Though a grounded dropship does not *use* weapon bays it has them (they have to exist so that they can be enabled at liftoff) , and that will make calcCounterAV run successfully when it shouldn't.     Part of my goal is to allow this code to work for mobile structures and large naval support vees later.

mkerensky: Ok, I think this is fixed now, with a couple of other errors affecting fighter squadrons with aero sanity that I found in testing.

PR #1528, opened 2019-10-25
Weight rounding

neoancient:
The purpose of this change is to provide a straightforward set of tools to round weights used in unit construction calculations with concise statements. In adds some overhead by always rounding to the nearest gram first in order to avoid inadvertently rounding up too much because of imprecision in floating point arithmetic.

Each way of rounding (e.g. next ton, closest kg, etc.) has a matching enum value in RoundWeight. This allows the method itself to be passed as a parameter, which prepares the way for eventual support of the optional fractional accounting rule. There are also static convenience methods for more concise code.

The changes here do not yet touch TestEntity and its subclasses. Calculations used in validation should be allowing the rounding default rounding method to be overridden by settings in data/mechfiles/UnitVerifierOptions.xml, but do not do so consistently, and I thought it better to undertake that as a separate task.

Throughout this process I have been generating lists of equipment weights for the section I was working on and comparing before and after as a check against typos. I also ran a final check of all unit weights.  Total of five units have changed calculated weights, all of which were verified as incorrect before the change:
Agrotera AGT-1A: extra half ton on partial wing due to floating point imprecision
Comititatus Jumpship (2995) and Olympus Recharge Station (2942): 120 tons armor calculated as 120.5 due to floating point imprecision
Death Trike: supercharger weight not correctly calculated for small support vehicles
M3 Drone CASPAR: incorrect formula for self-destruct system

Fixes Megamek/megameklab#280 (for real this time)


megamek/src/megamek/common/MiscType.java
```diff
@@ -864,30 +809,19 @@ public double getTonnage(Entity entity, int location) {
...
+            return defaultRounding.round(entity.getWeight() * 0.03, entity);
         } else if (hasFlag(MiscType.F_SDS_DESTRUCT)) {
-            double tWeight = 0;
-            tWeight = Math.ceil((entity.getWeight() * 0.1) * 2) / 2.0;
-            if (tWeight >10000) {
-                return 10000;
-            } else {
-                return Math.ceil((entity.getWeight() * 0.02) * 2) / 2.0;
-            }
+            return Math.min(RoundWeight.nextTon(entity.getWeight() * 0.1), 10000);
```
sixlettervariables: The old code looks like to the next half ton (now `defaultRounding.round(...)` I believe)

sixlettervariables: This is Okay, SDS Self Destruct had the wrong formula and was corrected.

PR #1540, opened 2019-11-11
EntityWeightClass cleanup

neoancient:
Include all SV types in getClassName(int, String, boolean). It currently limits the SV weight class values to TANK and VTOL, which leaves out naval and fixed wing types. No unit type checks should be needed anyway, since the isSupport parameter should be sufficient.

Also replaced string literals with constants and consolidated two case blocks with the same body.

Part of a fix for MegaMek/mekhq#1333


neoancient:
This won't affect the Simca. This is part of a general fix for all support vehicles, which needs to work for naval and fixed wing as well.

PR #1548, opened 2019-11-19
Preserve capital bay lookup names

neoancient:
When a weapon class extends another, the child class must be initialized first. When its constructor calls super() it will add the parent's internal name and additional lookup names to the EquipmentType lookup hash pointing to the child class. If the child initialized second, all the names for both classes point to the child. If the child is initialized first, the parent class with come behind and fix the values that should be pointing to the parent.

This fixes the relationship between CapitalMissileBayWeapon and TeleOperatedMissileBayWeapon, but removing the gotcha would require an extensive rework and I haven't figured out the best way to go about it yet. I did add a warning to the javadoc comment for the method, but as the method is 1000+ lines it's not likely to be seen.

Fixes MegaMek/megameklab#371

### DIFF
--- a/megamek/src/megamek/common/WeaponType.java
+++ b/megamek/src/megamek/common/WeaponType.java
@@ -1154,6 +1154,9 @@ public class WeaponType extends EquipmentType {
 
     /**
      * Add all the types of weapons we can create to the list
+     *
+     * When a weapon class extends another, the subclass must be listed first to avoid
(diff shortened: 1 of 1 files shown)


mkerensky:
Looks good to me.  I suspect I introduced this one with my previous PR on AMS Errata.  I did change the inheritance path of several capital/subcapital missile weapons in that one.

Issue #911, opened 2018-02-25
Feature Request: Flamer Heat&Damage Unofficial Option


Retry2000:
Currently, an option exists that is checked by default under basic options called "Flamers can deal heat instead of damage".  This gives Flamers some additional utility, however, players still may find the flamer and its variants lacking.

A second option under this option could be introduced which allows Flamer variants (Flamer, ER Flamer, Vehicle Flamer & Heavy Flamer) to inflict both heat _and_ damage instead of either heat _or_ damage, similar to how the Plasma Rifle works.  For instance, a flamer that targets (and hits) a heat-tracking unit like a Battlemech would deal 2 damage to an armor location & raise the battlemech's heat production by 2 points as well, instead of simply dealing 2 damage OR 2 heat.

It'd be a fairly simple, nice option to boost the utility of flamers and flamer-carrying units like the Firestarter.

TigerShark:
This is in the BattleMech Manual as an official, optional rule. Heat and damage are produced simultaneously in the text.

TigerShark:
Enhanced Flamers (Optional): If using this optional rule,
whenever a flamer of any kind strikes a target that tracks heat (such
as a ’Mech), apply both the flamer’s heat and damage to the target.

Page 99

binaryspica:
It would be a good addition I think. Normally flamers are deals lesser damage than its heat despite it is a point-blank range weapon.

TigerShark:
Any word on this one?

sixlettervariables:
None of the core developers have picked this up yet. I'll mark it as help wanted.

TigerShark:
This should probably be changed to an "official option," as this is in no way a house rule.

NickAragua:
The rule in BMM p. 99 has been implemented and will be out in the next release, although it doesn't really help flamers that much. ;)

PR #1550, opened 2019-11-21
Replace unofficial flamer heat option with official BMM rules

NickAragua:
Pretty straightforward. Replaces the unofficial "flamers do heat damage instead of damage" rule with the BMM page 99 rule allowing flamers to do both heat and damage. 

Implementation-wise, this means removing the ability to set flamer modes, consolidating flamer handling into one class, and fixing a minor error leading to reflective/heat dissipating armor not actual working against flamer heat damage.

Implements #911


neoancient:
The "flamers do heat instead of damage" isn't unofficial; it's the base rule. That's why it's checked by default.
TW, p. 304:
Screenshot from 2019-11-20 21-43-04

NickAragua:
Ok. In that case, I'm thinking that the checkbox state will be unchecked by default and specify the ability to declare heat/damage as previously. If checked, then it goes to BMM rules and does both. I don't see a need to have the "flamers do damage only" option.

NickAragua:
Going to re-do this one, went with a different and more comprehensive approach.

PR #1553, opened 2019-11-23
Enable easier editing of terrain features

NickAragua:
This fixes issue #920. For some reason, the TerrainHelper didn't implement the equals method, which made calling "setSelectedItem" on a dropdown full of those things somewhat problematic.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/BoardEditor.java
+++ b/megamek/src/megamek/client/ui/swing/BoardEditor.java
@@ -113,6 +113,11 @@ public class BoardEditor extends JComponent implements ItemListener,
         public int compareTo(TerrainHelper o) {
             return toString().compareTo(o.toString());
         }
+        
+        @Override
+        public boolean equals(Object o) {
+            return o instanceof TerrainHelper && this.getTerrainType() == ((TerrainHelper) o).getTerrainType();
+        }
 
     }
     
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/client/ui/swing/BoardEditor.java
```diff
@@ -113,6 +113,11 @@ public String getTerrainTooltip() {
         public int compareTo(TerrainHelper o) {
             return toString().compareTo(o.toString());
         }
+        
+        @Override
+        public boolean equals(Object o) {
+            return o instanceof TerrainHelper && this.getTerrainType() == ((TerrainHelper) o).getTerrainType();
```
sixlettervariables: Should probably add a hashCode implementation that returns the hash of terrainType.

NickAragua: What would be the point of adding a hashCode? It doesn't appear to be used at all in the JList implementation.

sixlettervariables: Consistency/OCD? [Rummaging through OpenJDK it doesn't appear they do to your point](https://github.com/AdoptOpenJDK/openjdk-jdk8u/blob/master/jdk/src/share/classes/javax/swing/JList.java).

NickAragua: Generally not a big fan of adding dead code, especially for a one-off internal class.

PR #1566, opened 2019-12-13
Rat generator editor updates

NickAragua:
Code changes necessary to get the rat generator editor up and running.

Example update (Bluehawk conventional fighter added)

The data files got changed a little bit, but appear to load normally after the change.


megamek/data/forcegenerator/factions.xml
```diff
@@ -454,12 +445,12 @@
...
 	<faction key='RR' name='Republic Remnant' minor='false' clan='false' periphery='false'>
 		<years>3135-</years>
-		<ratingLevels>Triarii Protectors,Principes Guards,Hastati Sentinels,Stone&apos;s Brigade</ratingLevels>
+		<ratingLevels>Triarii Protectors,Principes Guards,Hastati Sentinels,Stone's Brigade</ratingLevels>
 		<parentFaction>ROS</parentFaction>
 	</faction>
 	<faction key='ROS' name='Republic of the Sphere' minor='false' clan='false' periphery='false'>
 		<years>3081-</years>
-		<ratingLevels>Triarii Protectors,Principes Guards,Hastati Sentinels,Stone&apos;s Brigade</ratingLevels>
+		<ratingLevels>Triarii Protectors,Principes Guards,Hastati Sentinels,Stone's Brigade</ratingLevels>
```
sixlettervariables: Why the change? This really should be `&apos;` or wrapped in a CDATA.

megamek/src/megamek/utils/RATGeneratorEditor.java
```diff
@@ -849,41 +853,54 @@ public void setValueAt(Object value, int row, int col) {
...
             data.put(faction, list);
+            for (int i = 0; i < ERAS.length; i++) {
+                AvailabilityRating ar = new AvailabilityRating(getUnitKey(), ERAS[i],
+                            faction + ":" + 0);
+                if (mode == MODE_MODEL) {
+                    rg.setModelFactionRating(ERAS[i], unitRecord.getChassisKey(), unitRecord.getKey(), ar);
+                } else {
+                    rg.setChassisFactionRating(ERAS[i], getUnitKey(), ar);
+                }
+            }
```
neoancient: Zero and blank have different meanings. Assigning an explicit zero blocks inheritance. It's used for situations where a unit is used by, for example, all IS factions except DC. Or all DC commands except the Sword of Light.

megamek/src/megamek/client/ratgenerator/FactionRecord.java
```diff
@@ -602,72 +603,74 @@ public void writeToXml(PrintWriter pw) {
...
+                factionRecordBuilder.append("\t\t<techMargin>" + ((era - 3067) / 5) + "</techMargin>\n");
                 if (era > 3085) {
-                    pw.println("\t\t<upgradeMargin>" + ((era - 3085) / 5) + "</upgradeMargin>");
+                    factionRecordBuilder.append("\t\t<upgradeMargin>" + ((era - 3085) / 5) + "</upgradeMargin>\n");
                 }
             } else {
-                pw.println("\t\t<techMargin>" + ((era - 3067) / 5) + "</techMargin>");
+                factionRecordBuilder.append("\t\t<techMargin>" + ((era - 3067) / 5) + "</techMargin>\n");
             }
         }
```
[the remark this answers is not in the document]
NickAragua: The main reason is on line 702, because sometimes the faction has nothing in its interior and this was producing invalid XML.

megamek/src/megamek/client/ratgenerator/RATGenerator.java
```diff
@@ -215,16 +215,22 @@ public AvailabilityRating findModelAvailabilityRecord(int era, String unit, Fact
...
+            setChassisFactionRating(era, chassisKey, new AvailabilityRating(chassisKey, era, ar.getFaction() + ":1"));
         }
-        modelIndex.get(era).get(unit).put(ar.getFactionCode(), ar);
-        models.get(unit).getIncludedFactions().add(ar.getFactionCode());
+        
+        if (modelIndex.get(era).get(unitKey) == null) {
+            modelIndex.get(era).put(unitKey, new HashMap<String, AvailabilityRating>());
+        }
+        modelIndex.get(era).get(unitKey).put(ar.getFactionCode(), ar);
+        models.get(unitKey).getIncludedFactions().add(ar.getFactionCode());
```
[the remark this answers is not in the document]
NickAragua: The way it's currently implemented, every model record does have to be accompanied by a chassis record. The chassis saving code simply skips over the individual model records if there is no chassis record. A better approach may be to prevent adding a model record if there is no chassis record.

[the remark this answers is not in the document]
NickAragua: I'll take another look at it when I get the chance and see if there's a better approach.

megamek/src/megamek/client/ratgenerator/FactionRecord.java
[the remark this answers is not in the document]
neoancient: It was invalid because the attributes were not escaped and several of the names have apostrophes. But self-closing tags are not required for elements without children.`<faction></faction>` and `<faction/>` are equivalent and both perfectly legal. It doesn't really matter one way or te other, but if you're going to use a StringBuilder you should append each string individually rather than using string concatenation in the argument to append. That has the JVM instatiating an unnecessary additional StringBuilder. This also adds an additional newline between the last child and the end tag. You might consider StringJoiner instead.

megamek/src/megamek/client/ratgenerator/RATGenerator.java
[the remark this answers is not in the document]
neoancient: That would be a case of user error, but requiring that there be a chassis record first would guard against it. It might be simpler to validate the data and warn the user in a popup when saving. This was never intended to be a sophisticated tool.

NickAragua:
> Unless the changes in the data files are necessary to make it work, they should not be committed to this branch.

Any particular reason?

neoancient:
It bloats the diff with thousands of lines of irrelevant changes.

megamek/src/megamek/utils/RATGeneratorEditor.java
NickAragua: What should the default availability rating for a new row in that table be then?

neoancient: Nothing

megamek/src/megamek/client/ratgenerator/FactionRecord.java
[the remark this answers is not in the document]
NickAragua: For whatever reason, which I don't feel like investigating at the moment, <faction></faction> throws an "expected element" exception when loading the data. I got rid of the string concatenation where it was relevant.

megamek/src/megamek/utils/RATGeneratorEditor.java
```diff
@@ -745,13 +745,17 @@ public ModelRecord getUnitRecord(int row) {
...
+        
+        private String getUnitKey() {
+            return (mode == MODE_CHASSIS) ? unitRecord.getChassisKey() : unitRecord.getKey();
+        }
 
         public UnitEditorTableModel() {
             factions = new ArrayList<String>();
             data = new HashMap<String, ArrayList<String>>();
             weights = new HashMap<String, ArrayList<Integer>>();
-            unit = "";
```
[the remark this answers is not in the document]
NickAragua: In order to check for non-present chassis, we still need the chassis key, so I left this set of changes in place.

megamek/src/megamek/utils/RATGeneratorEditor.java
NickAragua: Pretty sure that winds up reading as 0 anyway in practice, since it's stored as an int in the AvailabilityRating data structure.

neoancient: You shouldn't be adding AvailablityRatings at all. This block shouldn't be here

megamek/data/forcegenerator/factions.xml
NickAragua: Added a function call to escape these.

megamek/src/megamek/utils/RATGeneratorEditor.java
NickAragua: Ok. I checked it without adding the availability ratings and it still works fine. I re-worked the validation since we're not actually guaranteed to have a chassis record for any particular era, so it just looks to see if there's any chassis record defined, period.

NickAragua:
All right, pretty sure I addressed most of the issues raised in the code review. Any thing else jump out?

PR #1574, opened 2019-12-20
Grounded dropship arc fix

NickAragua:
Fixes two issues:

1. Grounded spheroid dropships could fire nose weapons at ground targets.
2. Grounded spheroid dropships could fire aft weapons at targets.

Technically the rules say that aft weapons can be used at targets in the same hexes, but I don't see how that's possible as grounded dropships block movement into their hexes.


neoancient:
Since grounded dropships occupy multiple hexes, is it legal for them to straddle a canyon? That would allow another unit to be in the same hex, but I haven't checked whether the rules allow it.

NickAragua:
From what I've seen, the dropship "hugs" the ground, but I'll double check.

NickAragua:
After a few spot checks, the following findings:
Landing a spheroid dropship flattens terrain below it
Spheroid dropships cannot be deployed on the ground straddling a canyon
Spheroid dropships deployed on hexes with different levels "hug" the ground

So, there's no way during normal gameplay that a unit can get below a grounded spheroid dropship.

megamek/src/megamek/common/actions/WeaponAttackAction.java
```diff
@@ -1481,7 +1481,9 @@ private static String toHitIsImpossible(IGame game, Entity ae, int attackerId, T
         // for spheroid dropships in atmosphere (and on ground), the rules about
         // firing arcs are more complicated
         // TW errata 2.1
-        if (Compute.useSpheroidAtmosphere(game, ae) && weapon != null) {
+        if ((Compute.useSpheroidAtmosphere(game, ae) ||
+                (((IAero) ae).isSpheroid() && ae.getAltitude() == 0))
```
neoancient: I don't see anything that guarantees the Entity is an aerospace unit at this point, which means that this can throw a ClassCastException.

megamek/src/megamek/common/actions/WeaponAttackAction.java
```diff
@@ -1503,9 +1505,14 @@ private static String toHitIsImpossible(IGame game, Entity ae, int attackerId, T
...
-            if ((weapon.getLocation() == Aero.LOC_AFT)) {
-                if (ae.isAirborne() && (altDif > -1)) {
+            // For grounded spheroids, weapons can only be fired at targets in occupied hexes, 
+            // but it's not actually possible for a unit to occupy the same hex as a grounded spheroid so
+            // we simplify the calculation a bit
+            if ((weapon.getLocation() == Aero.LOC_AFT) && (altDif > -1)) {
+                if (ae.isAirborne()) {
                     return Messages.getString("WeaponAttackAction.TooHighForAft");
+                } else {
+                    return Messages.getString("WeaponAttackAction.GroundedSpheroidDropshipAftWeaponRestriction");
```
neoancient: The rules for landing dropships (TW, p. 87) says that the central hex is reduced by one level if not paved/road/water and the surrounding hexes are reduced to the same level if they started higher. This means that a dropship could land overhanging a cliff and a unit could get into the same hex underneath. This should only prevent the attack for range > 0.

NickAragua: That may be true, but currently, MegaMek does not implement overhanging dropship behavior - instead, the grounded spheroid "hugs" the ground if deployed over a cliff, and flattens the terrain out if it lands on uneven terrain regardless of terrain elevation differences.

That being said, I've introduced logic to check for the theoretical presence of another unit underneath one of the dropship's occupied hexes in case we do implement overhang behavior somehow.

megamek/src/megamek/common/actions/WeaponAttackAction.java
```diff
@@ -1503,9 +1505,28 @@ private static String toHitIsImpossible(IGame game, Entity ae, int attackerId, T
...
+                // if both targets are on the ground
+                // and the target is below the attacker 
+                // and the attacker is in one of the target's occupied hexes
+                // then we can shoot aft weapons at it
+                // note that this cannot actually happen in MegaMek currently but is left here for the possible eventuality
+                // that overhanging dropships are implemented
+                if(!ae.isAirborne() && !target.isAirborne()) {
+                    boolean targetInAttackerHex = ae.getOccupiedCoords().contains(target.getPosition()) ||
+                            ae.getPosition().equals(target.getPosition());
+                    boolean targetBelowAttacker = ae.getElevation() > target.getElevation();
```
neoancient: getElevation() returns the elevation relative to the surface of the hex rather than the absolute level.
```suggestion
                    boolean targetBelowAttacker = game.getBoard().getHex(ae.getPosition()).surface() >
                        game.getBoard().getHex(ae.getPosition()).surface() + target.getElevation();

```

NickAragua: Seems like, mathematically, this could be reduced to 0 > target.getElevation() (i.e. the target is below the surface of the hex), which I don't think is what we're looking for. Unless I'm missing something?

neoancient: That was my first thought, but then I realized that the dropship elevation is always zero over the center hex, but the situation implies that the other unit is in one of the surrounding hexes which has a lower surface elevation. This is why you need to compare `surface()` but also include the target rlevation because it could be a vtol flying above the dropship.

NickAragua: Ah, I think I get it. Probably the second surface call would be for the target's position then.

neoancient: Oh, I typed that wrong. Sorry for the confusion.

NickAragua: Done

PR #1585, opened 2020-01-04
Minor bot improvements

NickAragua:
Two bot behavior updates:

1) The bot will no longer obsessively target crippled units belonging to other bots, preferring to move on to more dangerous targets. This was due to a stale cache that wasn't being updated during phases that didn't have a "calculateMyTurn" method call.

2) First crack at getting the bot to try to avoid minefields, classifying them as hazards. It's not enough, as, occasionally, the paths that don't move over minefields/hazards are pruned prior to being evaluated for hazards. The damage estimate of walking over the minefield is the damage it will inflict * odds of it going off.

### DIFF
--- a/megamek/src/megamek/client/bot/BotClient.java
+++ b/megamek/src/megamek/client/bot/BotClient.java
@@ -380,6 +380,10 @@ public abstract class BotClient extends Client {
     @Override
     public void changePhase(IGame.Phase phase) {
         super.changePhase(phase);
+        
+        // clear out transient data
+        currentTurnEnemyEntities = null;
+        currentTurnFriendlyEntities = null;
 
         try {
             switch (phase) {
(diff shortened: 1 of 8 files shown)


NickAragua:
Hold off on this, still experiencing an obnoxious behavior where bot units will kick crippled units or fire at units that have become crippled during the movement phase.

NickAragua:
Introduced additional logic to refresh crippled unit lists and such. Basically, if a unit became crippled during an earlier phase, the bot would continue firing, stomping or shooting artillery at it during subsequent phases until the end of the turn. Now, it'll look for other targets immediately.

NickAragua:
Seems that I found another issue, which is that, when an aerospace fighter dropped bombs, it would subsequently refuse to carry out attacks if its alpha strike would cause it to overheat. Now we no longer consider a dive bomb plan if there are no bombs to drop. Because previous safeguards proved insufficient and there's more than one way to not have any bombs, apparently.

apokrovski:
I got the impression that the client's game object doesn't contain un-observed minefields, but I could be wrong. I'll check again when I get the chance.

Anyway, this one needs more time in the oven, because it looks like it introduced a bunch of transient NPEs (that go away if you kick and replace the bot) and doesn't fully address the main issue it was trying to address (the bot still targets crippled units from other bots).

NickAragua:
Will open at a later point, with minefield changes separate from "honor" changes.

PR #1591, opened 2020-01-10
Ba mines

### LINKED ISSUE #1590: Fa Shih unable to lay mines
Noticed while playing that I can't get the Fa Shih to deploy mines at all, tried both the Mac and Windows versions of 47.2

### Environment
This is with the bundled MekHQ version of the app

### Description
Step 1. Spawn a Fa Shih
Step 2. Attempt to lay mine during movement phase, click hex, click done, nada.

### Files
[FaShihMines.sav.gz](https://github.com/MegaMek/megamek/files/4031141/FaShihMines.sav.gz)
[megameklog.txt](https://github.com/MegaMek/megamek/files/4031142/megameklog.txt)

neoancient:
Usability fixes for using BA with mine dispensers. See rules on TO, p. 325. BA can lay a mine at the beginning or end of movement, but only when using jumping or VTOL movement. Currently you can plot a movement turn that combines walking and laying a mine with no indication that anything is wrong and the mine-laying command just does nothing. This PR changes the behavior to switch to jumping mode automatically when jump-capable BA lay a mine at the beginning of a turn, and disable the mine-laying button for BA that have moved using any mode other than jump or VTOL.

Fixes #1590

### DIFF
--- a/megamek/src/megamek/client/ui/swing/MovementDisplay.java
+++ b/megamek/src/megamek/client/ui/swing/MovementDisplay.java
@@ -911,6 +911,7 @@ public class MovementDisplay extends StatusBarPhaseDisplay {
         updateDumpButton();
         updateEvadeButton();
         updateBootleggerButton();
+        updateLayMineButton();
 
         updateStartupButton();
         updateShutdownButton();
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/client/ui/swing/MovementDisplay.java
```diff
@@ -2934,6 +2936,24 @@ private synchronized void updateLoadButtons() {
...
+        }
+
+        if (!ce.canLayMine()) {
+            setLayMineEnabled(false);
+        }
+        if (ce instanceof BattleArmor) {
+            setLayMineEnabled(cmd.getLastStep() == null
+                || cmd.isJumping()
+                || cmd.getLastStepMovementType().equals(EntityMovementType.MOVE_VTOL_WALK));
+        } else {
```
apokrovski: This looks like, if the current entity can't lay a mine and isn't battle armor, the lay mine button will become enabled. Maybe you want a return after line 2946?

neoancient: Yup

PR #1593, opened 2020-01-11
Bypass artillery hit calculation for non-artillery weapons.

### LINKED ISSUE #1576: Target on Predesignated Artillery Hex is automatically hit
### Environment
MekHQ 0.47.2 Windows
(

### Description
When firing at a target that is one a designated Artillery hex, the target is automatically hit.  I have pasted the round report below as an example.  

Weapons fire for Cestus CTS-6Y (Conflict Interstellar)
    Gauss Rifle at Galahad (Glass Spider) (Standard) (Clan Diamond Shark); , the shot is an automatic hit (Artillery firing at designated artillery target.), rolls 11 :  - Direct Blow - hits  LT
        Galahad (Glass Spider) (Standard) (Clan Diamond Shark) takes 17 damage to LT.
            3 Armor remaining.

### Files
[megameklog.txt](https://github.com/MegaMek/megamek/files/3993333/megameklog.txt)
[mekhqlog.txt](https://github.com/MegaMek/megamek/files/3993334/mekhqlog.txt)
[Artillery Bug.gz](https://github.com/MegaMek/megamek/files/3993331/Artillery.Bug.gz)

neoancient:
When calculating weapon to-hit numbers, all weapons are sent through the artillery attack calculation, which is supposed to return the starting value if none of the artillery attack categories apply. But the check for auto-hit of designated targets does not filter out non-artillery weapons. While I could have added a check there, I see no reason to call the method for non-artillery weapons in the first place.

Fixes #1576

### DIFF
--- a/megamek/src/megamek/common/actions/WeaponAttackAction.java
+++ b/megamek/src/megamek/common/actions/WeaponAttackAction.java
@@ -684,8 +684,10 @@ public class WeaponAttackAction extends AbstractAttackAction implements Serializ
         }
         
         // Is this an Artillery attack?
-        toHit = handleArtilleryAttacks(game, ae, target, ttype, losMods, toHit, wtype, weapon, atype, isArtilleryDirect,
-                isArtilleryFLAK, isArtilleryIndirect, isHoming, usesAmmo, srt);
+        if (isArtilleryDirect || isArtilleryIndirect) {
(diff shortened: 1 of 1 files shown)


mkerensky:
Looks good to me.  I tested it out with the two special cases I was concerned about - bearings only missiles and surface-to-surface capital (missile) weapons.

PR #1610, opened 2020-01-21
VSTOL Equipment Now Saves Properly and is shown on Unit Summary and TRO

Windchild292:
I've also done some refactoring as I've gone through the files, and added .idea to the .gitignore for those of us who use the Idea IDE

### DIFF
--- a/megamek/resources/megamek/common/templates/tro/aero.ftl
+++ b/megamek/resources/megamek/common/templates/tro/aero.ftl
@@ -4,10 +4,10 @@ ${fullName}
 <#include "fluff.ftl">
 </#if>
 
-<#if !isVSTOL || >
-Type: ${chassis}
-<#else>
+<#if isVSTOL && isConventional>
 Type: ${chassis} with VSTOL
+<#else>
+Type: ${chassis}
 </#if>
 Technology Base: ${techBase} 
 Tonnage: ${tonnage}
--- a/megamek/resources/megamek/common/templates/tro/aero.ftl
+++ b/megamek/resources/megamek/common/templates/tro/aero.ftl
@@ -4,11 +4,7 @@ ${fullName}
 <#include "fluff.ftl">
 </#if>
 
-<#if isVSTOL && isConventional>
-Type: ${chassis} with VSTOL
-<#else>
 Type: ${chassis}
-</#if>
 Technology Base: ${techBase} 
 Tonnage: ${tonnage}
 Battle Value: ${battleValue}
--- a/megamek/resources/megamek/common/templates/tro/aero.ftl
+++ b/megamek/resources/megamek/common/templates/tro/aero.ftl
@@ -14,7 +14,7 @@ ${formatBasicDataRow("Engine", engineName, engineMass)}
 	Safe Thrust: ${safeThrust}
 	Max Thrust: ${maxThrust}
 ${formatBasicDataRow("Structural Integrity:", si, "")}
-<#if isVSTOL && isConventional>
+<#if isVSTOL && isConventional && !isSupportVehicle>
 ${formatBasicDataRow("VSTOL Equipment:", "", vstolMass)}
 </#if>
 ${formatBasicDataRow("Heat Sinks:", hsCount, hsMass)}
@@ -44,8 +44,6 @@ None
 
 Weapons
 ${formatEquipmentRow("and Ammo", "Location", "Tonnage", "Heat", "SRV", "MRV", "LRV", "ERV")}
-
-
 <#list equipment as eq>
 ${formatEquipmentRow(eq.name, eq.location, eq.tonnage, eq.heat, eq.srv, eq.mrv, eq.lrv, eq.erv)}
 </#list>
--- a/megamek/resources/megamek/common/templates/tro/aero.ftl
+++ b/megamek/resources/megamek/common/templates/tro/aero.ftl
@@ -14,6 +14,9 @@ ${formatBasicDataRow("Engine", engineName, engineMass)}
 	Safe Thrust: ${safeThrust}
 	Max Thrust: ${maxThrust}
 ${formatBasicDataRow("Structural Integrity:", si, "")}
+<#if isVSTOL && isConventional>
+${formatBasicDataRow("VSTOL Equipment:", "", vstolMass)}
+</#if>
 ${formatBasicDataRow("Heat Sinks:", hsCount, hsMass)}
 ${formatBasicDataRow("Fuel:", fuelPoints, fuelMass)}
 ${formatBasicDataRow(cockpitType, "", cockpitMass)}
@@ -40,11 +43,13 @@ None
 </#if>
 
 Weapons
-${formatEquipmentRow("and Ammo", "Location", "Tonnage", "Heat", "SRV", "MRV", "LRV", "ERV")}	
+${formatEquipmentRow("and Ammo", "Location", "Tonnage", "Heat", "SRV", "MRV", "LRV", "ERV")}
+
+
 <#list equipment as eq>
 ${formatEquipmentRow(eq.name, eq.location, eq.tonnage, eq.heat, eq.srv, eq.mrv, eq.lrv, eq.erv)}
 </#list>
-	
+
 <#if quirks??>
 Features the following design quirks: ${quirks}
 </#if>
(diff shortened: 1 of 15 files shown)


megamek/src/megamek/client/ui/swing/MechViewPanel.java
```diff
@@ -137,7 +137,7 @@ public void setMech(Entity entity, TROView troView){
         txtMek.setCaretPosition(0);
         Image image = FluffImageHelper.getFluffImage(entity);
         icon = null;
-        if(null != image) {
+        if(image != null) {
```
neoancient: Why are you changing the order? The habit of putting the constant first is deliberate.

Windchild292: Readability of code. Having the constant first protects against bad coding habits and typos, but significantly increases the difficulty of reading the code.

This way is the preferred styling for Java, as shown in most standard coding guides.

sixlettervariables: We generally avoid stylistic changes outside the immediate area being changed, if only to decrease the cognitive burden reviewing changes.

That said, as a long time C/C++ programmer, I probably introduced a lot of those.

neoancient: It only affects the readability of the code if you haven't gotten used to it yet. Having the constant second is more typical, but it is not best practice.

megamek/resources/megamek/common/templates/tro/aero.ftl
```diff
@@ -40,11 +43,13 @@ None
 </#if>
 
 Weapons
-${formatEquipmentRow("and Ammo", "Location", "Tonnage", "Heat", "SRV", "MRV", "LRV", "ERV")}	
+${formatEquipmentRow("and Ammo", "Location", "Tonnage", "Heat", "SRV", "MRV", "LRV", "ERV")}
+
+
```
neoancient: This adds blank lines between the table header and the first row.

megamek/resources/megamek/common/templates/tro/aero.ftlh
```diff
@@ -71,6 +71,7 @@
 	<tr><td>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Safe Thrust:</td><td align="center">${safeThrust}</td><td></td></tr>
 	<tr><td>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Max Thrust:</td><td align="center">${maxThrust}</td><td></td></tr>
 	<tr><td>Structural Integrity:</td><td align="center">${si}</td><td></td></tr>
+	<tr><td>VSTOL Equipment:</td><td align="center"></td><td align="center">${vstolMass}</td></tr>
```
neoancient: This adds a line for VSTOL equipment on all aerospace units, but it's only relevant for conventional fighters.

megamek/resources/megamek/common/templates/tro/aero.ftl
```diff
@@ -14,6 +14,9 @@ ${formatBasicDataRow("Engine", engineName, engineMass)}
 	Safe Thrust: ${safeThrust}
 	Max Thrust: ${maxThrust}
 ${formatBasicDataRow("Structural Integrity:", si, "")}
+<#if isVSTOL && isConventional>
```
neoancient: `isVSTOL` is never defined in the template data model, causing the text version to fail. This also needs to exclude fixed wing support, which gain VSTOL capabilities through a chassis modification, which is listed elsewhere. That's probably best done by changing the definition of isConventional to exclude support vehicles.

megamek/src/megamek/client/ui/swing/UnitSelectorDialog.java
```diff
@@ -706,17 +703,16 @@ void refreshUnitView() {
             // error unit didn't load right. this is bad news.
             populateTextFields = false;
         }
-        if (populateTextFields && (mechView != null)) {
+        if (populateTextFields) {
```
neoancient: As indicated by the comment above, if an exception is thrown while loading the unit, mechView will still not be initialized. The exception is already caught and logged, but removing this null check will cause another exception to be thrown which may cause the ui to stop responding.

megamek/src/megamek/common/Aero.java
```diff
@@ -3375,6 +3373,14 @@ public void setVSTOL(boolean b) {
         vstol = b;
     }
 
+    public double getVSTOLGearTonnage(){
```
neoancient: This duplicates code already present in TestAero. Where possible, calculations related to construction rules, when not required to load the unit or required during gameplay, should be placed the megamek.common.verifier package rather than the Entity. This will also report an incorrect value for FixedWingSupport, which adds VSTOL capability through a chassis modification.

megamek/src/megamek/common/MechView.java
```diff
@@ -296,7 +295,14 @@ public MechView(Entity entity, boolean showDetail, boolean useAlternateCost, boo
                         .append(warningEnd());
                 }
             }
-            sBasic.add(new LabeledElement(Messages.getString("MechView.Movement"), moveString.toString())); //$NON-NLS-1$
+
+            // TODO : Add STOL message as part of the movement line
+            if (isConvFighter && ((Aero) entity).isVSTOL()) {
+                sBasic.add(new LabeledElement(Messages.getString("MechView.Movement"),
+                        moveString.toString().concat(" (VSTOL)"))); //$NON-NLS-1$
```
neoancient: Strings that are displayed to the user should go in the i18n resource file.

megamek/src/megamek/common/MechView.java
```diff
@@ -453,19 +459,19 @@ public MechView(Entity entity, boolean showDetail, boolean useAlternateCost, boo
...
             sFluff.add(new LabeledElement("Capabilities", entity.getFluff().getCapabilities()));
         }
         
-        if (entity.getFluff().getDeployment() != "") {
+        if (!entity.getFluff().getDeployment().equals("")) {
             sFluff.add(new LabeledElement("Deployment", entity.getFluff().getDeployment()));
         }
         
-        if (entity.getFluff().getHistory() != "") {
+        if (!entity.getFluff().getHistory().equals("")) {
```
neoancient: This is better, but I would suggest String#isEmpty

megamek/src/megamek/client/ui/swing/UnitSelectorDialog.java
Windchild292: If you look under that comment, it sets populateTextFields to false. That will also trigger this IF statement as false, so the mechview != null check is not required.

megamek/src/megamek/common/MechView.java
Windchild292: How do you handle untranslated string for German or Russian? Should I do VSTOL_de, de_VSTOL, leave it out?

megamek/src/megamek/common/MechView.java
Windchild292: I agree, changed them all to isEmpty calls instead.

megamek/resources/megamek/common/templates/tro/aero.ftlh
Windchild292: Okay, this is weird, I swear that was between two if statements with the VSTOL checks.... Will fix this now.

megamek/resources/megamek/common/templates/tro/aero.ftl
Windchild292: Okay, now I'm sure there's something up, cause I know I put in isVSTOL (mostly because I forgot it originally and got an error during testing). I think my last commit might not have gone through...

megamek/src/megamek/client/ui/swing/UnitSelectorDialog.java
neoancient: Sure enough

megamek/src/megamek/common/MechView.java
neoancient: Don't worry about the translations. Just put it in the default locale so if or when someone volunteers to update the translations it will be there for them.

megamek/src/megamek/common/Aero.java
Windchild292: Good to know, and fixed.

megamek/src/megamek/client/ui/swing/MechViewPanel.java
Windchild292: Modern best practice is to make it readable, and code review to catch any issues with assignment versus a comparison.
Also, constant first is normally only used for ==, not != comparisons, as issues are more likely to occur there.

neoancient:
I thought I already merged this, but somehow I missed the last commit.

PR #1611, opened 2020-01-22
Remove AMS processing for single AMS vs arrow IV

mkerensky:
Fixes a goof where single AMS could engage Arrow IV homing missiles against the rules.
AMS bays can still engage ArrowIV attacks using capital missile rules per SO errata.


apokrovski:
For my own education, where's the AMS bay processing?

mkerensky:
See the 'handleAMS' method in the two ArtilleryIndirectHomingHandlers that this PR modifies.  It calls calcCounterAV from WeaponHandler and then deals with the reporting as appropriate for arrowIV missiles.

PR #1612, opened 2020-01-23
Rat generator updates

NickAragua:
This PR updates the RAT Generator data files to the format currently output by the RAT Generator Editor. 

Also, for experimentation's sake, I introduced the Bluehawk/Protector, the Hover Scout and Falcon Hovertank. I recommend viewing those two commits separately. 

Let me know if I'm off on my estimation of their frequency, it's a bit of a guessing game. The Bluehawk/Protector are supposed to be relatively common in some periphery nations, less common in others, and somewhat occasionally feature in FedSuns/Combine forces.

The Hover Scout and Falcon are intended to just kind of show up once in a blue moon for a little variety in crappy hovercraft.

Also was able to cherry-pick/manually reconstruct the LAM entries I made. The intent here is for them to be somewhat uncommon during the Star League heyday. The high-tech versions drop off during the succession wars while the regular ones kind of plod along. Then, during the end of the Jihad, you get a few WOB LAMs as well, and then they go extinct completely shortly afterwards.


neoancient:
Most of it is a guessing game. The frequencies look like they should match what you are going for. For model frequencies when there's one standard or dominant variant I usually give it an 8, then common variants come in around 6 and go down from there, unless there's some info in the TRO that indicates otherwise. Originally I started with 6 as the default, but I found that 8 gives more room to control the rate of introduction/phase out.

For the LAMs, we'll just have to see how it goes. How often they show up will depend largely on varied a faction's force is. The factions involved are pretty close in that regard, so the proportion of LAMs to standard 'Mechs shouldn't vary too much.

For the Bluehawk and Protector, should they continue on in the Alliance Military Corps? If so, I would add RA.OA to those.

PR #1617, opened 2020-01-26
Fixing Sail Integrity Display for Space Stations

Windchild292:


neoancient:
It is intended that the presence or absence of a sail be made explicit.

Windchild292:
@neoancient
That leaves the underlying issue unfixed then, as TROView.NA does not exist.
Fixes are therefore adding TROView.NA with a string value ("N/A"), or switching it to TROView.None (which would display "None")

Also, why was that decision made, and would it make sense to make it non-standard to show for Space Stations?

mkerensky:
I haven't really been keeping close track of this fix, but has it been considered that Space Stations can actually carry Jump Sails for recharging their energy storage batteries (if so equipped)?

Per SO p148, the sail weight/integrity is calculated as for a jumpship of equal weight.

Windchild292:
@mkerensky 
That is correct, but none of the prebuilt stations mount them and all currently show Sail Integrity !TROView.NA! instead of a number, N/A, or None.
Hence the suggestion and original fix was to hide it if it showed up null, as that was something I found only for Space Stations and one (very early) JumpShip

mkerensky:
Ok, cool.

neoancient:
There was no mention of any issue in this PR. The description was left blank. It was presented as a fix for a non-existent problem. If the string resource is missing, the solution is to add it. The reason it is made explicit is so it's evident whether or not there is a sail without having to look at the whole block. This is particularly useful for primitive jumpships, where the lack of a sail is atypical but something that someone would necessarily look for.

neoancient:
There was no mention of any issue in this PR. The description was left blank. It was presented as a fix for a non-existent problem. If the string resource is missing, the solution is to add it. The reason it is made explicit is so it's evident whether or not there is a sail without having to look at the whole block. This is particularly useful for primitive jumpships, where the lack of a sail is atypical but not something that someone would necessarily look for.

Windchild292:
I just realized the lack of a definition, that's my bad (thought I had included one with a picture showing the issue) @neoancient.
And I saw the lack of a tag, found that it was superfluous for Space Stations and I only saw one other location where it shows up, and determined this was the most logical solution for the code base. Will instead add the description as N/A (although it being standard for Space Stations is still weird to me)

Issue #1348, opened 2019-05-19
RFE LRM Indirect Fire and water in update.


AskAgain:
### Environment
mekhq-0.44.0-windows and mekhq-windows-0.45.4
Java version 1.8.0_171
Platform Windows 10 10.0 (x86)

### Description
This was an old issue that had previously been deemed in keeping with the rules, but I believe the LRM indirect fire mechanics were updated in Total Warfare CAT35001v to reflect that partial cover given by water does protect the legs from indirect fire.

### Files
![Total Warfare CAT35001v](https://user-images.githubusercontent.com/50790862/57978824-27862180-79c9-11e9-9562-6d265ef7e26f.jpg)

![She breaches](https://user-images.githubusercontent.com/50790862/57978856-a54a2d00-79c9-11e9-8bb0-ea9780a1bccf.JPG)

PR #1620, opened 2020-01-28
Allow indirect LRMs to miss legs if target in water

mkerensky:
This resolves RFE #1348, which references current errata for indirect fire on TW p111.

### DIFF
--- a/megamek/src/megamek/common/weapons/WeaponHandler.java
+++ b/megamek/src/megamek/common/weapons/WeaponHandler.java
@@ -1560,7 +1560,10 @@ public class WeaponHandler implements AttackHandler, Serializable {
 
         //For indirect fire, remove leg hits only if target is in water partial cover
         //Per TW errata for indirect fire
-        if ((!isIndirect || (isIndirect && targetHex.containsTerrain(Terrains.WATER)))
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/weapons/WeaponHandler.java
```diff
@@ -1555,8 +1556,11 @@ protected void handleEntityDamage(Entity entityTarget,
         }
         boolean isIndirect = wtype.hasModes()
                 && weapon.curMode().equals("Indirect");
+        IHex targetHex = game.getBoard().getHex(target.getPosition());
 
-        if (!isIndirect
+        //For indirect fire, remove leg hits only if target is in water partial cover
+        //Per TW errata for indirect fire
+        if ((!isIndirect || (isIndirect && targetHex.containsTerrain(Terrains.WATER)))
```
neoancient: This would work in most situations, but a mech standing in depth 0 water (or a superheavy in 0-1) receiving partial cover from terrain would incorrectly have the partial cover hit removed.

mkerensky: Latest commit appears to fix this.

PR #1626, opened 2020-01-31
Add LargeCraft Roles to RatGen

mkerensky:
Adds various large craft roles to MissionRole so that search checkboxes work


megamek/src/megamek/client/ratgenerator/MissionRole.java
```diff
@@ -418,6 +419,132 @@ public static Double adjustAvailabilityByRole(double avRating, Collection<Missio
...
+                    } else if (mRec.getRoles().contains(POCKET_WARSHIP)) {
+                        avRating += avAdj[0];
+                    } else {
+                        return null;
+                    }
+                    break;
+				case MECH_CARRIER:
+				    if (mRec.getRoles().contains(MECH_CARRIER)) {
+                        avRating += avAdj[2];
+                    } else if (mRec.getRoles().contains(TROOP_CARRIER)) {
```
neoancient: TROOP_CARRIER is a combined forces transport, but it may just be vees and infantry. MECH_CARRIER should just be those with mech bays.

megamek/src/megamek/client/ratgenerator/MissionRole.java
```diff
@@ -418,6 +419,132 @@ public static Double adjustAvailabilityByRole(double avRating, Collection<Missio
...
+                        avRating += avAdj[0];
+                    } else {
+                        return null;
+                    }
+                    break;
+				case ASF_CARRIER:
+				    if (mRec.getRoles().contains(ASF_CARRIER)) {
+                        avRating += avAdj[2];
+                    } else if (mRec.getRoles().contains(ASSAULT)) {
```
neoancient: Similarly, this should probably be implemented as a filter that returns null for any unit that does not contain ASF_CARRIER.

megamek/src/megamek/client/ratgenerator/MissionRole.java
mkerensky: I had done this intentionally because a number of the ships listed as troop carriers include (substantial) mech carrier capabilities -  Colossus, Excalibur...
I recognize that every now and again it'll return a ship with no mech carrying capacity. 

For myself, I've updated the forcegen files to reflect each individual role as well where this is the case, but since my files are customized I haven't submitted those changes to the repo.

megamek/src/megamek/client/ratgenerator/MissionRole.java
mkerensky: Same as above, this is intentional since there are so few defined ASF carriers but so many assault ships carry about 6 fighters.

megamek/src/megamek/client/ratgenerator/MissionRole.java
neoancient: Ok, I'm good with that.

PR #1644, opened 2020-02-02
Add theme selection and dark themes

sixlettervariables:
This adds theme selection to MM and adds dark themes. Still TODO:

- Making the themes selectable by name rather than class name.
- Making tables use the correct row backgrounds
- Make all the various sub-UI's support theming

image

image

image

image


neoancient:
A few of the dialogs switch to the new theme (configure bot, configure unit, planetary conditions, at least), but most of them do not switch unless restarted. Is it feasible to get them all to change to the new theme?

sixlettervariables:
Interesting. I thought I had them all changing, but I'll check again. Hopefully this isn't a machine dependent problem!

PR #1654, opened 2020-02-07
Don't count multiple slots for armored cockpit

### LINKED ISSUE #422: Spotted units are highlighted
Can we get an indicator similar to the targeting modifiers, that gets attached to a unit when it is spotted for indirect fire?

neoancient:
Per rules for armored components (Tac Ops, p. 282), adding armor to any cockpit system (including command console) simply adds one ton to the cockpit weight. Currently we are adding one ton per slot, which results in an extra ton for multi-slot cockpits (command console, quadvee, unofficial dual cockpit).

Fixes #422

### DIFF
--- a/megamek/src/megamek/common/verifier/TestMech.java
+++ b/megamek/src/megamek/common/verifier/TestMech.java
@@ -848,7 +848,7 @@ public class TestMech extends TestEntity {
                     // Armored cockpit (including command console) adds 1 ton, regardless of number of slots
                     if ((cs.getType() == CriticalSlot.TYPE_SYSTEM)
                             && (cs.getIndex() == Mech.SYSTEM_COCKPIT)) {
-                        cockpitWeight = 1.0;;
+                        cockpitWeight = 1.0;
                     } else {
                         weight += 0.5;
                     }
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/verifier/TestMech.java
```diff
@@ -840,21 +840,22 @@ public String getName() {
...
         for (int location = Mech.LOC_HEAD; location < mech.locations(); location++) {
             for (int slot = 0; slot < mech.getNumberOfCriticals(location); slot++) {
                 CriticalSlot cs = mech.getCritical(location, slot);
                 if ((cs != null) && cs.isArmored()) {
-                    weight += 0.5;
-
+                    // Armored cockpit (including command console) adds 1 ton, regardless of number of slots
                     if ((cs.getType() == CriticalSlot.TYPE_SYSTEM)
                             && (cs.getIndex() == Mech.SYSTEM_COCKPIT)) {
+                        cockpitWeight = 1.0;;
```
NickAragua:
```suggestion
                        cockpitWeight = 1.0;
```

PR #1656, opened 2020-02-08
have rat generator wait a few seconds to load rats

NickAragua:
Addresses MekHQ #1455, and probably many other complaints about AtB civilian units and turrets not generating.


NickAragua:
To reproduce it, I had to put a breakpoint in the RAT loading thread.

On my machine it takes about 500ms to load the RATs. On Windchild's machine, according to the logs, it takes ~1000ms. I figure setting it to 5s will allow all but the most potato of computers to finish running the RAT loading thread.

PR #1660, opened 2020-02-10
Narrow/low profile quirk

NickAragua:
This is my crack at the updated narrow/low profile quirk. It wasn't nearly as bad as I initially thought, but still, a lot of changes.

The highlights:
- consolidated logic for determining whether a hit is a glancing blow or a narrow/low-profile glancing blow
- consolidated logic for adding glancing blow reports
- consolidated logic for applying glancing blow effects


NickAragua:
There are a few weapon types for which I'm a little confused as to the effect:
ACAPWeaponHandler (is this "autocannon with armor-piercing munitions"?)
MGAWeaponHandler (machine gun array?)
RifleWeaponHandler
SRMTandemChargeHandler

Some of these make calls to hit.setGlancingBlow(), which, I guess, modifies subsequent crit rolls? Not sure if those should stack for narrow/lowprofile glancing blows. I'm guessing yes, but I'm not at all familiar with most of those rules, so...

neoancient:
Where are the rules for interaction between narrow/low profile and glancing blows? I don't see anything in the glancing blows rules (TacOps, 80), or the quirk rules (StratOps, p. 195; BMM, p. 85).

NickAragua:
On BMM page 85, it says that the effects stack. 

So I guess means that, for the classes I mentioned, I should introduce a mechanism to explicitly update the 'glancing mod' for a ToHit roll, or to stack it up somehow.

HammerGS:
> There are a few weapon types for which I'm a little confused as to the effect:
> ACAPWeaponHandler (is this "autocannon with armor-piercing munitions"?) - yep
> MGAWeaponHandler (machine gun array?) - yep
> RifleWeaponHandler - Isn't this part of the infantry,the default attach?
> SRMTandemChargeHandler
> 
> Some of these make calls to hit.setGlancingBlow(), which, I guess, modifies subsequent crit rolls? Not sure if those should stack for narrow/lowprofile glancing blows. I'm guessing yes, but I'm not at all familiar with most of those rules, so...

Did some checking on the rules team area, and errata's. The Armor Piercing AC and Tandem charge looks like they retain their bonuses to crits even on the glancing blow.

NickAragua:
Ok, made the requested code changes. I'll be carrying out some testing over time and will keep an eye on whether the math works out on glancing blows. Good thing I've got some Marauders and Vulcans in my "test" campaign.

PR #1668, opened 2020-02-13
WIP - Personnel Improvements - MegaMek-side Changes

Windchild292:
DO NOT MERGE without the MekHQ branch, as it will break MekHQ.

This PR is being opened a bit earlier than the MekHQ one that I can get some advice on the changes in crew, specifically the extraData inclusion.


megamek/src/megamek/client/RandomNameGenerator.java
```diff
@@ -356,6 +348,59 @@ public String generate(boolean isFemale) {
...
+                if ((null != ethnicities) && (ethnicities.size() > 0)) {
+                    String eFirst = ethnicities.get(Compute.randomInt(ethnicities.size()));
+                    // ok now we can get the first and last name vectors
+                    if (isClan) {
+                        eFirst = eLast;
+                    }
+                    Vector<String> firstNames = firstM.get(eFirst);
+                    if (isFemale) {
+                        firstNames = firstF.get(eFirst);
+                    }
```
sixlettervariables: Candidate for ternary operator simplification.

megamek/src/megamek/common/Crew.java
```diff
@@ -1354,10 +1385,64 @@ public boolean doConsoleRoleSwap() {
         return false;
     }
 
+    //region extraData
+    public void initializeExtraDataAsHashMap(int crewNumber) {
```
sixlettervariables: Will every crew member have extra data? If not this should probably not assign a new hashmap for each crew member and just lazily initialize them in the add routines below.

megamek/src/megamek/common/Crew.java
```diff
@@ -1354,10 +1385,64 @@ public boolean doConsoleRoleSwap() {
...
+        this.extraData = new HashMap<>(crewNumber);
+        for (int i = 0; i < crewNumber; i++) {
+            this.extraData.put(i, new HashMap<>());
```
sixlettervariables: What is the expected size of these hash maps?

megamek/src/megamek/common/Crew.java
```diff
@@ -1354,10 +1385,64 @@ public boolean doConsoleRoleSwap() {
...
+    public void setExtraData(Map<Integer, Map<String, String>> extraData) {
+        this.extraData = extraData;
+    }
+
+    public void setExtraDataForCrewMember(int crewIndex, Map<String, String> dataMap) {
+        this.extraData.put(crewIndex, dataMap);
+    }
+
+    public void addExtraDataDataPoint(int crewIndex, String key, String value) {
+        this.extraData.get(crewIndex).put(key, value);
```
sixlettervariables: Should this create the hashmap if the crew member's slot doesn't exist?

megamek/src/megamek/common/Crew.java
```diff
@@ -1354,10 +1385,64 @@ public boolean doConsoleRoleSwap() {
         return false;
     }
 
+    //region extraData
```
sixlettervariables: Are all of these overloads used/necessary?

megamek/src/megamek/common/Crew.java
Windchild292: Maximum would be crew size, so in most cases (given the primary use case is for OpFor generation in MekHQ) under 10 instances for the first, then 2 to 4 values for the second set of maps.

megamek/src/megamek/common/Crew.java
Windchild292: This is the primary use case for MekHQ-side implementation, as every crew member will have a given name and gender as the minimum use case, which makes this an easier way than lazily initializing it.

megamek/src/megamek/common/Crew.java
Windchild292: If you mean the region tags, they make it far easier to parse through files in the IDE... so I would say yes.
If you mean the methods, probably not. I made a bunch just in case I need them, but wanted some advice on my implementation (to ensure I'm not doing something crazy) before I went and did the full implementation. I'll remove anything that I find useless following my implementation, or potentially leave in with documentation on how to use them.

megamek/src/megamek/common/Crew.java
Windchild292: Yes, it should. Fixing that.

megamek/src/megamek/common/Crew.java
sixlettervariables: I'd prefer removing anything unused.

megamek/src/megamek/common/Crew.java
sixlettervariables: My preference would be to not initialize the crew member slots to anything.

megamek/src/megamek/common/Crew.java
sixlettervariables: If we expect just 2 or 3 items per crew member, set the default capacity of the inner hash maps to 2 or 4.

megamek/src/megamek/client/RandomNameGenerator.java
Windchild292: I agree, fixed

megamek/src/megamek/common/MULParser.java
```diff
@@ -1183,6 +1161,15 @@ private void setPilotAttributes(Crew crew, int slot, Map<String,String> attribut
                 crew.setExternalIdAsString(attributes.get(EXT_ID), slot);
             }
 
+            if (attributes.containsKey(EXTRA_DATA)) {
+                Map<String, String> extraData = new HashMap<>();
+                String[] valuePairs = attributes.get(EXTRA_DATA).split(",");
+                for (int i = 0; i < valuePairs.length; i = i + 2) {
+                    //this is always in pairs, so it will not cause an out of bounds error
```
sixlettervariables: As MUL's can come from anywhere, this isn't a safe assumption.

megamek/src/megamek/common/Crew.java
```diff
@@ -1488,6 +1488,30 @@ public String getExtraDataValue(int crewIndex, String key) {
...
+
+            String[] keySet = new String[dataRow.keySet().size()];
+            dataRow.keySet().toArray(keySet);
+
+            for (int k = 0; k < keySet.length; k++) {
+                output.append(keySet[k]).append(",").append(dataRow.get(keySet[k]));
+                if (k < keySet.length - 1) {
+                    output.append(",");
+                }
+            }
```
sixlettervariables: Why not a `for` over the `entrySet`? This is pretty expensive in terms of allocations.

Windchild292: Easier parsing on load, and... This isn't that bad in terms of allocations?
EntrySet is a harder entity to parse on load and is more limiting to potential characters in the set.

megamek/src/megamek/common/MULParser.java
Windchild292: Given how they MUST be saved in String, String pairs, anything else should break on load.
Because anything else means you've got invalid data saved.

megamek/src/megamek/common/Crew.java
sixlettervariables: I mean:
```
boolean first = false;
for (EntrySet<K, V> row : dataRow.entrySet()) {
  if (!first) {
    output.append(",");
  } else {
    first = false;
  }
  output.append(row.getKey()).append(",").append(row.getValue());
}
```

megamek/src/megamek/common/MULParser.java
sixlettervariables: Then I'd rather it throw a specific exception than index out of range, but that's my druthers.

megamek/src/megamek/common/Crew.java
sixlettervariables: Also, would it be more natural to split these int pairs by `|` and key-values by `=`? This way reading the file is obvious as to the intent.

Side question: is `,`, `|`, or `=` valid in a key or value?

PR #1675, opened 2020-02-15
Replacing any Final Static references with Static Final

Windchild292:
This PR standardizes our declarations to use static final instead of a mixture of that and final static. This was a find all mentions and replace.


NickAragua:
For my education, what's the difference?

Windchild292:
Nothing, just switching between the two in the same project is considered to be a style issue and can be annoying to some people.

PR #1676, opened 2020-02-15
Removing potential of a divide by 0 error in Random Army Creator

Windchild292:
Also did some minor refactoring while I was in the file.


Windchild292:
Changed the ArrayList to List to get ahead of potential review comments

PR #1677, opened 2020-02-15
Freezers

neoancient:
This fixes a couple things related to prototype double heat sinks which I ran into while working on MegaMek/megameklab#438. The rules are found in IO, p. 71. The brief summary is that the unit may replace any number of single heat sinks with a prototype double heat sink, but the double heat sinks must always be assigned critical slots (on Mechs), even if this leaves some of the engine capacity unmet. The two types of prototype double heat sinks are identical in function, but not considered the same part, as the one developed in the 3020s (often called "freezers") were developed independently rather than being recovered Star League lostech.

1. During validation, ignore the prototype double heat sinks when checking whether all the engine internal heat sinks are accounted for.
2. Set the tech level to experimental, which is the same for all prototype equipment.

### DIFF
--- a/megamek/src/megamek/common/verifier/TestMech.java
+++ b/megamek/src/megamek/common/verifier/TestMech.java
@@ -553,7 +553,7 @@ public class TestMech extends TestEntity {
                         && (m.getUsableShotsLeft() <= 1)) {
                     continue;
                 }
-                if ((mech instanceof Mech) && (m.getType().getCriticals(mech) == 0)) {
+                if (m.getType().getCriticals(mech) == 0) {
                     continue;
                 }
                 if (!(m.getType() instanceof MiscType)) {
--- a/megamek/src/megamek/common/verifier/TestMech.java
+++ b/megamek/src/megamek/common/verifier/TestMech.java
@@ -541,19 +541,19 @@ public class TestMech extends TestEntity {
         return false;
     }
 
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/common/verifier/TestMech.java
```diff
@@ -589,21 +589,19 @@ public boolean checkCriticalSlotsForEquipment(Entity entity,
...
-        if ((countInternalHeatSinks > engine.integralHeatSinkCapacity(mech
-                .hasCompactHeatSinks()))
-                || ((countInternalHeatSinks < engine
-                        .integralHeatSinkCapacity(mech.hasCompactHeatSinks()))
-                        && (countInternalHeatSinks != ((Mech) entity)
-                                .heatSinks()) && !entity.isOmni())) {
-            heatSinks.addElement(Integer.valueOf(countInternalHeatSinks));
+        if ((countInternalHeatSinks > engine.integralHeatSinkCapacity(this.mech.hasCompactHeatSinks()))
+                || ((countInternalHeatSinks < engine.integralHeatSinkCapacity(this.mech.hasCompactHeatSinks()))
+                        && (countInternalHeatSinks != mech.heatSinks(false))
```
neoancient: This is the only logic change in this file. The boolean parameter to Mech#heatSinks tells it to ignore prototypes.

PR #1680, opened 2020-02-16
Fix #1679 Vertical Lift Off PSR incorrectly applying crater modifier

### LINKED ISSUE #1679: Vertical Lift Off PSR incorrectly applying crater modifier
### Environment
- MegaMek-0.47.4
- Standalone
- Linux
- openjdk version 1.8.0_232
- There's nothing relevant in megameklog.txt

### Description
When performing a "Vertical Lift Off" the resulting PSR is incorrectly having the +3 "Taking off from Crater" modifier applied. This applies to ASFs, LAMs in ASF mode, but NOT to spheroid dropships.

Looking at  megamek.common.IAero.java 366-388, "a crater" is defined as a hex where the level of all surrounding hexes is greater than what we are sitting on.

### Files
Crater_Takeoff

Crater_Takeoff_Dropship

RAldrich:
Make crater check logic include the Entity's current hex, to support single hex Entities (e.g. ASFs)

#1679

### DIFF
--- a/megamek/src/megamek/common/IAero.java
+++ b/megamek/src/megamek/common/IAero.java
@@ -367,7 +367,9 @@ public interface IAero {
         // TW doesn't define what a crater is, assume it means that the hex
         // level of all surrounding hexes is greater than what we are sitting on
         boolean allAdjacentHigher = true;
-        Set<Coords> positions = new HashSet<Coords>(((Entity) this).getSecondaryPositions().values());
+        Set<Coords> positions = new HashSet<>();
+        positions.add(pos);
+        positions.addAll((((Entity) this).getSecondaryPositions().values()));
         IHex adjHex;
         for (Coords currPos : positions) {
             hex = ((Entity) this).getGame().getBoard().getHex(currPos);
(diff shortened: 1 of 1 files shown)


RAldrich:
This was manually tested using both an ASF and a spheroid dropship in both flat terrain and using the craters on the Moonscape1 map. Everything works as expected. I didn't bother optimizing the dropship check, it redundantly checks that the area under the dropship footprint is flat, but that seems fairly harmless.

RAldrich:
> Could be simplified by retaining the original line and your `positions.add(pos)` (omitting the addAll)

True! Is that worth posting an update for? (Also, it looks like I don't have permissions to manually link this to the issue it's fixing - not sure why it didn't link automatically.)

And just because it's not documented, what's the process for merging? I assume a maintainer will pull it in when preparing to compile the next dev release?

NickAragua:
Yeah, go ahead and push the update. GitHub should update the pull request if you push from the same branch. Once we "pass" the code review, one of the devs will merge the PR and close the associated issue.

RAldrich:
> Yeah, go ahead and push the update. GitHub should update the pull request if you push from the same branch. Once we "pass" the code review, one of the devs will merge the PR and close the associated issue.

Pushed! Thanks for the guidance!

Issue #1468, opened 2019-09-02
Request: allow VTOL Battle Armor to set starting height on the lobby


binaryspica:
### Environment
Ver 0.47.0
Windows 7

### Description
As title says. Clans VTOL Battle Armors are able to move as VTOL, but they are not able to set the starting height so they need to begin the game on the ground.

PR #1685, opened 2020-02-17
Allow Starting Height to VTOL Infantry/Battle Armor 

binaryspica:
Resolving #1468. Also allows VTOL Infantry to set the starting height.

Also I need to have a question: How high they can ascend? I have checked Total Warfare but there is nothing so I set the maximum height as normal VTOLs, but is it correct?

### DIFF
--- a/megamek/src/megamek/client/ui/swing/CustomMechDialog.java
+++ b/megamek/src/megamek/client/ui/swing/CustomMechDialog.java
@@ -1098,7 +1098,7 @@ public class CustomMechDialog extends ClientDialog implements ActionListener,
             }
         }
 
-        if ((isVTOL) && (height > 50)
+        if ((isVTOL && height > 50)
                 || (isAirMech && height > 25)
                 || (isGlider && height > 12)) {
             msg = Messages.getString("CustomMechDialog.EnterCorrectHeight"); //$NON-NLS-1$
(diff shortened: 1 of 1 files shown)


binaryspica:
> The rules for VTOL movement (TW, p. 54) say that other units that share the VTOL movement mode use the same rules. So there is no need to distinguish between VTOL vehicles and infantry. The check for isVTOL should just check for movement mode and apply to all VTOL units.
> I don't know where the maximum of 50 comes from. TW says the maximum elevation for a VTOL is 500.

Will MegaMek support that much elevation? If so then there is no problem to change it, though.

neoancient:
I don't know of any reason it wouldn't, though I don't see a good reason to go that high. An elevation of 50 would put you out of weapons range. Let's just leave it alone and focus on one issue.

binaryspica:
Anyway thanks for the reference. Well, then what would be better, leave its maximum to 50 or expand it to 500 as rules says?

neoancient:
Leave it alone and just deal with the issue of setting starting height for VTOL BA and conventional infantry.

binaryspica:
Yes. Please note me if I miss anything else.

binaryspica:
Is there any problem? I thought it seems good to go....

HammerGS:
I still show requested changes, and right now lots of the devs are swamped.

binaryspica:
The requested change was about maximum height but I thought it was good to leave it to be. Was I misunderstood that?

NickAragua:
Right, I believe neo wanted you to adjust the logic so that all units with VTOL movement type behave identically.

neoancient:
Right now they do behave identically. There's no need to track that with three separate variables that have the same value.

binaryspica:
Got it. Well, is it possible to set the default height other than 0?

PR #1686, opened 2020-02-17
Allow units with C3 and Nova CEWS to link with the units with C3 and without Nova on lobby

### LINKED ISSUE #1664: A unit with C3 and Nova CEWS cannot access for C3 network of the unit without CEWS.
### Environment
Ver 0.47.4

### Description
As title says. If you made a custom unit that has C3(either C3 Master or C3 Slave) and Nova Combined Electronic Warfare System, for now they can only make the network with the other units that also have Nova CEWS. For example:

Unit **M1N** and **M2N** have C3 Master **and** Nova CEWS
Unit **M1** and **M2** have C3 Master but does not have Nova CEWS

Unit **S1N** has C3 Slave **and** Nova CEWS
Unit **S1** has C3 Slave but does not have Nova CEWS

Then, in the lobby, if you put **M1N** and **M1** as the company master, **M1N** only accepts **M2N** as its platoon master, and **M1** only accepts **M2** as its platoon master. It is not possible to connect **M1N** and **M2**, and **M1** and **M2N**.

The same also goes with the slave; You cannot connect **S1N** to **M2**, and **S1** and **M2N** also have the same problem.

I didn't checked Improved C3(C3i) but I suspect that it may have the same problem.

Also, I have checked IO and errata but nothing says you can't put C3 and Nova in the same unit.

[Test Master - Nova.zip](https://github.com/MegaMek/megamek/files/4184060/Test.Master.-.Nova.zip)

binaryspica:
I hope that it will fix #1664. 

Currently, you can't link Nova in the lobby unlike C3, and you can only do that by type /nova link numbers123 manually. And the check only see if both the unit and target have Nova or does not have Nova. So the phrase is unnecessary I think. Is there anything I have missed?

### DIFF
--- a/megamek/src/megamek/client/ui/swing/EquipChoicePanel.java
+++ b/megamek/src/megamek/client/ui/swing/EquipChoicePanel.java
@@ -1739,10 +1739,6 @@ public class EquipChoicePanel extends JPanel implements Serializable {
                 if (entity.hasNavalC3() != e.hasNavalC3()) {
                     continue;
                 }
-                // likewise can't connect c3 to nova
-                if (entity.hasNovaCEWS() != e.hasNovaCEWS()) {
-                    continue;
-                }
                 // maximum depth of a c3 network is 2 levels.
                 Entity eCompanyMaster = e.getC3Master();
                 if ((eCompanyMaster != null)


HammerGS:
See rules note on Bug report prior to merging.

megamek/src/megamek/client/ui/swing/EquipChoicePanel.java
```diff
@@ -1739,10 +1739,6 @@ private void refreshC3() {
                 if (entity.hasNavalC3() != e.hasNavalC3()) {
                     continue;
                 }
-                // likewise can't connect c3 to nova
```
neoancient: Removing this check lets a unit with a C3 slave but no Nova CEWS connect to a Nova network.

binaryspica: But isn't it only applies on the lobby? As I said above you can't link Nova on the lobby for now. Although I am not used to the structure of MegaMek but it seems that the part is only concern about config unit on the lobby but not the other parts.

It would be possible to only modify the sentence rather than remove it, though. Someone may make the option to start with Nova links later.

binaryspica: Also since it is the list of C3 links, modify it will allow a unit with Nova(but not C3) to link with an another unit with C3.

binaryspica: Well, this section(config unit on the lobby) have no ability to link Nova network at all, so the phrase is unnecessary anyways. For now you cannot link Nova network because this section provides nothing for link Nova networks. It only provides the links for C3, C3i, and Naval C3.

Else, is it better to make a new request that allows to set the Nova network BETWEEN THE ANOTHER UNIT WITH NOVA on the lobby?

PR #1690, opened 2020-02-18
Check equipment compatibility with motive type when validating vehicles

neoancient:
This is part of the fix for MegaMek/megameklab#440. The bug report only mentions combat vehicle flotation hull, but I have included the amphibious and dune buggy CV mods, as well as industrial equipment and a few other equipment types I noticed. The logic is in a public static method in TestTank where it can be used by TestSupportVehicle, as well as MML in equipment filtering.

### DIFF
--- a/megamek/src/megamek/common/verifier/TestTank.java
+++ b/megamek/src/megamek/common/verifier/TestTank.java
@@ -406,6 +406,10 @@ public class TestTank extends TestEntity {
                     correct = false;
                 }
             }
+            if (!legalForMotiveType(m.getType(), tank.getMovementMode())) {
+                buff.append(m.getType().getName()).append(" is incompatible with ").append(tank.getMovementModeAsString());
+                correct = false;
+            }
         }
         for (int loc = 0; loc < tank.locations(); loc++) {
             int count = 0;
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/common/verifier/TestTank.java
```diff
@@ -451,6 +458,71 @@ public boolean correctEntity(StringBuffer buff, int ammoTechLvl) {
...
+                        && !mode.equals(EntityMovementMode.WIGE);
+            }
+            if (eq.hasFlag(MiscType.F_AP_POD)) {
+                return !isNaval;
+            }
+        } else if (eq instanceof WeaponType) {
+            if (((WeaponType) eq).getAmmoType() == AmmoType.T_BPOD) {
+                return !isNaval;
+            }
+            if (((WeaponType) eq).getAmmoType() == AmmoType.T_NAIL_RIVET_GUN) {
```
NickAragua: What? VTOLs can't mount rivet guns? That's BS, I demand a refund! ::monocle::

neoancient: I think it's a total crock that submarines can't have combines. Hello? Kelp forests?

PR #1691, opened 2020-02-19
Keep BA AP weapons matched to their mount in the blk file.

neoancient:
BattleArmor anti-personnel weapons can be installed in either an AP mount or an armored glove. The relationship is tracked through Mounted::setLinked. When the BA unit is loaded the links are established during post-load initialization by linking each weapon to the first mount available in the list. For BA with a single AP mount this works fine, but BA with two armored gloves can find the weapon switching hands and BA with an AP mount and one or more armored gloves can have the weapon switched to a different type of mount. I used an approach similar to what is used for weapon bays, skipping the AP weapon when iterating through equipment and writing it directly after the mount that links to it. Then when loading any AP weapon is linked to the last eligible mount that was loaded.

Fixes MegaMek/megameklab#441.

### DIFF
--- a/megamek/src/megamek/common/loaders/BLKBattleArmorFile.java
+++ b/megamek/src/megamek/common/loaders/BLKBattleArmorFile.java
@@ -26,14 +26,7 @@
  */
 package megamek.common.loaders;
 
-import megamek.common.AmmoType;
-import megamek.common.BattleArmor;
-import megamek.common.Entity;
-import megamek.common.EntityMovementMode;
-import megamek.common.EquipmentType;
-import megamek.common.LocationFullException;
-import megamek.common.Mounted;
-import megamek.common.TechConstants;
+import megamek.common.*;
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/common/loaders/BLKBattleArmorFile.java
```diff
@@ -237,13 +231,24 @@ protected void loadEquipment(Entity t, String sName, int nLoc) throws EntityLoad
                     try {
                         Mounted m = t.addEquipment(etype, nLoc, false, 
                                 mountLoc, dwpMounted);
-                        if (numShots != 0 && m != null
```
neoancient: Entity#addEquipment never returns null. It can throw an exception, but it is guaranteed to be non-null in the try block. The test was confusing code analysis tools into flagging subsequent dereferences as potential NPEs.

PR #1693, opened 2020-02-19
Fix avoid faillure table of Radical Heat Sink

### LINKED ISSUE #1688: Radical Heat Sink starts from 3+ rather than 2+
### Environment
Ver 0.47.4

### Description
Per Radical Heat Sink Failure Table, pg. 89. IO, if number of turns used is 0 then the avoid failure on 2+, but for now it starts with 3+, both mechs and aerospace fighters.

binaryspica:
Fixes #1688. It increase the number of consecutive uses AFTER check the roll for the failure, and adds '0' section which avoid failure on 2+, as 89.IO says.

### DIFF
--- a/megamek/src/megamek/server/Server.java
+++ b/megamek/src/megamek/server/Server.java
@@ -19986,7 +19986,6 @@ public class Server implements Runnable {
             int radicalHSBonus = 0;
             Vector<Report> rhsReports = new Vector<Report>();
             if (entity.hasActivatedRadicalHS()) {
-                entity.setConsecutiveRHSUses(entity.getConsecutiveRHSUses() + 1);
                 if (entity instanceof Mech) {
                     radicalHSBonus = ((Mech) entity).getActiveSinks();
                 } else if (entity instanceof Aero) {
@@ -19997,6 +19996,9 @@ public class Server implements Runnable {
                 int rhsRoll = Compute.d6(2);
                 int targetNumber = 2;
                 switch (entity.getConsecutiveRHSUses()) {
+                    case 0:
+                        targetNumber = 2;
+                        break;
                     case 1:
                         targetNumber = 3;
                         break;
@@ -20032,6 +20034,7 @@ public class Server implements Runnable {
                 r.add(rhsRoll);
                 r.choose(rhsFailure);
                 rhsReports.add(r);
+                entity.setConsecutiveRHSUses(entity.getConsecutiveRHSUses() + 1);
 
                 if (rhsFailure) {
                     entity.setHasDamagedRHS(true);


megamek/src/megamek/server/Server.java
```diff
@@ -19997,6 +19996,9 @@ private void resolveHeat() {
                 int rhsRoll = Compute.d6(2);
                 int targetNumber = 2;
                 switch (entity.getConsecutiveRHSUses()) {
+                    case 0:
+                        targetNumber = 2;
+                        break;
```
NickAragua: This statement is kind of pointless, since the default value of targetNumber is 2.

megamek/src/megamek/server/Server.java
```diff
@@ -19997,6 +19996,9 @@ private void resolveHeat() {
                 int rhsRoll = Compute.d6(2);
                 int targetNumber = 2;
                 switch (entity.getConsecutiveRHSUses()) {
```
NickAragua: Is it just me or is does this switch statement have an unclosed curly bracket? How the hell does this even compile?

megamek/src/megamek/server/Server.java
binaryspica: But after that, you can found that...

>                     case 6:
>                     default:
>                         targetNumber = 13; // Auto-fail

Is it better to modify the default value and put =13 to Case 6?

NickAragua: Think I'm going to take a look at this after I get home, since I can't look at the full Server.java class via the web interface.

megamek/src/megamek/server/Server.java
binaryspica: I don't know, but I suspect that the problem is it cannot be open in the browser or you need to check it by raw data. I have to download GitHub Desktop to apply it.

Also, it seems the reason why I am keep failed to find this - I am constantly find Report(5000); but there is nothing on the search.

megamek/src/megamek/server/Server.java
mkerensky: I would use the constant TargetRoll.AUTOMATIC_FAIL instead of 13 for your default case.

Windchild292: Actually, it isn't because of a default value statement in the switch. Default value is 13.

megamek/src/megamek/server/Server.java
Windchild292: It just looks like it does because of Github struggling with the file size

megamek/src/megamek/server/Server.java
binaryspica: Could someone change 13 to TargetRoll.AUTOMATIC_FAIL ? I can't fix it by the web browser and I am still not used to GitHub Desktop.

Windchild292:
I'll fix and test this for you @binaryspica.

binaryspica:
Close, for #1688 will handle the issue instead.

PR #1697, opened 2020-02-19
Spot welder heat

neoancient:
Spot welders generate 2 points of heat when used as a weapon (TW, p. 148, footnote in the table). But unlike any other equipment, they are treated as energy weapons only when the unit does not have a fission or fusion engine (TM, p. 248). The special exception does not affect power amplifiers, other than having to loop through the misc equipment, since fusion and fission engines never require power amplifiers. But for heat sink calculations I had to add an engine check to determine whether to include it in the heat sink requirement.

Fixes MegaMek/megameklab#455

### DIFF
--- a/megamek/src/megamek/common/verifier/TestAero.java
+++ b/megamek/src/megamek/common/verifier/TestAero.java
@@ -603,12 +603,17 @@ public class TestAero extends TestEntity {
                 if ((linkedBy != null) && 
                         (linkedBy.getType() instanceof MiscType) && 
                         linkedBy.getType().hasFlag(MiscType.F_PPC_CAPACITOR)){
-                    weight += ((MiscType)linkedBy.getType()).getTonnage(aero);
+                    weight += linkedBy.getType().getTonnage(aero);
                 }
             }
-            // Power amp weighs: 
+            for (Mounted m : aero.getMisc()) {
+                if (m.getType().hasFlag(MiscType.F_CLUB) && m.getType().hasSubType(MiscType.S_SPOT_WELDER)) {
+                    weight += m.getType().getTonnage(aero);
+                }
+            }
+            // Power amp weighs:
             //   energy weapon tonnage * 0.1 rounded to nearest half ton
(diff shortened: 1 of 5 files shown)


megamek/src/megamek/common/verifier/TestAero.java
```diff
@@ -603,12 +603,17 @@ public double getWeightPowerAmp() {
                 if ((linkedBy != null) && 
                         (linkedBy.getType() instanceof MiscType) && 
                         linkedBy.getType().hasFlag(MiscType.F_PPC_CAPACITOR)){
-                    weight += ((MiscType)linkedBy.getType()).getTonnage(aero);
+                    weight += linkedBy.getType().getTonnage(aero);
                 }
             }
-            // Power amp weighs: 
+            for (Mounted m : aero.getMisc()) {
+                if (m.getType().hasFlag(MiscType.F_CLUB) && m.getType().hasSubType(MiscType.S_SPOT_WELDER)) {
```
NickAragua: Aerospace fighters can mount spot welders? wtf?

NickAragua: and clubs.

NickAragua: Well, never mind, just spot welders, which are apparently a subset of club.

Windchild292: I'm sorry, what?

Is that meant to only be for LAMs?

NickAragua: I guess that's a theoretically legit use case. Well, whatever, we're just implementing the rules.

neoancient: That's for conventional fighters, and I was just blindly thinking about how they use the same rules as vehicles for determining power amplifier weight without any consideration for the equipmrnt in question. Conventional fighter cannot in fact mount spot welders so while this doesn't hurt anything, the check is meaningless.

PR #1702, opened 2020-02-20
Refactoring and Code Cleanup In Server.java

Windchild292:
This is a low hanging fruit pass to fix some of the many issues in this file. All of the functionality changes will have a comment that can be used to track where they are.


megamek/src/megamek/server/Server.java
```diff
@@ -1095,6 +1080,7 @@ private IPlayer addNewPlayer(int connId, String name) {
     public void validatePlayerInfo(int playerId) {
         final IPlayer player = getPlayer(playerId);
 
+        // TODO : Windchild can this be removed?
         // maybe this isn't actually useful
         // // replace characters we don't like with "X"
         // StringBuffer nameBuff = new StringBuffer(player.getName());
```
Windchild292: Can we remove this area of commented out code?

megamek/src/megamek/server/Server.java
```diff
@@ -2887,9 +2832,7 @@ private void endCurrentPhase() {
                 changePhase(IGame.Phase.PHASE_EXCHANGE);
                 break;
             case PHASE_EXCHANGE:
-                game.addReports(vPhaseReport);
-                changePhase(IGame.Phase.PHASE_SET_ARTYAUTOHITHEXES);
-                break;
+                // TODO : Windchild this might have been a bug
```
Windchild292: Is this change correct or was it a bug that it was identical to the next Phase?

megamek/src/megamek/server/Server.java
```diff
@@ -3233,55 +3166,53 @@ private void receiveForwardIni(Packet packet, int connid) {
...
-                                         && !(turn instanceof GameTurn.UnloadStrandedTurn)
-                                         && !(turn instanceof GameTurn.TriggerBPodTurn) && !(turn instanceof GameTurn
-                    .TriggerAPPodTurn));
+            boolean isGeneralMoveTurn = !(turn instanceof GameTurn.SpecificEntityTurn)
+                    && !(turn instanceof GameTurn.UnitNumberTurn)
+                    && !(turn instanceof GameTurn.UnloadStrandedTurn)
+                    && !(turn instanceof GameTurn.TriggerBPodTurn)
+                    && !(turn instanceof GameTurn.TriggerAPPodTurn);
+            // TODO : Windchild my IDE has noticed that GameTurn.TriggerBPodTurn
+            // TODO : and GameTurn.TriggerAPPodTurn are always false, and can probably be removed
```
Windchild292: Thoughts on this change?

megamek/src/megamek/server/Server.java
```diff
@@ -4277,25 +4138,24 @@ private void applyDropshipProximityDamage(Coords centralPos,
...
                 }
 
-                alreadyHit = artilleryDamageHex(pos, centralPos, damageDice,
-                                                null, killer.getId(), killer, null, false, 0,
-                                                vPhaseReport, false, alreadyHit, true);
+                alreadyHit = artilleryDamageHex(pos, centralPos, damageDice, null, killer.getId(),
+                        killer, null, false, 0, vPhaseReport, false,
+                        alreadyHit, true);
 
+                // TODO : Windchild can this be removed?
```
Windchild292: Can this commented out code be removed?

megamek/src/megamek/server/Server.java
```diff
@@ -6297,36 +6057,36 @@ private void doVehicleFlipDamage(Tank entity, int damage, boolean startRight, in
...
-            hit = new HitData(startRight? Tank.LOC_LEFT : Tank.LOC_RIGHT);
-            break;
-        default:
-            hit = null; //Motive damage instead
+            case 0:
+                hit = new HitData(startRight ? Tank.LOC_RIGHT : Tank.LOC_LEFT);
+                break;
+            case 1:
+                hit = new HitData(Tank.LOC_TURRET);
+                break; //TODO : Windchild Ensure this change is correct
```
Windchild292: I'm pretty sure this is a bug, but wanted to check that the missing break wasn't on purpose.

megamek/src/megamek/server/Server.java
```diff
@@ -10058,6 +9670,7 @@ private void processTrailerMovement(Entity tractor, List<Coords> trainPath) {
                 } else if (tractor.isSuperHeavy()) {
                     // If the tractor is superheavy, we can put two trailers in each hex
                     // starting trailer 0 in the hex behind the tractor
+                    // TODO : Windchild the update to trailerPositionOffset is never used?
                     trailerPositionOffset = (Math.ceil((trailerPositionOffset /= 2.0)) + 1);
```
Windchild292: Why is this being updated if the update is never being used?

megamek/src/megamek/server/Server.java
```diff
@@ -10069,6 +9682,7 @@ private void processTrailerMovement(Entity tractor, List<Coords> trainPath) {
                     // Otherwise, we can put two trailers in each hex
                     // starting trailer 1 in the hex behind the tractor
                     trailerPositionOffset ++;
+                    // TODO : Windchild the update to trailerPositionOffset is never used?
                     trailerPositionOffset = Math.ceil((trailerPositionOffset /= 2.0));
```
Windchild292: Same as above, why is this being updated if the update is never being used.

megamek/src/megamek/server/Server.java
```diff
@@ -12125,6 +11660,7 @@ private boolean checkVibrabombs(Entity entity, Coords coords,
      * @param coords
      *            The <code>Coords</code> from which to remove minefields
      */
+    // TODO : Windchild Can this be removed?
```
Windchild292: Can this commented out method be removed?

megamek/src/megamek/server/Server.java
```diff
@@ -15888,13 +15303,9 @@ private void removeDuplicateAttacks(int entityId) {
         if (null != en) {
             allowed = en.getAllowedPhysicalAttacks();
         }
-        Vector<EntityAction> toKeep = new Vector<EntityAction>(/*
-                                                                * game.actionsSize
-                                                                * ()
-                                                                */);
+        Vector<EntityAction> toKeep = new Vector<>(/*game.actionsSize()*/); //TODO : Windchild Can this be removed?
```
Windchild292: can this commented out code be removed?

megamek/src/megamek/server/Server.java
```diff
@@ -17783,11 +17118,11 @@ private void resolveClubAttack(PhysicalResult pr, int lastEntityId) {
             }
         }
 
-        if (((MiscType) caa.getClub().getType())
-                    .hasSubType(MiscType.S_WRECKING_BALL)
-            && (ae instanceof SupportTank) && (te instanceof Mech)) {
+        if (caa.getClub().getType().hasSubType(MiscType.S_WRECKING_BALL)
+                && (ae instanceof SupportTank) && (te instanceof Mech)) {
             // forces a PSR like a charge
             if (ae instanceof LandAirMech && ae.isAirborneVTOLorWIGE()) {
```
Windchild292: My IDE suggests this is a bug, as it is always false. Is it supposed to be ||, not &&?

megamek/src/megamek/server/Server.java
```diff
@@ -23913,6 +23053,8 @@ private void resolveShutdownCrashes() {
                             / game.getOptions().intOption(
                                     OptionsConstants.ADVCOMBAT_VEHICLES_THRESHOLD_DIVISOR));
 
+                    // TODO : Windchild fix integer division in floating point context
```
Windchild292: I'm pretty sure this should be fixed, as it is integer division in floating point context. (can't remember why I wrote it as a comment, probably because it is a variable option)

megamek/src/megamek/server/Server.java
```diff
@@ -24823,6 +23938,7 @@ private void resolveShutdownCrashes() {
                 impactArmor = (te instanceof Mech)
                         && (te.getArmorType(hit.getLocation()) == EquipmentType.T_ARMOR_IMPACT_RESISTANT);
                 bar5 = te.getBARRating(hit.getLocation()) <= 5;
+                // TODO : Windchild the above is never used?
```
Windchild292: The above is never used, so can it be removed or is it a bug?

megamek/src/megamek/server/Server.java
```diff
@@ -25064,11 +24169,12 @@ private int damageExternalPassenger(Entity te, HitData hit, int damage,
...
-            int hits) {
-        if(!(en instanceof Mech) && !(en instanceof QuadMech)
-            && !(en instanceof BipedMech) && !(en instanceof Aero)
-            && !(en instanceof Tank)) {
+    private boolean checkEngineExplosion(Entity en, Vector<Report> vDesc, int hits) {
+        // TODO : Windchild can I be simplified? - en instance of QuadMech is always false, as is
+        // TODO : en instanceof BipedMech according to my IDE
+        if (!(en instanceof Mech) && !(en instanceof QuadMech)
+                && !(en instanceof BipedMech) && !(en instanceof Aero)
+                && !(en instanceof Tank)) {
```
Windchild292: Can I be simplified? - en instance of QuadMech is always false, as is  en instanceof BipedMech according to my IDE

megamek/src/megamek/server/Server.java
```diff
@@ -27786,49 +26771,46 @@ private void applyCargoCritical(Aero aero, int damageCaused, Vector<Report> repo
...
         return criticalEntity(en, loc, isRear, 0, false, false, damage);
     }
 
     /**
      * Rolls one critical hit
      */
-    public Vector<Report> oneCriticalEntity(Entity en, int loc,
-                                             boolean isRear, int damage, boolean damagedByFire) {
+    // TODO : Windchild can this be removed? - never called
+    public Vector<Report> oneCriticalEntity(Entity en, int loc, boolean isRear, int damage,
```
Windchild292: Can this be removed? It is never called

megamek/src/megamek/server/Server.java
```diff
@@ -30588,35 +29482,26 @@ public void addSmoke(ArrayList<Coords> coords, int windDir, boolean bInferno) {
...
-                    r.add(smokeCoords.getBoardNum());
-                    addReport(r);
                     smokeLevel = SmokeCloud.SMOKE_HEAVY;
                 } else {
-                    smokeLevel = Math.max(smokeLevel, SmokeCloud.SMOKE_HEAVY);
                     r = new Report(5185, Report.PUBLIC);
-                    r.add(smokeCoords.getBoardNum());
-                    addReport(r);
-                    smokeLevel = SmokeCloud.SMOKE_HEAVY;
+                    smokeLevel = Math.max(smokeLevel, SmokeCloud.SMOKE_HEAVY); // TODO : Windchild verify this change is correct
```
Windchild292: We were doing some math then overwriting it. Should be the math.max, or the SmokeCloud.SMOKE_HEAVY?

megamek/src/megamek/server/Server.java
```diff
@@ -33107,16 +31912,15 @@ public void run() {
         final String METHOD_NAME = "run()";
         Thread currentThread = Thread.currentThread();
         getLogger().info(getClass(), METHOD_NAME, "s: listening for clients...");
-        // HashSet<IConnection> toUpdate = new HashSet<IConnection>();
+        // HashSet<IConnection> toUpdate = new HashSet<IConnection>(); // TODO : Windchild Can this be removed?
```
Windchild292: Can this be removed?

megamek/src/megamek/server/Server.java
```diff
@@ -33129,11 +31933,11 @@ public void run() {
...
                     watchdogTimer.schedule(w, 1000, 500);
                 }
-            } catch (InterruptedIOException iioe) {
+            } catch (InterruptedIOException ignored) {
                 // ignore , just SOTimeout blowing..
-            } catch (IOException ex) {
+            } catch (IOException ignored) { }
 
-            }
+            // TODO : Windchild Can this be removed?
```
Windchild292: There's a bunch of commented out code below, can that be removed?

megamek/src/megamek/server/Server.java
```diff
@@ -33527,7 +32315,8 @@ private boolean checkBuildingCollapseWhileMoving(Building bldg,
         int toInf = Math.round(damage * percent);
 
         // some buildings scale remaining damage
-        // TODO: this isn't quite right for castles brian
+        // TODO : this isn't quite right for castles brian
+        // TODO : Windchild this is never used?
         damage = (int) Math.floor(bldg.getDamageToScale() * toInf);
```
Windchild292: This is never used, so can it be removed or is it a bug?

megamek/src/megamek/server/Server.java
```diff
@@ -33592,12 +32377,11 @@ private boolean checkBuildingCollapseWhileMoving(Building bldg,
         } // Handle the next entity
 
         // If we found any infantry, add a line to the phase report.
-        if (foundInfantry) {
+        if (foundInfantry) { // TODO : Windchild this is always false?
```
Windchild292: This is always false according to my IDE, is that meant to have happened or is it a bug?

megamek/src/megamek/server/Server.java
```diff
@@ -33784,6 +32561,8 @@ public boolean checkForCollapse(Building bldg,
                 recheckLoop = false;
                 // recheck if the basement collapsed since the basement falls
                 // might trigger a greater collapse.
+
+                // TODO : Windchild the recheck loop in this region all register as issues on my IDE
```
Windchild292: The recheck loop here all register as unused by my IDE. Is that a bug or just old code?

megamek/src/megamek/server/Server.java
```diff
@@ -37272,35 +35915,35 @@ private void handleAttacks(boolean pointblankShot) {
...
 
     /**
-     * create a <code>SmokeCloud object and add it to the server list</code>
+     * create a <code>SmokeCloud</code> object and add it to the server list
      *
-     * @param coords
+     * @param coords the location to create the cloud
      */
-    public void createSmoke(Coords coords) {
+    public void createSmoke(Coords coords) { // TODO : Windchild is this needed? No usages
```
Windchild292: Do we still need this method given it has zero usages?

megamek/src/megamek/server/Server.java
```diff
@@ -37272,35 +35915,35 @@ private void handleAttacks(boolean pointblankShot) {
...
-     * create a <code>SmokeCloud object and add it to the server list</code>
+     * create a <code>SmokeCloud</code> object and add it to the server list
      *
-     * @param coords
-     * @param level  1=Light 2=Heavy Smoke
+     * @param coords the location to create the smoke
+     * @param level  1=Light 2=Heavy Smoke 3:light LI smoke 4: Heavy LI smoke
      */
-    public void createSmoke(Coords coords, int level) {
+    public void createSmoke(Coords coords, int level) { // TODO : Windchild is this needed? No usages
```
Windchild292: Is this needed? No usages

megamek/src/megamek/server/Server.java
```diff
@@ -37490,20 +36129,22 @@ private void registerWithServerBrowser(boolean register, String urlString) {
                     }
                 }
             } else {
+                // TODO : Windchild Can this be removed?
```
Windchild292: This does nothing, can it just be removed?

megamek/src/megamek/server/Server.java
```diff
@@ -37490,20 +36129,22 @@ private void registerWithServerBrowser(boolean register, String urlString) {
...
-        } catch (Exception e) {
+        } catch (Exception ignored) {
         }
     }
 
     public Set<Coords> getHexUpdateSet() {
         return hexUpdateSet;
     }
 
+    // TODO : Windchild can this be removed?
```
Windchild292: This is never called, can it be removed?

megamek/src/megamek/server/Server.java
neoancient: Yes, that's a bug.

megamek/src/megamek/server/Server.java
neoancient: This is correct. In a regular game, PHASE_SET_ARTYAUTOHITHEXES comes after PHASE_EXCHANGE. In a scenario it comes after PHASE_STARTING_SCENARIO.

megamek/src/megamek/server/Server.java
neoancient: Yes, the quad and biped checks are redundant, as well as discriminatory toward TripodMech.

megamek/src/megamek/server/Server.java
neoancient: This was commented out over ten years ago. I'm going to go out on a limb and say it's safe to get rid of it.

megamek/src/megamek/server/Server.java
Windchild292: There are Tripod Mechs?

megamek/src/megamek/server/Server.java
neoancient: I'm not sure what was going on here, but it's been like that for over six years. Let's get rid of it.

megamek/src/megamek/server/Server.java
neoancient: That's been commented out for over ten years. Dump it.

megamek/src/megamek/server/Server.java
neoancient: yes

megamek/src/megamek/server/Server.java
neoancient: That looks like old debugging code. I don't see any reason to keep it.

Windchild292: Found a few other places with obvious debugging code, removed them too.

megamek/src/megamek/server/Server.java
neoancient: As near as I can tell with a method that's nearly 2k lines long, This is here to deal with the possibility of a different armor type when excess damage transfers to another location and was copied from the code that deals with the original hit. The only place it makes a difference is in the case of an acid warhead. I don't know the rules off the top of my head, but from what's in the code it does extra damage to some types of armor, including BAR <= 5, but the extra damage does not transfer. If that's correct, then it's not a bug and this line can be removed.

megamek/src/megamek/server/Server.java
neoancient: I don't see any need to maintain unused versions of overloaded methods.

megamek/src/megamek/server/Server.java
neoancient: I think this may be a bug. I suspect it should be changing the value of `toInf`. I don't know the rules involved off the top of my head.

megamek/src/megamek/server/Server.java
neoancient: I don't see any point to it.

megamek/src/megamek/server/Server.java
neoancient: It is a bug. It should be `te instanceof LandAirMech`. When an airborne LAM in AirMech mode is the target of a successful charge attack, it makes a control roll instead of a PSR. I don't know why it's restricting the attacker to a support vehicle, though. I also can't find a rule that says a successful hit is like a charge. The physical attacks table on TW, p. 214 says that on a hit roll of 2, the attacker deals damage to itself and has to make a PSR. I think this whole block may be wrong but it needs a more thorough rules check.

megamek/src/megamek/server/Server.java
neoancient: I don't know what the Math.max line is there for. We get to this block if there is a fire that generates heavy smoke and there is not already heavy smoke in that hex. So `smokeLevel = SmokeCloud.SMOKE_HEAVY` is correct.

megamek/src/megamek/server/Server.java
neoancient: That looks like it goes with the block of commented out code at the end of the method, so it can go with it.

megamek/src/megamek/server/Server.java
neoancient: My guess is that it should be set to true just after the Compute#isInBuilding check, which leads to reports generated. On the other hand, there's a new line added after the damage report if the infantry takes damage, and I don't think we need two blank lines. But there's no blank line between units if one of them takes no damage. So it's not too consistent. I would put a blank line in at the end if the block that starts with if (Compute.isInBuilding... and remove foundInfantry.

megamek/src/megamek/server/Server.java
neoancient: The two if statements that follow the for loop should be inside it.

megamek/src/megamek/server/Server.java
neoancient: Unused versions of overloaded methods. They can go.

megamek/src/megamek/server/Server.java
neoancient: Unused versions of overloaded methods. They can go.

megamek/src/megamek/server/Server.java
neoancient: Looking at how it's used, I don't think there should be a setter for this.

megamek/src/megamek/server/Server.java
```diff
@@ -23396,14 +22468,14 @@ private void resolveShutdownCrashes() {
...
-                    damage = te.getArmor(hit) >= 3 ? 3 : te.getArmor(hit);
+                    damage = Math.min(te.getArmor(hit), 3);
                     r = new Report(6061);
                     r.subject = te_n;
                     r.indent(2);
                     r.add(damage);
                     vDesc.addElement(r);
                 } else if (isPlatoon) {
-                    damage *= 1.5;
+                    damage = (int) Math.round(damage * 1.5);
```
neoancient: I double checked the rules on this. Acid damage to conventional infantry is rounded up.

megamek/src/megamek/server/Server.java
```diff
@@ -3885,32 +3766,31 @@ public boolean accept(Entity entity) {
                 // If there's only one team moving, we don't need to bother
                 // with the evenTracker, just make sure the even turns are
                 // evenly distributed
-                numEven += (teamEvenTurns / min) + 0.5;
+                numEven += (int) Math.round((teamEvenTurns / min) + 0.5);
```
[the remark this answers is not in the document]
Windchild292: Good to know, removed the addition and swapped to Math.ceil
I also round a place where we are using Math.round(Math.ceil(division)), so I removed the outer rounding.

megamek/src/megamek/server/Server.java
```diff
@@ -1379,22 +1350,22 @@ public boolean loadGame(File f) {
...
-     *         successfull
+     * @return A <code>boolean</code> value whether or not the loading was successful
      */
     public boolean loadGame(File f, boolean sendInfo) {
         final String METHOD_NAME = "loadGame(File,boolean)";
         getLogger().info(getClass(), METHOD_NAME, "s: loading saved game file '" + f + "'"); //$NON-NLS-1$
 
         IGame newGame;
-        try(InputStream is = new GZIPInputStream(new FileInputStream(f))) {
+        try {
```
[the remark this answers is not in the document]
Windchild292: Fixed this so it actually closes properly with a resources try-catch. My apologies, I forgot to update following comments on my LGTM PR for MekHQ.

I found pair of other locations that could use the same implementation, so I fixed them too.

megamek/src/megamek/server/Server.java
Windchild292: It is a bug, it should be scaling toInf.

megamek/src/megamek/server/Server.java
```diff
@@ -3766,7 +3745,7 @@ public boolean accept(Entity entity) {
                 // If there's only one team moving, we don't need to bother
                 // with the evenTracker, just make sure the even turns are
                 // evenly distributed
-                numEven += (int) Math.round((teamEvenTurns / min) + 0.5);
+                numEven += (int) Math.ceil(teamEvenTurns / min);
```
neoancient: What I was trying to say is that what was originally here (`teamEvenTurns / min + 0.5`) is the equivalent of `(int) Math.round(teamEvenTurns / min)`. Changing it to `(int) Math.round((teamEvenTurns / min) + 0.5)` is the equivalent of `(int) Math.ceil(teamEvenTurns / min)`, which changes the logic. Unless the original logic is wrong, this should be 
```suggestion
                numEven += (int) Math.round(teamEvenTurns / min);
```

Windchild292: Nope, I just did a late night change that doesn't make sense to me either. Fixing in the next few min.

PR #1711, opened 2020-02-22
Add Entity.getPriceMultiplier() method

RAldrich:
Refactor various Entity subtypes to move multiplier data from getCost() into getPriceMultiplier().

This enables external access to the multiplicative construction cost modifiers, which is needed to support
more accurate price calculations in MekHQ.

### DIFF
--- a/megamek/src/megamek/common/Infantry.java
+++ b/megamek/src/megamek/common/Infantry.java
@@ -1507,7 +1507,6 @@ public class Infantry extends Entity {
   */
     @Override
     public double getCost(boolean ignoreAmmo) {
-        double multiplier = 1;     //Cost Multiplier per TM
         double pweaponCost = 0;  //Primary Weapon Cost
         double sweaponCost = 0; // Secondary Weapon Cost
         double armorcost = 0; //Armor Cost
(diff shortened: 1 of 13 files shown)


RAldrich:
Manually tested, didn't notice any changes to costs (none are expected, this is just refactoring).

I have a MekHQ PR draft ready to go that's dependent on these changes.

Next step in this work would be to decouple the cost breakdown display from the cost calculations. Once that's done we'll have more freedom to redesign the cost calculations themselves, and maybe get Entity's cost structure into a form that plays nice with MekHQs concept of Parts.

megamek/src/megamek/common/Infantry.java
```diff
@@ -1577,75 +1576,79 @@ else if(nSneak == 3) {
...
+    }
+
+    @Override
+    public double getPriceMultiplier() {
+        double priceMultiplier = 1.0;
 
         //Anti-Mek Trained Multiplier
         if (isAntiMekTrained()) {
-            multiplier = 1;
+            priceMultiplier *= 5.0;
```
sixlettervariables: Why 5x for anti-mek trained? Previous code was 1x (unless this diff is odd).

RAldrich: Oh sorry, I should have called that out in the commit. 5.0x is the correct multiplier per the Tech Manual (p. 282).

It looks like this flag might not actually be getting used though - I see no cost difference between the regular and Anti-Mech foot platoons with this change applied.

rroby1: TM page 282 shows anti-mech training is x5 price multiplier

sixlettervariables: @RAldrich that'll be an existing bug then.

PR #1718, opened 2020-02-25
Move primary determination of equipment weight to Mounted.

neoancient:
This adds a getTonnage method to Mounted and changes all the places in MM where getTonnage is called on the mounted EquipmentType to use Mounted instead. The immediate reason for doing this is that the weight of the dumper depends on the size of the cargo bay it's connected to, and the EquipmentType itself has no way of determining that, even with the Entity and location provided. I have also moved some of the effects of BA mount types on weight into this method.

This is something I've been contemplating for some time, though it had nothing to do with the peculiar needs of dumpers. The next step in this process is to add a size field to Mounted so we can eliminate the need for multiple versions of equipment in different sizes such as communications equipment and ladders. It also provides an easier-to-use way of adding capacity to equipment like MASH theaters and drone operating systems without having to have a second piece of equipment just to expand.

### DIFF
--- a/megamek/src/megamek/common/verifier/TestProtomech.java
+++ b/megamek/src/megamek/common/verifier/TestProtomech.java
@@ -448,7 +448,7 @@ public class TestProtomech extends TestEntity {
                 continue;
             }
             slotsByLoc.merge(mount.getLocation(), 1, Integer::sum);
-            weightByLoc.merge(mount.getLocation(),  mount.getTonnage(), Double::sum);
+            weightByLoc.merge(mount.getLocation(), mount.getTonnage(), Double::sum);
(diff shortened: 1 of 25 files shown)


megamek/src/megamek/common/verifier/TestProtomech.java
```diff
@@ -449,8 +448,7 @@ public boolean hasIllegalEquipmentCombinations(StringBuffer buff) {
                 continue;
             }
             slotsByLoc.merge(mount.getLocation(), 1, Integer::sum);
-            weightByLoc.merge(mount.getLocation(),
-                    mount.getType().getTonnage(proto, mount.getLocation()), Double::sum);
+            weightByLoc.merge(mount.getLocation(),  mount.getTonnage(), Double::sum);
```
Windchild292: Weird spacing before mount

PR #1722, opened 2020-02-27
1477: Finished Implementing UserData in MegaMek

Windchild292:
This is being opened up for logical comments on the implementation of MegaMekFile handling directories before I start going down that rabbit hole.

It should be able to parse through files or directories, and find them in either place. It defaults to using userData for files, while subdirectories default to the standard file system.

For example, a user can change out the surnames.txt file, and only need to have that file in userdata while the rest of the files and folders are picked up from the standard file system.


megamek/src/megamek/common/util/MegaMekFile.java
```diff
@@ -1,43 +1,179 @@
...
+        }
+    }
+
+    /**
+     *
+     * @param files an array of files to split between subdirectories and files
+     * @param type true if you want to return files, false for subdirectories
+     * @return a list of either files or subdirectories based on the value of type
+     */
+    private List<File> splitFilesAndSubdirectories(File[] files, boolean type) {
```
sixlettervariables: This is not side effect free and is somewhat confusing when layered with the code calling it above. It is hard to reason about what exactly happens.

megamek/src/megamek/common/util/MegaMekFile.java
```diff
@@ -1,43 +1,179 @@
...
+        isDirectory = standardVersion.isDirectory();
+
+        if (isDirectory) {
+            // Create an array of files located in the standard file directory
+            List<File> standardFiles = splitFilesAndSubdirectories(standardVersion.listFiles(), true);
+
+            if (userDataVersion.exists()) {
+                // If we have a user data version of the directory, then we need to do some additional
+                // processing based on the files located within it
+                List<File> userDataSubdirectories = splitFilesAndSubdirectories(standardVersion.listFiles(), false);
```
sixlettervariables: This doesn't go but one layer deep, right?

Windchild292: Correct. Anything deeper should call this again

Windchild292: I'll put more thought into use cases to determine if this is the best way forward or if it would be better to immediately handle subdirectories.

Windchild292: I've decided that subdirectories should be handled here too, to clean up code elsewhere.

Windchild292:
I should probably open any zips in the MegaMekFile to determine between files and directories, and then search their contents, right? As then we can ensure that anything is being overwritten properly?

Just spitballing an idea/potential issue here.

Windchild292:
Another idea for this is a merge file contents, although that will probably be annoying to do.

Windchild292:
To make this reviewable and possible I'm going to need to implement this in pieces. I'm going to first rename the current MegaMekFile to OldeMegaMekFile, then start implementing the new version in MegaMek. This is because there are going to be significant changes to properly implement the directory changes.

PR #1727, opened 2020-02-28
Change the reference of unofficial Gatling Autocannons.

binaryspica:
Their reference are 207. TO, but that is the description of minefields. I think that it is the typo of 207. TM, the description of autocannons. These weapons are unofficial armaments so no one cares for its true reference, though.


HammerGS:
I would prefer something like Unofficial or a short form of it.

binaryspica:
If you allow it, then it is nothing wrong to do.

PR #1729, opened 2020-02-29
1688: Fixing Radical Heat Sink Target Number

Windchild292:
This PR is originally by @binaryspica, subsuming #1693 and fixing #1688.
I've also added a fix to the gitignore file to handle custom boards.


binaryspica:
I think that you will handle this better.

PR #1731, opened 2020-02-29
Mech zero crit equipment

neoancient:
This involves a series of exceptions to the way nearly everything else works in order to support the construction of 'Mechs with the smart robotic control system (SRCS; IO, p. 140-1). It occupies no critical slots, so it's not assigned a location. Since 'Mechs don't have a generic location like the vehicle body or aerospace hull/fuselage, it doesn't get written to the mtf file or displayed on the summary or TRO view.

I started with a bit of cleanup on the MTF file processing, including replacing String literals with named constants. I tried to make this generic rather than specific to the SRCS, so any equipment that takes up zero crits gets added to the MTF on a line that has the format "nocrits:[internalName]:[location abbrev]". Except Clan CASE, which gets handled differently for now.

I added some checks to the validation code to flag any unit that mounts multiple robotics systems, and added C3 system type counting at the same time.

The most awkward code is the part that shows the SRCS on the MechView summary and the TRO view.

Fixes MegaMek/megameklab#496

### DIFF
--- a/megamek/src/megamek/common/verifier/TestEntity.java
+++ b/megamek/src/megamek/common/verifier/TestEntity.java
@@ -1397,7 +1397,7 @@ public abstract class TestEntity implements TestEntityOption {
             illegal = true;
         }
         if (networks > 1) {
-            buff.append("Cannot multiple multiple network types on the same unit.\n");
+            buff.append("Cannot have multiple network types on the same unit.\n");
             illegal = true;
         }
         if (robotics > 1) {
@@ -1652,4 +1652,3 @@ class Structure {
     }
 
 } // End class Structure
-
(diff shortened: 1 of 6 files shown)


megamek/src/megamek/common/verifier/TestEntity.java
```diff
@@ -1375,7 +1396,14 @@ public boolean hasIllegalEquipmentCombinations(StringBuffer buff) {
             buff.append("Cannot mount HarJel repair system on non-Mech\n");
             illegal = true;
         }
-        
+        if (networks > 1) {
+            buff.append("Cannot multiple multiple network types on the same unit.\n");
```
sixlettervariables:
```suggestion
            buff.append("Cannot have multiple network types on the same unit.\n");
```

PR #1733, opened 2020-03-01
Creating static getLogger() in MegaMek.java

Windchild292:
This allows us to just call MegaMek.getLogger() instead of having a private variable, and it has been implemented in three files alongside a bunch of code cleanup.


megamek/src/megamek/MegaMek.java
```diff
@@ -121,23 +108,22 @@ public static void main(String[] args) {
...
         } catch (CommandLineParser.ParseException e) {
-            StringBuilder message = new StringBuilder(INCORRECT_ARGUMENTS_MESSAGE)
-                    .append(e.getMessage()).append('\n');
-            message.append(ARGUMENTS_DESCRIPTION_MESSAGE);
-            MegaMek.displayMessageAndExit(message.toString(),
-                                          "main(String[])");
+            String message = INCORRECT_ARGUMENTS_MESSAGE + e.getMessage() + '\n'
+                    + ARGUMENTS_DESCRIPTION_MESSAGE;
+            getLogger().log(MegaMek.class, "main(String[])", LogLevel.INFO, message);
+            TimerSingleton.getInstance().killTimer();
```
sixlettervariables: Is that needed if we're calling System.exit below?

megamek/src/megamek/MegaMek.java
```diff
@@ -255,41 +242,43 @@ public static String getMegaMekSHA256() {
...
-            if (!logDir.exists()) {
-                logDir.mkdir();
+        getLogger().info(MegaMek.class, "redirectOutput",
+                "Redirecting output to " + logFileName);
+        String sLogDir = PreferenceManager.getClientPreferences().getLogDirectory();
+        File logDir = new File(sLogDir);
+        if (!logDir.exists()) {
+            if (!logDir.mkdir()) {
+                getLogger().error(MegaMek.class, "redirectOutput",
+                        "Error in creating directory ./docs. We know this is annoying, and apologise. "
```
sixlettervariables: Is this really creating a directory called `./docs`? Wouldn't it be `./logs`?

megamek/src/megamek/MegaMek.java
Windchild292: Was copying origin functionality, but looking at it again that doesn't make sense to me.

megamek/src/megamek/MegaMek.java
Windchild292: Copy-paste error, fixing

megamek/src/megamek/MegaMek.java
Windchild292: It runs as a daemon thread, so not needed, as it will be garbage collected on exit.

megamek/src/megamek/MegaMek.java
```diff
@@ -73,33 +62,30 @@
...
-
         String logFileName = DEFAULT_LOG_FILE_NAME;
-
         CommandLineParser cp = new CommandLineParser(args);
 
         try {
             cp.parse();
             String lf = cp.getLogFilename();
             if (lf != null) {
-                if (lf.equals("none") || lf.equals("off")) { //$NON-NLS-1$ //$NON-NLS-2$
```
neoancient: Why are you removing this (here and elsewhere)?

Windchild292: Visually distracting and increases the parsing time of a line of code, for something that is only used by a single IDE while running a specific setting.

Windchild292:
Committed change

Windchild292:
@sixlettervariables 
For future PRs, can you please @me when you add the conflicts label? It does not notify me, nor does it show up in the Slack feed, so I had no idea this had any issues

PR #1735, opened 2020-03-03
Fix kg-standard weight calculations

neoancient:
I broke this somewhere along the line by not passing the entity to the rounding method, so it couldn't figure out whether to round to the half-ton or the kg. I added the argument, then removed the version of RoundWeight#round that doesn't have an Entity parameter to prevent the same mistake in the future. It's only used internally.

This purpose of the RoundWeight parameter to EquipmentType#getTonnage is to allow the standard rounding system to be overridden to support fractional accounting.

### DIFF
--- a/megamek/src/megamek/common/RoundWeight.java
+++ b/megamek/src/megamek/common/RoundWeight.java
@@ -80,17 +80,6 @@ public enum RoundWeight {
     }
 
     /**
-     * Applies the rounding operation to a weight. For methods that depend on whether the unit
-     * uses the ton or kilogram standard, assumes ton.
-     *
-     * @param weight The weight to be rounded, in metric tons.
-     * @return       The result of the rounding operation.
-     */
-    public double round(double weight) {
-        return round(weight, null);
-    }
-
-    /**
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/common/RoundWeight.java
```diff
@@ -118,7 +107,7 @@ public static double nearestHalfTon(double weight) {
      * @return       The weight in tons, rounded to the closest ton.
      */
     public static double nearestTon(double weight) {
-        return NEAREST_TON.round(weight);
+        return NEAREST_TON.round(weight, null);
```
NickAragua: What's the null do?

neoancient: The only one that needs an Entity argument is STANDARD, which will pass off to next half ton or nearest kg depending on the Entity type. If it's null, it defaults to next half ton. The rest ignore it.

Windchild292:
It appears that this is causing issues with the Clan Machine Gun, which has a weight of 0.25 tons.

neoancient:
It's related, but this isn't the cause.

Windchild292:
The is should have been 'might be', my bad. Good to know though.

PR #1740, opened 2020-03-05
Adding portraits to the .gitignore

Windchild292:
I missed this one in my gitignore merge, and figured it out when I moved portrait into MegaMek to fix a MekHQ issue.


NickAragua:
I don't think we can add the portraits folder to gitignore, because of the default portrait.

Windchild292:
That's why the !/portraits/ line is there, it specifically allows that one file.

PR #1754, opened 2020-03-10
fix 'next player' edge case

NickAragua:
This fixes an undesirable behavior I noticed when clicking 'next player'. Sometimes, the next player doesn't have any units that are eligible to move, but there's no mechanism to check a subsequent player on the same team. This PR introduces that mechanism.

### DIFF
--- a/megamek/src/megamek/server/Server.java
+++ b/megamek/src/megamek/server/Server.java
@@ -3135,71 +3135,78 @@ public class Server implements Runnable {
         if (game.getOptions().booleanOption(OptionsConstants.RPG_INDIVIDUAL_INITIATIVE)) {
             return;
         }
+        
         // get the next player from the team this player is on.
         IPlayer next = game.getTeamForPlayer(current).getNextValidPlayer(current, game);
-        // if the chosen player is a valid player, we change the turn order and
-        // inform the clients.
-        if ((next != null) && (game.getEntitiesOwnedBy(next) != 0)
-                && (game.getTurnForPlayer(next.getId()) != null)) {
-
-            int currentTurnIndex = game.getTurnIndex();
-            // now look for the next occurrence of player next in the turn order
-            List<GameTurn> turns = game.getTurnVector();
-            GameTurn turn = game.getTurn();
-            // not entirely necessary. As we will also check this for the
-            // activity of the button but to be sure do it on the server too.
-            boolean isGeneralMoveTurn = !(turn instanceof GameTurn.SpecificEntityTurn)
-                    && !(turn instanceof GameTurn.UnitNumberTurn)
-                    && !(turn instanceof GameTurn.UnloadStrandedTurn);
-            if (!isGeneralMoveTurn) {
-                // if this is not a general turn the player cannot forward his turn.
-                return;
-            }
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/server/Server.java
```diff
@@ -3135,71 +3135,78 @@ private void receiveForwardIni(int connectionId) {
...
-            // update turn order
-            if (switched) {
-                game.swapTurnOrder(currentTurnIndex, nextTurnId);
-                // update the turn packages for all players.
-                send(createTurnVectorPacket());
-                send(createTurnIndexPacket(connectionId));
-            }
-            // if nothing changed return without doing anything
+            
+            next = game.getTeamForPlayer(current).getNextValidPlayer(next, game);
```
Windchild292: Change is here, and can theoretically null error if the Player doesn't have a team assigned (according to a note in getTeamForPlayer, it can return null).

megamek/src/megamek/server/Server.java
```diff
@@ -3135,71 +3135,78 @@ private void receiveForwardIni(int connectionId) {
...
-                    nextTurnId = i;
-                    if (isEntityClassTurn) {
-                        // if we had an EntityClassTurn
-                        if ((turns.get(i) instanceof GameTurn.EntityClassTurn)) {
-                            // and found another EntityClassTurn
-                            if (!(((GameTurn.EntityClassTurn) turns.get(i)).getTurnCode() == classMask)) {
-                                // both have to refer to the SAME class(es) or
-                                // they need to be rejected.
+        
+        while(!next.equals(current)) {
```
Windchild292: Change is here

Windchild292:
```suggestion
        while (!current.equals(next)) {
```

megamek/src/megamek/server/Server.java
Windchild292:
```suggestion
            next = team.getNextValidPlayer(next, game);
```

megamek/src/megamek/server/Server.java
```diff
@@ -3135,71 +3135,78 @@ private void receiveForwardIni(int connectionId) {
         if (game.getOptions().booleanOption(OptionsConstants.RPG_INDIVIDUAL_INITIATIVE)) {
             return;
         }
+        
         // get the next player from the team this player is on.
         IPlayer next = game.getTeamForPlayer(current).getNextValidPlayer(current, game);
```
Windchild292:
```suggestion
        // get the team for the next player
        Team team = game.getTeamForPlayer(current)
        // Create the player
        IPlayer next;
        //the team can theoretically be null, so we need to check and ensure it is not a null value
        if (team != null) {
            // get the next player from the team this player is on.
            next = game.getTeamForPlayer(current).getNextValidPlayer(current, game);
        } else {
            next = null;
        }
```

Windchild292: This will properly handle the potential case of a null team, which shouldn't happen but the code for getTeamForPlayer says is a possibility.

NickAragua:
Made the suggested changes.

PR #1763, opened 2020-03-14
Adding resetLogFile to MMLogger

Windchild292:
This is used in https://github.com/MegaMek/mekhq/pull/1554 , and will be implemented shortly in MegaMek and MegaMekLab.


neoancient:
This added a method to the MMLogger interface without adding an implementation to megamek.common.logging.FakeLogger in the unit tests directory.

Windchild292:
I'll fix that in the morning, before D&D

PR #1765, opened 2020-03-14
Issue1583 map editor

SJuliez:
- Removed tileManager.loadAllHexes() from a boardview event handler. This line creates long load times (10s or more, was I the only one having this?). In everything I did (start games, load games, MapEditor, change tileset) it did not seem to make a difference. AFAIK the tileset images load individually whenever necessary. Tileset changes are handled elsewhere.

- Bridge:0 is now correctly marked as invalid and the MapEditor will not write it to the map.

- The board now reports a BUILDING hex that has forced exits and an exit to an adjacent BUILDING hex of different type (e.g. building:2 and building:1) or different class (BLDG_CLASS:) as invalid. The hexes themselves can't be marked easily because a hex cannot make this validation.

- Using board validation in the MapEditor will now call board.initializeHex on all hexes which will restructure all exitable terrains that do not have forced exits wherever necessary. This autocorrects buildings, roads and bridges and the like. If the user has checked the "Exits" checkbox (forcing the exits) on a hex then that hex will not be corrected but may report as invalid.

- Saving a board in the MapEditor (Save or Save As) will now first run board validation (as above). This means it will autocorrect building and road connection issues and report errors if there are any. It will then save the board anyway but at least the user has been notified of problems.
_Note: This autocorrection may change building and road hexes. This may remove things the user wanted to keep such as mixed buildings and cut off roads. But loading a board will run InitializeHex on all hexes removing these things anyway.  Therefore, as far as I can make out, all the changes that happen now before saving are unavoidable. So better make them visible to the user right away._

- Replaced tabs with white space and changed some comments and moved definitions around. And curiously I found Building.defaultCF but still dont know where in the rules the value of 120 for a hardened building comes from.


megamek/src/megamek/client/ui/swing/boardview/BoardView1.java
```diff
@@ -5120,7 +5120,7 @@ public synchronized void boardChangedHex(BoardEvent b) {
      */
     public synchronized void boardChangedAllHexes(BoardEvent b) {
         clearHexImageCache();
-        tileManager.loadAllHexes();
+//        tileManager.loadAllHexes(); Creates long load times and seems unnecessary
```
Windchild292: Then why keep the code?

megamek/src/megamek/common/Building.java
```diff
@@ -33,13 +33,21 @@
...
 public class Building implements Serializable {
 
-    // Private attributes and helper functions.
-
-    /**
-     *
-     */
     private static final long serialVersionUID = -8236017592012683793L;
 
+    // The Building Types
```
Windchild292: UNKNOWN should probably be moved up here from line 235.

megamek/src/megamek/common/Building.java
```diff
@@ -51,16 +59,19 @@
...
      * The Basement type of the building.
      */
     private Map<Coords,BasementType> basement = new HashMap<Coords,BasementType>();
+    
     /**
-     * the class of the building
+     * The Building Class of the building; equal to the terrain elevation of the BUILDING CLASS terrain of a hex.
+     * Types: STANDARD, HANGAR, FORTRESS, GUN_EMPLACEMENT 
      */
     private int bldgClass = Building.STANDARD;
```
Windchild292: Maybe move this to be listed under the classes instead of having a comment that might go out of date?

megamek/src/megamek/common/Building.java
```diff
@@ -51,16 +59,19 @@
     private Vector<Coords> coordinates = new Vector<Coords>();
 
     /**
-     * The construction type of the building.
+     * The Construction Type of the building; equal to the terrain elevation of the BUILDING terrain of a hex.
+     * Types: LIGHT, MEDIUM, HEAVY, HARDENED, WALL
      */
     private int type = Building.UNKNOWN;
```
Windchild292: Maybe move this to be under the list of types instead of a comment?

megamek/src/megamek/client/ui/swing/BoardEditor.java
```diff
@@ -1440,12 +1440,32 @@ public String getDescription() {
...
      * "save as"; otherwise, saves the board to the specified file.
      */
     private void boardSave() {
+        // First, correct connection issues and do a validation.
+        correctExits();
+        StringBuffer errBuff = new StringBuffer();
+        board.isValid(errBuff);
+        if (errBuff.length() > 0) {
+            showBoardValidationReport(errBuff);
+        }
```
Windchild292: Why use a StringBuffer? Does this require thread safety?

megamek/src/megamek/common/Board.java
```diff
@@ -971,12 +971,40 @@ private boolean isValid(IHex[] data, int width, int height, StringBuffer errBuff
...
+                
+                // Multi-hex problems 
+                // A building hex must only have exits to other building hexes of the same Building Type and Class
+                if (hex.containsTerrain(Terrains.BUILDING) && hex.getTerrain(Terrains.BUILDING).hasExitsSpecified()) {
+                    for (int dir = 0; dir < 6; dir++) {
+                        IHex adjHex = getHexInDir(x, y, dir);
+                        if (adjHex != null 
+                                && adjHex.containsTerrain(Terrains.BUILDING) 
+                                && hex.containsTerrainExit(Terrains.BUILDING, dir)) 
+                        {
```
Windchild292:
```suggestion
                                && hex.containsTerrainExit(Terrains.BUILDING, dir)) {
```

megamek/src/megamek/common/Board.java
```diff
@@ -971,12 +971,40 @@ private boolean isValid(IHex[] data, int width, int height, StringBuffer errBuff
                 }
                 StringBuffer currBuff = new StringBuffer();
                 boolean valid = hex.isValid(currBuff);
+                
+                // Multi-hex problems 
+                // A building hex must only have exits to other building hexes of the same Building Type and Class
+                if (hex.containsTerrain(Terrains.BUILDING) && hex.getTerrain(Terrains.BUILDING).hasExitsSpecified()) {
+                    for (int dir = 0; dir < 6; dir++) {
+                        IHex adjHex = getHexInDir(x, y, dir);
+                        if (adjHex != null
```
Windchild292:
```suggestion
                        if ((adjHex != null) 
```

megamek/src/megamek/common/Board.java
```diff
@@ -971,12 +971,40 @@ private boolean isValid(IHex[] data, int width, int height, StringBuffer errBuff
...
+                                && adjHex.containsTerrain(Terrains.BUILDING) 
+                                && hex.containsTerrainExit(Terrains.BUILDING, dir)) 
+                        {
+                            if (adjHex.getTerrain(Terrains.BUILDING).getLevel() != hex.getTerrain(Terrains.BUILDING).getLevel()) {
+                                valid = false;
+                                currBuff.append("Building has an exit to a building of another Building Type (Light, Medium...).\n");
+                            }
+                            if (hex.containsTerrain(Terrains.BLDG_CLASS) 
+                                    && ((adjHex.containsTerrain(Terrains.BLDG_CLASS) 
+                                            && adjHex.getTerrain(Terrains.BLDG_CLASS).getLevel() != hex.getTerrain(Terrains.BLDG_CLASS).getLevel())
```
Windchild292:
```suggestion
                                            && (adjHex.getTerrain(Terrains.BLDG_CLASS).getLevel() != hex.getTerrain(Terrains.BLDG_CLASS).getLevel()))
```

megamek/src/megamek/common/Board.java
```diff
@@ -971,12 +971,40 @@ private boolean isValid(IHex[] data, int width, int height, StringBuffer errBuff
...
+                                            || (!adjHex.containsTerrain(Terrains.BLDG_CLASS) )  )) {
```
Windchild292:
```suggestion
                                            || (!adjHex.containsTerrain(Terrains.BLDG_CLASS)))) {
```

megamek/src/megamek/common/Board.java
```diff
@@ -971,12 +971,40 @@ private boolean isValid(IHex[] data, int width, int height, StringBuffer errBuff
...
+                                valid = false;
+                                currBuff.append("Building has an exit in direction "+dir+" to a building of another Building Class.\n");
```
Windchild292: Replace the + dir + with chained append calls.

megamek/src/megamek/common/Building.java
```diff
@@ -167,7 +178,7 @@ protected void include(Coords coords, IBoard board, int structureType) {
         }
 
         if (structureType == Terrains.BUILDING) {
-            // Error off if the building type, or CF is off.
+            // Error if the Building Type (Light, Medium...) or Building Class (Standard, Hangar...) is off.
```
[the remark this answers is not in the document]
SJuliez: Don't know much about building rules but a building with different CFs in different hexes at least loads without error. Therefore I'd like to keep it like that for now.

megamek/src/megamek/client/ui/swing/BoardEditor.java
SJuliez: I didn't introduce it and I didn't really think about that.

megamek/src/megamek/common/Board.java
SJuliez: Why? much longer. Effortlessly readable as is.

Windchild292: Faster processing

SJuliez: For the sake of not arguing about irrelevant issues I've changed it. I also followed most of your suggestions above (I think I just didnt add the parenthesis clutter in Terrain.java as this would be inconsistent :D but everything else). Thanks for looking at it.

Issue #377, opened 2016-10-21
Are You Sure You Want To Quit MegaMek?


BLOODWOLF333:
May we please have a "are you sure you want to quit MM" pop up......i just accidently closed my MM game because I thought I was in the MS paint window which I pasted a screenshot into (honest mistake).  HQ does it and is more forgiving for the players.

Qwertronix:
I would also add the option to save the current game/MUL to the prompt. Basically the usual "Would you like to save <blah> before quitting MegaMek? Yes/No/Cancel".

MGM175:
Definitely the second option would be better.

PR #1777, opened 2020-03-19
Add save game prompt when closing MM via Window "X"

SJuliez:
Resolves issue #377 

Pressing Cancel will stop MM from exiting:
image

Pressing Cancel here or not choosing a file somehow will also stop MM from exiting:
image

### DIFF
--- a/megamek/src/megamek/client/ui/swing/ClientGUI.java
+++ b/megamek/src/megamek/client/ui/swing/ClientGUI.java
@@ -65,6 +65,7 @@ import javax.swing.JPanel;
 import javax.swing.JScrollPane;
 import javax.swing.JTextPane;
 import javax.swing.ScrollPaneConstants;
+import javax.swing.WindowConstants;
 import javax.swing.filechooser.FileFilter;
 import javax.swing.filechooser.FileNameExtensionFilter;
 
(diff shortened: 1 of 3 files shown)


megamek/src/megamek/client/ui/swing/ClientGUI.java
```diff
@@ -460,12 +461,29 @@ public void initialize() {
...
+        frame.setDefaultCloseOperation(WindowConstants.DO_NOTHING_ON_CLOSE);
         frame.addWindowListener(new WindowAdapter() {
             @Override
             public void windowClosing(WindowEvent e) {
-                frame.setVisible(false);
-                saveSettings();
-                die();
+                ignoreHotKeys = true;
+                int savePrompt = JOptionPane.showConfirmDialog(null,
+                        "Do you want to save the game before quitting MekHQ?",
```
NickAragua: Maybe "before quitting MegaMek"?

megamek/src/megamek/client/ui/swing/ClientGUI.java
```diff
@@ -1641,6 +1641,30 @@ public void deleteAllUnits(Client c) {
         }
         c.sendDeleteEntities(ids);
     }
+    
+    private boolean saveGame() {
```
NickAragua: I'm surprised we don't already have a save game dialog.

SJuliez: Couldn't find a class for it. The menu bar command also routes to ClientGUI. I just moved it to its own method because I needed it twice.

megamek/src/megamek/client/ui/swing/ClientGUI.java
SJuliez: Ah, now everyone knows I just copied it from MekHQ... ok changed it and changed the icon to a warning icon.

PR #1781, opened 2020-03-20
Random Map theme is kept for editing

### LINKED ISSUE #1767: 0.47.5-SNAPSHOT - Board Editor: Theme isn't kept from the board generation window
Raised on discord, the theme is not set to the theme used when generating the initial board.

SJuliez:
After generating a random map with a theme that theme is also set for editing hexes.

Resolves #1767 (as I understand it).


Windchild292:
Just tested, this is exactly what I meant by the story

Issue #1789, opened 2020-03-22
RFE: Random Map Settings: Theme should be a dropdown of potential Themes


Windchild292:
Another fix for themes, it shouldn't require them to be memorized but instead be a dropdown list as seen in the map editor hex settings.

PR #1790, opened 2020-03-22
Various small changes around MapEditor/RndMapDialog

SJuliez:
- Corrections for the RandomMapDialog behaviour when calling it from the ChatLounge (Generated Map) and a correction for the Saxarba tileset as it was not showing buildings correctly.

- The RandomMapDialog will now show a dropdown of Themes, resolving #1789 

- The MapEditor Open Board dialog will now remember the directory

- The MapEditor has a new button that will open a map's source file in the system's standard text editor if one is configured for .board files (I can't test that on anything but Win10)

### DIFF
--- a/megamek/src/megamek/client/ui/swing/ClientGUI.java
+++ b/megamek/src/megamek/client/ui/swing/ClientGUI.java
@@ -324,10 +324,6 @@ public class ClientGUI extends JPanel implements WindowListener, BoardViewListen
         return bv;
     }
     
-    public BoardView1 getBoardView1() {
-        return bv;
-    }
-
     /**
      * Try to load the "bing" sound clip.
      */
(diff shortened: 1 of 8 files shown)


megamek/src/megamek/client/ui/swing/ClientGUI.java
```diff
@@ -323,6 +323,10 @@ public ClientGUI(Client client, MegaMekController c) {
     public IBoardView getBoardView() {
         return bv;
     }
+    
+    public BoardView1 getBoardView1() {
```
NickAragua: Probably a better approach is to add "getTileSetManager()" to the interface. 

While I think that using an interface here is an unfortunate choice (we are unlikely to swap out to different BoardView implementations), having these two methods side by side is inconsistent and will cause confusion for future developers (which one do I use?)

SJuliez: OK done. (I'd honestly like to remove all that ITileManager and IBoard stuff. It gets in the way all the time)

neoancient: While it's unlikely that most of them will ever have alternate implementations, my plans for multiple-map engagements require IBoard as a common interface between single and compound maps.

PR #1795, opened 2020-03-25
Corrections for obvious board errors in unofficial boards

SJuliez:
Corrects various board errors; only in the unofficial boards and only where it seemed obvious that it was not meant that way (incorrect buildings, pavement:0, validation errors, uneven sea surface in lucho's maps etc.).


HammerGS:
Out of curiosity has the boardvalidator.java been updated in megamek.utils.  I run it when I add map boards.  I suspect it's not catch the errors in this PR.

Windchild292:
Not yet

PR #1799, opened 2020-03-28
Fix unit file equipment entry modifier order

neoancient:
When mtf or blk files are parsed, any modifiers to an equipment entry (rear facing, pod-mounted, etc) are checked starting from the end of the string, stripping the modifier off after it is noted. This requires the checks to be checked in the reverse order of the way they are written. I've reordered those that were out of position.

Fixes MegaMek/megameklab#529.

### DIFF
--- a/megamek/src/megamek/common/loaders/BLKFile.java
+++ b/megamek/src/megamek/common/loaders/BLKFile.java
@@ -81,18 +81,21 @@ public class BLKFile {
         if (saEquip[0] != null) {
             for (String s : saEquip) {
                 String equipName = s.trim();
+                boolean isOmniMounted = false;
                 boolean isTurreted = false;
                 boolean isPintleTurreted = false;
-                if (equipName.toUpperCase().endsWith("(ST)")) {
-                    isTurreted = true;
-                    equipName = equipName.substring(0, equipName.length() - 4).trim();
+                if (equipName.toUpperCase().endsWith(":OMNI")) {
+                    isOmniMounted = true;
+                    equipName = equipName.substring(0, equipName.length() - 5).trim();
                 }
                 if (equipName.toUpperCase().endsWith("(PT)")) {
                     isPintleTurreted = true;
                     equipName = equipName.substring(0, equipName.length() - 4).trim();
                 }
-                boolean isOmniMounted = equipName.toUpperCase().endsWith(":OMNI");
-                equipName = equipName.replace(":OMNI", "");
+                if (equipName.toUpperCase().endsWith("(ST)")) {
+                    isTurreted = true;
+                    equipName = equipName.substring(0, equipName.length() - 4).trim();
+                }
 
                 int facing = -1;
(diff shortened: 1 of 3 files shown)


megamek/src/megamek/common/loaders/BLKFile.java
```diff
@@ -81,18 +81,21 @@ protected void loadEquipment(Entity t, String sName, int nLoc)
...
                 if (equipName.toUpperCase().endsWith("(PT)")) {
                     isPintleTurreted = true;
                     equipName = equipName.substring(0, equipName.length() - 4).trim();
                 }
-                boolean isOmniMounted = equipName.toUpperCase().endsWith(":OMNI");
-                equipName = equipName.replace(":OMNI", "");
+                if (equipName.toUpperCase().endsWith("(ST)")) {
+                    isTurreted = true;
+                    equipName = equipName.substring(0, equipName.length() - 4).trim();
+                }
```
Windchild292: Why aren't these else if statements?

megamek/src/megamek/common/loaders/BLKFile.java
```diff
@@ -81,18 +81,21 @@ protected void loadEquipment(Entity t, String sName, int nLoc)
         if (saEquip[0] != null) {
             for (String s : saEquip) {
                 String equipName = s.trim();
+                boolean isOmniMounted = false;
```
Windchild292: Is this ever used or just assigned?

megamek/src/megamek/common/loaders/BLKFile.java
neoancient: Because the whole point of this is that more than one can apply.

Windchild292: Wouldn't the endsWith prevent that?

megamek/src/megamek/common/loaders/BLKFile.java
neoancient: Did you actually look at the rest of the code?

Windchild292: I did, and saw it assigned and not used.

I did miss that it was a previously existing variable, hence the question.

megamek/src/megamek/common/loaders/BLKFile.java
neoancient: No, because the string is modified before the next check.

megamek/src/megamek/common/loaders/BLKFile.java
neoancient: It's used in a part of the code that wasn't modified for this PR. Hence my question about actually looking at the rest of the code.

Windchild292: I scan read, so I can miss stuff because of it.
I always take a look at the rest of the code when reviewing, to try to avoid dumb mistakes like this. My apologies on this one

Issue #1803, opened 2020-03-29
BV Discrepancy


AskAgain:
Respectfully while I am glad to know that the first issue was indeed the official rules change, that still doesn't explain the second one.

The second issue I came across several months ago but hadn't really thought it enough of a shift to mention, basically when the configure window on certain 'Mechs is opened and closed by selecting OK (whether or not I change anything that should have no affect on BV) the result is a small discrepancy in BV between the before and after. Was it already addressed in the Developmental builds?

![Customizing changes BV](https://user-images.githubusercontent.com/50790862/77855092-06b25f00-718a-11ea-95d3-f707109a59d0.JPG)
![Customizing changes BV2](https://user-images.githubusercontent.com/50790862/77855093-09ad4f80-718a-11ea-9c12-27f1c28bfe7d.JPG)

This seems to have continued on into 0.46.1 from 0.46.0.

![Customizing changes BV3](https://user-images.githubusercontent.com/50790862/77855095-0ca84000-718a-11ea-925f-a8f0cdf6e8b4.JPG)
![Customizing changes BV4](https://user-images.githubusercontent.com/50790862/77855103-129e2100-718a-11ea-9a7a-95f3f19f461a.JPG)

While I have nothing concrete I have reason to suspect this is only the case for Clan 'Mechs and even then it doesn't happen with every single one that I have needed to configure over the last month. I think it might be tied to a weapon system or a family of them. As the Stormcrow A I use regularly is one of the affected I suspect it might have something to do with Clan MD Pulse Lasers, Streak SRM6s, or LRM20s.

Files
If more files are needed I am willing to try to provide them.

HammerGS:
I just tested this running directly from the repo (so would be code in the next Dev release) and I can't duplicate it.

Are there changes being made in the equipment tab?

AskAgain:
In the examples above I made no changes. Just opened the configure window and when I closed them the BV had shifted slightly (32-33 BV). I have seen it on other Clan Mechs as I track the roster on WolfNet. I can try to send examples of all the ones I can find that do it compared to the ones that don't. Will likely have to be later though. 

Promising that the newest code seems to have caught it. I might start with 0.47.4 and see if if happens there and if not then drop it altogether.

AskAgain:
The following was using 0.47.4, loading the previously saved .mul and opening and closing the configure window by pressing OK (not closing the window).

[...]

PR #1805, opened 2020-03-29
Adding BV Calculation to ChatLounge Entity Right Click

Windchild292:
This add the ability to view the BV calculation in the Chat Lounge through a right click, as (roughly) requested by @HammerGS.
I'd recommend reviewing with whitespace differences off.


HammerGS:
Thanks for doing this. I think it's something we should have. It's a help to trouble shoot issues like https://github.com/MegaMek/megamek/issues/1803#issuecomment-605691380

PR #1806, opened 2020-03-29
Moving the delete command in chat lounge, and adding mnemonics for the lounge

Windchild292:
@HammerGS


HammerGS:
Noticed the new BV was ahead of config, and asked to move to to prevent player confusion, since they are used to having config second on the list.

Windchild292:
And I've added some missing mnemonics, to try to hit most of the missing shortcuts

PR #1816, opened 2020-04-03
Fix infantry destruction rate of Fuel-Air explosives

### LINKED ISSUE #1814: Fuel-Air Explosives not automatically destroying infantry correctly.
Megamek 47.5

### Description
Interstellar Ops Pg 165 and 166, Fuel-Air munitions game rules, Infantry.
"In addition to this, all infantry units must also roll 2D6 and add their distance from the
FAM’s point of impact (in hexes) from the result. If this modified roll is
9 or less for a conventional infantry unit, or is less than 7 for a battle
armor unit, the remaining troops in the unit are also destroyed."

Thus, A standard infantry unit at ground zero would need to roll a 10+ on 2D6 to not be automatically destroyed.  A Unit 1 Hex away would need to roll 9+ to not be automatically destroyed and so on.

Megamek seems to have this reversed.  It is automatically destroying units that roll high, and declaring low rolls to survive automatic destruction.

### Files
There should be a zip file containing an HTML gamelog, and the saved game.

[gamelog.zip](https://github.com/MegaMek/megamek/files/4426813/gamelog.zip)
[Fuelair.sav.gz](https://github.com/MegaMek/megamek/files/4426821/Fuelair.sav.gz)

binaryspica:
It will fix #1814. The units are destroyed if the result is under the target number, but it was reversed and the units are destroyed if the result is over than the target number

### DIFF
--- a/megamek/src/megamek/common/weapons/AreaEffectHelper.java
+++ b/megamek/src/megamek/common/weapons/AreaEffectHelper.java
@@ -289,7 +289,7 @@ public class AreaEffectHelper {
         
         int roll = Compute.d6(2);
         int result = roll + distFromCenter;
-        boolean destroyed = result < rollTarget;
+        boolean destroyed = result <= rollTarget;
         
         Report r = new Report(9987);
         r.indent(1);
--- a/megamek/src/megamek/common/weapons/AreaEffectHelper.java
+++ b/megamek/src/megamek/common/weapons/AreaEffectHelper.java
@@ -289,7 +289,7 @@ public class AreaEffectHelper {
         
         int roll = Compute.d6(2);
         int result = roll + distFromCenter;
-        boolean destroyed = result > rollTarget;
+        boolean destroyed = result < rollTarget;
         
         Report r = new Report(9987);
         r.indent(1);


megamek/src/megamek/common/weapons/AreaEffectHelper.java
```diff
@@ -289,7 +289,7 @@ public static void checkInfantryDestruction(Entity entity, int distFromCenter, E
         
         int roll = Compute.d6(2);
         int result = roll + distFromCenter;
-        boolean destroyed = result > rollTarget;
+        boolean destroyed = result < rollTarget;
```
NickAragua, line 292:
```suggestion
        boolean destroyed = result <= rollTarget;
```

binaryspica: >If the modified roll is **9 or less** for a conventional infantry unit, or is less than 7 for a battle armor unit, ..

What a confusing description. Thanks for point it out.

NickAragua:
I'm pretty sure it's literally the only place in battletech rules where you're looking for "less than"

sixlettervariables:
@binaryspica @nickaragua this broke Fire Control Tests.

https://github.com/MegaMek/megamek/runs/559144352#step:6:1326

NickAragua:
lol, how the hell did it do that, the thing that got changed wasn't even being unit tested.

binaryspica:
Sorry, but I have no idea what to do. Isn't it just change an operator? Is the pull request related to this?

NickAragua:
I'll take care of it, don't worry

PR #1823, opened 2020-04-07
Dependency update

neoancient:
Updates the following dependencies:
1. Library
    a. JAXB from 2.3.0 to 2.3.2. This deals with the warning that there has been an illegal reflective access. There was also a change in architecture following 2.3.0 so the group and module for the api module is different.
    b. XStream from 1.40.10 to 1.40.11. This deals with a vulnerability when the security framework is not initialized. XStream has not yet dealt with the changes in Java 9 that throw out a warning when reflection is used across modules. At some point in the future Java will disable that ability and MM will not run on later versions unless XStream figures out another way to serialize and deserialize java.* classes or we move to another library for saves.
2. Added a `test` dependency to the `assembleDist` gradle task so unit tests must pass before the `release` task can run.

### DIFF
--- a/megamek/build.gradle
+++ b/megamek/build.gradle
@@ -28,17 +28,15 @@ configurations {
 }
 
 dependencies {
-    implementation 'com.thoughtworks.xstream:xstream:1.4.10'
+    implementation 'com.thoughtworks.xstream:xstream:1.4.11'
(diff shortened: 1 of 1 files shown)


neoancient:
The two removed runtimeOnly dependencies are now inherited from jaxb-runtime.

PR #1826, opened 2020-04-07
SV fuel requirements with propeller mod

neoancient:
Fixed wing and airship support vehicles do not require fuel when they have a solar, fusion, or fission engine and the propeller chassis modification. I encoded the logic in FixedWingSupport#kgPerFuelPoint (returns 0.0 if fuel is not required) but forgot about adding the check when writing the validation code.

N.B. When implemented, Airship will be a child class of FixedWingSupport.

Fixes MegaMek/megameklab#536


Windchild292:
Question: what does N.B. mean?

neoancient:
N.B. stands for nota bene, which is Latin for note well. It is used to draw attention to something of particular note or, in this case, to introduce an aside.

Windchild292:
Ah, I do know the term nota bene, but never seen it shortened.

PR #1835, opened 2020-04-10
Turntimer

Graysho:
Added Basic Option to set a turn time limit for movement and fire phase in second.
If set to 0, no limit is imposed.

### DIFF
--- a/megamek/i18n/megamek/common/options/messages.properties
+++ b/megamek/i18n/megamek/common/options/messages.properties
@@ -1,7 +1,7 @@
 GameOptionsInfo.group.basic.displayableName=Basic Options
 GameOptionsInfo.option.friendly_fire.displayableName=Friendly fire
 GameOptionsInfo.option.friendly_fire.description=If checked,\nthe game will allow attacks against friendly units,\ncontrary to the rules.\nUnchecked by default.
-GameOptionsInfo.option.turn_timer.displayableName=Turn timer
+GameOptionsInfo.option.turn_timer.displayableName=Turn timer in minutes
 GameOptionsInfo.option.turn_timer.description=If value is greater 0,\n the game will end a players current action turn after the time is up.\n deactivated by 0 value by default.
 GameOptionsInfo.option.skip_ineligable_movement.displayableName=Skip ineligible during movement
 GameOptionsInfo.option.skip_ineligable_movement.description=If checked,\nthe game will skip a unit during the movement phase if it is immobile or otherwise inactive.\nUnchecked by default.
--- a/megamek/i18n/megamek/common/options/messages.properties
+++ b/megamek/i18n/megamek/common/options/messages.properties
@@ -1,6 +1,8 @@
 GameOptionsInfo.group.basic.displayableName=Basic Options
 GameOptionsInfo.option.friendly_fire.displayableName=Friendly fire
 GameOptionsInfo.option.friendly_fire.description=If checked,\nthe game will allow attacks against friendly units,\ncontrary to the rules.\nUnchecked by default.
+GameOptionsInfo.option.turn_timer.displayableName=Turn timer
+GameOptionsInfo.option.turn_timer.description=If value is greater 0,\n the game will end a players current action turn after the time is up.\n deactivated by 0 value by default.
 GameOptionsInfo.option.skip_ineligable_movement.displayableName=Skip ineligible during movement
 GameOptionsInfo.option.skip_ineligable_movement.description=If checked,\nthe game will skip a unit during the movement phase if it is immobile or otherwise inactive.\nUnchecked by default.
 GameOptionsInfo.option.skip_ineligable_firing.displayableName=Skip ineligible during firing
--- a/megamek/i18n/megamek/common/options/messages_de.properties
+++ b/megamek/i18n/megamek/common/options/messages_de.properties
@@ -1,6 +1,8 @@
 GameOptionsInfo.group.base.displayableName=Grundoptionen
 GameOptionsInfo.option.friendly_fire.displayableName=Friendly fire
 GameOptionsInfo.option.friendly_fire.description=Diese Option deaktivieren,\num bei Verwendung der Schaltfläche 'nächstes Ziel' eigene Einheiten zu überspringen.\nVoreinstellung: Deaktiviert.
+GameOptionsInfo.option.turn_timer.displayableName=Zug Zeitbegrenzung
+GameOptionsInfo.option.turn_timer.description=Ist der Wert größer als 0,\n wird das Spiel die aktuelle Aktionsphase des Spielers beenden, sobald die Zeit abgelaufen ist.\n Wert 0 standardmäßig deaktiviert.
 GameOptionsInfo.option.skip_ineligable_movement.displayableName=Bewegungsunfähige Einheiten überspringen
 GameOptionsInfo.option.skip_ineligable_movement.description=Wenn aktiviert,\nso werden bewegungsunfähige Einheiten für die Bewegungsinitiative nicht berücksichtigt.\nVoreinstellung: Deaktiviert.
 GameOptionsInfo.option.skip_ineligable_firing.displayableName=In Feuerphase unnötige Einheiten überspringen
--- a/megamek/i18n/megamek/common/options/messages_de.properties
+++ b/megamek/i18n/megamek/common/options/messages_de.properties
@@ -1,7 +1,7 @@
 GameOptionsInfo.group.base.displayableName=Grundoptionen
 GameOptionsInfo.option.friendly_fire.displayableName=Friendly fire
 GameOptionsInfo.option.friendly_fire.description=Diese Option deaktivieren,\num bei Verwendung der Schaltfläche 'nächstes Ziel' eigene Einheiten zu überspringen.\nVoreinstellung: Deaktiviert.
-GameOptionsInfo.option.turn_timer.displayableName=Zug Zeitbegrenzung
+GameOptionsInfo.option.turn_timer.displayableName=Zug Zeitbegrenzung in Minuten
 GameOptionsInfo.option.turn_timer.description=Ist der Wert größer als 0,\n wird das Spiel die aktuelle Aktionsphase des Spielers beenden, sobald die Zeit abgelaufen ist.\n Wert 0 standardmäßig deaktiviert.
 GameOptionsInfo.option.skip_ineligable_movement.displayableName=Bewegungsunfähige Einheiten überspringen
 GameOptionsInfo.option.skip_ineligable_movement.description=Wenn aktiviert,\nso werden bewegungsunfähige Einheiten für die Bewegungsinitiative nicht berücksichtigt.\nVoreinstellung: Deaktiviert.
--- a/megamek/src/megamek/client/ui/swing/util/TurnTimer.java
+++ b/megamek/src/megamek/client/ui/swing/util/TurnTimer.java
@@ -62,9 +62,14 @@ public class TurnTimer {
                 remaining.setText(minutes+":"+seconds);
                 progressBar.setValue(counter);
                 if (counter<1) {
+                    // get the NagForNoAction setting here
+                  boolean nagSet =   GUIPreferences.getInstance().getNagForNoAction();
                     // prevent the popup dialog from breaking time limit
                     GUIPreferences.getInstance().setNagForNoAction(false);
                     phaseDisplay.ready();
+                    // reset NagForNoAction to the value it was before to preserve the user experience
+                    // for use cases outside the timer
+                    GUIPreferences.getInstance().setNagForNoAction(nagSet);
                     timer.stop();
                     display.setVisible(false);
                     display.dispose();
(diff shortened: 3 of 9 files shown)


binaryspica:
Will it also works for bots?

NickAragua:
Not from looking at the code it won't.

megamek/i18n/megamek/common/options/messages.properties
```diff
@@ -1,6 +1,8 @@
 GameOptionsInfo.group.basic.displayableName=Basic Options
 GameOptionsInfo.option.friendly_fire.displayableName=Friendly fire
 GameOptionsInfo.option.friendly_fire.description=If checked,\nthe game will allow attacks against friendly units,\ncontrary to the rules.\nUnchecked by default.
+GameOptionsInfo.option.turn_timer.displayableName=Turn timer
+GameOptionsInfo.option.turn_timer.description=If value is greater 0,\n the game will end a players current action turn after the time is up.\n deactivated by 0 value by default.
```
neoancient, line 5: The description should say what the number means. From reading the rest of the code and testing I know it's the number of seconds, but there's nothing that tells this to the user.

megamek/i18n/megamek/common/options/messages_de.properties
```diff
@@ -1,6 +1,8 @@
 GameOptionsInfo.group.base.displayableName=Grundoptionen
 GameOptionsInfo.option.friendly_fire.displayableName=Friendly fire
 GameOptionsInfo.option.friendly_fire.description=Diese Option deaktivieren,\num bei Verwendung der Schaltfläche 'nächstes Ziel' eigene Einheiten zu überspringen.\nVoreinstellung: Deaktiviert.
+GameOptionsInfo.option.turn_timer.displayableName=Zug Zeitbegrenzung
+GameOptionsInfo.option.turn_timer.description=Ist der Wert größer als 0,\n wird das Spiel die aktuelle Aktionsphase des Spielers beenden, sobald die Zeit abgelaufen ist.\n Wert 0 standardmäßig deaktiviert.
```
neoancient, line 5: Wie oben

megamek/src/megamek/common/options/GameOptions.java
```diff
@@ -55,6 +55,7 @@ public GameOptions() {
     public synchronized void initialize() {
         IBasicOptionGroup base = addGroup("basic"); //$NON-NLS-1$
         addOption(base, OptionsConstants.BASE_FRIENDLY_FIRE, false); //$NON-NLS-1$
+        addOption(base, OptionsConstants.BASE_TURN_TIMER, 0); //$NON-NLS-1$
```
neoancient, line 58: I would prefer newer options go at the bottom, unless there are similar options for it to be grouped with. People get used to looking in a given option in a certain place.

megamek/src/megamek/common/util/TurnTimer.java
```diff
@@ -0,0 +1,119 @@
...
+        progressBar.setValue(timeLimit);
+        display.setVisible(true);
+
+    }
+
+
+    public static TurnTimer init(AbstractPhaseDisplay phaseDisplay, Client client){
+
+        //check if there should be a turn timer running
+        if (timerShouldStart(client)==true){
```
neoancient, line 92: Comparing a boolean expression to a boolean constant is superfluous.
```suggestion
        if (timerShouldStart(client)) {
```

megamek/src/megamek/common/util/TurnTimer.java
```diff
@@ -0,0 +1,119 @@
...
+            Option timer = (Option) client.getGame().getOptions().getOption("turn_timer");
+           TurnTimer tt = new TurnTimer(timer.intValue(), phaseDisplay);
```
neoancient, line 94: Indentation is inconsistent.

megamek/src/megamek/common/util/TurnTimer.java
```diff
@@ -0,0 +1,119 @@
...
+
+        SwingUtilities.invokeLater( new Runnable() {
+            public void run() {
+                remaining = new JLabel(timeLimit+" sec");
+                timer = new Timer(1000, listener);
+                timer.start();
+                display.setLayout(new BorderLayout());
+                display.getContentPane().add(remaining, BorderLayout.NORTH);
+                display.getContentPane().add(progressBar, BorderLayout.CENTER);
+                display.setSize(50,100);
```
neoancient, line 65: This is a bit narrow for the text in the label. Instead of setting an absolute size, try calling pack() to set the container dimensions to what is required by its child components.
```suggestion
                display.pack();
```

megamek/src/megamek/common/util/TurnTimer.java
```diff
@@ -0,0 +1,119 @@
+package megamek.common.util;
```
neoancient, line 1: Since this is only used by the client, it should go somewhere in megamek.client rather than megamek.common.

neoancient, line 1: Every file needs a copyright header.

megamek/src/megamek/common/util/TurnTimer.java
```diff
@@ -0,0 +1,119 @@
...
+    private int timeLimit ;
+
+
+    public TurnTimer(int limit, AbstractPhaseDisplay phaseDisplay) {
+
+        timeLimit = limit;
+        display = new JDialog();
+        progressBar = new JProgressBar(JProgressBar.VERTICAL, 0, limit);
+        progressBar.setValue(limit);
+        progressBar.setString(String.valueOf(timeLimit)+"s");
```
neoancient, line 35: String#valueOf is redundant in string concatenation.
```suggestion
        progressBar.setString(timeLimit + "s");
```

megamek/src/megamek/common/util/TurnTimer.java
```diff
@@ -0,0 +1,119 @@
...
+     * Checks if a turn time limit is set in options
+     * limit is only imposed on movement, firing
+     *  @todo: add timer to physical and targeting phase currently it is only in movement and fire
+     */
+    private static boolean timerShouldStart(Client client) {
+        // check if there is a timer set
+        Option timer = (Option) client.getGame().getOptions().getOption("turn_timer");
+        // if timer is set to 0 in options, it is disabled so we only create one if a limit is set in options
+        if (timer.intValue() > 0 ) {
+            Enum phase = client.getGame().getPhase();
```
neoancient, line 111: Use the specific enum class.
```suggestion
            IGame.Phase phase = client.getGame().getPhase();
```

megamek/src/megamek/common/util/TurnTimer.java
```diff
@@ -0,0 +1,119 @@
...
+
+            // turn timer should only kick in on firing, targeting, movement and physical attack phase
+            if (phase == IGame.Phase.PHASE_MOVEMENT || phase == IGame.Phase.PHASE_FIRING || phase == IGame.Phase.PHASE_PHYSICAL || phase == IGame.Phase.PHASE_TARGETING)
+              return true;
```
neoancient, line 115: This can be simplified:
```suggestion
            return phase == IGame.Phase.PHASE_MOVEMENT || phase == IGame.Phase.PHASE_FIRING || phase == IGame.Phase.PHASE_PHYSICAL || phase == IGame.Phase.PHASE_TARGETING;
```

megamek/src/megamek/common/util/TurnTimer.java
```diff
@@ -0,0 +1,119 @@
...
+            return tt;
+        }
+        return null;
+    }
+
+    /**
+     * Checks if a turn time limit is set in options
+     * limit is only imposed on movement, firing
+     *  @todo: add timer to physical and targeting phase currently it is only in movement and fire
+     */
```
neoancient, line 105: @todo is not a JavaDoc tag. This should be outside the JavaDoc block.
```suggestion
    /**
     * Checks if a turn time limit is set in options
     * limit is only imposed on movement, firing
     */
    //TODO: add timer to physical and targeting phase currently it is only in movement and fire
```

megamek/src/megamek/common/util/TurnTimer.java
```diff
@@ -0,0 +1,119 @@
...
+        progressBar.setValue(limit);
+        progressBar.setString(String.valueOf(timeLimit)+"s");
+
+        listener = new ActionListener() {
+            int counter = limit;
+            public void actionPerformed(ActionEvent ae) {
+                counter--;
+                remaining.setText(counter+" sec");
+                progressBar.setValue(counter);
+                progressBar.setString(String.valueOf(counter)+"s");
```
neoancient, line 43: Redundant call to String#valueOf in concatenation.
```suggestion
                progressBar.setString(counter + "s");
```

megamek/src/megamek/common/util/TurnTimer.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+    public void stopTimer() {
+        display.setVisible(false);
+        display.dispose();
+        timer.stop();
+    }
+    public boolean isRunning(){
+        return timer.isRunning();
+    }
+    public void resetTimer(){
```
neoancient, line 82: This method is never used. Did you intend to use it somewhere, or is it included because it might be useful in the future?

Graysho:
> This looks pretty good. Are you planning to add the timer to the targeting and physical displays as well? The "shouldTimerStart()" method mentions those, but I only see changes to firing and movement display.

There is a @todo, yes. I wanted to see how  it works and performs in gameplay, before distributing it to more phases

megamek/i18n/megamek/common/options/messages.properties
Graysho: Yes, it should better be in minutes and say so in the description. Will change it that way.

As for the Style-Guide, you are right too ...

megamek/src/megamek/common/util/TurnTimer.java
Graysho: No, if all other phases work like movement and firing did, It won't be needed. Need to dig into other game phases, if I dont need it. Then it can be removed.

megamek/src/megamek/common/options/GameOptions.java
Graysho: ok, will put new stuff to bottom in future, thanks for the advise.

megamek/src/megamek/common/util/TurnTimer.java
Graysho: moved it into client.swing.util

Graysho: done

megamek/src/megamek/client/ui/swing/util/TurnTimer.java
```diff
@@ -60,6 +62,8 @@ public void actionPerformed(ActionEvent ae) {
                 remaining.setText(minutes+":"+seconds);
                 progressBar.setValue(counter);
                 if (counter<1) {
+                    // prevent the popup dialog from breaking time limit
+                    GUIPreferences.getInstance().setNagForNoAction(false);
```
neoancient, line 66: This is a global setting that persists across sessions and should not be changed by the settings for a particular game. Even within the same game, it will cause it not to show if the user clicks done before the timer runs out, and it should show for that. The best solution I can think of is to add a boolean parameter to AbstractPhaseDisplay#ready that can be used to disable the nag popup.

Graysho: Well, or I could disable it, call ready() and enable it again, if it was enabled before
Whichever you prefer

Graysho:
resolved conflicts

megamek/src/megamek/client/ui/swing/util/TurnTimer.java
```diff
@@ -0,0 +1,136 @@
...
+        SwingUtilities.invokeLater( new Runnable() {
+            public void run() {
+                remaining = new JLabel(timeLimit+" sec");
+                timer = new Timer(1000, listener);
+                phaseDisplay.getClientgui().getMenuBar().add(display);
+                timer.start();
+                display.setLayout(new FlowLayout());
+                display.add(remaining);
+                display.add(progressBar);
+                display.setVisible(true);
```
neoancient, line 96: There's no need to rebuild the timer panel each time it's displayed. It should be initialized once in the constructor and just set to visible or invisible as needed.

neoancient: I'm sorry it's taking me so long to get back to you. I have so little time these days.

Graysho: done

HammerGS:
@Graysho Could you give a brief write up on this I can include as a document. I'm sure it will generate lots of questions on how it's used.

Graysho:
> @Graysho Could you give a brief write up on this I can include as a document. I'm sure it will generate lots of questions on how it's used.

The turn timer works as a count down for movement and fire phase, visible as a progress bar with label on the menu bar in movement and fire phase

If set to 0 in game options, it is deactivated.

Once the time, defined in options runs out, the turn of the player ends with the current state. It is as if the player had clicked the "Fire" or "Move" button.

In fire phase, if time runs out while you are still selecting weapons to fire, those selected until time runs out will fire.
In movement, if you have selected a path on the map, the unit will move the path, as selected at the moment of time end.

HammerGS:
Thanks!

HammerGS:
Hi, Getting a few questions on this one.

1) Does the timer have to be set to full minutes or can seconds be set.
2) Had a couple of people report that if they run it on Linux the timer doesn't show.

PR #1838, opened 2020-04-11
rat generator update; fix rat generator panel;

NickAragua:
I'm really disliking how the RAT generator save mechanism randomly swaps around the order of units for no reason, which clutters the diff. I may look at it in the future. 

This PR basically adds the silverfin and silverback to the RAT Generator's tables, fixes a bug where the unit type dropdown wouldn't show up in the RATGen UI, and adds ultra-light weight class to the RATGen UI for naval vessels.


neoancient:
Ultralight is a category for BattleMechs and PA(L) BattleArmor. The Silverfin is showing up as ultralight because of a bug in EntityWeightClass because Hydrofoil support vehicles are missing from the method that determines support vehicle weight classes and leaves it at zero.

NickAragua:
Ok, rolled back the ultra-light checkbox for naval units.

PR #1839, opened 2020-04-11
Include hydrofoil in naval support vehicle weight class determination

neoancient:
Hydrofoil support vehicles were missing from the method that determines support vehicle weight class. I also reworked the method to remove a lot of code repetition and fragile reliance on String literal matching. The type parameter passed to EntityWeightClass#getSupportWeightClass comes from Entity#getMovementModeAsString. I took advantage of the lookup aliases in EntityMovementMode to match to the enum value.

### DIFF
--- a/megamek/src/megamek/common/EntityWeightClass.java
+++ b/megamek/src/megamek/common/EntityWeightClass.java
@@ -177,85 +177,51 @@ public class EntityWeightClass {
      * @return        The weight class
      */
     public static int getSupportWeightClass(double tonnage, String type) {
-        int i = 0;
-
-        switch (type) {
-            case "Wheeled":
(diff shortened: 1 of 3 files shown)


megamek/src/megamek/common/EntityWeightClass.java
```diff
@@ -177,83 +177,51 @@ public static int getWeightClass(double tonnage, String type) {
...
+            case AIRSHIP:
+                weightLimits = airshipSupportVehicleWeightLimits;
+                break;
+            case STATION_KEEPING:
+                weightLimits = satelliteSupportVehicleWeightLimits;
+                break;
+            default:
+                return WEIGHT_MEDIUM_SUPPORT;
+        }
+        for (int i = WEIGHT_SMALL_SUPPORT; i < weightLimits.length; i++) {
```
Windchild292, line 218: Maybe add a space here?

PR #1843, opened 2020-04-12
Saxarba roads

SJuliez:
Making this a PR cause it's not obivous from the file lists what this does...
I added a few road options to the Saxarba tileset (which imo is by far the best). This will not change the appearance of any existing maps but I hope that some mapmaker makes use of it some day.
road:1;fluff:20 makes corners more rounded (bottom of the screenshot - the corners form a circle instead of a hexagon)
road:1;fluff:21 makes roads appear horizontal when possible.
In other tilesets these fluffed roads appear as normal roads.
image


HammerGS:
I concur with comments on Saxarba tileset.  I think it would also be amazing to find a way to expand on the use of SirMegaV's - https://forum.megamek.org/showthread.php?tid=2111 

His buildings are amazing.

Link to dropbox folder - https://www.dropbox.com/s/lrkxmrhve0mjjuc/hexes.7z?dl=0

PR #1845, opened 2020-04-12
AbstractUnitSelector: MegaMek Implementation

Windchild292:
This implements the Unit Selector as an Abstract class that is modified for each of the three projects, as part of the addition of the selector to MekHQ as requested by @HammerGS.


megamek/src/megamek/client/ui/swing/dialog/AbstractUnitSelectorDialog.java
```diff
@@ -0,0 +1,934 @@
...
+     */
+    private void setUserPreferences() {
+        GUIPreferences guiPreferences = GUIPreferences.getInstance();
+        setSize(guiPreferences.getMechSelectorSizeWidth(), guiPreferences.getMechSelectorSizeHeight());
+
+        comboUnitType.setSelectedIndex(guiPreferences.getMechSelectorUnitType());
+
+        comboWeight.setSelectedIndex(guiPreferences.getMechSelectorWeightClass());
+
+        updateTypeCombo(guiPreferences.getMechSelectorRulesLevels().replaceAll("\\[", ""));
```
sixlettervariables, line 176: Why is this replaced here and not below where the right bracket is replaced?

Windchild292: Yep, inherited code... but I'm not sure.

I've reworked it so that it just sends the guiPreferences to updateTypeCombo and if needed it gets the rules and does the replacements, as that's just a better way to handle it.

Windchild292:
Anything else for this @sixlettervariables?

PR #1862, opened 2020-04-18
Fix BA SSW weight multiplier.

neoancient:
The SSW is supposed to be lighter, not heavier (per suit, at least).

Fixes MegaMek/megameklab#578.

### DIFF
--- a/megamek/src/megamek/common/Mounted.java
+++ b/megamek/src/megamek/common/Mounted.java
@@ -572,7 +572,7 @@ public class Mounted implements Serializable, RoundUpdated, PhaseUpdated {
         if (isDWPMounted()) {
             return defaultRounding.round(retVal * 0.75, getEntity());
         } else if (isSquadSupportWeapon()) {
-            retVal *= getEntity().isClan() ? 1.4 : 1.5;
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/Mounted.java
```diff
@@ -572,7 +572,7 @@ public double getTonnage(RoundWeight defaultRounding) {
         if (isDWPMounted()) {
             return defaultRounding.round(retVal * 0.75, getEntity());
         } else if (isSquadSupportWeapon()) {
-            retVal *= getEntity().isClan() ? 1.4 : 1.5;
```
Windchild292, line 575: Should clan be the opposite side, as clan units are 5 member versus IS' 4?

neoancient: It's not based on squad size. It's based on tech base. Each SSW mount on a Clan suit is 40% of the total weapon weight and on an IS suit it's 50% of the weapon weight.

neoancient: See the rules TM, 270.

PR #1863, opened 2020-04-18
Don't update Maven repo if just docs updates

sixlettervariables:


neoancient:
We can also ignore data. Really anyting that doesn't go into the jar, which off the top of my head is src and i18n.

PR #1881, opened 2020-04-24
Fixed wing equipment eligibility

neoancient:
Fixed wing support vehicles are implemented as a subclass of ConventionalFighter, which makes sense for game play purposes, but for construction purposes they are vehicles and have more in common with tanks than aerospace units. This PR corrects the checks used for determining which equipment they can use, basing on the same checks used by other support vehicles and adding fixed wing, airship, and satellite motive types to the the movement mode filtering.

I also added the checks for MiscType and WeaponType validity to the entity validator for support vehicles and fixed a few other things I noticed while checking the rules. These include rules reference page numbers and switching the environmental sealing in the unit files from the CV version to the SV version.

The deleted Partisan Heavy Tank (C3) was a duplicated file. I noticed it when running the entity validator.


megamek/src/megamek/common/loaders/BLKDropshipFile.java
```diff
@@ -45,6 +45,21 @@ public BLKDropshipFile(BuildingBlock bb) {
         dataFile = bb;
     }
 
+    @Override
+    public int defaultVGLFacing(int location, boolean rearFacing) {
```
NickAragua, line 49: Two questions:

1) Is there a way to avoid this copy-paste across many files?
2) Will this for spheroid dropships?

neoancient: 1. The loaders have a lot of repeated code. They really should be in a class hierarchy appropriate to the unit type rather than all inheriting directly from BlkFile. But that's way beyond the scope of this.
2. Yes. Rear side arcs are implemented as rear-mounted equipment on the wings.

PR #1886, opened 2020-04-25
Round advanced aero fuel tank and pump weight to next ton.

neoancient:
Advanced aerospace units add an additional 2% of the fuel weight for fuel tanks and pump, rounded up to the nearest ton (StratOps, 2018 printing, p. 147). This is a change from older editions which only required it to be rounded to the half ton.

Fixes MegaMek/megameklab#562.

### DIFF
--- a/megamek/src/megamek/common/verifier/TestAdvancedAerospace.java
+++ b/megamek/src/megamek/common/verifier/TestAdvancedAerospace.java
@@ -22,20 +22,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
-import megamek.common.Aero;
-import megamek.common.AmmoType;
-import megamek.common.Bay;
-import megamek.common.CriticalSlot;
-import megamek.common.Entity;
-import megamek.common.EquipmentType;
-import megamek.common.ITechManager;
-import megamek.common.Jumpship;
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/verifier/TestAdvancedAerospace.java
```diff
@@ -524,8 +511,9 @@ public String printWeightSail() {
     
     @Override
     public double getWeightFuel() {
-        // Add 2% for pumps and round up to the half ton
-        return ceil(vessel.getFuelTonnage() * 1.02, Ceil.TON);
+        // Fuel tanks and pumps add an additional 2%, rounded to the nearest ton
+        double pumpWeight = RoundWeight.nextTon(vessel.getFuelTonnage() * 0.02);
+        return vessel.getFuelTonnage() + pumpWeight;
```
Windchild292, line 516: Why not just return RoundWeight.nextTon(vessel.getFuelTonnage() * 1.02) ?

neoancient: The tanks and pumps have to be rounded to the next ton, but the fuel itself can come in half ton lots.

PR #1894, opened 2020-04-27
ATAC/DTAC construction rules

neoancient:
Completes the construction rules for the Autonomous Tactical Analysis Computer (ATAC) and Direct Tactical Analysis Computer (DTAC) (see IO, pp. 145-7). Each system has its weight and crew requirements increased by the number of drones it can control. Following the pattern of the remote carrier control system and MASH extra theaters, these are implemented as a separate MiscType that expands the capabilities.

Besides adding ATAC and DTAC to the special handling in the TRO summary I also added the ability to handle plurals correctly. It could be more succinct, but I did it this way to make it easier to handle i18n if we ever get template translations.

I also looked over the similar drone (remote) carrier control system and fixed the tech advancement of the extra drone instance to match the base unit (removing extraneous declaration of default values) and changed the display name to place it near the base unit in an equipment list sorted by name.

See MegaMek/megameklab#570
Fixes MegaMek/megameklab#309


neoancient:
Made obsolete by #1902.

PR #1895, opened 2020-04-27
Remove all but primitive from list of armors for primitive jumpships

neoancient:
MML currently presents primitive and standard aerospace armor for primitive jumpships if the date is late enough for standard. This is a vestige of the original implementation based on the text of IO, which was not clear on whether the reduced point value was a function of the ship or the armor. The errata has since made it clear that the armor points per ton are reduced as part of the unit construction itself, so there is no option to mount standard and get more points. This changes the method that MML uses to get the available armors to reflect that.

### DIFF
--- a/megamek/src/megamek/common/Aero.java
+++ b/megamek/src/megamek/common/Aero.java
@@ -3852,11 +3852,7 @@ public class Aero extends Entity implements IAero, IBomber {
         }
     }
 
-    /**
-     * Is this a primitive ASF?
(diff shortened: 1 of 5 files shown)


neoancient:
The relevant text from the errata:

> All  armor mounted on a Primitive JumpShip—including the “free” armor provided by the vessel’s structural integrity value—must be multiplied by 0.66, rounding down to the nearest whole number.

PR #1902, opened 2020-04-30
Variable equipment size, part 1

neoancient:
This is a change I've been putting off for a while because of its potential to have hidden side effects, but there's some equipment that won't match the construction rules without it. It affects items that have a variable size or weight. So far we have taken two approaches:
1. Where the weight is a linear progression, we have created an EquipmentType instance for each possible value. This includes cargo for mechs, communications equipment, and ladders. It fails for BA mission equipment, which does not come in discrete sizes, making it impossible to create or modify a unit with a size other than the two we have created for canon units.
2. Where the weight is non-linear, we have created one EquipmentType instance for the base unit and an additional instance for expanded capacity. The two current examples are the drone (remote) control console and MASH equipment. Two drone control systems, ATAC and DTAC, have the base units implemented but no ability yet to expand the number of drones they can control. While this system works adequately for the most part, it is non-intuitive for designing units in MML and there is no way to determine which base unit the expanded capacity instance goes with, if there is more than one. The rules for the DCCS explicitly allow multiples on the unit, though no canon unit has them.

To rework this I broadened the use of the ammoCapacity field of Mounted, currently used for weapon bay ammo bin size, for use with any variable-sized equipment. I added new methods to access the name and cost of the item through Mounted to be able to account for the size. Many of the files that were changed were a result of replacing EquipmentType#getName with Mounted#getName where it made sense to do so. The extra drone/MASH theater instances have been removed, and I added code in the unit loaders to handle legacy entries.

This is the first phase, and deals with DCCS, MASH, ATAC, and DTAC. Some of the code is in place for handling the equipment listed under #1 above for illustration purposes, but as this change is fairly broad I wanted to limit it. This also does not show the equipment capacity in the TRO summaries. That required some more involved changes, which I worked on in a separate branch.

There will be accompanying MML and MekHQ PRs.


megamek/src/megamek/common/MiscType.java
```diff
@@ -435,13 +434,51 @@ public double getDamageDivisor() {
...
+                    || hasFlag(MiscType.F_ATAC) || hasFlag(MiscType.F_DTAC)) {
+                return String.format(" (%d %s)", (int) size, size > 1 ?
+                        Messages.getString("MiscType.drones") :
+                        Messages.getString("MiscType.drone"));
+            } else if (hasFlag(MiscType.F_MASH)) {
+                return String.format(" (%d %s)", (int) size, size > 1 ?
+                        Messages.getString("MiscType.theaters") :
+                        Messages.getString("MiscType.theater"));
+            } else if (hasFlag(MiscType.F_LADDER)) {
+                return String.format(" (%d m");
```
sixlettervariables, line 449: Missing the format argument.

PR #1906, opened 2020-05-01
Random Names Rework Part 2

Windchild292:
This is part two of my rework of the random name system.

Main changes:
1. Name data is now saved in comma-spaced csv files.
2. All duplicated names within the same historical ethnicity list have been merged.
3. Gender is now an Enum
4. Gender is now specifically defined as biological gender/sex, with display options ready for future expansion in MekHQ.
5. Default names have been moved to the RandomNameGenerator
6. Names are now generated based on the entity's owner's name unless using the RandomNameDialog (which assigns based on the selected faction, as normal).
7. Organized Generators into their own folder

This looks far worse a PR than it is because of the data file changes. Most of the files that are changed are just a few lines of changes, but gender and the generators do touch on a bunch of files.


megamek/docs/Rules and Technical Stuff/Random Name Generator.txt
```diff
@@ -1,33 +1,45 @@
...
-fld1 - the id for the ethnic group
-fld2 - the ethnic group name. Not currently read in, just for easy reference.
-fld3 - The relative frequency of this ethnic surname in the faction.
-fld4-fldn - These fields identify the relative frequency of first names from an ethnic group given the surname listed in fld1.
-
-@author Jay Lawson
\ No newline at end of file
+Random Name Generator:
+Author (updated version - April 29th, 2020): Justin "Windchild" Bowen
+Author (original version):
```
sixlettervariables, line 3: ?
```suggestion
Author (original version): Jay Lawson
```

megamek/src/megamek/client/generator/RandomGenderGenerator.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+    //endregion Constructors
+
+    //region Getters/Setters
+    public static int getPercentFemale() {
+        return percentFemale;
+    }
+
+    public static void setPercentFemale(int i) {
+        percentFemale = i;
```
sixlettervariables, line 43: Do we need sanity checks here?

megamek/docs/Rules and Technical Stuff/Random Name Generator.txt
Windchild292: How did I miss that...

megamek/src/megamek/client/generator/RandomGenderGenerator.java
Windchild292: Hmm, the only chance of this not fitting between the standard limits is that the person manually edited the save file, and even then it will generate properly based on their input value (although the options might not load because of an out of bounds error)

I'm leaning towards saying no

PR #1907, opened 2020-05-02
put bot turrets on top floor of building

NickAragua:
Pretty straightforward, and I'm surprised I never noticed it. The bot would routinely place its turrets on the ground floor of a building rather than the roof, which would lead to the player being able to fire on the containing building but the bot being unable to fire back.

### DIFF
--- a/megamek/src/megamek/client/bot/princess/Princess.java
+++ b/megamek/src/megamek/client/bot/princess/Princess.java
@@ -520,7 +520,7 @@ public class Princess extends BotClient {
      * Gun Emplacements should deploy on the rooftop of the building for maximum visibility.
      */
     private int getDeployElevation(Entity deployEntity, IHex deployHex) {
-        if(deployEntity instanceof GunEmplacement) {
+        if (deployEntity instanceof GunEmplacement) {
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/client/bot/princess/Princess.java
```diff
@@ -505,7 +505,7 @@ protected void calculateDeployment() {
             final IHex deployHex = game.getBoard().getHex(deployCoords);
 
             // Entity.elevatoinOccupied performs a null check on IHex
```
sixlettervariables, line 507: Should move this comment into the method below.

megamek/src/megamek/client/bot/princess/Princess.java
```diff
@@ -514,6 +514,18 @@ protected void calculateDeployment() {
             methodEnd(getClass(), METHOD_NAME);
         }
     }
+    
+    /**
+     * Calculate the deployment elevation for the given entity.
+     * Gun Emplacements should deploy on the rooftop of the building for maximum visibility.
+     */
+    private int getDeployElevation(Entity deployEntity, IHex deployHex) {
+        if(deployEntity instanceof GunEmplacement) {
```
Windchild292, line 523:
```suggestion
        if (deployEntity instanceof GunEmplacement) {
```

PR #1920, opened 2020-05-06
Don't loop twice when applying area damage to BA

### LINKED ISSUE #1915: Battle Armor taking Damage multiples from artillery attacks.
MM 47.6

A Point of Battle armor takes 4 points of Damage from Artillery (BA Tube, Radius 1 damage to be precise.).
The 4 points of damage was applied 5 times in a row, dropping the unit from 18+1 hit points per suit to zero from this single hit.

Its not just BA tube artillery.
Same thing happened when a Thumper Thermobaric round landed 1 hex away also.

neoancient:
When calling Server#damageEntity with areaSatArty == true, the method is called again for each suit location with areaSatArty false. Each suit is receiving the full damage value once per suit in the squad.

Fixes #1915


neoancient:
I didn't look close enough at the if condition and thought it was checking for loc > 0, so I just changed the starting index and removed two lines of code. Except that's not what it was.

PR #1921, opened 2020-05-06
Addition of a new AmmoType equals method that only compares AmmoType …

ChaoticInsanity:
…and not rack size.

This fixes MekHQ issue 1665.

This is the MegaMek side fix for MekHQ PR 1730.

### DIFF
--- a/megamek/src/megamek/common/AmmoType.java
+++ b/megamek/src/megamek/common/AmmoType.java
@@ -397,22 +397,9 @@ public class AmmoType extends EquipmentType {
      */
     @Override
     public boolean equals(Object other) {
-        if (!(other instanceof AmmoType)) {
-            return false;
-        }
-        // there a couple of flags that need to be checked
-        if (getAmmoType() == T_MML) {
-            if (!matchingMMLFlags((AmmoType) other)) {
-                return false;
-            }
-        }
-        if (getAmmoType() == T_AR10) {
-            if (!matchingAR10Flags((AmmoType) other)) {
-                return false;
-            }
-        }
-        return ((getAmmoType() == ((AmmoType) other).getAmmoType()) && (getRackSize() == ((AmmoType) other)
-                .getRackSize()));
+
+        return equalsAmmoTypeOnly(other) && (getRackSize() == ((AmmoType) other)
+                .getRackSize());
     }
 
     /**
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/AmmoType.java
```diff
@@ -402,30 +402,43 @@ public boolean equals(Object other) {
...
+        }
+        if (getAmmoType() == T_AR10) {
+            if (!matchingAR10Flags((AmmoType) other)) {
                 return false;
             }
         }
-        return ((getAmmoType() == ((AmmoType) other).getAmmoType()) && (getRackSize() == ((AmmoType) other)
-                .getRackSize()));
+        return getAmmoType() == ((AmmoType) other).getAmmoType();
     }
```
Windchild292, line 442: I'm almost certain this will cause issues relating to being able to reuse AC2 ammo for AC20, etc.

ChaoticInsanity: Actually you might be right I forgot to test that use case. Let me setup a new test campaign. My debug sessions were following the code path of missile racks and not ACs. The Acs might eventually reach the same predicate and use the new equals method which would cause a different bug.

ChaoticInsanity: You were absolutely correct about the AC ammo. I have made a change in MekHQ so the predicate is based on the AMMO_BY_TYPE set. With this adjustment AC ammo is being compared correctly.

neoancient:
This is only used in MekHQ. Why complicate AmmoType in MM?

ChaoticInsanity:
MekHQ relies on AmmoType from MM to compare ammunition. MM will only ever use the original Equals method, while MHQ in the case of the 'Use Ammo by Type' campaign option will use the new method when it makes sense. This new ammo type comparison should stay in the AmmoType class, hence the change to MM.

neoancient:
All the new logic should go in AmmoBin where it's actually used.

Windchild292:
I disagree. This would be best implemented in this format, with the code for equals being merely:
```
return equalsAmmoTypeOnly(other) && (getRackSize() == ((AmmoType) other).getRackSize()));
```
This would keep the code in the same place while enabling MekHQ to use the logic it needs, and avoids one variant being out of date with the other.

neoancient:
If you have code that is only ever going to be needed in one place, it belongs in the module where it is going to be used. Not in some other module in a different repository.

megamek/src/megamek/common/AmmoType.java
```diff
@@ -397,6 +397,20 @@ public AmmoType() {
      */
     @Override
     public boolean equals(Object other) {
+
```
Windchild292, line 400:
```suggestion
```

ChaoticInsanity:
Windchild, I really like your solution and it is the extra step in the refactoring that my brain has a tendency to skip. With the change you suggested the helper methods are no longer needed and equalsAmmoTypeOnly becomes the original equals without the rack size check and equals becomes the exactly what you suggested.

This is the more elegant fix, doesn't duplicate any code and resolves the bug. I have already committed this change.

Windchild292:
I strongly disagree when the code is a direct copy of code from another repository with no changes, as that would best be done by breaking down the code into usable sub methods instead. However, in most cases your point is the best practice neo.

I'm reviewing the changes now.

Windchild292:
Waiting for the CI to run, then I'll merge the MegaMek side of this PR.

Issue #1930, opened 2020-05-09
Sandblaster SPA requires pilot also have Weapon specialist SPA


SBBurzmali:
### Environment
MegaMek 0.47.6

### Description
In MegaMek 0.47.6, the Sandblaster special piloting ability (SPA) requires that the Weapon specialist SPA also be set in order for the pilot to gain the benefit from Sandblaster. Per the rules, CamOps 79 and CamOps 82, these two SPAs are independent whereas it is not currently possible for a pilot in MegaMek to have the clustering bonus from Sandblaster without also having the to hit bonus from Weapon specialist.

### Files
I've attached a save game with 4 archers facing off with an atlas, archer 1 has both SPAs, archer 2 has the Sandblaster SPA, archer 3 had the Weapon specialist (LRM 20) SPA, and archer 4 has no SPAs. By firing on the atlas with each archer, you can see that only 1 is gaining the benefit of the Sandblaster SPA, while 1 and 3 are gaining the -2 to hit from Weapon specialist. I've also attached a gamelog if required.
[SandblasterBug.sav.gz](https://github.com/MegaMek/megamek/files/4604026/SandblasterBug.sav.gz)
[gamelog.tar.gz](https://github.com/MegaMek/megamek/files/4604033/gamelog.tar.gz)

SBBurzmali:
The issue appears to be in that GUNNERY_WEAPON_SPECIALIST is used in both WeaponAttackAction and WeaponHandle, meaning that the bonuses are always paired. The solution would seem to be to create a new entry in PilotOptions for a Sandblaster specialization, duplicate the setup in CustomMechDialog for the new option and use that new option in place of GUNNERY_WEAPON_SPECIALIST in WeaponHandle.

SBBurzmali:
Here's a patch that corrects the issue. I'm not sure what the best way to submit it is.
[Fix-for-1930-Changing-Sandblaster-SPA-from-a-boolean.tar.gz](https://github.com/MegaMek/megamek/files/4604854/Fix-for-1930-Changing-Sandblaster-SPA-from-a-boolean.tar.gz)

HammerGS:
Can you please review this and submit as a PR. It will create a PR that we can better review and merge.  Thanks.

https://github.com/MegaMek/megamek/wiki/Creating-a-Pull-Request%28PR%29-if-an-outside-contributor

SBBurzmali:
All set. PR created.

sixlettervariables:
[...]

PR #1933, opened 2020-05-10
Fix for issue #1930 - Changing Sandblaster SPA from boolean to list

SBBurzmali:
Fix in Megamek for Sandblaster SPA issue #1930. Issue does not look to be present in MekHQ, and should work the same as the Weapon Specialist SPA when a MegaMek game is launched from MekHQ, I didn't see any special code for that, but it may be present. I'm not exactly sure how save games are formatted, but this change would potentially mean a game saved with a pilot with a Sandblaster SPA set to True would try to load with MegaMek expecting a value from a list.


sixlettervariables:
We don't guarantee save games work between versions of MM, so no worries there.

Issue #1942, opened 2020-05-12
Princess seems to lock up or grab memory


bczretribution:
win 10
47.6 (through MHQ)

running a large space battle, about half way through after casualties have been taken.  windows reported memory usage spikes to the maximum allocated and the game locks up and droes not progress a single units turn for an hour or more.  eventually it completes all actions over ~5-10 minutes time, and interface is jittery and laggy.  process starts over at the start of each new turn.

attached below is the JvisualVM snapshots of 2 instances of memory before the lockup (1&2) and 1 during the lockup (3)

issue might or might not be related to the memory, that is just a current best guess.  issue might be related to large battle size, or space battles

[snapshot.zip](https://github.com/MegaMek/megamek/files/4613000/snapshot.zip)

sixlettervariables:
Looks like we're hanging onto MoveSteps perhaps longer than we need to? Interesting that the top three bytes are:
1. Object[] (5Gb)
2. Integer (1Gb)
3. MoveStep (0.5Gb)

NickAragua:
Out of curiosity, are you using newtonian physics or basic spaceflight rules?

bczretribution:
basic spaceflight

bczretribution:
here is another snapshot.  left for work with it "thinking" came back to it completely locked up.  restarted from autosave and this is from only a few minutes after loading.
[snapshot new.zip](https://github.com/MegaMek/megamek/files/4615319/snapshot.new.zip)

sixlettervariables:
One improvement would be to not automatically instantiate `launched`. I can't tell if that's used very often, but I don't think it is.

NickAragua:
I guess the other useful thing would be a save game. My computer's pretty beefy so I never get into "takes several minutes" situations (maybe 20-30 seconds at most on really big maps with super fast units).

bczretribution:
i have an i7-8700K, and have currently allocated 12gb of 64gb of available ram.  this is not a question of available resources
[autosave.sav.zip](https://github.com/MegaMek/megamek/files/4625611/autosave.sav.zip)

NickAragua:
Thanks for the save, it always helps to have a "working example".

NickAragua:
lol, looks like the problem is with all the ejected pilots, it generates endlessly long paths for them for some reason.

bczretribution:
that explains why it doesn't become an issue until several turns in

NickAragua:
Fixed it for next release. 

[...]

PR #1947, opened 2020-05-14
minor performance improvement

NickAragua:
Very minor performance/memory usage improvement - avoid initializing infrequently-used fields in MoveStep, initializing them only on access. Also remove usage of deprecated constructor, avoiding passing around an extra int.

Normally this would be considered "squeezing blood from a stone", but MoveStep gets initialized quite a bit during bot pathfinding operations: imagine how many possible paths and steps a 12/18 aerospace fighter goes through.

Update: now with actual fix for #1942: apparently, ejected pilots were being evaluated for movement and coming up with endless paths. Implemented blanket fix that prevents path generation for units that have moved already.

### DIFF
--- a/megamek/src/megamek/common/MoveStep.java
+++ b/megamek/src/megamek/common/MoveStep.java
@@ -143,7 +143,7 @@ public class MoveStep implements Serializable {
     // for optional vector movement
     private int[] mv;
     private int recoveryUnit = -1;
-    TreeMap<Integer, Vector<Integer>> launched = new TreeMap<Integer, Vector<Integer>>();
+    TreeMap<Integer, Vector<Integer>> launched;
     private boolean isEvading = false;
     private boolean isShuttingDown = false;
     private boolean isStartingUp = false;
@@ -171,7 +171,7 @@ public class MoveStep implements Serializable {
      * A collection of buildings that are crushed during this move step. This is
      * used for landed Aerodyne Dropships and Mobile Structures.
      */
-    private ArrayList<Coords> crushedBuildingLocs = new ArrayList<Coords>();
+    private ArrayList<Coords> crushedBuildingLocs;
 
     /**
      * Global logging instance.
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/common/MoveStep.java
```diff
@@ -3962,6 +3964,10 @@ public boolean useSpheroidAtmosphere(IGame game, Entity en) {
      * @return An {@link ArrayList} of {@link Coords} containing buildings within a dropship's landing zone.
      */
     public ArrayList<Coords> getCrushedBuildingLocs() {
```
sixlettervariables, line 3966: Does the caller add to it? Or could the caller just receive say Collections.emptyList instead?

megamek/src/megamek/common/MoveStep.java
```diff
@@ -475,6 +473,10 @@ public Coords getTargetPosition() {
     }
 
     public TreeMap<Integer, Vector<Integer>> getLaunched() {
```
sixlettervariables, line 475: Does the caller add to this? Or could we return an empty map to them via Collections.emptyMap?

megamek/src/megamek/common/MoveStep.java
NickAragua: No, but it looks like I introduced a few bugs with that optimization.

NickAragua: Found a place where this collection was being added to without being initalized.

megamek/src/megamek/common/MoveStep.java
NickAragua: I think we better initialize it anyway, out of an abundance of caution. That function is invoked very infrequently, and in places where the minute performance hit of checking if the underlying var has been initialized doesn't matter (user interactions, movement processing on server).

Issue #1898, opened 2020-04-28
Boards of sizes xx:yyy don't load


SJuliez:
Tested with a few versions from current development to MM 0.45:

Boards with a two-digit x size and a three digit y size don't load. The issue is I think with board.indexFor which doesn't work for that situation. There is currently only one such board, SimonLandmine's Valley sample board (30x150).

The whole board loading seems a bit arcane, probably for historical reasons. The coords are unfortunately written in the board file without a separator or enough 000000 to separate them safely. The indexFor function tries to find the real coordinates from the coord string. The board saves the hex in an 1D-array based on those coordinates and retrieves hexes from this but assigns that hex Coords based on count variables (which could be different from the coord read from the file if the file has been altered).

SimonLandmine:
Oops. Sorry!

PR #1970, opened 2020-05-21
Change tileset additional changes

SJuliez:
Originally a fix for the hex shifting error but this contains a few additional changes concerning events and moves the "change tileset" function to the Board.


SJuliez:
Not sure what to make of the board_validator....

HammerGS:
Arlith added the Board Validator a few years ago, to prevent crashing and find and fix some of the handmade boards. I use it a lot when adding maps from mapmakers I don't now.

SJuliez:
Well, I didn't change any of the boards...
Anyway, why does it fail now? Instead of yesterday...

Windchild292:
It was just added yesterday

sixlettervariables:
The new board validation check runs whenever a change is made which could impact loading boards, including code which reads boards.

megamek/src/megamek/client/ui/swing/TilesetManager.java
```diff
@@ -126,12 +128,18 @@
...
+    
+    // The TileManager is associated with a game passed down from the Boardview
+    // and must pass this on to the hextileset for it to register listeners
+    private IGame game;
 
     /**
      * Creates new TilesetManager
      */
-    public TilesetManager(BoardView1 bv) throws IOException {
+    public TilesetManager(BoardView1 bv, IGame g) throws IOException {
```
NickAragua, line 139: I'd like to point out that BoardView already has a game object, so there's not a need to pass it around.

SJuliez: OK, I have another solution. If there is some obviously better one, please do point me to it. I'm open towards at least a megamek-internal philosophy on that. I made boardview's game public and final and removed the pass to the Tilesetmanager.

SJuliez: Is there something I should do because of the board validator?

NickAragua: Yeah, if you click on the details for the Validate Boards check, you can look at which boards fail as a result of the changes made and address those.

SJuliez:
@NickAragua OK, I corrected the boards. The one board reporting as not working is ok but doesnt load because of #1898

PR #1984, opened 2020-06-01
MekHQ 1529: Reimplementing negative armour check

Windchild292:
This fixes MegaMek/MekHQ#1529


sixlettervariables:
Did the commit which commented it out explain why?

Windchild292:
It did not

sixlettervariables:
How did this affect runs of the unit verifier? I haven't quite gotten to adding that to the CI system.

Windchild292:
It will cause failures on illegal units that have been released would be my bet... although we probably want to fail on those?

PR #1986, opened 2020-06-02
Lobby: Allow saving quirks override for units reduced to no quirks

### LINKED ISSUE #1115: Unable to save quirks in lobby if all quirks removed.
Pretty simple, the interface to save quirk changes to the override file disappears if you take a unit with quirks and wish to save it with no quirks. 

eg. if a madcat has both weak head armor, and imp targeting med removed, these changes cannot be saved without adding a new quirk of some kind.

SJuliez:
Resolves #1115 

The code explicitly checked if the selected unit(s) had at least one quirk and otherwise did not show the menu options to save the quirk override. The QuirksHandler could load a unit with zero quirks override (using <quirk>none</quirk> as in the examples) but it had no code to save this. It somehow feels intentional but I fail to see what it was good for.

### DIFF
--- a/megamek/src/megamek/common/QuirksHandler.java
+++ b/megamek/src/megamek/common/QuirksHandler.java
@@ -452,35 +452,36 @@ public class QuirksHandler {
                     output.write("\t\t" + getOpenTag(QUIRK));
                     output.write(NO_QUIRKS);
                     output.write(getCloseTag(QUIRK) + "\n");
-                }
-                for (QuirkEntry quirk : quirks) {
-                    // Write Weapon Quirk
-                    if (quirk.isWeaponQuirk()) {
-                        output.write("\t\t" + getOpenTag(WEAPON_QUIRK) + "\n");
-                        // Quirk Name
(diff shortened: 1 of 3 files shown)


megamek/src/megamek/common/QuirksHandler.java
```diff
@@ -446,6 +448,11 @@ public static void saveCustomQuirksList() throws IOException {
 
                 // Write out quirks
                 List<QuirkEntry> quirks = customQuirkMap.get(unitId);
+                if (quirks.isEmpty()) {
+                    output.write("\t\t" + getOpenTag(QUIRK));
+                    output.write(NO_QUIRKS);
+                    output.write(getCloseTag(QUIRK) + "\n");
+                }
                 for (QuirkEntry quirk : quirks) {
```
Windchild292, line 456: I know this is a bit superfluous, but would this be best placed in an else statement for future readability?

SJuliez: Sounds right, yes

Issue #1078, opened 2018-08-04
Advanced Munitions Available In Standard Tech


BLOODWOLF333:
### Environment
Megamek 0.45.0
Java 1.8
Windows 8.1

### Description
The ammo variant list is now displaying what looks to be IS and Clan ammo variants when tech level is set to standard, variable tech is off, and ignore clan ammo limitations is not checked.  Related to #1017 .  And streaks are showing duplicate ammo variants.

Duplicate streak ammo
![github](https://user-images.githubusercontent.com/16950832/43681220-6e533622-9813-11e8-95c8-796c47427d07.png)

Looks like IS and Clan ammo variants.  Some variants are advanced tech level (heat-seeking)
![github2](https://user-images.githubusercontent.com/16950832/43681221-71d7562a-9813-11e8-839c-1b2a1c3b8145.png)

![image](https://user-images.githubusercontent.com/16950832/43681241-bbcf3af4-9813-11e8-8e0f-0a85f64bbbe6.png)
### Files
Clan Jade Falcon mul
[CLJF.zip](https://github.com/MegaMek/megamek/files/2259884/CLJF.zip)
Standard tech level game options
[Standard.zip](https://github.com/MegaMek/megamek/files/2259885/Standard.zip)

KyleM5:
Looks like this bug no longer is the case. Can be closed.

KyleM5:
This bug no longer exists. Can be closed.

PR #1991, opened 2020-06-04
Ammo corrections

SJuliez:
Adds and corrects a few techlevels in the Ammos. 

I was looking at #1078. This doesn't really resolve the issue because when "variable tech level" is off, all ammo within the Tech Level is available regardless of year and clan/IS and it seems intentional. So the question is rather if this is how it should be or if even without the "variable tech" setting IS weapons should be restricted to IS ammo. Comments welcome.

### DIFF
--- a/megamek/src/megamek/common/AmmoType.java
+++ b/megamek/src/megamek/common/AmmoType.java
@@ -1360,7 +1360,7 @@ public class AmmoType extends EquipmentType {
                 .setISApproximate(true,false,false,false,false)
                 .setPrototypeFactions(F_TH)
                 .setProductionFactions(F_TH)
-                .setStaticTechLevel(SimpleTechLevel.STANDARD),"373,TO"));
+                .setStaticTechLevel(SimpleTechLevel.ADVANCED),"373,TO"));
         
         munitions.add(new MunitionMutator("Anti-personnel", 1, M_ANTI_PERSONNEL,
                 new TechAdvancement(TECH_BASE_IS)
@@ -1373,7 +1373,7 @@ public class AmmoType extends EquipmentType {
                 .setPrototypeFactions(F_TH)
                 .setProductionFactions(F_TH)
                 .setReintroductionFactions(F_FS,F_LC)
-                .setStaticTechLevel(SimpleTechLevel.STANDARD),"373,TO"));
+                .setStaticTechLevel(SimpleTechLevel.ADVANCED),"373,TO"));
         
         //Armor Piercing is the base ammo type see further down.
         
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/AmmoType.java
```diff
@@ -1373,7 +1373,7 @@ public static void initializeTypes() {
                 .setPrototypeFactions(F_TH)
                 .setProductionFactions(F_TH)
                 .setReintroductionFactions(F_FS,F_LC)
-                .setStaticTechLevel(SimpleTechLevel.STANDARD),"373,TO"));
```
Windchild292, line 1376: How do we handle things on two pages? This is on pg 373-374 of TO

megamek/src/megamek/common/AmmoType.java
```diff
@@ -1388,7 +1388,7 @@ public static void initializeTypes() {
...
+                .setStaticTechLevel(SimpleTechLevel.ADVANCED),"373,TO"));
```
Windchild292, line 1391:
```suggestion
                .setStaticTechLevel(SimpleTechLevel.ADVANCED), "374,TO"));
```

megamek/src/megamek/common/AmmoType.java
```diff
@@ -1443,7 +1443,7 @@ public static void initializeTypes() {
                 .setClanApproximate(true,false,false,false,false)
                 .setPrototypeFactions(F_TH)
                 .setProductionFactions(F_TH)
-                .setStaticTechLevel(SimpleTechLevel.STANDARD),"373,TO"));
+                .setStaticTechLevel(SimpleTechLevel.ADVANCED),"373,TO"));
```
Windchild292, line 1446: Same question as the IS one

megamek/src/megamek/common/AmmoType.java
```diff
@@ -1457,7 +1457,7 @@ public static void initializeTypes() {
...
```
Windchild292, line 1460:
```suggestion
                .setStaticTechLevel(SimpleTechLevel.ADVANCED), "374,TO"));
```

megamek/src/megamek/common/AmmoType.java
```diff
@@ -2130,13 +2131,13 @@ public static void initializeTypes() {
...
                 .setProductionFactions(F_TH)
-                .setStaticTechLevel(SimpleTechLevel.STANDARD),"207,TM"));
+                .setStaticTechLevel(SimpleTechLevel.STANDARD),"230,TM"));
               
         munitions.add(new MunitionMutator("Dead-Fire", 2, M_DEAD_FIRE,
                 new TechAdvancement(TECH_BASE_IS).setTechRating(RATING_C)
                 .setAvailability(RATING_X, RATING_X, RATING_E, RATING_E)
                 .setISAdvancement(3052).setPrototypeFactions(F_DC)
-                .setStaticTechLevel(SimpleTechLevel.UNOFFICIAL), "131,IO"));
+                .setStaticTechLevel(SimpleTechLevel.EXPERIMENTAL), "131,IO"));
```
Windchild292, line 2140: Same here, pages 131-132

megamek/src/megamek/common/AmmoType.java
HammerGS: We’ve always just used one page and tried to be the first page.

SJuliez: As far as I can see this is never referenced, only saved. So I'd just give the first page.

SJuliez:
Corrected the Mortar page numbers.

SJuliez:
Ah, and if we don't want to change the system of having all ammos within the tech level available (regardless of Clan/IS) when "variable tech level" is off, then #1078 should be closed.

PR #1992, opened 2020-06-06
Inclines and cliffs

SJuliez:
Adding incline & sheer cliff graphics and rules according to TO p.39.

- [x] Implement hidden & automatic terrains in editor 
- [x] Add Saxarba gfx to saxarba tileset
- [x] Add Saxarba gfx to hq_saxarba tileset
- [x] Add GUI option
- [x] Don't save automatic terrains
- [x] Remove incline when cliff at the same exit
- [x] Cliff rules: no backwards movement across sheer cliffs at all
- [x] Cliff rules: Mechs up cliff: PSR, +1 for lvl1 cliff +2 for lvl2 cliff, if lost, nothing happens, MP lost
- [x] Cliff rules: Mechs down cliff: PSR, +0 for lvl1 cliff +1 for lvl2 cliff, if lost, fall, movement over
- [x] Cliff rules: Veh (exc VTOL/WIGE) up cliff: not possible; 
- [x] Cliff rules: up cliff WIGE: 1MP extra 
- [x] Cliff rules: Veh (exc VTOL/WIGE) down cliff: DSR, fail: Motive roll, move over; 
- [x] Cliff rules: down cliff VTOL/WIGE: no effect at all
- [x] Cliff rules: Inf: Forbidden exc Mountaineer
- [x] Order of PSRs when turning on pavement then walking up a cliff
- [x] Climbing actions don't seem to be implemented in MM
- [x] AO shadows only on cliffs and lvl2 incs
- [x] Shorten lvl1 shadows
- [x] show PSRs as danger in move paths 
- [X] Underwater cliffs, remove any underwater inclines for now
- [x] Test QuadVees
- [x] Test LAMs
- [x] Test Hover
- [x] Test Mechs
- [x] Test VTOL, Aero, WIGE
- [ ] Spider in Water, cliff at water's edge, fails, now has 6 MP left but all moves illegal -> incorrect-> reproducible, maybe connected to not getting up in water after hull down fail
- [x] Make inclines transparent in other tilesets

Open points:
- Should the inclines be added to other tilesets? Currently they are simply invisible in other tilesets and switching on the GUI Pref does not change anything except for a slight change of shadows.
- What do Protomechs do with cliffs?
- Roads across normal level changes do not change anything. Should roads across a cliff allow normal movement across the level change? (Making the cliff act as a normal level change)
- Hovers can slip down 1 level without effect, 2 levels results in a fall. Should a skid/slip down a 1 level cliff result in a fall? Or an automatically failed PSR with motive damage chance (more along cliff rules)?


HammerGS:
@SJuliez I'd like to post a screen shot and tease this on the forums. I don't ever but a date. Any objections?

SJuliez:
Im not sure what you mean with "I don't ever but a date" but I certainly have no objections!

HammerGS:
Typing to fast like I always go....should have be "put a date"

SJuliez:
Example map...
Can I attach the board file here somehow?

cliffmap

HammerGS:
You'd need to zip the board file.

SJuliez:
[35x17_TropicalCoast.zip](https://github.com/MegaMek/megamek/files/4746706/35x17_TropicalCoast.zip)

SJuliez:
> > * Roads across normal level changes do not change anything. Should roads across a cliff allow normal movement across the level change? (Making the cliff act as a normal level change)
> >   Wouldn't that basically be a bridge?
> No, I mean just a road on the ground level on both sides of a cliff, 1 or 2 levels. On a normal level change the road across doesnt change anything, as the units still have to ascend or descend the level change. Across a cliff it makes sense to assume that the road doesnt just end at the cliff face and continue 6m below. When there is a road across a cliff then it will presumably change level in a non-vertical fashion. It seems to make sense that a unit traveling across a cliff would experience just a normal level change. The road would effectively cancel the cliff.
> 
> > * Should there be a GUI Pref for it at all? It's not like MM has too few of those.
> >   What would it be like, how would it be used?
> In the Client Settings, Graphics panel a point saying
> [ ] Show edge highlights/shadows at elevation changes
Having it activated will show the incline gfx, otherwise things would be as before (with cliffs of course, these are always active)

SJuliez:
- [x] ProtoMechs
- [x] Incline gfx applied to all tilesets

HammerGS:
I think the GUI pref makes sense, 

and 

"Or an automatically failed PSR with motive damage chance (more along cliff rules)? Letting it just slide over would mean that sliding across the cliff is better than going down on purpose which wouldn't feel right." this was the liked option.

I agree the road would cancel the cliff.

SJuliez:
@HammerGS : After looking at server.java I think I'll have to come back to the damage ruling (sorry, should have looked first). There are rules in place for skidding and side-slipping down levels. For vehicles, TW p.62/63 and 153 say that they only take damage when they slide down more than the allowed level change, i.e. more than 1 level. Again, looking at vehicles driving off a lvl1 cliff possibly suffering a motive dmg chance, it seems to make sense to (mentally) reduce the allowed level change for vehicles across sheer cliffs to 0. This also reflects that vehicles cant go up cliffs at all and down only with dmg potential. In this way, slipping down a lvl1 sheer cliff will damage a vehicle with weight/10*(lvl+1) = 2*weight/10 (and all possible follow-up motive dmgs and so on). To make this work I have to adjust the server skidding/sideslipping processing only very little. Does that sound ok?

Meanwhile I noticed that vehicles cannot go up a lvl2 incline (not cliff) that is crossed by a ground road. I suppose no one has thought about that very much. I just noticed. Roads don't really help units much with level differences.

I'll have to send a few more units on round trips along a cliff for testing skids but otherwise I'm pretty much done here.

HammerGS:
Switched to review per request from SJuliez on slack.

megamek/src/megamek/common/Compute.java
```diff
@@ -548,6 +548,42 @@ public static boolean isPilotingSkillNeeded(IGame game, int entityId,
                 && !isPavementStep) {
             return true;
         }
+        
+        // Sheer Cliffs, TO p.39
+        // Roads over cliffs cancel the cliff effects for units that move on roads
+        boolean quadveeVehMode = entity instanceof QuadVee
```
NickAragua, line 554: Seems to me like this logic and the logic in SharedUtility.NagPSR should be unified somehow to avoid inconsistencies if/ when it needs to be updated. Keeping it in a distinct method will also allow it to be referenced by other code that may need it (e.g. the bot)

megamek/src/megamek/common/Compute.java
```diff
@@ -3160,72 +3196,72 @@ public static float getExpectedDamage(IGame g, WeaponAttackAction waa,
...
-                        	} else if (rangeToTarget > 40) {
-                			av += bayWType.getExtAV();
-                        	} else if (rangeToTarget > 25) {
-                			av += bayWType.getLongAV();
-                        	} else if (rangeToTarget > 12) {
-                			av += bayWType.getMedAV();
-                        	} else {
-                			av += bayWType.getShortAV();
-                        	}
+                            threat = 12;
```
NickAragua, line 3208: Normally I don't bother about white space formatting, but it looks like the tabs got completely clobbered here.

megamek/src/megamek/common/Hex.java
```diff
@@ -590,20 +590,23 @@ public void getUnstuckModifier(int elev, PilotingRollData rollTarget) {
             terrain.getUnstuckModifier(elev, rollTarget);
         }
     }
+    
+    public boolean hasCliffTopTowards(IHex otherHex) {
```
NickAragua, line 594: Let's put a comment on this public method.

megamek/src/megamek/common/MechSummaryCache.java
```diff
@@ -489,8 +489,8 @@ private boolean loadMechsFromDirectory(Vector<MechSummary> vMechs,
             Set<String> sKnownFiles, long lLastCheck, File fDir,
             boolean ignoreUnofficial) {
         boolean bNeedsUpdate = false;
-        loadReport.append("  Looking in ").append(fDir.getPath())
-                .append("...\n");
+//        loadReport.append("  Looking in ").append(fDir.getPath())
```
NickAragua, line 492: Should probably uncomment these guys?

megamek/src/megamek/common/MoveStep.java
```diff
@@ -3356,24 +3364,76 @@ public boolean isMovementPossible(IGame game, Coords src, int srcEl, CachedEntit
                 return false;
             }
         }
+        
+        // Sheer Cliffs, TO p.39
+        // Roads over cliffs cancel the cliff effects for units that move on roads
+        boolean vehicleAffectedByCliff = entity instanceof Tank
```
NickAragua, line 3370: It looks like a triplication of the 'is this affected by a cliff' code, so yet another argument to consolidate that logic.

megamek/src/megamek/server/Server.java
```diff
@@ -8720,6 +8721,69 @@ else if (bldgExited.equals(bldgEntered)
                     }
                 }
             }
+            
+            // Sheer Cliffs, TO p.39
+            boolean vehicleAffectedByCliff = entity instanceof Tank
```
NickAragua, line 8726: A quadruplication!

megamek/src/megamek/common/Compute.java
SJuliez: OK, autoindented it.

megamek/src/megamek/common/Hex.java
SJuliez: Did.

megamek/src/megamek/common/MechSummaryCache.java
SJuliez: Right :) Although these clutter up the log.

SJuliez:
@NickAragua I agree that in principle all those four sections where movement stuff is checked would ideally reference some single place where the logic is kept. But I don't want to do this just for the cliffs and leave the other 5000 lines of it unchanged and changing the whole thing is not for this PR. Therefore let's please keep it simple for now. I may be doing this at some point but I won't promise.

PR #1999, opened 2020-06-10
fix localload command

NickAragua:
A probable fix for the localload command not working. The Server.loadGame() function was updated to change some properties of the XStream object, but the game.sendLoadGame() command wasn't updated to match that and so was failing.

It was also complaining about Victory being non-serializable.

While running this from a gui client, I observed that the unit images don't really update properly when loading the file (an Atlas showed up as a Vedette), but I'll address that in another pull request if necessary.


NickAragua:
Renamed the class to reflect its functionality a little better while I was at it.

PR #2007, opened 2020-06-15
Variable equipment size, part 2

neoancient:
This builds on #1902 by using the size field in Mounted to consolidate cargo, liquid cargo, communications gear, ladders, and BA mission equipment to one MiscType each. This allows BA mission equipment in any arbitrary size (in 1 kg lots) and liquid cargo as a single equipment mount rather than a multiple of single-ton mounts with a possible half-ton. Because cargo and communications equipment size determine the number of critical slots on a mech, the slot count has been changed to query Mounted instead of the EquipmentType, which accounts for most of the changes. There are also some changes in MtfFile that allocate any additional critical slots needed after the size has been set, and the code that sets the size has been moved to the correct place.

### DIFF
--- a/megamek/i18n/megamek/common/messages.properties
+++ b/megamek/i18n/megamek/common/messages.properties
@@ -452,3 +452,5 @@ MiscType.drones=drones
 MiscType.drone=drone
 MiscType.theaters=theaters
 MiscType.theater=theater
+MiscType.tons=tons
+MiscType.ton=ton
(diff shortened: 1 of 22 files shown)


megamek/i18n/megamek/common/messages.properties
```diff
@@ -451,4 +451,6 @@ MovementType.StationKeeping=Station Keeping
 MiscType.drones=drones
 MiscType.drone=drone
 MiscType.theaters=theaters
-MiscType.theater=theater
\ No newline at end of file
+MiscType.theater=theater
+MiscType.tons=tons
```
NickAragua, line 455: Did we really not have a token for "ton" and "tons?

NickAragua:
That poor MiscType class, heh.

megamek/i18n/megamek/common/messages.properties
neoancient: We might. I didn'tcheck because I thought the advantages of keeping this in one place were worth the possibility of duplication.

PR #2011, opened 2020-06-17
fix several minor npes and pathfinding issues

NickAragua:
what it says on the label.

- random NPE when trying to customize infantry unit in mekhq's meklab
- stack overflow when bot attempting to move legged mech with no LOS to anything
- efficiency improvement to skip pathfinding calculation for prone units that are technically not immobile but are practically immobile

### DIFF
--- a/megamek/src/megamek/common/templates/InfantryTROView.java
+++ b/megamek/src/megamek/common/templates/InfantryTROView.java
@@ -185,7 +185,7 @@ public class InfantryTROView extends TROView {
             notes.add(String.format(Messages.getString("TROView.InfantryNote.SingleFieldGun"),
                     fieldGuns.get(0).getName(), shots, (int) fieldGuns.get(0).getTonnage(inf)));
         }
-        if (inf.getSecondaryN() > 1) {
+        if (inf.getSecondaryWeapon() != null) {
             if (inf.getSecondaryWeapon().hasFlag(WeaponType.F_INF_BURST)) {
                 notes.add(Messages.getString("TROView.InfantryNote.Burst"));
             }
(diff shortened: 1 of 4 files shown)


megamek/src/megamek/common/templates/InfantryTROView.java
```diff
@@ -185,7 +185,7 @@ private void addWeaponNotes(List<String> notes) {
             notes.add(String.format(Messages.getString("TROView.InfantryNote.SingleFieldGun"),
                     fieldGuns.get(0).getName(), shots, (int) fieldGuns.get(0).getTonnage(inf)));
         }
-        if (inf.getSecondaryN() > 1) {
+        if (inf.getSecondaryWeapon() != null) {
```
NickAragua, line 188: checks "secondaryN" but then looks at getSecondaryWeapon(), which results in an NPE when secondaryN > 1 but there's no secondary weapon.

neoancient: Do you know what unit was causing that? This fix is fine, but it suggests another problem somewhere else.

NickAragua: No idea. I was trying to customize an infantry unit with the goal of installing field guns, so (roughly):
- pull up 'customize in mek lab'
- actually click on the 'mek lab' because there's a bug with the command center where it goes to the infirmary tab instead
- swap motive type to tracked
- attempt to use the field gun drop down, fail

NickAragua: Oh yeah, it was a customized ER support laser unit with some kind of sealed environment suit.

PR #2012, opened 2020-06-17
fix one more NPE

NickAragua:
Fix a subsequent NPE on the infantry TRO view, now it doesn't barf when setting # of secondary weapons to 0 without removing the secondary weapon (no idea why you'd want to do that, but there it is).


NickAragua:
Ok, I'll have it do both the SecondaryN check and the null check, in that case.

NickAragua:
That ought to do it.

PR #2014, opened 2020-06-18
Fix spikes implementation

### LINKED ISSUE #2004: Spikes - Not working correctly
The rules for Spikes may be found in Tactical Operations, page 291.
Version 47.6

First off, Megamek currently only has the Spikes roll to break whenever the location mounting them is hit by a physical attack. However, there should be a break chance every time the location that mounts them suffers damage from any source at all.

Secondly, Megamek currently has Spikes only deal +2 damage to their target during a charge if the Spikes on a location damaged by the charge are "impacted" - legs included. TacOps states here that it should be an automatic +2 damage for every cluster of Spikes on a Torso location the attacker has.

Thirdly, Megamek does not have Spikes reduce physical damage to the attacker during a charge. Here TacOps says that 4 points of damage should be absorbed as per normal on any location that suffers damage but mounts Spikes.

Attached is an autosave that shows all three problems in effect.
[autosave.sav.gz](https://github.com/MegaMek/megamek/files/4776648/autosave.sav.gz)

neoancient:
Fixes the implementation of spikes. From the bug report that prompted this:
1. Spikes should have a chance of breaking whenever their location takes damage from any source, not just from a physical attack.
2. A mech with spikes deals +2 damage on a charge for each torso location with spikes, not just locations that take damage from the charge.
3. Spikes should reduce the damage to the attacker on a charge.
Additional issues:
4. When making a physical attack against a target with spikes, the spike damage is reduced if the attacker also has spikes in the location that made the attack.
5. A push attack can deal damage to the attacking mech's arms if the defender has spikes in the location that was pushed.
6. Physical attacks that deal damage in clusters have the reduction to each cluster that hits a protected location rather than a maximum of once per location.

See https://bg.battletech.com/forums/index.php?topic=47190.msg1088629#msg1088629 for clarification on items 4 and 6.

Fixes #2004

### DIFF
--- a/megamek/i18n/megamek/common/report-messages.properties
+++ b/megamek/i18n/megamek/common/report-messages.properties
@@ -564,8 +564,8 @@
 4330=Defender protected by spikes! (-4 damage, attacker takes 2 damage)
 4331=Defender protected by spikes! (-4 damage)
 4335=Attacker has spikes! (+<data> charge damage)
-4440=, but the spikes broke off (rolled <data>)
-4445=, the spikes remain intact (rolled <data>)
+4440=The spikes broke off (rolled <data>)
+4445=The spikes remain intact (rolled <data>)
 4450=<data> tries to trip <data>. Needs <data> rolls <data> 
 4455=Retractable blade extended:
 4456=The retractable blade is destroyed!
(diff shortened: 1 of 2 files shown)


neoancient:
Set as draft pending clarification from the rules committee on how to split the two points of spike damage among the attacker's legs when a quad or tripod mech makes a DFA attack.

HammerGS:
> Set as draft pending clarification from the rules committee on how to split the two points of spike damage among the attacker's legs when a quad or tripod mech makes a DFA attack.

Results are in, damage should be applied as 1 point each to the rear legs on a quad, and the Left and Right leg on a Tripod.

Issue #2015, opened 2020-06-19
ArrayIndexOutOfBoundsException: -1 in Configure unit


HammerGS:
When have a bunch of units in the lobby and you use the next to move between them, when you get to the bottom of the list the next instead of taking you to the top of the list. Freezes the window and you have to use the X to close it.

### Environment
What version of MegaMek does your issue apply to?  47.7 running from the Repo
What operating system are you using?  Windows,

### Description
```
Exception in thread "AWT-EventQueue-0" java.lang.ArrayIndexOutOfBoundsException: -1
	at java.util.ArrayList.elementData(ArrayList.java:422)
	at java.util.ArrayList.get(ArrayList.java:435)
	at megamek.client.ui.swing.EquipChoicePanel$WeaponAmmoChoicePanel.applyChoice(EquipChoicePanel.java:1423)
	at megamek.client.ui.swing.EquipChoicePanel.applyChoices(EquipChoicePanel.java:401)
	at megamek.client.ui.swing.CustomMechDialog.actionPerformed(CustomMechDialog.java:1266)
	at javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:2022)
        at javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2348)
```

### Files
[megameklog.txt](https://github.com/MegaMek/megamek/files/4805522/megameklog.txt)

sixlettervariables:
Closed by #2016.

PR #2016, opened 2020-06-20
Only update ammo bin if within array list bounds #2015

sixlettervariables:
Resolves the out-of-bounds condition seen in #2015, however, there may be more to it so I'm leaving this as a draft while I investigate further. The stacktrace bubbled up from an unexpected source given where @HammerGS said he was pressing 'next'.


sixlettervariables:
I couldn't reproduce the exception, but my guess is that the last unit ("last" being a red herring) in @HammerGS's case was a unit with a certain ammo loadout that did not match. I tried re-setting and re-ordering things to see if there was a specific order of operations that best caused the failure but I could not.

My guess is this "fix" will at worst allow a customization to not occur even though it visually may have.

PR #2017, opened 2020-06-21
Allow units to flee if they made it into prohibited terrain #2005

### LINKED ISSUE #2005: Mechwarrior in Water cannot flee the board.
Witnessed in a Game.  MW ejects and lands in water next to the board edge. Mechwarrior is stuck in water and can't flee board.

image

### Environment
What version of MegaMek does your issue apply to?  47.6

sixlettervariables:
This fixes #2005 by allowing any unit to flee that _could_ flee. This ignores how they may have gotten themselves into prohibited terrain.

This is one possibility for that issue. Another would be to disable the flee button for units which would be immobile (due to prohibited terrain).

### DIFF
--- a/megamek/src/megamek/common/MoveStep.java
+++ b/megamek/src/megamek/common/MoveStep.java
@@ -3166,6 +3166,12 @@ public class MoveStep implements Serializable {
             return true;
         }
 
+        // If you want to flee, and you can flee, flee.
+        if ((type == MoveStepType.FLEE)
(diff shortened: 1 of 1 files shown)


HammerGS:
I'd agree if they want to flee, they'll find a way.

sixlettervariables:
Maybe I'll add an option to disable this for folks who'd rather make their ejected pilots tread water waiting to be picked up.

Issue #2024, opened 2020-06-24
MekHQ 47.7 won't load all units.


seekerofshadowlight:
### Environment
MekHq 47.7
Windows 10 ver 10.0 (x86)
Java version 1.8.0_251

### Description
The game will not load all the units. It loads my unit, then then the others controlled by the bots are not loaded at all or only some loaded, such as the opponent, but not my help, or one opponent group but not the second group
![MM11](https://user-images.githubusercontent.com/67355237/85499942-8d6fb000-b5b0-11ea-9309-730245d1d8ef.jpg)

sometimes exiting and reloading works, or it will get some units and other times I can do this more than ten times and no dice.

### Files

[Goldern knights30620205.zip](https://github.com/MegaMek/megamek/files/4823182/Goldern.knights30620205.zip)

HammerGS:
NPE from Log
```
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
	at megamek.client.ui.swing.AccessibilityWindow$1.gamePlayerDisconnected(AccessibilityWindow.java:46)
	at megamek.common.event.GamePlayerDisconnectedEvent.fireEvent(GamePlayerDisconnectedEvent.java:39)
	at megamek.common.Game.processGameEvent(Game.java:3153)
	at megamek.client.Client.disconnected(Client.java:337)
	at megamek.client.Client$1$1.run(Client.java:203)
```

sixlettervariables:
Potentially closed by #2025.

PR #2025, opened 2020-06-25
Add null-proofing to avoid #2024

sixlettervariables:
This will avoid the NRE in the AccessibilityWindow if for whatever reason a player or entity is missing, fixing the exception seen in the logs of #2024. I was unable to recreate the exception or the bug as described, but the call stack is plausible and can be null-proofed.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/AccessibilityWindow.java
+++ b/megamek/src/megamek/client/ui/swing/AccessibilityWindow.java
@@ -38,7 +38,7 @@ public class AccessibilityWindow extends JDialog implements KeyListener {
         client.getGame().addGameListener(new GameListenerAdapter() {
             @Override
             public void gamePlayerConnected(GamePlayerConnectedEvent e) {
-                String name = e != null && e.getPlayer() != null 
+                String name = (e != null) && (e.getPlayer() != null)
                             ? e.getPlayer().getName()
                             : "UNNAMED";
                 systemEvent("New player has connected. Their name is " + name + ".");
@@ -46,7 +46,7 @@ public class AccessibilityWindow extends JDialog implements KeyListener {
 
             @Override
             public void gamePlayerDisconnected(GamePlayerDisconnectedEvent e) {
-                String name = e != null && e.getPlayer() != null 
+                String name = (e != null) && (e.getPlayer() != null)
                             ? e.getPlayer().getName()
                             : "UNNAMED";
                 systemEvent("The player " + name + " has disconnected.");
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/client/ui/swing/AccessibilityWindow.java
```diff
@@ -38,12 +38,18 @@ public AccessibilityWindow(ChatterBox cb, ClientGUI clientgui) {
         client.getGame().addGameListener(new GameListenerAdapter() {
             @Override
             public void gamePlayerConnected(GamePlayerConnectedEvent e) {
-                systemEvent("New player has connected. Their name is " + e.getPlayer().getName() + ".");
+                String name = e != null && e.getPlayer() != null
```
Windchild292, line 41: This might be easier to read if it is in brackets?

megamek/src/megamek/client/ui/swing/AccessibilityWindow.java
```diff
@@ -66,20 +75,29 @@ public void gameEntityNewOffboard(GameEntityNewOffboardEvent e) {
...
+                    String name = ent.getOwner() != null ? ent.getOwner().getName() : "UNNAMED";
+                    systemEvent("Removed " + ent.getDisplayName() + " from player " + name + ".");
+                }
             }
 
             @Override
             public void gameEntityChange(GameEntityChangeEvent e) {
-                final Entity ent = e.getEntity();
-                systemEvent(e.toString());
+                if (e != null && e.getEntity() != null) {
```
Windchild292, line 87: Brackets are missing here

megamek/src/megamek/client/ui/swing/AccessibilityWindow.java
```diff
@@ -66,20 +75,29 @@ public void gameEntityNewOffboard(GameEntityNewOffboardEvent e) {
...
+                    systemEvent(e.getEntity().toString());
+                }
             }
 
             @Override
             public void gameNewAction(GameNewActionEvent e) {
-                final Entity ent = client.getEntity(e.getAction().getEntityId());
-                systemEvent( ent.getDisplayName() + " from player " + ent.getOwner().getName() + " is doing " + e.getAction().toString() + ".");
+                if (e != null && e.getAction() != null) {
```
Windchild292, line 94: Missing brackets here

megamek/src/megamek/client/ui/swing/AccessibilityWindow.java
sixlettervariables: I don't usually use brackets in simple cases.

Issue #2020, opened 2020-06-22
RFE/Bug - Delayed deployment currently not available for off-board artillery


SimonLandmine:
At present in MegaMek, if you designate a unit (artillery) to deploy off-board, it is available from the beginning of the game, regardless of whether you've set a delayed deployment for it, although there is nothing to indicate this.

It is therefore impossible to set off-board artillery to only become available in the "Before Turn X" deployment (which can be useful for some escalating scenarios).

Would it be possible to either ...
a) Indicate that delayed deployment isn't available for units deployed off-board, or
b) Apply the deployment delay to off-board units as well as on-board units.

sldfgunslinger2766:
I can confirm this behavior in Dev 0.47.6. The bot controlled Quad A 4 turret deploys off board as advertised, but is available to fire on Turn 1 even after multiple attempts to set and save the deployment round for round 7.

sldfgunslinger2766:
![deployment](https://user-images.githubusercontent.com/40328392/85343121-4fd42f80-b4a9-11ea-85ef-1c7912368e45.PNG)

sldfgunslinger2766:
![deployment II](https://user-images.githubusercontent.com/40328392/85343237-9590f800-b4a9-11ea-9ef5-9dcbff1263fa.PNG)

sixlettervariables:
From `Entity.java`:
```java
    /**
     * Deploy this offboard entity at the previously specified distance and
     * direction. This should only be invoked by the <code>Server</code> after
     * the board has been selected and all the players are ready to start. The
     * side effects of this methods set the unit's position and facing as
     * appropriate (as well as deploying the unit).
     * <p/>
     * Onboard units (units with an offboard distance of zero and a direction of
     * <code>Entity.NONE</code>) will be unaffected by this method.
     */
    public void deployOffBoard() {
        if (null == game) {
            throw new IllegalStateException(
                    "game not set; possible serialization error");
        }
        // N.B. 17 / 2 = 8, but the middle of 1..17 is 9, so we
        // add a bit (because 17 % 2 == 1 and 16 % 2 == 0).
        switch (offBoardDirection) {
            case NONE:
                break;
            case NORTH:
                setPosition(new Coords((game.getBoard().getWidth() / 2)
                        + (game.getBoard().getWidth() % 2),
                        -getOffBoardDistance()));
                setFacing(3);
                setDeployed(true);
                break;
            case SOUTH:
[...]

PR #2027, opened 2020-06-25
Respect deployment round for offboard entities #2020

sixlettervariables:
This implements #2020 by respecting the deployment round for offboard entities. I added a `shouldDeployOffboard` method to compliment `shouldDeploy` but covering only offboard entities (to avoid scattering isOffboard checks and remembering to use them in the future).

### DIFF
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -9344,7 +9344,9 @@ public abstract class Entity extends TurnOrdered implements Transporter,
      * Returns true if the entity should be deployed
      */
     public boolean shouldDeploy(int round) {
-        return (!deployed && (getDeployRound() <= round) && !isOffBoard());
+        return !isDeployed() 
+            && (getDeployRound() <= round)
+            && !isOffBoard();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


megamek/src/megamek/server/Server.java
```diff
@@ -2366,6 +2376,8 @@ public boolean accept(Entity entity) {
             case PHASE_PHYSICAL:
             case PHASE_TARGETING:
             case PHASE_OFFBOARD:
+                deployOffBoardEntities();
```
NickAragua, line 2379: Is this going to loop through every entity in every phase?

megamek/src/megamek/common/Entity.java
```diff
@@ -9347,6 +9347,16 @@ public boolean shouldDeploy(int round) {
...
     }
 
+    /**
+     * Returns true if the offboard entity should be deployed this round.
+     * @param round The current round number.
+     * @return True if and only if the offboard entity should deploy this
+     *         round, otherwise false.
+     */
+    public boolean shouldOffboardDeploy(int round) {
+        return (isOffBoard() && (!deployed && (getDeployRound() <= round)));
```
NickAragua, line 9357: A little too many parens here.

sixlettervariables: I used the same number as the `shouldDeploy` routine (my last PR had too few; so I guess I'm binary searching).

NickAragua: :)

megamek/src/megamek/server/Server.java
sixlettervariables: Yes, but only take action on offboard entities.

PR #2028, opened 2020-06-25
Damage decals showing the unit damage level on icons

### LINKED ISSUE #1910: RFE: Ability to apply damage decal to sprite.
Summary from the Slack Conversation.  The idea would be to have a decal (like camo) that could be applied to a sprite to reflect the damage it's taken.  

**Piratical Initial Comment**
_Is it possible in the code with some tweaking to make it so you can apply multiple sets of camo to a unit? Basically the first set would be the base camouflage coat and then another on top of that. I wanted to be able to add damage decals to units in game over top of the units base coat._

**BullseyeSmith Question**
_How could you see the bullet holes on such a small scale?  Would you want the image to have small cut-outs on it like it was missing a corner or an arm?  Do you want smoke coming off it from outside the unit outline (which isn't part of the shown camo)?  Do you have a before & after example,_

**Piratical Comment**
_I was thinking of patches of damaged armor and internal structure that looked chewed up in places for the top decal, scorched by PPC/AC blasts. Was something I thought of last night randomly.  Not much of an artist so no examples off the top of my head._

@neoancient comment from Slack
_It may not be that complicated. It we had an overlay for each damage level (light, moderate, heavy, critical) it could be applied to the sprite when the level changes._

I'm confident that either of the artists that do sprites (Deadborder or Saxarba) could do this up for us. But would each unit type need it's own decal or could one set of decals be used from all unit types

SJuliez:
Resolves #1910 

Adds damage decals to units based on entity.getDamageLevel(), showing different decals for light, moderate, heavy and crippled. Since this is pixel-added anyway, we can use any calculation for it; currently, it uses some sort of overlay mode with the example decals provided by saxarba. It currently looks a little dark-ish, I'll maybe have to change the calculation.

I found no game event that will trigger whenever a unit receives damage, not even gameEntityChanged seems to do that, so currently the update of the unit icons happens at each phase change for units that received damage this turn and have at least light damage. 

Since it will be difficult to distinguish the various damage levels, especially on small units, I suggest using a slight smoke effect for heavy dmg and a more dense one for crippled, preferably starting somewhere near the center.

Notes to self:
- [x] loading and adding decals
- [x] GUI Pref setting
- [x] actual decals (have two)
- [x] test multihex units, add decal to damaged hex only
- [x] use multiple different decals for each level? Requires more work on loading/storing/choosing them

image


NickAragua:
I think updating damage levels after phase change is fine. 

Agreed about differentiating damage levels with smoke or whatever other visible indicator. 

Not sure about needing a GUI pref, but it probably won't hurt (let's make it on by default though).

It would be pretty neat if we could just apply the damage decals to the damaged "sides" only of multi-hex units.

HammerGS:
This fits firmly into the run before we walk section. What about something like a decal for a unit hit with inferno's?

Saxarba1:
I did a first version of a smoke and a smoke + fire decals that could be applied over the moderate damage to indicate heavy and crippled respectively. Only tested with the locust and behemoth graphics. They look a bit over the top in the locust and a bit scarce in the behemoth, so I guess they should work fine for all sizes.
CrippledDamage
HeavyDamage
DamageTest

NickAragua:
Just want to say that looks freaking awesome.

HammerGS:
Those look fantastic.  Have a couple of suggestions.  The decals with smoke and fire look a little linear maybe stagger them a little like a scalene triangle.  The books describe autocannons are pock marked (which you've got) but lasers and PPC's leave scars. 

But I'll echo NickArgua - freaking awesome and can't wait to see this come together.

Saxarba1:
Yes, good idea. I'll move the central smoke column/fire a bit up so they are not all aligned horizontally.

Saxarba1:
Also, I'll try replacing some of the pock marks with linear scars.

SJuliez:
Thanks @Saxarba1 !
I divided the smoke overlays. Units over 70 tons get three smokestacks, over 40 tons get 2 and the smalls get only 1.
The IMMOBILE text kinda gets in the way. I'm tempted to move that to the unit tag. I should probably exclude infantry ;)
Smoke1
Smoke2
Smoke3
SmokeFire1
SmokeFire2
SmokeFire3







image

NickAragua:
I like the idea of moving the immobile label (for consistency, I'd almost say we should move all such labels to the unit tag).

HammerGS:
Having played around in the branch, this is gonna be awesome. I also agree the "Immobile" type tags got to move.

SJuliez:
Ok, I'll take a loot at the tags, but I'll keep it separate from this one.

SJuliez:
Some more changes: 
- made slim units get only 2 smokes/fires at most (tanks and Aero)
- made very slim units get only one fire/smoke (VTOL)
- only scars in the lobby
- no damage on the icons in the unit overview (they are smaller and too easily obscured)
- and refactored the tilemanager by extracting the EntityImage internal class and put tileset things into its own package

Should give this a test IMO. Possible things to add would be:
- multiple different images for the same damage (smoke and fire mostly, with the scars it's not very visible that it's always the same image)
- inferno hits (already have nice images from Saxarba, these could also be applied as a heat indicator)
- add scars to wrecks; let wrecks burn for a round, then smoke for another round...

HammerGS:
I know Saxarba was going to work on updated decals. Should we hold off until we have those?

megamek/src/megamek/client/ui/swing/tileset/EntityImage.java
```diff
@@ -0,0 +1,405 @@
...
+        this.tint = tint;
+        this.camo = camo;
+        parent = comp;
+        this.wreck = wreck;
+        this.dmgLevel = entity.getDamageLevel();
+        this.weight = entity.getWeight();
+        isInfantry = entity instanceof Infantry;
+        isSecondaryPos = secondaryPos != 0 && secondaryPos != -1;
+        isPreview = preview;
+        isSlim = entity instanceof Tank || entity instanceof Aero;
```
NickAragua, line 115: Is there a way of determining this based on the unit image (without any too crazy computations)? e.g. checking width/height?  A Demolisher tank might warrant three smoke stacks, while a Maultier APC only really needs one.

megamek/src/megamek/client/ui/swing/tileset/TilesetManager.java
```diff
@@ -0,0 +1,556 @@
...
+import megamek.common.util.ImageUtil;
+import megamek.common.util.MegaMekFile;
+
+/**
+ * Handles loading and manipulating images from both the mech tileset and the
+ * terrain tileset.
+ *
+ * @author Ben
+ */
+public class TilesetManager implements IPreferenceChangeListener, ITilesetManager {
```
NickAragua, line 49: Did you just move this file, or did you make changes to it? If the latter, could you point to where the changes are?

SJuliez: I'm really a git noob. Does it show everything as changed because I moved it?
Well in reality, I changed almost everything because there was so much redundant and (to me) unorganized code. I also moved out EntityImage to its own class and removed the default hex Tileset that was in there as a 1000-line long list of println (instead that default is now present as a simple text file in case something with the actual tileset files goes wrong... and I guess there's a few other files that can crash MM when they're not present so why worry too much about this one).

megamek/src/megamek/client/ui/swing/tileset/EntityImage.java
SJuliez: Possible by checking the pixels in the center row. I just think that it wouldn't make a big difference; two or three smoke stacks on a unit that could hold three but burns alright with two - in other words, all the pixel-counting effort wouldn't really show in the result. I wouldnt really want to do that to increase the number of smoke stacks. If on the other hand many units showed smoke stacks where there shouldnt be any and a weight test doesnt help, then that's be a good solution.

NickAragua: I'm not aware of too many skinny 100-ton tanks. I guess the Long Tom + Trailer currently? Probably let's leave it for a later PR.

megamek/src/megamek/client/ui/swing/tileset/TilesetManager.java
NickAragua: Ok, I'll review it as "fresh".

megamek/src/megamek/client/ui/swing/tileset/TilesetManager.java
```diff
@@ -0,0 +1,556 @@
...
+        // A new Hex Tileset has been selected
+        if (e.getName().equals(IClientPreferences.MAP_TILESET)) {
+            HexTileset hts = new HexTileset(boardview.game);
+            try {
+                hexTileset.incDepth = 0;
+                hts.loadFromFile((String) e.getNewValue());
+                hexTileset = hts;
+                boardview.clearHexImageCache();
+            } catch (IOException ex) {
+                return;
```
NickAragua, line 148: Is there a chance that something meaningful could happen here? If so, let's log it.

megamek/src/megamek/client/ui/swing/tileset/TilesetManager.java
```diff
@@ -0,0 +1,556 @@
...
+            hexTileset.incDepth = 0;
+            hexTileset.loadFromFile(PreferenceManager.getClientPreferences().getMapTileset());
+        } catch (Exception FileNotFoundException) {
+            System.out.println("Error loading tileset, "
+                    + "reverting to default hexset! " + "Could not find file: "
+                    + PreferenceManager.getClientPreferences().getMapTileset());
+            if (new MegaMekFile(Configuration.hexesDir(), FILENAME_DEFAULT_HEX_SET).getFile().exists()){
+                hexTileset.loadFromFile(FILENAME_DEFAULT_HEX_SET);
+            } else {
+                System.err.println("Critical Error loading default tileset!");
```
NickAragua, line 130: Do we want to avoid system.out.println and instead log the error?

megamek/src/megamek/client/ui/swing/tileset/TilesetManager.java
```diff
@@ -0,0 +1,556 @@
...
+            return (generic != null) ? ImageUtil.getScaledImage(generic, 56, 48) : null;
+        }
+        return entityImage.getIcon();
+    }
+
+    /** Retrieve a wreck icon for the unit. */
+    public Image wreckMarkerFor(Entity entity, int secondaryPos) {
+        EntityImage entityImage = getFromCache(entity, secondaryPos);
+        if (entityImage == null) {
+            System.out.println("Unable to load wreckMarker image for entity: " + entity.getShortNameRaw()); //$NON-NLS-1$
```
NickAragua, line 173: Another system.out.println which will not be seen by anyone.

megamek/src/megamek/client/ui/swing/tileset/TilesetManager.java
```diff
@@ -0,0 +1,556 @@
...
+            return imageFor(entity, entity.getSecondaryFacing(), secondaryPos);
+        }
+        return imageFor(entity, entity.getFacing(), secondaryPos);
+    }
+
+    /** Retrieve an image for the unit. */
+    public Image imageFor(Entity entity, int facing, int secondaryPos) {
+        EntityImage entityImage = getFromCache(entity, secondaryPos);
+        if (entityImage == null) {
+            System.err.println("Unable to load image for entity: " + entity.getShortNameRaw()); //$NON-NLS-1$
```
NickAragua, line 199: I'm going to stop calling out system.println and just say that it would be useful to go through this file and replace them with calls to actual logging.

Saxarba1:
Here are some reworked damage decals. The light and moderate decals have laser scars and fill the whole hex now, so in multi-hex units they don't appear as isolated in each hex. Also I moved the smoke and fire so they are not aligned horizontally in the cases when the three smoke stacks are used. There's also a big smoke and fire version for the central hex of multi-hex units.
LightDamage
ModerateDamage
CentralMultiHexCrippled
CentralMultiHexHeavy
CrippledDamage
HeavyDamage

HammerGS:
@Saxarba1 Those are fantastic.

SJuliez:
@Saxarba1 Thanks!
Which blending mode were you using when painting these?
image
When just overlaying it I find the laser scars don't look like scars. Using "multiply" in a gfx program works better for the scars but makes the shot marks much darker:
image
Also, I think the scars are slightly visible outside the unit in the drop shadow but I can deal with that.

Saxarba1:
I was using a normal overlay. The yellow/orange color was trying to achieve a residual glow in the scars. I'll make a version with the scars using the same neutral colors as the shot marks and making the scars thinner, to see if they look better.

Saxarba1:
@SJuliez See if these look better. I did a couple of versions of each in case you want to add some variation. The scars repetition will probably be more noticeable than the small shots alone.
Light1
Light2
Moderate1
Moderate2

SJuliez:
Using the MMLogger now. Also removed the decals from drop shadows and added support for the bigger multi-hex image for dropships.

SJuliez:
Not quite finished... Need to add the changed small smoke/fire, decide if its worth adding support for multiple images

HammerGS:
You had some feedback tossed up in Slack on the test images you posted.

SJuliez:
Saw it, already changed the scar images used to the latest version from Saxarba. But he provided two of each and I could make up a few different smoke/fire images from the ones I now have. But it's not a one-liner code addition to load and choose between an arbitrary number of images. Will have to sleep on it (and get through a work monday).

SJuliez:
OK, now it supports multiple images for everything. The images are organized into directories, with Crippled/ being for the scars of crippled units and smoke1/ for the smoke images of very slim/light units and so on. The names of the images don't matter and we can put in as many as we like. I created a few from Saxarba's images.

From my side this is ready for review. @NickAragua maybe you want to take a look at the changes again. The DirectoryItems class was very practical for this, but I noticed that the ImageFileFactory (=FilenameFilter) that is often used with it scales all images to hex size (a little evil with such a generic name). I didnt want to change the places where it is used so I renamed it to ScaledImageFileFactory and added ImageFileFactory which now doesnt scale images.

image

megamek/src/megamek/client/ui/swing/util/ScaledImageFileFactory.java
```diff
@@ -0,0 +1,215 @@
...
+            private ZipEntry itemEntry = zipEntry; // copy the ZipEntry
+            private Image image = null; // cache the Image
+
+            public Object getItem() throws Exception {
+
+                // Cache the image on first use.
+                if (null == image) {
+
+                    // Get ready to read from the item.
+                    InputStream in = new BufferedInputStream(zipFile
```
NickAragua, line 138: According to LGTM (and it seems to be right), you should close this input stream regardless of whether the call succeeds or fails.

Issue #2042, opened 2020-07-02
Flak Artillery Vs Airborn Aerospace  Minimum range


Logajam:
### Environment
MM 47.7

### Description
When firing at an aerospace fighter using Flak Artillery, Minimum range for a direct fire artillery attack is ignored.  TO pg 185

Tested By flying an altitude 1 aerospace fighter so its flight path was 1 hex away from the artillery unit.  The game considers this to be range 3, and brings up the direct fire artillery minimum range error message.

PR #2046, opened 2020-07-03
no minimum range restriction for arty flak attacks vs ASF

NickAragua:
Two changes:
- per TacOps page 185, there is no "minimum range" for artillery flak attacks vs airborne aerospace units
- allow the bot to properly compute expected damage for direct-fire artillery attacks, so she can potentially use artillery in an AA role

Addresses #2042

### DIFF
--- a/megamek/src/megamek/common/actions/WeaponAttackAction.java
+++ b/megamek/src/megamek/common/actions/WeaponAttackAction.java
@@ -1780,7 +1780,7 @@ public class WeaponAttackAction extends AbstractAttackAction implements Serializ
                     return Messages.getString("WeaponAttackAction.NoDirectCruiseMissile");
                 }
                 // Direct fire artillery cannot be fired at less than 6 hexes
-                if (isArtilleryDirect && (Compute.effectiveDistance(game, ae, target) <= 6)) {
+                if (isArtilleryDirect && !target.isAirborne() && (Compute.effectiveDistance(game, ae, target) <= 6)) {
                     return Messages.getString("WeaponAttackAction.TooShortForDirectArty");
                 }
                 // ...or more than 17 hexes
--- a/megamek/src/megamek/client/bot/princess/WeaponFireInfo.java
+++ b/megamek/src/megamek/client/bot/princess/WeaponFireInfo.java
@@ -405,6 +405,13 @@ public class WeaponFireInfo {
             return 7D;
         }
 
+        // artillery and cluster table use the rack size as the base damage amount
+        // a little inaccurate, but better than ignoring those weapons entirely       
+        if((weaponType.getDamage() == WeaponType.DAMAGE_BY_CLUSTERTABLE) ||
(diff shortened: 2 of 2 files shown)


megamek/src/megamek/common/actions/WeaponAttackAction.java
```diff
@@ -1780,7 +1780,7 @@ private static String toHitIsImpossible(IGame game, Entity ae, int attackerId, T
                     return Messages.getString("WeaponAttackAction.NoDirectCruiseMissile");
                 }
                 // Direct fire artillery cannot be fired at less than 6 hexes
-                if (isArtilleryDirect && (Compute.effectiveDistance(game, ae, target) <= 6)) {
+                if (isArtilleryDirect && !target.isAirborne() && (Compute.effectiveDistance(game, ae, target) <= 6)) {
                     return Messages.getString("WeaponAttackAction.TooShortForDirectArty");
                 }
```
Windchild292, line 1785: Should this check for Flak ammunition?

Also, I don't believe isAirborne handles landed VTOLs (which do seem to be a valid target based on my read of the rules in question). I'm not even going to ask about the other stuff expending VTOL MPs during a turn they are landed, for example, as that would also be a valid target based on my read of the rules.

NickAragua: The arty actually uses standard ammo to do this type of attack.

Per TacOps page 185, the minimum range exception only applies to attacks targeting "airborne aerospace units", so VTOLs still need to be out of minimum range to be targeted.

Windchild292: First line states any ordnance with the "F" Damage type in the brackets?

My bad, I missed that bit. Although that raises a different problem, namely isAirborne returns true if the altitude is greater than 0 or the movement mode is either aerodyne or spheroid.

NickAragua: I checked it, and MegaMek only allows targeting of airborne aerospace units using F-type ammo (which basically means standard ammo) with direct-fire artillery - cluster and FAE don't show as allowing a direct attack vs airborne ASF. 

I'm not really sure if there is any other problem with using isAirborne(), that's the standard test we use to check if a unit is an ASF and currently in the air. The only point of this specific check is to see if we need to declare the attack as impossible due to being within minimum range.

Windchild292: Wouldn't that still return true for a vtol hovering at a higher elevation though?

NickAragua: Specifically no. It's only for "airborne units in the manner of an aerospace fighter".

NickAragua: There's a separate "isAirborneVTOLorWIGE" method that handles the VTOL situation.

NickAragua:
Anything else on this?

PR #2052, opened 2020-07-06
Account for variable equipment size and armored components in TRO entries

neoancient:
This is part of the variable equipment size changes that kept separate then forgot about. I didn't realize it wasn't merged until I tried to update it to indicated armored components.

The equipment table is built by sorting the equipment list by location and mapping the EquipmentType to the number found in that location. I have replaced the EquipmentType key with a data class that distinguishes between different sizes (e.g. a 20' ladder and a 40' ladder in the same location won't be combined into a single ladder) and also between armored and non-armored instances of the same equipment type in the same location.

There's also some code cleanup.

Fixes MegaMek/megameklab#636


megamek/src/megamek/common/templates/InfantryTROView.java
```diff
@@ -226,10 +226,8 @@ private void addAugmentationNotes(List<String> notes) {
         }
         if (!options.isEmpty()) {
             notes.add(Messages.getString("TROView.InfantryNote.Augmented"));
-            options.forEach(o -> {
-                notes.add(o.getDisplayableName().replaceAll("\\s+\\(Not Implemn?ented\\)", "") + ": "
-                        + o.getDescription().replaceAll("See IO.*", ""));
-            });
+            options.forEach(o -> notes.add(o.getDisplayableName().replaceAll("\\s+\\(Not Implemented\\)", "") + ": "
```
NickAragua, line 229: Yikes.

megamek/src/megamek/common/templates/TROView.java
```diff
@@ -439,26 +429,30 @@ protected int addEquipment(Entity entity, boolean includeAmmo) {
...
-                String name = stripNotes(eq.getName());
+            for (final Map.Entry<EquipmentKey, Integer> entry : equipment.get(loc).entrySet()) {
+                final EquipmentType eq = entry.getKey().getType();
+                final int count = equipment.get(loc).get(entry.getKey());
+                String name = stripNotes(entry.getKey().name());
+                if (entry.getKey().isArmored()) {
+                    name += " (Armored)";
+                }
                 if (eq instanceof AmmoType) {
                     name = String.format("%s (%d)", name, ((AmmoType) eq).getShots() * count);
```
NickAragua, line 446: The name variable appears to get overwritten if the equipment is not ammo but takes up more than one slot?

megamek/src/megamek/common/templates/InfantryTROView.java
neoancient: I was lazy the first time around and accounted for the spelling error rather that fixing it.

megamek/src/megamek/common/templates/TROView.java
neoancient: Actually it's if there's more than one of them. It adds the count and appends an 's' to the name. But the formation of a plural is not correct in cases where the equipment name itself ends in 's' and the possible addition of "(Armored)" throws it off in any case. So I removed the grammatical inflection as well as including the armored tag.

PR #2070, opened 2020-07-20
Fix spheroid extra fire control weight arc indices

neoancient:
The virtual aft side locations on spheroid dropships and small craft have a different index offset in the named constants and the method that calculates whether extra tonnage is needed for fire control. The constant takes the hull location into account while the calculation method does not. The fix is in the first commit. The other two commits are code cleanup, mostly involving removing String concatenation from StringBuilder#append calls.

Fixes MegaMek/megameklab#668

### DIFF
--- a/megamek/src/megamek/common/verifier/TestSmallCraft.java
+++ b/megamek/src/megamek/common/verifier/TestSmallCraft.java
@@ -47,11 +47,11 @@ public class TestSmallCraft extends TestAero {
     public static final int ARC_LWING = SmallCraft.LOC_LWING;
     public static final int ARC_RWING = SmallCraft.LOC_RWING;
     public static final int ARC_AFT = SmallCraft.LOC_AFT;
-    public static final int ARC_NONE = SmallCraft.LOC_HULL;
+    public static final int REAR_ARC_OFFSET = SmallCraft.LOC_HULL;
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/verifier/TestSmallCraft.java
```diff
@@ -518,22 +519,24 @@ public SmallCraft getSmallCraft() {
...
-        return " is greater than " + Integer.toString(wert) + "!";
+        return " is greater than " + wert + "!";
     }
 
     /**
      * Checks to see if this unit has valid armor assignment.
      * 
-     * @param buff
-     * @return
+     * @param buff A buffer that collects messages about validation failurs
```
Windchild292, line 530:
```suggestion
     * @param buff A buffer that collects messages about validation failures
```

megamek/src/megamek/common/verifier/TestSmallCraft.java
```diff
@@ -542,16 +545,17 @@ public boolean correctArmor(StringBuffer buff) {
     
     /**
      * Checks that the heatsink type is a legal value.
-     * 
-     * @param buff
-     * @return
+     *
+     * @param buff A buffer that collects messages about validation failurs
```
Windchild292, line 549:
```suggestion
     * @param buff A buffer that collects messages about validation failures
```

PR #2076, opened 2020-07-22
Add former lookup names

neoancient:
Along the way apparently the internal name of some equipment changed and the old value was not added as a lookup name.

Fixes MegaMek/megameklab#682


Windchild292:
I think you meant to link to MegaMek/megameklab#682.

PR #2080, opened 2020-07-22
Unit Label Team Coloring

SJuliez:
Adds a team coloring option for the unit label borders and updates the Client settings. The local player's unit labels are by default green, allied units (same team) cyan and all enemy units red. 
Currently there's no GUI to change the colors. Wanted to keep this separate as there's quite a few other colors that could use some sort of UI.
Because I stumbled over it after dealing with some conflicts I also adjusted the Client Settings dialog to immediately update the icons when the damage decals are activated or deactivated (preview function)

image

image


megamek/src/megamek/client/ui/swing/boardview/EntitySprite.java
```diff
@@ -510,9 +510,23 @@ public void prepare() {
                 graph.fillRoundRect(labelRect.x, labelRect.y, labelRect.width,
                         labelRect.height, 5, 10);
 
-                if (guip.getEntityOwnerLabelColor()) {
-                    graph.setColor(PlayerColors.getColor(
-                            entity.getOwner().getColorIndex(), false));
+                // Draw a label border with player colors or team coloring
+                if (guip.getUnitLabelBorder()) {
+                    if (guip.getUnitLabelBorderTeam())
+                    {
```
Windchild292, line 516:
```suggestion
                    if (guip.getUnitLabelBorderTeam()) {
```

PR #2083, opened 2020-07-25
fix npe when displaying summary/TRO for gun emplacements

### LINKED ISSUE #2022: Watchdog ECM Bunker - Empty TRO card
### Environment
- MegaMek 0.47.7
- Java:
openjdk version "11.0.7.2" 2020-04-26
OpenJDK Runtime Environment (Alibaba Dragonwell) (build 11.0.7.2+9)
OpenJDK 64-Bit Server VM (Alibaba Dragonwell) (build 11.0.7.2+9, mixed mode)
(I don't think that Java was a matter; all other things were OK).
- OS: Linux (Ubuntu 20.04 Focal Fossa)

### Description
So I've selected the Watchdog ECM Bunker, and TRO card was empty, like on the picture.

### Files
Provided screenshot
bug

Now the question is - is it intended or not?

NickAragua:
This fixes #2022 in two ways:

1) isClanArmor no longer throws an exception if passed an invalid location, returning false instead
2) Created new TRO templates for gun emplacements which omit armor and structure information and selectively use those for the TRO display.

### DIFF
--- a/megamek/src/megamek/common/templates/VehicleTROView.java
+++ b/megamek/src/megamek/common/templates/VehicleTROView.java
@@ -19,6 +19,7 @@ import java.util.Map;
 
 import megamek.common.Entity;
 import megamek.common.EntityFluff;
+import megamek.common.GunEmplacement;
 import megamek.common.Messages;
 import megamek.common.Mounted;
 import megamek.common.SuperHeavyTank;
@@ -44,6 +45,13 @@ public class VehicleTROView extends TROView {
 
     @Override
     protected String getTemplateFileName(boolean html) {
+        if(tank instanceof GunEmplacement) {
+            if (html) {
+                return "gunemplacement.ftlh";
+            }
+            return "gunemplacement.ftl";
+        }
+        
         if (html) {
             return "vehicle.ftlh";
         }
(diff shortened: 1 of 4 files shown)


megamek/src/megamek/common/templates/VehicleTROView.java
```diff
@@ -44,6 +45,13 @@ public VehicleTROView(Tank tank) {
 
     @Override
     protected String getTemplateFileName(boolean html) {
+        if(tank instanceof GunEmplacement) {
```
Windchild292, line 48:
```suggestion
        if (tank instanceof GunEmplacement) {
```

megamek/src/megamek/common/templates/VehicleTROView.java
```diff
@@ -44,6 +45,13 @@ public VehicleTROView(Tank tank) {
...
+            if (html) {
+                return "gunemplacement.ftlh";
+            }
+            return "gunemplacement.ftl";
```
Windchild292, line 52:
```suggestion
            if (html) {
                return "gunemplacement.ftlh";
            } else {
                return "gunemplacement.ftl";
            }
```

Windchild292: I find it easier to read if the return is in an else in statements like this, but this is very much a suggestion.

PR #2088, opened 2020-07-27
Oververbose log

SJuliez:
I wanted to remove those many lines of MM log coming from the hexTileset and the MekSummary cache, so I changed it all to use the MMLogger and set them to Info and the log level to Warning.

The calls to MMLogger are very bulky (certainly worse than //NON-NLS1) so I added some convenience functions. These only need a `this` as the first parameter (instead of `getClass()`) and don't need the method name which is taken from the stack trace instead. There's also a convenience function for setting the log level that uses `this` as a parameter instead of a String giving the class name, removing the chance to misspell and not notice.

Together with the constants
```
private static final MMLogger LOG = DefaultMmLogger.getInstance();
private static final LogLevel LOGLVL = LogLevel.WARNING;
```
one can call
`LOG.info(this, "Unable to load wreck image for entity: " + entity.getShortNameRaw());`
and
`LOG.setLogLevel(this, LOGLVL);`

which I think is an improvement.


megamek/src/megamek/common/logging/DefaultMmLogger.java
```diff
@@ -279,6 +294,23 @@ public void info(final Class<?> callingClass,
...
+    @Override
+    public void info(Object callingObject, String message) {
+        log(callingObject.getClass(), getCallingMethod(), LogLevel.INFO, message);
+    }
+    
+    /** 
+     * Retrieves the name of the method calling log/info/error etc. 
+     * from the stack trace. 
+     */
+    private String getCallingMethod() {
```
sixlettervariables, line 306:
```suggestion
    /** 
     * Retrieves the name of the method calling log/info/error etc. 
     * from the stack trace. 
     *
     * This method MUST be called from a top-level method (such as info or fatal)
     * as it depends on the call stack depth of exactly three methods between
     * internally calling getStackTrace and the user's code.
     */
    private String getCallingMethod() {
```

SJuliez:
Very true. Thanks.

HammerGS:
@SJuliez  left you a message on Slack not sure if you saw.

Windchild292:
Also, this has merge conflicts that need to be fixed before being merged.

PR #2094, opened 2020-07-29
PPC Capacitors also add 5 heat in the first charge turn

### LINKED ISSUE #2065: PPC capacitors not building up heat properly
Version 0.47.7

PPC Capacitors are not generating any heat on the turn that they are charged.
As per the BattleMech Manual page 105, the action of charging a PPC should generate the same +5 heat as on any turn where the charge is held but not fired.

Attached is a save file where this was tested.

[Capacitor.sav.gz](https://github.com/MegaMek/megamek/files/4943269/Capacitor.sav.gz)

SJuliez:
Resolves #2065

### DIFF
--- a/megamek/src/megamek/common/Mounted.java
+++ b/megamek/src/megamek/common/Mounted.java
@@ -29,7 +29,6 @@ import java.util.Vector;
 
 import megamek.common.actions.WeaponAttackAction;
 import megamek.common.logging.DefaultMmLogger;
-import megamek.common.logging.MMLogger;
 import megamek.common.options.OptionsConstants;
 import megamek.common.options.WeaponQuirks;
 import megamek.common.weapons.AmmoWeapon;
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/common/Mounted.java
```diff
@@ -979,35 +978,60 @@ public void setHotLoad(boolean hotload) {
...
+     * Returns 1 or 2 if this Mounted has a linked
+     * and charged (= set to charge in an earlier turn) 
+     * PPC Capacitor.
      */
-
     public int hasChargedCapacitor() {
+        if (getCrossLinkedBy() != null
+                && isWorkingCapacitor(getCrossLinkedBy())
+                && getCrossLinkedBy().curMode().equals("Charge")) {
+            return 2;
```
NickAragua, line 998: Can we take these magic numbers and make them into constants?

NickAragua: Or, even better, an enum?

SJuliez: That is a problem of the original code. I only wanted to correct the error.

Issue #2095, opened 2020-07-31
NPE in Ruler.buildAttackInfo


HammerGS:
From Discord
I have a unit that's refusing to deploy properly? Or something just finished repairing it today (mostly, still need a heat sink delivered ...) it shows up in the proper lance in the ToE, and in the table of deployed units in the Briefing Room screen, and in the Megamek lobby when I Start Game

but then when I actually enter the game it's not there

### Environment
What version of MegaMek does your issue apply to?  47.7 MekHQ
What operating system are you using?  Windows, 

### Description
In line MekHQ file line 471
```
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
	at megamek.client.ui.swing.Ruler.buildAttackInfo(Ruler.java:436)
	at megamek.client.ui.swing.Ruler.setText(Ruler.java:382)
	at megamek.client.ui.swing.Ruler.checkBoxSelectionChanged(Ruler.java:508)
	at megamek.client.ui.swing.Ruler$4.itemStateChanged(Ruler.java:173)
```

### Files
[megameklog.txt](https://github.com/MegaMek/megamek/files/5004461/megameklog.txt)
[mekhqlog.txt](https://github.com/MegaMek/megamek/files/5004462/mekhqlog.txt)

Windchild292:
I'm pretty certain this is two different issues, but it is still an NPE that should be fixed at least.

PR #2110, opened 2020-08-05
fix ruler exception

NickAragua:
Somehow, the person who was using the "ruler" for #2095 managed to select an out-of-bounds hex. This will prevent that from erroring out.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/Ruler.java
+++ b/megamek/src/megamek/client/ui/swing/Ruler.java
@@ -375,7 +375,7 @@ public class Ruler extends JDialog implements BoardViewListener {
             // leave at default value
         }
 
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/client/ui/swing/Ruler.java
```diff
@@ -375,6 +375,10 @@ private void setText() {
             // leave at default value
         }
 
+        if(!client.getGame().getBoard().contains(start) || !client.getGame().getBoard().contains(end)) {
```
Windchild292, line 378:
```suggestion
        if (!client.getGame().getBoard().contains(start) || !client.getGame().getBoard().contains(end)) {
```

PR #2112, opened 2020-08-05
ARTS construction data

neoancient:
Support for including Advanced Robotic Transport System on units (IO, p. 147). It's a fully automated fighter or small craft bay for drone ships that allows it to launch, recover, and make limited repairs with no bay personnel. It can also be used with standard naval repair facilities. It has no effect on play (though it could be added as a repair location in MekHQ), but increases the bay weight by 25% and adds 1 million C-bills (per bay) to the cost.

Closes MegaMek/megameklab#658.

### DIFF
--- a/megamek/src/megamek/common/ASFBay.java
+++ b/megamek/src/megamek/common/ASFBay.java
@@ -154,14 +154,20 @@ public final class ASFBay extends Bay {
 
     @Override
     public String getUnusedString(boolean showrecovery) {
-        if (showrecovery) {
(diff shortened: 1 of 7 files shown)


neoancient:
A unit that was built with the old placeholder MiscType will load correctly and the ARTS will be added to the parts that couldn't load. The unit will need to be rebuilt anyway. There's no way to determine in every case which bay it should belong to, so we wouldn't be able to add conversion code. Even if it were possible, I don't think it's enough of an issue to bother with. MekHQ will log that the part couldn't be restored and remove it from the campaign.

PR #2132, opened 2020-08-11
add Java object look-ahead library to mitigate deserialization vulnerabilities

Algebro7:
Megamek unsafely deserializes untrusted raw objects from files in `megamek/common/MechSummaryCache.java` and over the wire in `megamek/common/net/marshall/NativeSerializationMarshaller.java`, which could allow attackers to gain remote code execution on the game host machine. I don't believe this is currently exploitable due to a lack of well-known gadgets in the classpath (such as those within CommonsCollections), but it could still be possible to construct a viable gadget chain with enough effort. Future dependencies that get added could also make it exploitable.

This commit wraps the InputStream in these instances with SerialKiller, a Look-Ahead Java Object Deserialization library. I generated a whitelist of allowed objects using SerialKiller's profiling mode, which is stored in mmconf/serialkiller.xml. This file will need to be updated if more objects are added that need to be deserialized. Note that I've whitelisted all internal classes (`megamek.common.*`), so this should only affect external library classes.

I've tested by playing a couple of games locally on Linux and verifying that no benign classes are being blocked. I also confirmed that the deserialization proofs-of-concept I developed are blocked by SerialKiller (with an error in megameklog.txt). I would love help testing to make sure everything works smoothly in real multiplayer games.


megamek/src/megamek/common/net/marshall/NativeSerializationMarshaller.java
```diff
@@ -48,7 +50,7 @@ public void marshall(Packet packet, OutputStream stream) throws Exception {
      */
     @Override
     public Packet unmarshall(InputStream stream) throws Exception {
-        ObjectInputStream in = new ObjectInputStream(stream);
+        ObjectInputStream in = new SerialKiller(stream, Objects.requireNonNull(getClass().getClassLoader().getResource("./megamek/serialkiller.xml")).toString());
```
Algebro7, line 53: I just noticed the game hangs when I launch using the unixDistTar package, but not when I run with gradlew. Do I need to move the shipped serialkiller.xml file to the mmconf folder instead of the resources folder to use it here?

Algebro7: This should be fixed with 3cc0cbd

NickAragua:
How does this interact with MekHQ?

Algebro7:
Good question--I believe it shouldn't have any impact on MekHQ as long as the mmconf folder is still accessible at the same path, but I can test that

NickAragua:
Try megamek.common.util.SharedConfiguration or SerializationHelper

Algebro7:
c47f9e0 moved the SerialKiller config to the JAR (where it belongs) and fixes the resource paths so it's loaded both when ran from a class file and from the DistTars.

NickAragua:
Even better.

Algebro7:
Regarding the SerialKiller source repo itself, I'm currently using jitpack to build the master branch since there were a number of fixes merged after the last release but the maintainer never got around to tagging a new release (0.5). Should we keep using master or would you prefer to pin the current commit hash so it doesn't change on us (probably unlikely)? We can always switch back to the maven central release if 0.5 gets published.

NickAragua:
Looks like it hasn't changed for two years, so I think we're ok with using the master branch.

Algebro7:
I'm currently trying to get it to build against mekhq to make sure there it doesn't break anything there...I'd also be more comfortable playing a couple of real life multiplayer games to make sure there aren't any external classes I missed in my whitelist while profiling.

Algebro7:
MekHQ and MML build and appear to build and run normally after adding the SK jitpack repo to their build.gradle files (which I've opened PRs for). I'll try to play some games this weekend to catch any additional classes that need to be whitelisted, but I think I've exercised most of the common functionality.

Algebro7:
I've played a few games with the new SK configuration with no issues/benevolent classes being blocked, both as the client and the host. I'm pretty sure this is good to merge now.

NickAragua:
I've been thinking about it, and why not just whitelist everything in the megamek namespace?

Algebro7:
Rather than megamek.common?

NickAragua:
Correct. It's theoretically possible that some game data may wind up not being from the common namespace.

Algebro7:
That's certainly doable, I just didn't because I didn't run into anything other than megamek.common. I'll update the whitelist to reflect that in a bit.

NickAragua:
Ok, looks good.

Issue #2099, opened 2020-07-31
RFE: Unit sprites on Combat Log to reduce wall of text


MrVexxar:
My suggestion is:

The implementation of unit sprites a line above the start of each attack, displaying the attacking unit with an arrow (or anything else) pointing to the defending unit. You could keep the rest of the text as it is, since the sprites would be enough to fix the "cramped wall of text" feel the Combat Log has, when there are too many units.

This would reduce the 'wall of text' feel of the Combat Log, which very often causes you to misread the attacks (and end up cheering up for a critical hit on your own unit. xD), and also keep you from getting lost in the combat log. Very often, I scroll down a bit too much and end up losing which line I was reading, when there are too many units in battle.

If the implementation of unit sprites is too much of a hassle to be worth it, I would at least suggest for a revision on the Combat Log layout. Further separation of each attack, and more visual cues to understand where each unit attack starts and to which player they belong, and who are they attacking.

Not saying the current layout is bad. It's just it can get confusing when there are too many units, specially if they have missiles.

--
On a side note: Thanks for this amazing game! You guys are incredible!

-edit for better wording on a few things-

PR #2135, opened 2020-08-12
Unit sprites in the Combat Log

Krashner:
This addresses #2099 by adding unit sprites to the combat log. It's meant to help illustrate more clearly what is happening to whom during a battle. By default, it doesn't display images for reports that are indented more than once, specifically because having an image for every entry takes up a ton of vertical space. However, there is a function to override that if you wanted to add the images to a specific report. I didn't try to add wreckage or damage decals for the sake of readability and consistency, but I don't imagine it would be too difficult to implement if someone wanted it.

The way it works is by first generating the composite art, i.e. base art, camo and tint, and then it converts that image to a base64 string and creates an \ object with that string as the src. By default, the HTMLEditorKit doesn't support decoding the base64 string, so there's a custom toolkit and image view that handles it. All of the entity art is created, encoded and cached when "receiveEntities" is called on the client. It knows where to put the \ object because it adds the entity's ID when "addDesc" is called from the report. Currently it adds the ID between two vertical bars, e.g. |1|, and then the client extracts the ID and swaps the whole sequence with the desired \ object.

This is what it looks like with the sprites.
ex1

It also displays in the mini report display and the movement alert. I believe that I accounted for all of the various ways of displaying reports, but I would not be surprised if I missed something.
ex4
ex5


megamek/src/megamek/client/Client.java
```diff
@@ -161,6 +129,15 @@
 
     public Map<String, Client> bots = new TreeMap<String, Client>(StringUtil.stringComparator());
 
+    //get the image data
+    private MechTileset mechTileset = new MechTileset(Configuration.unitImagesDir());
```
NickAragua, line 133: Why not use the TilesetManager class instead of basically rolling your own implementation of it? It's present in the BoardView class and contains pretty much all of the information you could need.

megamek/src/megamek/client/ui/swing/util/BASE64ImageView.java
```diff
@@ -0,0 +1,84 @@
+package megamek.client.ui.swing.util;
```
NickAragua, line 1: This class needs a standard megamek header comment.

megamek/src/megamek/client/ui/swing/util/BASE64ImageView.java
```diff
@@ -0,0 +1,84 @@
...
+import java.awt.*;
+import java.awt.image.BufferedImage;
+import java.io.ByteArrayInputStream;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.Base64;
+import java.util.Dictionary;
+import java.util.Hashtable;
+
+public class BASE64ImageView extends ImageView {
```
NickAragua, line 16: Let's put a javadoc-style comment on the class and any public-facing methods.

megamek/src/megamek/client/ui/swing/util/BASE64ToolKit.java
```diff
@@ -0,0 +1,33 @@
+package megamek.client.ui.swing.util;
+
+import javax.swing.text.*;
+import javax.swing.text.html.HTML;
+import javax.swing.text.html.HTMLEditorKit;
+
+public class BASE64ToolKit extends HTMLEditorKit {
```
NickAragua, line 7: "Copyright" header, javadoc comment on class and public-facing methods.

Windchild292:
I'd like to add that this should be optional, as some people might not want to clutter up the screen in question.

NickAragua:
Agreed, I'd suggest making it an "advanced" setting. Going to think it over a bit whether we want it enabled by default or not.

HammerGS:
I really really like this, but something about the sprite placement is jarring to my eye. Maybe at the front of the line always?

Just musings. Cause I think this is a great concept.

SJuliez:
I think this is very much an improvement. I would scale the icons down, which is already done for the UnitOverview (EntityImage.getIcon()). I suppose better placement would require using a <TABLE> in the report and then the icon would have to be placed separately from the text. Is that possible currently?

megamek/src/megamek/client/Client.java
Krashner: Initially I was having trouble getting an image to show, so after I got it working initially I just kept building around it. Now that I have a better understand of how it works, I'll definitely go back and do it that way and address the other changes as well.

Krashner:
I had experimented with several placements for the sprites, but I wasn't sure what people would prefer. I'll adjust it around some more and post some screenshots to see what everyone thinks looks best.

HammerGS:
More musings, I wonder if it would be possible to do a screen shot of the sprites involved in the attack and use that.

Krashner:
Using screenshots could definitely be done, you'd just have to encode the screenshot instead of the unit sprite. All you'd really be changing is where the image comes from. I might try that later and see how it looks.

I've swapped over to using the entity icons and tried a few different layouts. I think I'm partial to the first one currently, but I'm open to more suggestions.
newLineEx7

NickAragua:
Yeah, I like the 1st screenshot best. I think SJuliez suggestion of scaling the images down a little bit may be worthwhile as well. Maybe start with 75% and see how that looks?

HammerGS:
I like the first one as well. Looking forward to seeing how this develops.

sixlettervariables:
Love it. First choice would be the in game screen shot, second would be reduced size sprites. Either way, this is an outstanding addition.

Krashner:
In the last screenshot, the sprites were about 67% their regular size, which I believe is the same size used in the UnitOverview. I've also tried adjusting the scale to 50% and 75% to get an idea of what that looks like by comparison. I haven't quite figured out the best approach to get screenshots of the units yet, so for the time being it still just displays the unit sprite.

ex9

NickAragua:
I don't see an issue with using the unit sprites.

HammerGS:
Prefer the 67% (middle one) like the sprite rotation as well. That indicting actual facing in game?

Krashner:
That's how I was leaning as well and It does use the sprites actual facing. Initially I wanted to lock that down, but I think it's kind of neat so I left it. It also currently shows damage decals and I decided to go ahead and add wrecks also.

image

megamek/src/megamek/client/Client.java
```diff
@@ -1125,7 +1115,85 @@ public String receiveReport(Vector<Report> v) {
...
+        if(imgCache == null || !imgCache.containsKey(id)) {
+            return null;
+        }
+        return imgCache.get(id);
+    }
+
+    /**
+     * Hashtable for storing image tags containing base64Text src
+     */
+    private void cacheImageData(Entity entity){
```
NickAragua, line 1155: I'm fairly certain that the TilesetManager has its own caching mechanism. I don't think you need to have a separate caching mechanism here.

Krashner: I looked at that, but I'm storing strings after the image is encoded as a Base64 string and put into an \ tag. the data looks something like the following, but just quite a bit longer. 

``

I do actually use TilesetManager's image cache when I call `imageFor` though. Perhaps I should rename `cacheImageData` so that it doesn't sound like I'm just recaching the images?

NickAragua: Now that I look at it again, I get what you're doing.

Issue #2142, opened 2020-08-13
MM 47.9 Princess locking up during pathfinding - IllegalArgumentException


SimonLandmine:
Standalone MegaMek 47.9
Windows10
Java 1.8.0_261

Princess locked up as player 'Mi' with Missile Boat configuration while selecting move. Resolved by kick-and-replace.

> Exception in thread "Princess-precognition (Mi)" java.lang.IllegalArgumentException: Entity invalid. ID 8
> 	at megamek.common.Compute.isPilotingSkillNeeded(Compute.java:491)
> 	at megamek.common.MovePath.addStep(MovePath.java:497)
> 	at megamek.common.MovePath.addStep(MovePath.java:267)
> 	at megamek.common.MovePath.addStep(MovePath.java:179)
> 	at megamek.common.BulldozerMovePath.addStep(BulldozerMovePath.java:80)
> 	at megamek.common.pathfinder.DestructionAwareDestinationPathfinder.generateChildNodes(DestructionAwareDestinationPathfinder.java:171)
> 	at megamek.common.pathfinder.DestructionAwareDestinationPathfinder.findPathToCoords(DestructionAwareDestinationPathfinder.java:114)
> 	at megamek.common.pathfinder.DestructionAwareDestinationPathfinder.findPathToCoords(DestructionAwareDestinationPathfinder.java:55)
> 	at megamek.client.bot.princess.PathEnumerator.updateLongRangePaths(PathEnumerator.java:380)
> 	at megamek.client.bot.princess.PathEnumerator.recalculateMovesFor(PathEnumerator.java:317)
> 	at megamek.client.bot.princess.Precognition.run(Precognition.java:438)

[megameklog20200813-0304.zip](https://github.com/MegaMek/megamek/files/5066798/megameklog20200813-0304.zip)

gsparks3:
Just got a similar lockup/exception: 

```
Exception in thread "Princess-precognition (Lyran Commonwealth)" java.lang.IllegalArgumentException: Entity invalid. ID 82
	at megamek.common.Compute.isPilotingSkillNeeded(Compute.java:491)
	at megamek.common.MovePath.addStep(MovePath.java:497)
	at megamek.common.MovePath.addStep(MovePath.java:267)
	at megamek.common.MovePath.addStep(MovePath.java:179)
	at megamek.common.BulldozerMovePath.addStep(BulldozerMovePath.java:80)
[...]

PR #2148, opened 2020-08-14
make bot turn calculations more tolerant of threading-related errors fixable by kicking and replacing the bot

### LINKED ISSUE #2141: [0.47.9] ConcurrentModificationException while calculating bot firing turn
### Environment
MekHQ integrated Megamek - v0.47.9
Windows 10
Java version 1.8.0_261

### Description
Princess bot stalled due to a ConcurrentModificationException while calculating fire for a unit. See attached logs and savegame for details. 

Edit: After further checking, loading the savegame does not reproduce the bug on my end, so sadly it probably won't on yours either. Hopefully the logs are helpful. 

### Files
[megameklog.txt](https://github.com/MegaMek/megamek/files/5066284/megameklog.txt)
[mekhqlog.txt](https://github.com/MegaMek/megamek/files/5066286/mekhqlog.txt)
[indirectFireBug.sav.gz](https://github.com/MegaMek/megamek/files/5066304/indirectFireBug.sav.gz)
Player name: Coppercoats

NickAragua:
Theoretically fixes #2141 

I'm guessing this is caused when the bot is calculating a firing turn and receives an update for the entity being examined. Why or how this would happen, I don't know.

Not really sure what the impact will be on performance, or if this the right way of going about it (will this change actually prevent the concurrent modification exception?)


megamek/src/megamek/common/EquipmentType.java
```diff
@@ -553,10 +555,16 @@ public boolean hasModeType(String modeType) {
     		return false;
     	}
     	
-    	for(EquipmentMode mode : modes) {
-    		if(mode.getName().equals(modeType)) {
-    			return true;
-    		}
+    	// sometimes, the list of modes may be modified while we're looping through it
+    	// so let's do this in a thread-safe way
+    	List<EquipmentMode> threadSafeModes = Collections.synchronizedList(modes);
```
sixlettervariables, line 560: `Collections.synchronizedList` protects the given methods on `List` from concurrent access, however, they do not protect iteration. So you'd need to use `synchronized`.

Except, `threadSafeModes` isn't being used for all of your mutable uses of `modes` (`add`, `remove`). As it stands, you'd likely be better served by simply wrapping every usage of `modes` in a `synchronized` block on `modes` itself.

sixlettervariables: Also it looks like `modes` is a `Vector`, which already has the same behavior as a `synchronizedList`. So basically, either `addMode` or `removeMode` was called while this was going on. My guess is a network delay or GC pause gave it enough timing to fail.

NickAragua:
Went with a different approach - now, the two areas of bot code that exhibit crashes (as described in #2141 and #2142) which can be resolved by kicking and replacing the bot have a retry mechanism. The methods will attempt to retry up to three times, waiting between 500 and 1500 milliseconds between retries.

NickAragua:
Also fixed an observed bug during testing where a spheroid dropship in atmosphere that was out of control appeared to allow the player to move it.

megamek/src/megamek/client/bot/princess/PathEnumerator.java
```diff
@@ -168,166 +170,173 @@ void updateUnitLocations(Entity entity, List<MovePath> paths) {
...
-            getOwner().methodEnd(getClass(), METHOD_NAME);
+        
+        int retryCount = 0;
+        boolean success = false;
+        
+        while ((retryCount < BotClient.BOT_TURN_RETRY_COUNT) && !success) {
+	        try {
+	
+	            // Record it's current position.
+	            getLastKnownLocations().put(
```
sixlettervariables, line 181: Same as above.

NickAragua:
Refactored. Found another bug while I was in there (seems that superheavy mechs can't unjam rotary autocannons for some reason), but I'll deal with that one separately.

Issue #2150, opened 2020-08-16
Princess seems to stop with an error when abandoning a vehicle


mhjacks:
### Environment
What version of MegaMek does your issue apply to?  Latest Megamek Master branch (0.47.10-SNAPSHOT 553f6b6a13)
What operating system are you using?  Windows, Mac, or Linux? Linux (Fedora)
What java version are you using?  If unsure it will state your java version at the top of the megameklog.txt file.
Java 11

### Description
Princess stops when some vehicles are abandoned.  This makes it hard to use the abandon when immobile option.
My test case uses hawk moth 2's with Inferno MMLs to maximize crit potential

The relevant error in the log seems to be:
`Ejecting crew size: 3
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
        at megamek.client.Client.getTargetImage(Client.java:1192)
        at megamek.client.Client.cacheImgTag(Client.java:1167)
        at megamek.client.Client.receiveEntityRemove(Client.java:993)
        at megamek.client.Client.handlePacket(Client.java:1322)
        at megamek.client.bot.princess.Princess.handlePacket(Princess.java:2014)
        at megamek.client.Client$1$2.run(Client.java:192)
        at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
        at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:758)
        at java.awt.EventQueue.access$500(EventQueue.java:97)
        at java.awt.EventQueue$3.run(EventQueue.java:709)
        at java.awt.EventQueue$3.run(EventQueue.java:703)
        at java.security.AccessController.doPrivileged(Native Method)
        at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:74)
        at java.awt.EventQueue.dispatchEvent(EventQueue`

### Files
[megameklog.txt](https://github.com/MegaMek/megamek/files/5079961/megameklog.txt)
[autosave.sav.gz](https://github.com/MegaMek/megamek/files/5079962/autosave.sav.gz)

NickAragua:
Looks like a problem with PR #2135  @Krashner

Krashner:
I'll take a look at it today.

Windchild292:
Fixed in #2156

PR #2156, opened 2020-08-17
Fixed null pointer exception causing bot to break

Krashner:
Fix for #2150. In some instances, the BoardView for the client is being called before it gets initialized. So I just added a null check to fix the error. I also refactored the null check prior to trying to encode an image to better protect from a potential null reference. I didn't notice any other problems, but if more issues arise in regards to PR #2135 then just notify me and I'll take a look at them too.

### DIFF
--- a/megamek/src/megamek/client/Client.java
+++ b/megamek/src/megamek/client/Client.java
@@ -1151,7 +1151,7 @@ public class Client implements IClientCommandHandler {
     private void cacheImgTag(Entity entity){
 
         if(entity == null) {
-            System.out.println("Null entity reference in cacheImgTag()");
+            MegaMek.getLogger().error(this, "Null entity reference");
             return;
         }
 
--- a/megamek/src/megamek/client/Client.java
+++ b/megamek/src/megamek/client/Client.java
@@ -1184,14 +1184,12 @@ public class Client implements IClientCommandHandler {
      * Gets the current mech image
      */
     private Image getTargetImage(Entity e){
-        if(bv == null){
-            return  null;
-        }else {
-            if (e.isDestroyed()) {
-                return bv.getTilesetManager().wreckMarkerFor(e, -1);
-            } else {
-                return bv.getTilesetManager().imageFor(e);
-            }
+        if (bv == null) {
+            return null;
+        } else if (e.isDestroyed()) {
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/client/Client.java
```diff
@@ -1151,33 +1151,29 @@ private String getCachedImgTag(int id){
...
-                    ImageIO.write((RenderedImage) image, "png", baos);
-                    baos.flush();
-                    base64Text = Base64.getEncoder().encodeToString(baos.toByteArray());
-                    baos.close();
-                    String img = "<img src='data:image/png;base64," + base64Text + "'>";
-                    imgCache.put(entity.getId(), img);
-                } catch (final IOException ioe) {
-                    throw new UncheckedIOException(ioe);
-                }
+        if(getTargetImage(entity) != null){//convert image to base64, add to to <img> tag and store in cache
```
Windchild292, line 1164:
```suggestion
        if (getTargetImage(entity) != null) {
            //convert image to base64, add to to <img> tag and store in cache
```

megamek/src/megamek/client/Client.java
```diff
@@ -1186,10 +1182,14 @@ private void cacheImgTag(Entity entity){
...
+        if(bv == null){
+            return  null;
         }else {
-            return bv.getTilesetManager().imageFor(e);
+            if (e.isDestroyed()) {
+                return bv.getTilesetManager().wreckMarkerFor(e, -1);
+            } else {
+                return bv.getTilesetManager().imageFor(e);
+            }
         }
```
Windchild292, line 1193:
```
if (bv == null) {
    return null;
} else if (e.isDestroyed()) {
    return bv.getTilesetManager().wreckMarkerFor(e, -1);
} else {
    return bv.getTilesetManager().imageFor(e);
}
```

Windchild292: Suggested rewrite to improve readability.

Krashner: Apparently I'm sloppy when it's early. I appreciate the suggestions.

Windchild292: It happens to us all, and welcome :)

PR #2158, opened 2020-08-18
Keyboard Shortcuts Overlay

SJuliez:
Adds an overlay to the board showing various keyboard shortcuts. I didnt even know that "Next Valid Target (No Allies)" existed, so I figure other people might not know either. The overlay has a menu item and a shortcut (Ctrl-K) to toggle it and it will remember when it's turned off.
The shortcuts shown change with the phase and active player.

Updated the IDisplayable interface with default methods and comments. 

I seem to have committed a change to the default keybinds xml therefore this is currently draft.

Movement Phase:
image
Firing Phase:
image
Other player's turn:
image
Menu item:
image


HammerGS:
Hey, do any of these change existing keybindings? If so we should make sure to update the text file.

SJuliez:
No changes to the keybinds.

PR #2170, opened 2020-08-23
MekHQ 1892: MegaMek Standardization of Launch Game Dialogs

Windchild292:
This PR standardizes and reduces repeated code by turning the Launch/Connect to Game dialogs into inheriting a single method that has a verification method within.

This is the MegaMek-side of fixing https://github.com/MegaMek/mekhq/issues/1892 and standardizing the dialogs between the two programs.


megamek/src/megamek/client/ui/swing/MegaMekGUI.java
```diff
@@ -559,68 +533,48 @@ public String getDescription() {
...
-                    ex.getMessage() + ").";
-            JOptionPane
-                    .showMessageDialog(
-                            frame,
-                            error,
-                            Messages.getString("MegaMek.HostGameAlert.title"), JOptionPane.ERROR_MESSAGE); //$NON-NLS-1$
+            MegaMek.getLogger().error(this, "could not create server socket on port " + hd.getPort());
+            JOptionPane.showMessageDialog(frame,
+                    "Error: could not start server at localhost" + ":" + hd.getPort() + " (" + ex.getMessage() + ").",
+                    Messages.getString("MegaMek.HostGameAlert.title"), JOptionPane.ERROR_MESSAGE);
```
SJuliez, line 556: Is it intentional not to log the ex.getMessage()? Otherwise, the String error of line 596 could be re-used in 553 and 555

megamek/src/megamek/client/ui/swing/gameConnectionDialogs/ConnectDialog.java
```diff
@@ -0,0 +1,108 @@
...
+        
+        add(middlePanel, BorderLayout.CENTER);
+        
+        createButtons();
+        
+        pack();
+        setResizable(false);
+        center();
+    }
+    //endregion Initialization
```
SJuliez, line 79: My feeling would be (not because it's important right here, more as a matter principle) that InitComponents might rather be 
"protected JPanel createMainPanel() { ... }"
Everything from "add(middlePanel..." to center() (lines 71-77) would then move to the AbstractGCD and in the subclass it would just "return middlePanel;" instead. In this way the subclass can create whatever it wants in that panel and just return it and it does not have to "remember" doing anything (e.g. to createButtons()).
Of course, this forces createbuttons() on the subclasses but I suppose that's the buttons we need for a game conection dialog...
I mean this purely as a suggestion.

SJuliez (approved):
This is OK by me. Just 2 comments/thoughts.

megamek/src/megamek/client/ui/swing/gameConnectionDialogs/ConnectDialog.java
Windchild292: I concur, that's better code. Implement.

megamek/src/megamek/client/ui/swing/MegaMekGUI.java
Windchild292: Ended up finding a bunch of localization issues in this, so I've fixed that now.

Windchild292:
@SJuliez 
This is ready for review with the previous suggestions implemented.

megamek/src/megamek/client/ui/Messages.java
```diff
@@ -53,4 +53,12 @@ public static String getString(String key) {
...
         return MessageFormat.format(getString(key), args);
     }
+
+    /**
+     * Returns the formatted internationalized text for the given key in the resource bundle,
+     * replacing occurrences of %s in the message with the contents of args.
+     */
+    public static String getFormattedString(String key, Object... args) {
+        return String.format(getString(key), args);
+    }
```
SJuliez, line 63: So far this was done using "message=SomeError {0}: {1}". The call to Messages.getString looks the same in both cases I think. Is there an advantage to the %s version?

SJuliez (approved):
Just one little question, otherwise all OK.

megamek/src/megamek/client/ui/Messages.java
Windchild292: This does not format numbers, which is necessary for ports.... and I apparently forgot to put that in the comment.

PR #2174, opened 2020-08-24
write out kills to mul result file

NickAragua:
Companion to https://github.com/MegaMek/mekhq/pull/1964 - writes out external ID to kill credit in summary MUL file as long as it exists.

### DIFF
--- a/megamek/src/megamek/common/EntityListFile.java
+++ b/megamek/src/megamek/common/EntityListFile.java
@@ -105,7 +105,7 @@ public class EntityListFile {
                 if (mount.getEntity().usesWeaponBays() 
                         || mount.getEntity() instanceof Dropship) {
                     output.append("\" capacity=\"")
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/EntityListFile.java
```diff
@@ -105,7 +105,7 @@ private static String formatSlot(String index, Mounted mount,
                 if (mount.getEntity().usesWeaponBays() 
                         || mount.getEntity() instanceof Dropship) {
                     output.append("\" capacity=\"")
-                        .append(String.valueOf(mount.getAmmoCapacity()));
+                        .append(String.valueOf(mount.getSize()));
```
sixlettervariables, line 108: This is related?

NickAragua: No, but it's listed as deprecated so I figured I'd change it while I was there.

PR #2179, opened 2020-08-26
MegaMek 2176:  Add Quad Type Turrets for Most Weapons

### LINKED ISSUE #2176: RFE:  Some weapon types seem to be missing quad turret types
For example, IS LB-X turrets are all missing a quad variant, as are Clan LB-X turrets.
Clan HAG and some Gauss types also seem to be missing quad variants.

Is there an in-game reason for this, or is it an oversight?

If it's an oversight, what are the limitations of sizing and ammo?  I would be willing to work on a PR to add them.

mhjacks:
Fixes #2176 
This standardizes all of the weapon types that have triple turrets to also have Quad turrets, across IS and Clan tech families.  (Some had quads, some didn't before.)  I also added chemlaser ammo to the Clan chem laser turrets (since they need that.)

I can load MegaMek and see the units show up.  (The icons are wrong, I'd like to know how to fix that.)  I have not yet attempted to add these units to the RATs but that's a definite enhancement I could work on if you want that prior to merge


NickAragua:
To get the right icons you have to modify data\images\units\mechset.txt

I'm not sure that there are images for all of those you added, so you'll either have to draw some yourself or improvise.

Additionally, given my notes on the issue re: maximum tonnage of weapons (from TacOps page 266), do we want to limit these things to 15 tons of guns?

HammerGS:
That was one of the reasons some of these turrets where limited. Others were just cause.

mhjacks:
Yeah, but that's kind of arbitrary, isn't it?  The Quad Arrow IV is way over, so is the dual Gauss, and frankly most of the triples.  I was considering just re-using the Triple icon where it was available

mhjacks:
I edited the previous comment quite a bit (Sorry, I tend to either write a lot of small comments or edit one to death), to make sure the notification triggers fire:

I think the relevant section of TacOps says: "Heavy Weapons: Only Fortress-type Mobile Structures may
mount Heavy weapons (weapons that individually weigh half a
ton or more). The maximum tonnage of such weapons that may
be mounted per hex—discounting ammunition, turret mecha-
nisms and heat sinks or power amplifiers—is equal to the Fortress’
CF divided by 10. The weight of these weapons, plus any turrets,
power amplifiers, heat sinks or ammunition, counts against the
structure’s weight capacity as per normal equipment."

Note that the structure record sheet envisions multi-hex, multi-level structures. So why not have some quad HAG/40 or Imp Heavy Gauss turrets?

The fortress can be multi-hex; it appears that there's not a per-hex weight limit but there is an overall limit based on the CF of the fortress. In the case of Forward Base, the whole fortress could be several thousand CF, couldn't it? I agree it's kind of...excessive in some cases. (7 turret slots on Forward base means I could within megemak mount 28 HAG/40s). I will stipulate that is abuse of the system.

But as it is, some of the heaviest systems (Arrow IV) gets quad options, and things like plasma rifles (or, say, MML 3's) don't.

Another factor here is that megameklab doesn't edit gun emplacements. (Since they're defined in terms of "the whole building"). So for players that don't want to crack open the source code and hand-edit .blk files this is the easiest way to give them those turrets.

HammerGS:
We don't have buildings coded per the rules in Megamek. When we made expanded the turrets a few years ago it was to replace the 10 odd turrets we had. 

The choice of size and type of equipment was a mix of rules and what seemed cool. Lots of quad and things that felt over powered where left out. Cause a quad mount Imp. Heavy Gauss isn't going to be fun for a lot of players on the receiving end of that. 

I think some of the turrets in this PR make sense others like the over powered ones I'd rather see the tech level set at unofficial.

mhjacks:
I completely agree about some of these things being overpowered.  Especially with the way BV is calculated, some of these are completely destabilizing (the Quad Imp Heavy Gauss is 847 BV!) from a pure BV standpoint.  The Gausszilla Annihilator was supposed to be a joke...and where there's one turret, there can be two, etc.  My group is planning on doing "mowing down Princess" things with this, but I definitely see how this could rapidly drain the joy from PvP play if used unscrupulously.

So, I kind of like the idea of making some of these unofficial.  Are there objective criteria we can agree on for where that threshold should be?  Like, damage per round or something?  But that would require re-classifying a number of units that are already included, and that would probably require RAT work too.  (One could argue that these new units might not belong in the RATs, in general.)

Also, icon-wise, my art skills are...lacking.  I could specify the dual version (implying the weapons are "stacked") where those exist.  Would that work for now?

NickAragua:
I'd say anything that breaks the 15 tons of guns limit should be unofficial. 

I think dual versions are fine for quads for now.

mhjacks:
ACK on dual versions of icons.  Thanks!

I hear you, but I think that limitation would make too many turrets "unofficial".  I don't think the "15" limit is real - some guns by themselves are more than 15 tons (our friend the IHGR for example).  The limitation TacOps states is 10% of the mobile fortress's weight by total CF (which we have no way to calculate in megamek now since we can't represent an aggregate structure).  150 CF is not a limit TacOps places on a Mobile Fortress structure.  As a practical matter, I think even the Calliope is > 15 tons with guns and gear, right?

`<Guns Equipment>
ISERLargeLaser
ISERLargeLaser
ISMediumPulseLaser
ISMediumPulseLaser
ISLRM5
ISLRM5
ISLRM5 Ammo
ISLRM5 Ammo
ISLRM5 Ammo
ISLRM5 Ammo
ISLRM5 Ammo
ISLRM5 Ammo
</Guns Equipment>
`
Many of the existing turrets have ~10 tons of Ammo (at least) and their single versions have guns > 5 tons.

Would it work to make all of the turrets being added in this PR Unofficial and we leave anything existing alone?  (Also the "new" turrets would not be added to the RATs; should I be concerned about the interaction of mixed tech level for different cardinality turrets and RATs?

NickAragua:
Yeah, I think sticking to making the turrets in this PR unofficial is fine. No need to rock the boat for the other stuff. Go ahead and add the clan/artemis turrets if you want. 

Ammo doesn't count towards the 15 ton limit, it's literally only weapons (not even Artemis, just weapons). But anyway, now that I read it again, I think that you could successfully argue that it's the Fortress' total CF/10 rather than individual hex CF/10. We'd probably need a rules clarification on it, but can probably hold off on that until someone actually starts implementing turret construction rules.

Let's hold off the RAT modifications for another PR.

mhjacks:
OK, fair enough on all counts.  I'll keep working through this and will push more commits later.  Thanks!

mhjacks:
Ready for review, I believe.  No errors loading and I eyeballed the new icons in the add unit dialog.

PR #2189, opened 2020-08-30
Restore mapfiles from SimonLandmine

mhjacks:
I just unzipped the map-pack where it seems to belong in the source tree.  If this approach is acceptable, I can proceed by restoring the other map packs (Derv's Map Pack 2 and deadboarder).


HammerGS:
I said I would handle this. I don’t need a PR.

PR #2205, opened 2020-09-09
Sync boards with mekhq

mhjacks:
Copied boards directory from current MekHQ head.


HammerGS:
As I’ve said before. I don’t need a PR I will handle this.

PR #2213, opened 2020-09-12
L1 Foliage

### LINKED ISSUE #1719: RFE: Level 1 Foliage
Would it be possible to add "Level 1 Foliage" as a terrain option (as per TO, p36) at some point?

(It would better represent hedgerows, bocage, etc., as well as allowing 'mechs to tower dramatically over the scenery while vehicles and infantry skulk below.)

### LINKED ISSUE #2212: 47.9 Possible error in Diagramming LOS calculation
Standalone MM 47.9
Windows 10
Java 1.8.0_261

Receiving a "LOS blocked by terrain" calculation, when I believe that the Rampage should be able to see the Jenner.

image

image

Rampage at 1108 is on a Level 1 hex, behind a Level 2 parapet. I believe it should be able to see (and be seen by) the Jenner in hex 1809, as there is only one hex of Light Woods between them, other than the parapet.

[QuerySave_2020-09-11_14-45-00.sav.gz](https://github.com/MegaMek/megamek/files/5212000/QuerySave_2020-09-11_14-45-00.sav.gz)
(Player name: SimonLandmine - other 'players' are Princess 'role behaviour' instances)

[megameklog.txt](https://github.com/MegaMek/megamek/files/5212003/megameklog.txt)

SJuliez:
Implements Level 1 Foliage of TO p.36. I called them "low woods" and "low jungle"

- Mechs or anything else that counts as being 2 levels high for LOS (e.g. VTOLs at elev 1) in or behind L1F receive no bonus (TO says Mechs never receive a partial cover bonus. Since a mech is a full level higher than the woods I suppose they don't receive any bonus from standing in it or standing behind it)
- The Board Editor got two new buttons to place low woods and jungle. 
- The terrain is represented with a new terrain type called foliage_elev that gives the elevation of the woods with the allowed values of 1, 2 for L/H and 3 for U woods/jungles. 
- This new terrain is required but the board loader adds it with regular foliage height when it's not present
- The boardview shows "LOW" in any hex with L1F. I think its difficult to represent this purely with gfx, and in any case I didnt have any new gfx. I created a separate image for the standard grass Saxarba light and heavy woods but no more (for now).

Notes:
- Atmospheric map altitude of low foliage is 1 like with regular foliage
- Adjusted the TFs of jungles to the values given in TO, 10 more than woods
- Made woods and jungles in the same hex invalid
- I'm not sure if special chainsaw attacks on woods are in MM. In any case, I didnt change anything in that direction.
- Resolves #2212 
- Resolves #1719 

LowWoods


HammerGS:
This is great. We’ve had lots of requests for this over the years.

NickAragua:
We really gotta get some art for low woods/jungle so that we don't have to have LOW in green letters written on it. We already have enough text labels on those hexes.

megamek/src/megamek/common/Hex.java
```diff
@@ -631,82 +631,102 @@ public boolean isClearHex() {
...
+            int wl = terrainLevel(Terrains.WOODS);
+            int jl = terrainLevel(Terrains.JUNGLE);
+            int el = terrainLevel(Terrains.FOLIAGE_ELEV);
+            
+            boolean isLightOrHeavy = wl == 1 || jl == 1 || wl == 2 || jl == 2;
+            boolean isUltra = wl == 3 || jl == 3;
+            
+            if (! ((el == 1) || (isLightOrHeavy && el == 2) || (isUltra && el == 3))) {
+                valid = false;
+                errBuff.append("Foliage elevation is wrong, must be 1 or 2 for Light/Heavy and 1 or 3 for Ultra Woods/Jungle!\n");
```
NickAragua, line 689: I'm confused by this.

NickAragua (approved):
Looks fine at first glance, other than one question I have re: "low" ultra-heavy terrain. I think it will require some heavy testing, and a good subsequent feature is to introduce the possibility of low terrain to the map generator.

megamek/src/megamek/common/Hex.java
SJuliez: The elevation can always be 1 (regardless of light, heavy or ultra), it can also be 2 but only in light or heavy woods/jungle and in ultra woods/jungle 3 is an allowed value. Not quite sure how to answer ...

SJuliez:
Re: map generator: I understand, many people play AtB and they won't ever see low foliage without the map generator creating it...
Re: gfx and "LOW": I'd love to cut the LOW text, but how would one paint the foliage height from 90° above in a way that is obvious  enough to really not mark these woods at all? I mean so obvious that people don't end up parking their mechs "behind" a thick tree clump and end up getting shot down ;)

NickAragua:
The map generator thing is definitely for a subsequent PR, as for the "LOW" thing... are there any actual table top maps out there that have low terrain?

HammerGS:
> The map generator thing is definitely for a subsequent PR, as for the "LOW" thing... are there any actual table top maps out there that have low terrain?

Not yet, but the Tukayyid map pack will have a bunch of them with level 1.

NickAragua:
We can probably revisit the issue then (unless they just put "LOW" on all their low terrain)

megamek/src/megamek/common/Hex.java
NickAragua: Ok, I just didn't realize that ultra terrain is either 1 or 3. Kind of weird, but it is what it is.

HammerGS:
I have it on good authority CGL is going to use "Foliage" on the hexes to indicate level. We could just use that instead of "Low"

SJuliez:
Eh, how do I get past those boards validator results? Or can I ignore them? The big Valley_sample cant be opened at all currently because its higher than 100 hexes, so I could only delete it. And I guess the Church/ map isnt supposed to be changed right now...

sixlettervariables:
@sjuliez you're only responsible for boards that broke as part of this PR. If they were already broken it isn't your issue to solve (but you could).

SJuliez:
@sixlettervariables Thanks!
Keeping track...
- [ ] Change the BoardEditor buttons to read "foliage"
- [ ] Add Foliage to the RandomMapGen
- [ ] Repair boards
- [ ] Test if the Server agrees with the client on LOS issues

megamek/src/megamek/common/LosEffects.java
```diff
@@ -1099,35 +1099,44 @@ private static LosEffects losForCoords(IGame game, AttackInfo ai,
...
+        // The interpolated elevation for TacOps LOS diagramming
+        double losElevation = (ai.targetAbsHeight * ai.attackPos.distance(coords) 
+                + ai.attackAbsHeight * ai.targetPos.distance(coords))
+                    / (ai.targetPos.distance(coords) + ai.attackPos.distance(coords))
+                    + 1;
+        
+        // The higher of the attacker's height and defender's height
+        int maxUnitHeight = Math.max(ai.attackAbsHeight, ai.targetAbsHeight);
+        boolean attackerAdjc = ai.attackPos.distance(coords) == 1;
+        boolean targetAdjc = ai.attackPos.distance(coords) == 1;
```
NickAragua, line 1127: Did you mean to have this be equivalent to "attackerAdjc"? Also, what does Adjc stand for?

NickAragua (changes requested):
One minor adjustment needs to be made to "target adjacent(?)" calculation logic, but looks good otherwise.

megamek/src/megamek/common/LosEffects.java
Windchild292: Adjacent would be my bet

SJuliez: Good you caught that! ai.targetPos indeed - wonder if I would have found that during testing.
And yes it's adjacent. Wanted to abbreviate it. Started with "adj", found it reads like adjustment, therefore the -c

SJuliez:
OK Finished from my side... Currently foliage will not show up as the default foliage settings for the random map generator is 0. I would leave that to the AtB masters to maybe change the current map gens or add new variants to the scenario generator(? I have not played AtB recently). Or I can change the default so that part of the woods can show up as foliage, no problem. Note that foliage and woods are additive, so setting both to "High" will create quite the forest.

NickAragua (approved):
Think we're good to go then.

PR #2224, opened 2020-09-17
Logging: Creating Static Logging Methods, Deprecating Old Logging Methods, and Removing Unused Methods

Windchild292:
To review: whitespace differences off.


Windchild292:
The alert is a false positive because I put in commented out code that will handle the last version of error logging but cannot uncomment it because of an overlap with an existing method.

SJuliez (approved):
All OK. I couldn't see the differences in FactionRecord (if there were any) as it is shown as completely replaced.

Windchild292:
The difference is I swapped a single logger call to use this so i could get rid of one of the older versions. I also fixed the tab spacing, which is why it shows up like that.

PR #2225, opened 2020-09-19
Fixing RandomNameGenerator Historical Ethnicity Non-Specified Generation Title

Windchild292:
This PR contains refactoring as part of modifying the MekHQ GM tools, and some older minor changes to files as part of my MekHQ dev work

### DIFF
--- a/megamek/src/megamek/common/util/sorter/NaturalOrderComparator.java
+++ b/megamek/src/megamek/common/util/sorter/NaturalOrderComparator.java
@@ -18,7 +18,6 @@
  */
 package megamek.common.util.sorter;
 
-import java.io.Serializable;
 import java.text.Collator;
 import java.util.Comparator;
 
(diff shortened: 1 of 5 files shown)


megamek/src/megamek/common/util/sorter/NaturalOrderComparator.java
```diff
@@ -36,7 +37,8 @@
  * would be sorted into { "Atlas 0", "Atlas 1", "Atlas 2", "Atlas 5", "Atlas 15" }
  * instead of ASCII's { "Atlas 0", "Atlas 1", "Atlas 15", "Atlas 2", "Atlas 5" }
  */
-public class NaturalOrderComparator implements Comparator<String> {
+public class NaturalOrderComparator implements Comparator<String>, Serializable {
+    private static final long serialVersionUID = -5116813198443091269L;
```
NickAragua, line 41: Why do we need this to be serializable?

Windchild292: This is preventative...  From IDEA's Inspection Reasoning: "Reports on subclasses of java.lang.Comparator which are not also Serializable. Objects of java.util.TreeMap or java.util.TreeSet will become non-Serializable if instantiated with such Comparators. This can result in unexpected and difficult-to-diagnose bugs. Since subclasses of java.lang.Comparator are often stateless, simply marking them Serializable is a small cost to avoid such issues."

NickAragua (approved):

megamek/src/megamek/common/util/sorter/NaturalOrderComparator.java
NickAragua: That certainly is a strange behavior.

PR #2226, opened 2020-09-20
on phase change, update gun emplacements

NickAragua:
Because gun emplacements don't actually take damage, there's no 'entity update' when they take a crit during weapons fire. They do get updated during a phase change, so we'll refresh the sprite at that point to avoid extra network chattiness (it's possible that an emplacement gets crit during the movement phase, but fairly unlikely).

### DIFF
--- a/megamek/src/megamek/client/ui/swing/boardview/BoardView1.java
+++ b/megamek/src/megamek/client/ui/swing/boardview/BoardView1.java
@@ -5356,8 +5356,8 @@ public class BoardView1 extends JPanel implements IBoardView, Scrollable,
                 default:
             }
             for (Entity en: game.getEntitiesVector()) {
-                if (en.getDamageLevel() != Entity.DMG_NONE && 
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/client/ui/swing/boardview/BoardView1.java
```diff
@@ -5355,7 +5356,8 @@ public void gamePhaseChange(GamePhaseChangeEvent e) {
                 default:
             }
             for (Entity en: game.getEntitiesVector()) {
-                if (en.getDamageLevel() != Entity.DMG_NONE && en.damageThisRound != 0) {
+                if (en.getDamageLevel() != Entity.DMG_NONE && 
+                        (en.damageThisRound != 0 || en instanceof GunEmplacement)) {
```
Windchild292, line 5360:
```suggestion
                if ((en.getDamageLevel() != Entity.DMG_NONE) && 
                        ((en.damageThisRound != 0) || (en instanceof GunEmplacement))) {
```

Windchild292 (approved):

PR #2236, opened 2020-09-23
MekHQ 2044: Fixing Omni Aero TRO

Windchild292:
This fixes https://github.com/MegaMek/mekhq/issues/2044 , where all omni aerospace fighters fail to display their TRO.


neoancient (changes requested):
This change doesn't correct the issue.
Screenshot from 2020-09-23 12-54-09

NickAragua (commented):
What was the stuff that got removed?

neoancient:
The problem is that the fixed equipment rows are missing the attack values. Instead of adding the data, this just leaves the cells blank.

Windchild292:
MekHQ-side this does fix the issue of all omnifighters not displaying the TRO, just not the deeper issue. I've received word this is required for @RexPearce and thus works as a temporary fix for his testing. I'll open up an additional issue to fix the actual underlying issue, but would prefer to merge the overlying fix now @neoancient.

neoancient (approved):

PR #2238, opened 2020-09-23
Board editor fixes (#2233,#1762)

### LINKED ISSUE #1762: 0.47.4 Map Editor - CTRL+scroll change to Bridge class
When doing 'CTRL+Scroll' over the bridge icon to change the Bridge Class value, if you've used 'ALT+click' to select an existing bridge, you lose the 'Exits' value for the existing bridge.

### LINKED ISSUE #2233: Map Editor - new terrain picks up 'exits' setting from the 'Terrain Settings' window
MM 0.47.10 and earlier
Windows 10
Java 8

When a new terrain type is selected by clicking on the WYSIWYG buttons at the top of the right-hand menu, it inherits any 'exits' value that is currently defined in the 'Terrain Settings' fields (e.g. from previously configuring a Bridge), even if exits are irrelevant to the terrain type (e.g. the Light Woods in the screenshot below).

This seems to happen with most terrain types (e.g. Roads, Woods, Mud, Swamp, Pavement, Water), except for Buildings, Fuel Tanks, and, ironically given this example, Bridges.

The simplest solution might be to, when a new terrain type is selected via a WYSIWYG button, the 'exits' checkbox is unchecked (except when ALT-clicking Buildings to enforce single-hex construction).

image

SJuliez:
Resolves #2233 
Resolves #1762 
... and various small issues.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/BoardEditor.java
+++ b/megamek/src/megamek/client/ui/swing/BoardEditor.java
@@ -1571,13 +1571,13 @@ public class BoardEditor extends JComponent
                 return false;
             }
         }
-        butSourceFile.setEnabled(true);
 
         // write the board
-        try {
-            OutputStream os = new FileOutputStream(curBoardFile);
+        try (OutputStream os = new FileOutputStream(curBoardFile)) {
             board.save(os);
-            os.close();
+            
+            // Adapt to successful save
+            butSourceFile.setEnabled(true);
             savedUndoStackSize = undoStack.size();
             hasChanges = false;
             setFrameTitle();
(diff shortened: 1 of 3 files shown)


megamek/src/megamek/client/ui/swing/BoardEditor.java
```diff
@@ -1622,47 +1554,69 @@ private void boardSaveImage(boolean ignoreUnits) {
...
+        if (curBoardFile == null || saveAs) {
+            if (!chooseSaveBoardFile()) {
+                return false;
+            }
+        }
+        butSourceFile.setEnabled(true);
+
+        // write the board
+        try {
+            OutputStream os = new FileOutputStream(curBoardFile);
```
sixlettervariables, line 1578: This should be moved into the try statement itself (`try (...)`) so if board.save throws, `os` will be cleaned up.

sixlettervariables (commented):
Small improvement requested otherwise looks good!

megamek/src/megamek/client/ui/swing/BoardEditor.java
SJuliez: Done.

Windchild292 (approved):

PR #2239, opened 2020-09-23
Fix units getting stuck when jumping into water and getting hull down result

### LINKED ISSUE #705: [43.4] Unable to get up from 'hull down' in depth 1 water
[hulldownbug.sav.gz](https://github.com/MegaMek/megamek/files/1391204/hulldownbug.sav.gz)

No errors in the log. 

Select the "Hull down" Wasp and attempt to get up. He *should* be able to, having 6 walking MP, but he's not.

NickAragua:
Finally fixes #705 

When a unit goes hull down, it's set as 'hull down', but none of the other fall behaviors occur - including elevation adjustment for sinking to the bottom. This PR addresses the issue.

### DIFF
--- a/megamek/src/megamek/server/ServerHelper.java
+++ b/megamek/src/megamek/server/ServerHelper.java
@@ -395,4 +395,22 @@ public class ServerHelper {
             vPhaseReport.addAll(s.destroyEntity(entity, "pilot death", true));
         }
     }
+    
+    /**
+     * Helper function that causes an entity to sink to the bottom of the water
+     * hex it's currently in.
+     */
+    public static void sinkToBottom(Entity entity) {
+        IHex fallHex = entity.getGame().getBoard().getHex(entity.getPosition());
+        int waterDepth = 0;
+        
+        // we're going hull down, we still sink to the bottom if appropriate
+        if (fallHex.containsTerrain(Terrains.WATER)) {
+            // *Only* use this if there actually is water in the hex, otherwise
+            // we get ITerrain.LEVEL_NONE, i.e. Integer.minValue...
+            waterDepth = fallHex.terrainLevel(Terrains.WATER);
+        }
+        
+        entity.setElevation(-waterDepth);
+    }
 }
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/server/ServerHelper.java
```diff
@@ -395,4 +395,22 @@ else if ((entity.heat >= 14) && !entity.isShutDown()) {
             vPhaseReport.addAll(s.destroyEntity(entity, "pilot death", true));
         }
     }
+    
+    /**
+     * Helper function that causes an entity to sink to the bottom of the water
+     * hex it's currently in.
+     */
+    public static void sinkToBottom(Entity entity) {
+        IHex fallHex = entity.getGame().getBoard().getHex(entity.getPosition());
```
sixlettervariables, line 404: Based on our historical bug reports you should probably return early if entity or fallHex is null.

megamek/src/megamek/server/ServerHelper.java
```diff
@@ -395,4 +395,22 @@ else if ((entity.heat >= 14) && !entity.isShutDown()) {
...
+        int waterDepth = 0;
+        
+        // we're going hull down, we still sink to the bottom if appropriate
+        if (fallHex.containsTerrain(Terrains.WATER)) {
+            // *Only* use this if there actually is water in the hex, otherwise
+            // we get ITerrain.LEVEL_NONE, i.e. Integer.minValue...
+            waterDepth = fallHex.terrainLevel(Terrains.WATER);
+        }
+        
+        entity.setElevation(-waterDepth);
```
sixlettervariables, line 414: Is this always appropriate? Should they always be set to elevation 0 if not in water, or should this just be skipped?

NickAragua: I think I better limit it to just in water, and also it occurs to me I forgot to check for the entity being on top of a bridge.

sixlettervariables (approved):

Issue #1093, opened 2018-08-11
[0.45.0 : Bug] Hover Tank Dropping onto Liquid Magma Not Destroyed


SirMegaV:
### Environment
0.45.0

### Description
Hover tank dropping onto liquid magma from magma crust not destroyed. Image below. (But it's hovering........)

### Files
![hovertankcannotmoveaboveliquidmagma-2](https://user-images.githubusercontent.com/17064665/43987492-1e996944-9d54-11e8-9120-7f1aaa42b821.png)

BLOODWOLF333:
It looks like you mixed mine and ColSandersLite's tilesets together.

SirMegaV:
Nope. I’m using your tileset then modify to use mine.

PR #2248, opened 2020-09-25
update magma rules to match tacops

NickAragua:
Makes the following changes to current megamek behavior regarding magma (TacOps page 36):

- hovercraft do not collapse magma crust, nor is the user nagged about it any longer
- jumping onto magma crust breaks through on 4+, added appropriate nag
- added nag for jumping into liquid magma
- consolidated "magma crust breaking" logic

Addresses #1093, although it's a little bit better for hovertanks.

Only thing I'm not sure about is how jumping + hover movement interplays here - if, let's say, a Kanga were to jump onto a lava crust, does it have a chance to break through?


megamek/i18n/megamek/common/report-messages.properties
```diff
@@ -195,6 +195,7 @@
 2388=Have a nice trip!
 2390=<data> is landing on top of <data>!
 2395=<newline><data> (<data>) breaks through magma crust on a 6, rolls <data>.
```
sixlettervariables, line 197: Noticed this line doesn't match it's nag. Is it a 6+ or a 6?

sixlettervariables (approved):
One comment about related-ish messages.

megamek/i18n/megamek/common/report-messages.properties
NickAragua: It's a 1d6, so there's no +

sixlettervariables: The nag states 6+

NickAragua: Ah, I see. Updated.

PR #2252, opened 2020-09-26
Board Loading fix

### LINKED ISSUE #1898: Boards of sizes xx:yyy don't load

SJuliez:
Changes board loading to allow loading xx * yyy boards (2 digit width and 3 digit height). This is done by ignoring the coordinates that are written in the board file and assuming the hexes appear in the order they are written by the board editor. I suppose it's not a big loss that manually changing coords in a board file won't have any effect anymore.

While at it, added in the Logger and moved a few streams to the try () statement.

Resolves #1898

### DIFF
--- a/megamek/src/megamek/common/Board.java
+++ b/megamek/src/megamek/common/Board.java
@@ -374,7 +374,7 @@ public class Board implements Serializable, IBoard {
                             // Log the error and remove the
                             // building from the board.
                             if (errBuff == null) {
-                                System.err.println("Unable to create building.");
+                                MegaMek.getLogger().error(this, "Unable to create building.");
                                 excep.printStackTrace();
                             } else {
                                 errBuff.append("Unable to create building at " + coords.toString() + "!\n");
@@ -404,7 +404,7 @@ public class Board implements Serializable, IBoard {
                             // Log the error and remove the
                             // fuel tank from the board.
                             if (errBuff == null) {
-                                System.err.println("Unable to create fuel tank.");
+                                MegaMek.getLogger().error(this, "Unable to create fuel tank.");
                                 excep.printStackTrace();
                             } else {
                                 errBuff.append("Unable to create fuel tank at " + coords.toString() + "!\n");
@@ -434,7 +434,7 @@ public class Board implements Serializable, IBoard {
                             // Log the error and remove the
                             // bridge from the board.
                             if (errBuff == null) {
-                                System.err.println("Unable to create bridge.");
+                                MegaMek.getLogger().error(this, "Unable to create bridge.");
                                 excep.printStackTrace();
                             } else {
                                 errBuff.append("Unable to create bridge at " + coords.toString() + "!\n");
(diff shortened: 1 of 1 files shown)


sixlettervariables (commented):
Does this break the `note` keyword?

megamek/src/megamek/common/Board.java
```diff
@@ -956,16 +951,10 @@ public void load(final String filename) {
...
-            // tell the board to load!
+        try (InputStream is = new FileInputStream(filepath)) {
             load(is);
-            // okay, done!
-            is.close();
-        } catch (java.io.IOException ex) {
-            System.err.println("error opening file to load board!");
-            System.err.println(ex);
+        } catch (IOException ex) {
+            MegaMek.getLogger().error(this, "IO Error opening file to load board! " + ex);
```
Windchild292, line 957: Curious why you decided to append the ex instead of have it as `MegaMek.getLogger().error(this, "IO Error opening file to load board!", ex);`

megamek/src/megamek/common/Board.java
```diff
@@ -1093,10 +1075,10 @@ public void load(InputStream is, StringBuffer errBuff, boolean continueLoadOnErr
         if (isValid(nd, nw, nh, errBuff) && ((nw > 1) || (nh > 1) || (di == (nw * nh)))) {
             newData(nw, nh, nd, errBuff);
         } else if (continueLoadOnError && ((nw > 1) || (nh > 1) || (di == (nw * nh)))) {
-            System.err.println("Error reading board, data invalid.");
+            MegaMek.getLogger().error(this, "Invalid board data!");
             newData(nw, nh, nd, errBuff);
         } else if (errBuff == null){
-            System.err.println("Error reading board, data invalid.");
+            MegaMek.getLogger().error(this, "Invalid board data!");
```
Windchild292, line 1081: Same error message twice in the same method without a way to tell them apart, which could cause issues in the future.

Windchild292 (commented):
Two comments about logging.

megamek/src/megamek/common/Board.java
SJuliez: I just replicated what was done before without much thought. Didn't even notice that ex was the exception, not a message... I'll change it.

SJuliez:
@sixlettervariables No, notes must bring their own coords and there's no interaction with the hex ordering. But I guess from the code that they won't work well on large boards either. But Im not sure what to make of them, apparently they're not called up anywhere currently and they don't seem to be the same as the notes I can place on the map in-game.

HammerGS:
@SJuliez Comment for you on Slack regarding this one. Just seeing if it's ready for todays release? Thanks.

Windchild292 (approved):

PR #2255, opened 2020-09-27
don't apply vibra bomb damage to moving unit not in hex

### LINKED ISSUE #1978: Vibrabomb Minefield deals damage with the wrong timing.
### Environment
Testing on MM46.1

### Description
Place a mech in a hex with a Vibrabomb, or use Thunder Vibrabombs to deploy a minefield in the mechs hex.  Set the vibrabomb to have a detection threshold of 20, so it will activate.

Move the mech out of the vibrabomb minefield.  The mech will take damage.

This is incorrect.

There are several places in tactical operations where it is stated that a unit leaving a hex that is mined does not set off the mines.  For example on pg 373

In addition the example for Vibrabomb mine usage on pg 209 says that the check to see if the vibramine activates happens when a mech enters a hex.

Since in my example above the mech that is activating the mines is the one that just entered another hex entirely, it can not be in the hex it just left to be hit by the mines.

NickAragua:
Basically, a unit moving out of a vibrabomb hex *can* still detonate the vibrabomb field if it's heavier than the field's tonnage setting by at least 10 tons, but it shouldn't take any damage.

Fixes #1978

### DIFF
--- a/megamek/src/megamek/server/Server.java
+++ b/megamek/src/megamek/server/Server.java
@@ -11582,7 +11582,7 @@ public class Server implements Runnable {
                 // if the moving entity is not actually moving into the vibrabomb
                 // hex, it won't get damaged
                 Integer excludeEntityID = null;
-                if(!coords.equals(mf.getCoords())) {
+                if (!coords.equals(mf.getCoords())) {
                     excludeEntityID = entity.getId();
                 }
                     
@@ -11757,7 +11757,7 @@ public class Server implements Runnable {
             if (entity.isAirborne()) {
                 continue;
             }
-            
+
             // check for the OptionsConstants.ADVGRNDMOV_NO_PREMOVE_VIBRA option
             // If it's set, and the target has not yet moved,
             // it doesn't get damaged.
@@ -11771,7 +11771,7 @@ public class Server implements Runnable {
             }
             
             // the "currently moving entity" may not be in the same hex, so it needs to be excluded
-            if((entityToExclude != null) && (entity.getId() == entityToExclude)) {
+            if ((entityToExclude != null) && (entity.getId() == entityToExclude)) {
                 // report not hitting vibrabomb
                 r = new Report(2157);
                 r.subject = entity.getId();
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/server/Server.java
```diff
@@ -11578,7 +11578,15 @@ private boolean checkVibrabombs(Entity entity, Coords coords, boolean displaced,
                 r.add(entity.getShortName(), true);
                 r.add(mf.getCoords().getBoardNum(), true);
                 vMineReport.add(r);
-                explodeVibrabomb(mf, vMineReport, false);
+                
+                // if the moving entity is not actually moving into the vibrabomb
+                // hex, it won't get damaged
+                Integer excludeEntityID = null;
+                if(!coords.equals(mf.getCoords())) {
```
Windchild292, line 11585:
```suggestion
                if (!coords.equals(mf.getCoords())) {
```

megamek/src/megamek/server/Server.java
```diff
@@ -11749,7 +11757,7 @@ private void explodeVibrabomb(Minefield mf, Vector<Report> vBoomReport, boolean
             if (entity.isAirborne()) {
                 continue;
             }
-
+
```
Windchild292, line 11760:
```suggestion

```

megamek/src/megamek/server/Server.java
```diff
@@ -11761,12 +11769,23 @@ private void explodeVibrabomb(Minefield mf, Vector<Report> vBoomReport, boolean
...
             }
-            // report hitting vibrabomb
-            r = new Report(2160);
-            r.subject = entity.getId();
-            r.add(entity.getShortName(), true);
-            vBoomReport.add(r);
-
+            
+            // the "currently moving entity" may not be in the same hex, so it needs to be excluded
+            if((entityToExclude != null) && (entity.getId() == entityToExclude)) {
```
Windchild292, line 11774:
```suggestion
            if ((entityToExclude != null) && (entity.getId() == entityToExclude)) {
```

Windchild292 (approved):
Looks good bar three formatting issues.

PR #2256, opened 2020-09-27
fix minefield clearance eligibility; bulletproof attack handling

### LINKED ISSUE #1813: Clearing minefields with weapon in 0.47.4
Linux
openjdk version "1.8.0_242"
OpenJDK Runtime Environment (build 1.8.0_242-b08)
OpenJDK 64-Bit Server VM (build 25.242-b08, mixed mode)

In MegaMek 0.46.1 everything is OK:

https://i.imgur.com/SIy98Ya.png

Is OK, because I can't clear mine with Medium Laser but I can with LRM-20.

But in MegaMek 0.47.4

https://i.imgur.com/Tf8wPp8.png

I can fire using bad weapon and if I do this java goes to 100% CPU and stays there even after closing the client window.
[megameklog-0.47.4-clearing-minefield-with-bad-weapon.txt](https://github.com/MegaMek/megamek/files/4424405/megameklog-0.47.4-clearing-minefield-with-bad-weapon.txt)

NickAragua:
Fixes #1813 

"can clear minefield" was checking ammo types, but lasers don't have ammo, so they were reporting as being able to clear minefields.

Doing that was causing the WeaponAttackHandler to go into an endless loop as the number of hits would never get reduced. Added a fallback case to reduce the number of hits anyway and log it, to at least avoid going into an endless loop.

### DIFF
--- a/megamek/src/megamek/common/weapons/WeaponHandler.java
+++ b/megamek/src/megamek/common/weapons/WeaponHandler.java
@@ -1173,33 +1173,34 @@ public class WeaponHandler implements AttackHandler, Serializable {
                         r.subject = subjectId;
                         vPhaseReport.addElement(r);
                         hits = 0;
-                    }
-                    // targeting a hex for igniting
-                    if ((target.getTargetType() == Targetable.TYPE_HEX_IGNITE)
+                    // targeting a hex for igniting    
+                    } else if ((target.getTargetType() == Targetable.TYPE_HEX_IGNITE)
                             || (target.getTargetType() == Targetable.TYPE_BLDG_IGNITE)) {
                         handleIgnitionDamage(vPhaseReport, bldg, hits);
                         hits = 0;
-                    }
                     // targeting a hex for clearing
-                    if (target.getTargetType() == Targetable.TYPE_HEX_CLEAR) {
+                    } else if (target.getTargetType() == Targetable.TYPE_HEX_CLEAR) {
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/common/weapons/WeaponHandler.java
```diff
@@ -1173,33 +1173,34 @@ public boolean handle(IGame.Phase phase, Vector<Report> returnedReports) {
...
                         handleEntityDamage(entityTarget, vPhaseReport, bldg,
                                 hits, nCluster, bldgAbsorbs);
                         server.creditKill(entityTarget, ae);
                         hits -= nCluster;
                         firstHit = false;
+                    } else {
+                        // we shouldn't be here, but if we get here, let's set hits to 0
+                        // to avoid infinite loops
+                        hits = 0;
+                        getLogger().error(getClass(), "Unexpected target type: " + target.getTargetType());
```
Windchild292, line 1203:
```suggestion
                        getLogger().error(this, "Unexpected target type: " + target.getTargetType());
```

Windchild292 (approved):

megamek/src/megamek/common/weapons/WeaponHandler.java
NickAragua: Under the hood, the 'error' method just calls 'getClass()' anyway, so I don't see why this is necessary.

PR #2263, opened 2020-09-28
disable bot name field when editing existing bot

### LINKED ISSUE #2171: MM 0.47.9 - Can break game in Lobby by using Bot Settings right-click
Standalone MegaMek 47.9
Windows10
Java 1.8.0_261

In the Lobby, create a bot player.
Assign some 'meks to that bot player.
Then, in the player list on the left, right-click the bot player and click 'Bot Settings...' to amend the bot.
It is possible to change the bot name at this point, but if you do so and click [Okay], the bot will become unselectable (cannot change team or start point, nor edit 'Bot Settings...'), and any units that have been allocated to that bot will be locked to that phantom bot, and cannot be transferred to another player.

(Does this mean that the player name is used as an ID key value, rather than simply being a label attached to a hidden player ID?)

NickAragua:
Fixes #2171 by disabling the option to edit the bot name if configuring an existing bot, as the "proper" fix is a lot more involved.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/BotConfigDialog.java
+++ b/megamek/src/megamek/client/ui/swing/BotConfigDialog.java
@@ -101,10 +101,11 @@ public class BotConfigDialog extends JDialog implements ActionListener, KeyListe
     public BotConfigDialog(JFrame parent, BotClient existingBot) {
         this(parent, new HashSet<>());
         
+        nameField.setEnabled(false);
+        
         if (existingBot instanceof Princess) {
             this.princessBehavior = ((Princess) existingBot).getBehaviorSettings();
             nameField.setText(existingBot.getName());
-            nameField.setEnabled(false);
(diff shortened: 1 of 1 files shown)


Windchild292:
I know most people no longer use it, but what about test bot? Further, given this is a bandaid, we should create an issue once this is done for the deeper fix

NickAragua:
Good point about TestBot. I'll disable the name editing there too.

NickAragua:
Disabled name editing for both Princess and TestBot.

Windchild292 (approved):

Windchild292:
I do think we should open an issue for the proper fix in the future, if possible.

NickAragua:
It occurs to me that this will mess up the ability to edit bot names in MekHQ. Going to need to bake it a little more.

NickAragua:
This will be made redundant shortly.

Issue #2273, opened 2020-09-28
Aerospace and Conventional Fighters change BV on save/load


RexPearce:
Confirmed on versions: 0.46.1, 0.47.10, and current master.

Steps to reproduce:
 1) start a new campaign
 2) add a conventional fighter, e.g. Guardian Fighter (Standard)
 3) add an aero fighter, e.g. Transit TR-9
 4) Note their BVs
 5) save/load the campaign
 6) Note their BVs have changed.

Doesn't seem to be related to the pilot - assigning a regular pilot the save/loading has no effect on BV.

neoancient:
Capital fighters weapon groups combine left and right wing weapons into a single wing location. That's what the WNG is for. It doesn't represent an actual physical location. The question is where is it being marked destroyed?

sixlettervariables:
MekHQ uses the location string logic from MM which checks the armor amount being `<= 0` in a location and if so it's destroyed. It uses all 6 locations for Aeros which includes the Wings and Fuselage.

SJuliez:
From the linked PRs I assume this is fixed.

PR #2274, opened 2020-09-29
Address instability in BV calcs for Aeros #2273

sixlettervariables:
This attempts to address the instability in BV calcs for Aeros seen in #2273. I'm definitely new to this area of the code base, but it looks like the BV calcs were not updated when LOC_WINGS or LOC_FUSELAGE was added (not sure which). It also appears that the location string details were not updated either. I updated both to detect these as "pseudo" locations not appropriate for armor, but otherwise appropriate for equipment. This may not be the best means of resolving this issue.


neoancient (approved):
I completely missed seeing that you had a fix for this. I was planning looking at it today.

Issue #2215, opened 2020-09-12
RasterFormatException: y lies outside raster


HammerGS:
Playing around with SNAPSHOT built Sept 11th.

### Environment
What version of MegaMek does your issue apply to?  Megamek
What operating system are you using?  Windows, 

### Description
Load a game, grab a couple of mechs and change the map to Beginner Box Desert 1. 

Let the fun begin

### Files
[megameklog.txt](https://github.com/MegaMek/megamek/files/5213400/megameklog.txt)

From the log
```
boardview1: loading images for board
Loading atlas: data\images\hexes\saxarba\theme_mars\theme_mars_atlas.png
Loading atlas: data\images\hexes\saxarba\theme_mars\theme_mars_atlas.png
Loading atlas: data\images\hexes\saxarba\theme_mars\theme_mars_atlas.png
Loading atlas: data\images\hexes\saxarba\theme_mars\theme_mars_atlas.png
Loading atlas: data\images\hexes\saxarba\theme_mars\theme_mars_atlas.png
Loading atlas: data\images\hexes\saxarba\theme_mars\theme_mars_atlas.png
Loading atlas: data\images\hexes\saxarba\theme_mars\theme_mars_atlas.png
Loading atlas: data\images\hexes\saxarba\theme_mars\theme_mars_atlas.png
Loading atlas: data\images\hexes\saxarba\theme_mars\theme_mars_atlas.png
Loading atlas: data\images\hexes\saxarba\theme_mars\theme_mars_atlas.png
Loading atlas: data\images\hexes\saxarba\theme_mars\theme_mars_atlas.png
Loading atlas: data\images\hexes\hexes_atlas.png
Loading atlas: data\images\hexes\transparent\transparent_atlas.png
Loading atlas: data\images\hexes\transparent\transparent_atlas.png
Loading atlas: data\images\hexes\hexes_atlas.png
Loading atlas: data\images\hexes\hexes_atlas.png
Loading atlas: data\images\hexes\hexes_atlas.png
Exception in thread "AWT-EventQueue-0" java.awt.image.RasterFormatException: y lies outside raster
	at java.desktop/sun.awt.image.IntegerInterleavedRaster.createWritableChild(IntegerInterleavedRaster.java:460)
	at java.desktop/java.awt.image.BufferedImage.getSubimage(BufferedImage.java:1202)
	at megamek.client.ui.swing.boardview.BoardView1.paintComponent(BoardView1.java:1262)
[...]

PR #2275, opened 2020-09-30
expand boundary checking in background image drawing

NickAragua:
Explicit fix for #2215

Log entries after exception handling show the following error:

13:39:15,692 ERROR [megamek.client.ui.swing.boardview.BoardView1] {AWT-EventQueue-0} 
paintComponent() : Error drawing background image. Raster Bounds: 0.00, 0.00, width:447.00, height:447.00, Attempted Draw Coordinates: 0, -1, width:447, height:448

So the rendering code goes into the block starting at line 1262 because -1 is technically not equal to 0. But it getSubImage doesn't like being passed negative numbers for coordinates, so we'll avoid going there in more cases. Leaving the try/catch block in case some other geometry error crops up.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/boardview/BoardView1.java
+++ b/megamek/src/megamek/client/ui/swing/boardview/BoardView1.java
@@ -1259,7 +1259,7 @@ public class BoardView1 extends JPanel implements IBoardView, Scrollable,
                     if (x == 0) {
                         xRem = clipping.x % w;
                     }
-                    if (xRem >= 0 || yRem >= 0) {
+                    if (xRem > 0 || yRem > 0) {
                         try {
                             g.drawImage(
                                     bvBgImage.getSubimage(xRem, yRem, w - xRem,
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/client/ui/swing/boardview/BoardView1.java
```diff
@@ -1259,7 +1259,7 @@ public synchronized void paintComponent(Graphics g) {
                     if (x == 0) {
                         xRem = clipping.x % w;
                     }
-                    if (xRem != 0 || yRem != 0) {
+                    if (xRem > 0 || yRem > 0) {
```
Windchild292, line 1262:
```suggestion
                    if ((xRem > 0) || (yRem > 0)) {
```

Windchild292 (approved):

megamek/src/megamek/client/ui/swing/boardview/BoardView1.java
Windchild292: This one felt weird to suggest, but it is part of our code style guide

PR #2287, opened 2020-10-01
isCrippled/isDmgLight/isDmgModerate/isDmgHeavy Debug Logging

Windchild292:
This adds some more debug logging for these methods in Aero and Tank, which I've found is required for testing.


megamek/src/megamek/common/GunEmplacement.java
```diff
@@ -612,18 +594,18 @@ public void setDeployed(boolean deployed) {
...
         
         // we'll consider undeployed emplacements to be fully intact
-        if((occupiedStructure == null) || (initialBuildingCF + initialBuildingArmor == 0)) {
+        if ((occupiedStructure == null) || (initialBuildingCF + initialBuildingArmor == 0)) {
             return 1.0;
         }
         
-        return (occupiedStructure.getCurrentCF(getPosition()) + occupiedStructure.getArmor(getPosition())) / 
-                (initialBuildingCF + initialBuildingArmor);
+        return (1.0 * (occupiedStructure.getCurrentCF(getPosition()) + occupiedStructure.getArmor(getPosition())))
```
sixlettervariables, line 608: Is this a bug fix?

Windchild292: It is, it was trying to return a double from an int/int.

sixlettervariables: More obvious to cast the denominator to double. I presume this means that gun emplacements never showed the correct amount of armor remaining? (always 0)

Windchild292: Swapped, and I believe so. This was a catch from my IDE.

sixlettervariables (approved):
Looks good, sorry I missed this update!

Windchild292:
No worries at all.

PR #2293, opened 2020-10-03
Small fixed wing support vehicles implementation

neoancient:
This should finish the implementation of the aerospace small support vehicles - currently only fixed wing, but airships and satellites in the future. There are two basic areas covered:
1. Set up ammo bins for any weapons. This is done for ground vehicles in BlkFile, but I failed to check whether BLKFixedWIngSupportFile overrides loadEquipment.
2. Provide aerospace stats to infantry weapons. This includes determining AV at each range band and calculating the maximum range band for the weapon based on the infantryRange. Infantry weapons don't use the standard S/M/L values, but instead have a single value. Small support vehicle weapon range is 1x/2x/3x this value, for a maximum range of infantryRange * 3.

There were several places in the code that use the maxRange field directly rather than using the accessor method, which bypasses the different calculation for infantry weapons. I fixed those and changed the visibility of all the aerotech fields to protected.


megamek/src/megamek/common/weapons/infantry/InfantryWeaponHandler.java
```diff
@@ -157,12 +157,15 @@ protected int calcnClusterAero(Entity entityTarget) {
...
+        int av;
         //Sigh, another rules oversight - nobody bothered to figure this out
         //To be consistent with other cluster weapons we will assume 60% hit
-        if ((ae instanceof Infantry) && !(ae instanceof BattleArmor)) {
+        if (ae.isConventionalInfantry()) {
             double damage = ((Infantry) ae).getDamagePerTrooper();
             av = (int) Math.round(damage * 0.6 * ((Infantry) ae).getShootingStrength());
+        } else {
+            // Small fixed wing support
+            av = super.calcAttackValue();
```
NickAragua, line 168: I hate to ask, but what about battle armor?

NickAragua (approved):
Looks fine, just one question.

megamek/src/megamek/common/weapons/infantry/InfantryWeaponHandler.java
neoancient: I'm not really sure what that code is about, though I have some ideas. It's beyond the scope of what I'm working on so I made it more concise but otherwise left it alone.

NickAragua: Ah, I think I see what you're saying - that check before was explicitly "is this infantry but not battle armor".

PR #2294, opened 2020-10-04
Flag for filtering small SV equipment

neoancient:
Adds a flag to MiscType that is used for flagging equipment as illegal to small support vehicles. This comes from the following rule clarification:
https://bg.battletech.com/forums/techmanual/answered-small-support-vehicles-and-special-equipment/

Basically anything on the heavy equipment tables in TM is forbidden, as is any C3 equipment or anything that has a  damage value in the other books. I also tagged equipment that small support vehicles cannot use for indirect reasons, namely Artemis, PPC capacitors, etc. (because they can't use the equipment it modifies). This helps MML filter the equipment list.

### DIFF
--- a/megamek/src/megamek/common/MiscType.java
+++ b/megamek/src/megamek/common/MiscType.java
@@ -270,13 +270,14 @@ public class MiscType extends EquipmentType {
     public static final BigInteger F_CAPITAL_ARMOR = BigInteger.valueOf(1).shiftLeft(209);
     public static final BigInteger F_FERRO_CARBIDE = BigInteger.valueOf(1).shiftLeft(210);
     public static final BigInteger F_IMP_FERRO = BigInteger.valueOf(1).shiftLeft(211);
+    // Not usable by small support vehicles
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/common/MiscType.java
```diff
@@ -270,13 +270,14 @@
     public static final BigInteger F_CAPITAL_ARMOR = BigInteger.valueOf(1).shiftLeft(209);
     public static final BigInteger F_FERRO_CARBIDE = BigInteger.valueOf(1).shiftLeft(210);
     public static final BigInteger F_IMP_FERRO = BigInteger.valueOf(1).shiftLeft(211);
+    // Not usable by small support vehicles
+    public static final BigInteger F_HEAVY_EQUIPMENT = BigInteger.valueOf(1).shiftLeft(212);
```
NickAragua, line 274: Question, is this going to affect saves (especially mekhq ones)?

neoancient: No. The only part of MiscType that gets serialized anywhere is te equipment lookup name.

NickAragua (approved):

PR #2297, opened 2020-10-04
Icon Rework Phase 2: MegaMek Static Directory Manager

Windchild292:
This is phase 2 of my icon standardization. I merged the CamoManager and PortraitManager, then added the MechTileset to it. I then implemented the name swapover. This is inherited by the MekHQ-side, and cleans up a few duplicated file loads along the way.


megamek/src/megamek/client/ui/swing/tileset/MMStaticDirectoryManager.java
```diff
@@ -0,0 +1,537 @@
...
+        initializeCamouflage();
+        initializeMechTileset();
+    }
+
+    /**
+     * Parses MM's portraits folder when first called or when it was refreshed.
+     *
+     * @see #refreshPortraitDirectory()
+     */
+    private static void initializePortraits() {
```
sixlettervariables, line 73: Should these initializers should be synchronized? (If so, not `initialize` itself, to avoid deadlocks).

sixlettervariables (commented):
Just one comment.

megamek/src/megamek/client/ui/swing/tileset/MMStaticDirectoryManager.java
Windchild292: There haven't been any issues with it as of yet, but that may become an issue in future. Maybe @SJuliez can comment on why it wasn't done that way originally?

SJuliez: I didn't think about it. I think MM uses only few extra threads, for the RATs, boardview refresh and pathfinding. None of these will request an image. So perhaps we are safe now...

SJuliez (approved):
Looks good to me. Such a long name, MMStaticDirectoryManager. Maybe MMImageManager? Or icon instead of image? Don't know, just a thought for later...

PR #2298, opened 2020-10-04
VTOL pathfinding and movement corrections

### LINKED ISSUE #2280: Odd behaviour from barnstorming VTOLS. Bug?
0.47.10 standalone
Windows 10
Java 11.0.8

Having been making a new board featuring a canyon with a high bridge above it, I was testing the feasibility of flying a VTOL down the canyon like an old _Airwolf_ episode (but with the rather less impressive _Ferret_). But when it came to flying under the bridge, something strange happened. I re-tested a few times to make sure, and it happened every time.

The canyon is 6 levels deep, and the bridge is height 6, stretching from one side to the other. The VTOL is at elevation 3, so nicely in the middle of the canyon.

When you attempt to move the VTOL under the bridge, it seems to force you to carry out a Controlled Sideslip under the bridge, instead (more dramatic, but a waste of MP). The pathfinder slews the VTOL sideways by one hex, sideslips under the bridge, and then straightens up again to continue the path.

You can see this in the attached screenshot. What should have been a simple 3 MP forward move costs 6 MP, as the VTOL moves forward one hex, turns right, sideslips, turns left again, then moves another hex.

20200930-1806-VTOL+Bridge

Is there any reason for this behaviour?

So that this can be replicated, as I'm not sure how many existing boards feature gratuitously high bridges, I've attached the board (which was in location /data/boards/SimonLandmine/64x68/ while I was working on it), and a saved game with the VTOL ready to show off. Player name = 'SimonLandmine'.

[64x68 DesertCanyonRiver.zip](https://github.com/MegaMek/megamek/files/5307235/64x68.DesertCanyonRiver.zip)
[Test_20200930-1801.sav.gz](https://github.com/MegaMek/megamek/files/5307237/Test_20200930-1801.sav.gz)

(I also tested to see whether it afflicted WiGEs, but they were fine.)

SJuliez:
Some love for VTOL movement:
- Makes the Go Up and Go Down buttons behave when over and under bridges
- Allows going under bridges without acrobatics
- Decouples the terrain checks for woods, buildings, bridges and so on so that they can be used in the same hex without error. Having buildings or woods in water should also work better than before. Bridges under water are not allowed tho.
- The pathfinder will now provide paths that scale hills, terrain and buildings when necessary (and able). Limited this function to VTOL movement tho. It doesn't seem to have much use for infantry in buildings. I tested it for submarines but this will make them surface when the player might not really want them to. I think the depth of a sub should be directly controlled by the player even if thats a little more effort.

image
image

Resolves #2280

### DIFF
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -2242,6 +2242,7 @@ public abstract class Entity extends TurnOrdered implements Transporter,
             case VTOL:
                 maxAlt = hex.surface() + 50;
                 // When under a bridge, restrict upward movement
+                // "- 1" to correct that height() reports one less than the rules (TW p.99) say
                 if (hex.containsTerrain(Terrains.BRIDGE_ELEV)
                         && assumedElevation < hex.terrainLevel(Terrains.BRIDGE_ELEV)) {
                     maxAlt = hex.terrainLevel(Terrains.BRIDGE_ELEV) - height() - 1;   
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -2129,15 +2129,9 @@ public abstract class Entity extends TurnOrdered implements Transporter,
         if (!getGame().getBoard().contains(assumedPos)) {
             return false;
         }
-        boolean inWaterOrWoods = false;
         IHex hex = getGame().getBoard().getHex(assumedPos);
         int assumedAlt = assumedElevation + hex.surface();
         int minAlt = hex.surface();
-        if (hex.containsTerrain(Terrains.WOODS)
-            || hex.containsTerrain(Terrains.WATER)
-            || hex.containsTerrain(Terrains.JUNGLE)) {
-            inWaterOrWoods = true;
-        }
         switch (getMovementMode()) {
             case INF_JUMP:
             case INF_LEG:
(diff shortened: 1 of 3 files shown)


megamek/src/megamek/common/Entity.java
```diff
@@ -2233,6 +2241,11 @@ public boolean canGoUp(int assumedElevation, Coords assumedPos) {
                 break;
             case VTOL:
                 maxAlt = hex.surface() + 50;
+                // When under a bridge, restrict upward movement
+                if (hex.containsTerrain(Terrains.BRIDGE_ELEV)
+                        && assumedElevation < hex.terrainLevel(Terrains.BRIDGE_ELEV)) {
+                    maxAlt = hex.terrainLevel(Terrains.BRIDGE_ELEV) - height() - 1;
```
sixlettervariables, line 2247: If this takes into account `height()` being one less, as below, can you include that note here too?

megamek/src/megamek/common/Entity.java
```diff
@@ -2287,23 +2300,34 @@ public boolean isElevationValid(int assumedElevation, IHex hex) {
...
+                // VTOLs can be anywhere on or above ground and fly beneath bridges,
+                // land on buildings and ignore planted fields and industrial zone 
+                // but cannot land on water or trees
+                // As always, height() reports one less than the rules (TW p.99) say
+                // Units may move under a bridge if their top is 
+                // lower than or equal to the bridge height (TW p.62)
+                boolean allowed = (assumedElevation <= 50) && (assumedElevation >= 0);
+                if (hex.containsTerrain(Terrains.BRIDGE_ELEV)) {
+                    allowed &= (assumedElevation >= hex.terrainLevel(Terrains.BRIDGE_ELEV))
+                            || (assumedElevation + height() + 1 <= hex.terrainLevel(Terrains.BRIDGE_ELEV));
```
sixlettervariables, line 2318: Is `<=` proper for "under" a bridge?

megamek/src/megamek/common/pathfinder/PathDecorator.java
```diff
@@ -160,37 +162,58 @@
     }
     
     /**
-     * Given a path, adds "UP" steps to it so that a forward movement can pass over intervening terrain
+     * For units using VTOL movement, add "UP" steps to the end of the MovePath source 
+     * so that a forward movement can pass over intervening terrain
      */
     public static void AdjustElevationForForwardMovement(MovePath source) {
+        // Do this only for VTOLs
+        if (!(source.getEntity().getMovementMode() == EntityMovementMode.VTOL)) {
```
sixlettervariables, line 170:
```suggestion
        if (source.getEntity().getMovementMode() != EntityMovementMode.VTOL) {
```

sixlettervariables (commented):
Few questions, but mostly due to my lack of experience with VTOLs in our code.

megamek/src/megamek/common/Terrain.java
```diff
@@ -637,6 +637,10 @@ public boolean isValid(StringBuffer errBuff) {
             rv = false;
         } else if (type == Terrains.FOLIAGE_ELEV && (level < 1 || level > 3)) {
             rv = false;
+        } else if (type == Terrains.BLDG_ELEV && level < 1) {
+            rv = false;
+        } else if (type == Terrains.BRIDGE_ELEV && level < 0) {
```
Windchild292, line 642:
```suggestion
        } else if ((type == Terrains.BLDG_ELEV) && (level < 1)) {
            rv = false;
        } else if ((type == Terrains.BRIDGE_ELEV) && (level < 0)) {
```

megamek/src/megamek/common/Entity.java
SJuliez: TW says a unit may move under a bridge so long as the height of the unit is less than or equal to the bridge level. I.E. a mech fits under a level 2 bridge. A heli, elevation 1 and height 1 (0 in MM for whatever reason), should also fit under a bridge. I think (hope) this one is correct.

SJuliez:
Suggestions are in.

megamek/src/megamek/common/Entity.java
SJuliez: Done.

sixlettervariables (approved):

PR #2307, opened 2020-10-05
Fix cost calculation for tank and conventional fighter heat sinks and power amplifiers

neoancient:
The cost calculation for Tank does not include all equipment that can affect the cost of power amplifiers and heat sinks. Since there are methods elsewhere that perform these calculations I used them instead. For heat sink calculations I added a static method to TestEntity to avoid having to create an instance that involves the additional overhead of parsing the options file.

Since the turret weight is calculated in the same loop, I added some fixes to that as well. A PPC capacitor is included in the weight of a turret mounted PPC, and small support vehicle turrets should be rounded to the kg and not the half ton.

Fixes MegaMek/megameklab#762

### DIFF
--- a/megamek/src/megamek/common/verifier/TestEntity.java
+++ b/megamek/src/megamek/common/verifier/TestEntity.java
@@ -870,7 +870,7 @@ public abstract class TestEntity implements TestEntityOption {
             // Spot welders are treated as energy weapons on units that don't have a fusion or fission engine
             if (m.getType().hasFlag(MiscType.F_CLUB) && m.getType().hasSubType(MiscType.S_SPOT_WELDER)
                 && entity.hasEngine() && (entity.getEngine().isFusion()
-                                                || entity.getEngine().getEngineType() == Engine.FISSION)) {
+                                                || (entity.getEngine().getEngineType() == Engine.FISSION))) {
                 continue;
             }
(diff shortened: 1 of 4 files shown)


megamek/src/megamek/common/verifier/TestEntity.java
```diff
@@ -855,16 +866,16 @@ protected int heatNeutralHSRequirement() {
...
             }
         }
-        for (Mounted m : getEntity().getMisc()) {
+        for (Mounted m : entity.getMisc()) {
             // Spot welders are treated as energy weapons on units that don't have a fusion or fission engine
             if (m.getType().hasFlag(MiscType.F_CLUB) && m.getType().hasSubType(MiscType.S_SPOT_WELDER)
-                && getEntity().hasEngine() && (getEntity().getEngine().isFusion()
-                                                || getEntity().getEngine().getEngineType() == Engine.FISSION)) {
+                && entity.hasEngine() && (entity.getEngine().isFusion()
+                                                || entity.getEngine().getEngineType() == Engine.FISSION)) {
```
Windchild292, line 873:
```suggestion
                                                || (entity.getEngine().getEngineType() == Engine.FISSION))) {
```

Windchild292 (approved):

PR #2308, opened 2020-10-05
Normalizes Data Files

rjhancock:
All this PR does is normalize all Unit files to the following.

1) (BLK) ##'s and d#'s were converted to # .
2) All line endings converted from CRLF -> LF
3) All file encodings were enforced to UTF-8.

These were issues discovered while working on an importer of data into another project.


sixlettervariables (changes requested):
Can you put up a gist or something with the tool or script used? Or if possible include it in the project. We certainly will need to run this in the future.

rjhancock:
A simple ruby script that is easily modified for other file types.

https://gist.github.com/rjhancock/eccbee2203148b1b1954cc5b168ddc1b

sixlettervariables:
What handled normalizing the comments in the file?

rjhancock:
VSCode Regex search and replace.

`^.#`

sixlettervariables (approved):
I approve based on the scripts used to update the files. Leaving the final Okay to @HammerGS.

PR #2310, opened 2020-10-06
Implement Static Ammo Feed Quirk

NickAragua:
What it says on the tin. The Static Ammo Feed quirk prevents ammo-based weapons from switching to a different ammo type. 

Also make Princess respect both caseless and static ammo feed ammo switching restrictions when switching weapons.

Includes unit test update to account for parameter changes.


sixlettervariables (approved):
Grab the two Spurious Javadoc @param tags, but otherwise looks good!

PR #2311, opened 2020-10-06
Properly locating the RandomCallsignGenerator

Windchild292:


sixlettervariables (approved):
Honestly I like generators better, but When in Rome(tm)!

Windchild292:
Same, but there would be far more changes for that one.

PR #2315, opened 2020-10-07
fix bot counterbattery NPE

NickAragua:
Fixes an observed NPE when the bot was targeting off-board units for counterbattery fire with on-board artillery. We do this by introducing a number of null checks, and short-circuiting some computations.

Also improves exception logging, as it was only logging the exception and not even the line number.

### DIFF
--- a/megamek/src/megamek/client/bot/BotClient.java
+++ b/megamek/src/megamek/client/bot/BotClient.java
@@ -588,7 +588,14 @@ public abstract class BotClient extends Client {
             
             return true;
         } catch (Throwable t) {
-            MegaMek.getLogger().error(t.toString());
+            StringBuilder sb = new StringBuilder();
+            sb.append(t.toString()).append("\r\n");
+           
+            for (StackTraceElement traceElement : t.getStackTrace()) {
+                sb.append(traceElement.toString());
+                
+            }
+            MegaMek.getLogger().error(sb.toString());
             
             return false;
         }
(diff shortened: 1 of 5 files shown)


megamek/src/megamek/client/bot/BotClient.java
```diff
@@ -588,7 +588,14 @@ private synchronized boolean calculateMyTurnWorker() {
...
         } catch (Throwable t) {
-            MegaMek.getLogger().error(t.toString());
+            StringBuilder sb = new StringBuilder();
+            sb.append(t.toString()).append("\r\n");
+           
+            for (StackTraceElement traceElement : t.getStackTrace()) {
+                sb.append(traceElement.toString());
+                
+            }
+            MegaMek.getLogger().error(sb.toString());
```
sixlettervariables, line 598: Can we not just pass `t` to error?

sixlettervariables (approved):
One comment, but otherwise approved.

sixlettervariables:
It would be nice if instead of `null` we had an "offboard" hex we could use.

megamek/src/megamek/client/bot/BotClient.java
NickAragua: Yep. Much easier.

NickAragua:
Yeah, although I'm reluctant to make that kind of change, as there's probably a good amount of logic that depends on the offboard hex being null when it's offboard.

PR #2324, opened 2020-10-08
Fix TRO exception

neoancient:
Adds a check for location before attempting to get the armor type for that location. That this point in the code, items in LOC_NONE have been skipped. The robotic control system and the DNI are exceptions.

Fixes MegaMek/megameklab#777


NickAragua (approved):

NickAragua:
From a long term perspective, maybe we want to introduce an ARMOR_NONE or INVALID_LOCATION constant or something similar, and have getArmorType return that if the passed-in location is invalid, as I've seen quite a few bugs involving that.

PR #2329, opened 2020-10-10
MM Logging: Removing Unused Methods

Windchild292:
This removes any deprecated logging methods currently unused in the suite


SJuliez (approved):
Yay Cleanup!

PR #2334, opened 2020-10-11
Housekeeping

### LINKED ISSUE #2259: [RFE] Color code minimap based on team, not player
This is a feature request, not a bug.

When playing team games with large numbers of players (eg:8) it would be useful to have an option to configure the minimap so that each team gets a color (eg:red and blue) as opposed to each player's color appearing.

SJuliez:
Even youtubers sometimes don't seem to configure anything in MM which makes me think that the MM default config should be improved already. This and other things:

- Replaced the unit label border teamcolor sub-option with a general team coloring option that also applies to the minimap. Also changed the ally color to more of a blue. While player coloring may have applications, I believe teamcoloring has no downside as a default option and is therefore on by default. Resolves #2259 
- Removed the option from the Client Settings to not show a hex tooltip in hexes without units (i.e., never show the terrain). This prevents a player from seeing a bridge's or building CF, armor or any other complicated terrain with (I think) no remedy but to set this precise option in the Client Settings. Not useful on any but the simplest maps.
- Set the tooltip to show weapons, armor and the portrait by default. These are all very useful (or nice in the case of portraits) and players shouldn't have to first turn them on.
- Added the keybind to toggle the Keybinds Board Overlay directly into the header text to clarify that this can be easily toggled and made it more opaque to improve readability.
- The Keybinds Board Overlay will now show some relevant keybinds in the Board Editor
- Set the "Darken Map at night" option to default off. I noticed with a youtuber that this doesn't read well ("Oh this is a blue map set") and it's probably more annoying than helpful after a while (i.e., 10 seconds).
- Set the terrain shadows to default on. People don't see to have many issues with them and I do think they improve the overall look.
- Removed the option "Minimap can be shown" from the Client Settings. Instead the minimap will be shown and re-shown when it was on before and not be opened by MM again if it was closed by the player.
- Set "Mouse Wheel Zoom" to default true, as zooming with the wheel instead of scrolling is standard game behavior.
- Set "Flip mouse wheel" to default true as zooming in with wheel up is also standard game behavior.
- The minimap now respects the "Flip mouse wheel" setting.
- Added a keybind for toggling hex coords, Ctrl-G. Some games such as Civ use this shortcut to show a grid; we dont have a grid, but coords are the next best thing.
- Added a menu shortcut for the Client Settings, Alt-C. Should have done this a million clicks before.
- In the Client Settings, set the font of the advanced settings value (the text field) to the SANS_SERIF logical font which will make the unicode characters appear there correctly (at least on my system). 
- Made the Client Settings and Boardview use PreferenceChangeEvents to show changes directly instead of unnecessary boardview calls.
- Made the Client Settings actually revert changes when it's cancelled.

image image

### DIFF
--- a/megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
+++ b/megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
@@ -315,6 +315,7 @@ public class CommonSettingsDialog extends ClientDialog implements
     private boolean savedTeamColoring;
     private boolean savedUnitLabelBorder;
     private boolean savedShowDamageDecal;
+    private boolean savedShowDamageLabel;
     private String savedFovHighlightRingsRadii;
     private String savedFovHighlightRingsColors;
     private int savedFovHighlightAlpha;
@@ -359,7 +360,6 @@ public class CommonSettingsDialog extends ClientDialog implements
         settingsPane.getVerticalScrollBar().setUnitIncrement(16);
         panTabs.add("Main", settingsPane);
         
-
         JScrollPane graphicsPane = new JScrollPane(getGraphicsPanel());
         graphicsPane.getVerticalScrollBar().setUnitIncrement(16);
         panTabs.add("Graphics", graphicsPane);
@@ -427,6 +427,7 @@ public class CommonSettingsDialog extends ClientDialog implements
         // --------------        
 
         showDamageLevel = new JCheckBox(Messages.getString("CommonSettingsDialog.showDamageLevel")); //$NON-NLS-1$
+        showDamageLevel.addItemListener(this);
         row = new ArrayList<>();
         row.add(showDamageLevel);
         comps.add(row);
@@ -912,6 +913,7 @@ public class CommonSettingsDialog extends ClientDialog implements
             savedTeamColoring = gs.getTeamColoring();
             savedUnitLabelBorder = gs.getUnitLabelBorder();
             savedShowDamageDecal = gs.getShowDamageDecal();
+            savedShowDamageLabel = gs.getShowDamageLevel();
             savedFovHighlightRingsRadii = gs.getFovHighlightRingsRadii();
             savedFovHighlightRingsColors = gs.getFovHighlightRingsColorsHsb();
             savedFovHighlightAlpha = gs.getFovHighlightAlpha();
@@ -941,6 +943,7 @@ public class CommonSettingsDialog extends ClientDialog implements
         guip.setTeamColoring(savedTeamColoring);
         guip.setUnitLabelBorder(savedUnitLabelBorder);
         guip.setShowDamageDecal(savedShowDamageDecal);
+        guip.setShowDamageLevel(savedShowDamageLabel);
         guip.setFovHighlightRingsRadii(savedFovHighlightRingsRadii);
         guip.setFovHighlightRingsColorsHsb(savedFovHighlightRingsColors);
         guip.setFovHighlightAlpha(savedFovHighlightAlpha);
@@ -1263,6 +1266,8 @@ public class CommonSettingsDialog extends ClientDialog implements
             
         } else if (source.equals(showDamageDecal)) {
             guip.setShowDamageDecal(showDamageDecal.isSelected());
+        } else if (source.equals(showDamageLevel)) {
+            guip.setShowDamageLevel(showDamageLevel.isSelected());
         }
     }
 
--- a/megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
+++ b/megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
@@ -345,7 +345,6 @@ public class CommonSettingsDialog extends ClientDialog implements
         setLayout(new BorderLayout());
         getContentPane().add(panTabs, BorderLayout.CENTER);
         getContentPane().add(getButtonsPanel(), BorderLayout.PAGE_END);
-        
         // Close this dialog when the window manager says to.
         addWindowListener(new WindowAdapter() {
             @Override
@@ -594,7 +593,6 @@ public class CommonSettingsDialog extends ClientDialog implements
         row = new ArrayList<>();
         row.add(getFocus);
         comps.add(row);
-        
         mouseWheelZoom = new JCheckBox(Messages.getString("CommonSettingsDialog.mouseWheelZoom")); //$NON-NLS-1$
         row = new ArrayList<>();
         row.add(mouseWheelZoom);
--- a/megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
+++ b/megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
@@ -1795,7 +1795,7 @@ public class CommonSettingsDialog extends ClientDialog implements
 
     /** Constructs the button ordering panel for one phase. */ 
     private JScrollPane getButtonOrderPane(DefaultListModel<PhaseCommand> list, 
-            StatusBarPhaseDisplay.PhaseCommand commands[]) {
+            StatusBarPhaseDisplay.PhaseCommand[] commands) {
         JPanel panel = new JPanel();
         Arrays.sort(commands, cmdComp);        
         for (StatusBarPhaseDisplay.PhaseCommand cmd : commands) {
--- a/megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
+++ b/megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
@@ -194,7 +194,6 @@ public class CommonSettingsDialog extends ClientDialog implements
     private JCheckBox animateMove;
     private JCheckBox showWrecks;
     private JCheckBox soundMute;
-    private JCheckBox showMapHexPopup;
     private JCheckBox showWpsinTT;
     private JCheckBox showArmorMiniVisTT;
     private JCheckBox showPilotPortraitTT;
@@ -226,8 +225,6 @@ public class CommonSettingsDialog extends ClientDialog implements
     private JCheckBox mmSymbol;
     private JCheckBox entityOwnerColor;
     private JCheckBox teamColoring;
-//    private JRadioButton borderTeamColor;
-//    private JRadioButton borderPlayerColor;
     private JCheckBox useSoftCenter;
     private JCheckBox levelhighlight;
     private JCheckBox shadowMap;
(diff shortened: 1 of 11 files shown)


megamek/src/megamek/client/ui/swing/GUIPreferences.java
```diff
@@ -312,14 +312,20 @@ protected GUIPreferences() {
 
         store.setDefault(FOV_HIGHLIGHT_RINGS_RADII, "5 10 15 20 25");
         store.setDefault(FOV_HIGHLIGHT_RINGS_COLORS_HSB, "0.3 1.0 1.0 ; 0.45 1.0 1.0 ; 0.6 1.0 1.0 ; 0.75 1.0 1.0 ; 0.9 1.0 1.0 ; 1.05 1.0 1.0 ");
-
+        store.setDefault(FOV_HIGHLIGHT,false);
+        store.setDefault(FOV_HIGHLIGHT_ALPHA, 40);
+        store.setDefault(FOV_DARKEN,true);
```
Windchild292, line 317:
```suggestion
        store.setDefault(FOV_DARKEN, true);
```

megamek/src/megamek/client/ui/swing/GUIPreferences.java
```diff
@@ -312,14 +312,20 @@ protected GUIPreferences() {
 
         store.setDefault(FOV_HIGHLIGHT_RINGS_RADII, "5 10 15 20 25");
         store.setDefault(FOV_HIGHLIGHT_RINGS_COLORS_HSB, "0.3 1.0 1.0 ; 0.45 1.0 1.0 ; 0.6 1.0 1.0 ; 0.75 1.0 1.0 ; 0.9 1.0 1.0 ; 1.05 1.0 1.0 ");
-
+        store.setDefault(FOV_HIGHLIGHT,false);
```
Windchild292, line 315:
```suggestion
        store.setDefault(FOV_HIGHLIGHT, false);
```

megamek/src/megamek/client/ui/swing/GUIPreferences.java
```diff
@@ -312,14 +312,20 @@ protected GUIPreferences() {
...
+        store.setDefault(FOV_HIGHLIGHT_ALPHA, 40);
+        store.setDefault(FOV_DARKEN,true);
+        store.setDefault(FOV_DARKEN_ALPHA, 100);
+        store.setDefault(FOV_STRIPES, 35);
+        store.setDefault(FOV_GRAYSCALE, "false");
```
Windchild292, line 320: I know this is old, but why is this saved as a string "false"? Any ideas?

Windchild292, line 320: Why were these moved?

megamek/src/megamek/client/ui/swing/boardview/BoardView1.java
```diff
@@ -1050,6 +1050,28 @@ public void performAction() {
...
+        controller.registerCommandAction(KeyCommandBind.TOGGLE_HEX_COORDS.cmd,
+                new CommandAction() {
+
+                    @Override
+                    public boolean shouldPerformAction() {
+                        if (shouldIgnoreKeyCommands()) {
+                            return false;
+                        } else {
+                            return true;
+                        }
```
Windchild292, line 1064:
```suggestion
                        return !shouldIgnoreKeyCommands();
```

megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
```diff
@@ -303,6 +301,27 @@ public void mouseDragged(MouseEvent e) {
...
+    private boolean savedTeamColoring;
+    private boolean savedUnitLabelBorder;
+    private boolean savedShowDamageDecal;
+    private boolean savedShowDamageLabel;
+    private String savedFovHighlightRingsRadii;
+    private String savedFovHighlightRingsColors;
+    private int savedFovHighlightAlpha;
+    private int savedFovDarkenAlpha;
+    private int savedNumStripesSlider;
+    HashMap<String, String> savedAdvancedOpt = new HashMap<>();
```
Windchild292, line 324: Before I continue I have to get clarification on this: Why are we saving values for a revert and not just having the update only occur when the dialog is saved?

Windchild292 (commented):
One important clarification about CommonSettingsDialog is required before I can continue my review.

megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
SJuliez: Because these settings immediately show on the board or minimap and to do this they must be saved to the GUIPrefs.

megamek/src/megamek/client/ui/swing/GUIPreferences.java
SJuliez: Hmmmm no :)

megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
Windchild292: That makes sense. I'll finish my review post-D&D

megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
```diff
@@ -320,34 +339,13 @@ public CommonSettingsDialog(JFrame owner, ClientGUI cg) {
...
-        JScrollPane advancedSettingsPane = new JScrollPane(advancedSettingsPanel);
-        panTabs.add("Advanced", advancedSettingsPane);
+        super(owner, Messages.getString("CommonSettingsDialog.title"), true);
 
+        JTabbedPane panTabs = new JTabbedPane();
         setLayout(new BorderLayout());
         getContentPane().add(panTabs, BorderLayout.CENTER);
         getContentPane().add(getButtonsPanel(), BorderLayout.PAGE_END);
-
+
```
Windchild292, line 348:
```suggestion

```

megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
```diff
@@ -605,7 +594,7 @@ private JPanel getSettingsPanel() {
         row = new ArrayList<>();
         row.add(getFocus);
         comps.add(row);
-
+
```
Windchild292, line 597:
```suggestion

```

megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
```diff
@@ -1776,98 +1765,50 @@ private JPanel getButtonOrderPanel(){
...
+        targetingPhaseCommands = new DefaultListModel<StatusBarPhaseDisplay.PhaseCommand>();
+        phasePane.add("Targeting", getButtonOrderPane(targetingPhaseCommands,
+                TargetingPhaseDisplay.TargetingCommand.values()));
+        
+        return buttonOrderPanel;
+    }
+
+    /** Constructs the button ordering panel for one phase. */ 
+    private JScrollPane getButtonOrderPane(DefaultListModel<PhaseCommand> list, 
+            StatusBarPhaseDisplay.PhaseCommand commands[]) {
```
Windchild292, line 1798: This is not a Java-style array.

megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
```diff
@@ -1202,10 +1224,6 @@ public void itemStateChanged(ItemEvent event) {
             stampFormatLabel.setEnabled(stampFilenames.isSelected());
         } else if (source.equals(fovInsideEnabled)) {
             guip.setFovHighlight(fovInsideEnabled.isSelected());
-            if ((clientgui != null) && (clientgui.bv != null)) {
-                clientgui.bv.clearHexImageCache();
-                clientgui.bv.repaint();
-            }
```
Windchild292, line 1208: Just checking, but we don't need to do either of these to properly update the board with the newly changed state?

Windchild292 (commented):
Two more comments, but I think this looks good.

megamek/src/megamek/client/ui/swing/GUIPreferences.java
SJuliez: It's actually treated as a boolean everywhere else, so this is just wrong. I'll correct it.

SJuliez: I moved a few of the lines around to group them up better. May not have been terribly important for these specific lines, I don't remember.

megamek/src/megamek/client/ui/swing/boardview/BoardView1.java
SJuliez: Oh indeed :) Never noticed. I'll correct this for all the other binds too.

megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
SJuliez: Not sure what's wrong with this line. No tabs. A bit of breathing space for the code ;)

megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
SJuliez: This empty line separates the option blocks. Not different from many before and after. Dont think this should be changed.

megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
SJuliez: No, these lines are actually nonsense I may have put in myself a few years ago. The boardview does preferencechange-listening and can update when such an event is fired. It just needs the right checks for what pref has changed to update accordingly. It could even update without checking what has changed but I think that's bad form, as all hexes have to be redrawn etc.

megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
SJuliez: PhaseCommand[] commands

SJuliez:
OK, added in the suggestions as commented.

megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
Windchild292: This removes the extra spacing from it, which would be automatically removed in a further commit. Just dealing with future blankspace changes now.

megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
Windchild292: This removes the extra spacing from it, which would be automatically removed in a further commit. Just dealing with future blankspace changes now.

Windchild292 (approved):

PR #2351, opened 2020-10-22
don't use potential crit roll to determine actual crit check

NickAragua:
Adds an explicit flag to indicate that an aerospace unit's fuel tank was hit. This fixes an observed behavior where, sometimes, for no apparent reason, an aerospace unit would display as crippled.

The actual cause is that there's this "get/setPotCrit" method which stores the potential crit roll result when rolling a critical hit on an aero. That's fine, but the 'isCrippled()' method uses that to check whether the fuel tank has been hit. Sometimes, the potential crit hit result would be the fuel tank, but no actual crit hits would get rolled, and then the unit would show as crippled.


sixlettervariables (approved):
Looks good.

PR #2352, opened 2020-10-22
File Factory Standardization

Windchild292:
Following up on https://github.com/MegaMek/megamek/pull/2165, this relocates and standardizes the file factories.

This requires the MekHQ-side to merge.


megamek/src/megamek/common/util/fileUtils/ScaledImageFileFactory.java
```diff
@@ -0,0 +1,135 @@
...
+
+            @Override
+            public Object getItem() throws Exception {
+                // Cache the image on first use.
+                if (image == null) {
+                    image = createZippedImage(itemEntry, zipFile);
+                }
+                // Return a copy of the image.
+                return ImageUtil.getScaledImage(image, getWidth(), getHeight());
+            }
```
sixlettervariables, line 132: Should we be caching the scaled image as long as height/width don't change? Or do we scale them a lot?

sixlettervariables (approved):
One question but no blockers.

megamek/src/megamek/common/util/fileUtils/ScaledImageFileFactory.java
Windchild292: Pre-existing... but good point. Will swap that over

Windchild292: It has to do with cloning the image. I'll check and see if that's required anywhere else.

Windchild292: From testing it does not appear to be required, and thus I've pushed that change.

sixlettervariables (approved):

PR #2353, opened 2020-10-22
AbstractIcon: Standardizing GUI and Initial Implementations

Windchild292:
This is the next phase of the AbstractIcon implementation, during which the graphical components are standardized and now use AbstractIcon::getImage.


megamek/src/megamek/client/ui/swing/ChatLounge.java
```diff
@@ -3483,7 +3483,7 @@ public Component getTableCellRendererComponent(JTable table, Object value, boole
                         if (compact) {
                             clearImage();
                         } else {
-                            setPortrait(entity.getCrew());
+                            setImage(entity.getCrew().getPortrait(0).getImage(50));
```
sixlettervariables, line 3486: Those are oddly specific values. Should they be named?

megamek/src/megamek/client/ui/swing/dialog/imageChooser/AbstractIconRenderer.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Override
+    public String getToolTipText() {
+        return tip;
+    }
+
+    /** Sets the label text of the image, removing the file extension. */
+    private void setText(String text) {
+        // Remove the file extension
+        for (String ext: EXTENSIONS) {
```
sixlettervariables, line 64: I'm sure this was inherited, but this doesn't remove a file extension.

sixlettervariables (commented):
Couple of questions/notes.

megamek/src/megamek/client/ui/swing/ChatLounge.java
Windchild292: We don't name sizes currently, so I would say not

sixlettervariables: Wasn't obvious that was the square size, makes sense now.

megamek/src/megamek/client/ui/swing/dialog/imageChooser/AbstractIconRenderer.java
Windchild292: From testing it does, however I'd say the extension is required because it is entirely valid to have Male_1.gif and Male_1.jpg as two distinct files in our setup. I've pushed a commit to handle this.

sixlettervariables: More I meant it just rips out any text match rather than the actual file extension.

sixlettervariables (approved):

PR #2357, opened 2020-10-23
Ejection damage

NickAragua:
Implements the ruling here: https://bg.battletech.com/forums/strategic-operations/answered-repairing-an-ejected-cockpit/

Basically, when ejecting from a mech, the cockpit is treated as having been hit (maybe repairable in MekHQ, maybe not). For mechs with full-head ejection systems, the head is treated as being blown off.

Also adds full-head ejection to the various Wolfhounds and Hatchetmen that were missing it. As a consequence, they're all now 'advanced rules'. Review with whitespace off. 

Mech edits were made with MML.


sixlettervariables (approved):
First look at the code is good. Still needs data sign-offs.

HammerGS (approved):
Data wise looks good to me.

PR #2368, opened 2020-10-30
destroyed head should still let torso cocpkit mechs fire weapons

### LINKED ISSUE #2300: Sensor hits with torso-mounted cockpits working incorrectly
Version 47.10

Megamek currently prevents units featuring torso-mounted cockpits from firing any weapons after their head is destroyed (specifying "attacker sensors destroyed")

However, page 49 of the BMM states that "Instead of the standard modifiers from sensor damage, the first sensor hit incurs a +2 Target Number modifier to all weapon attacks, while the second sensor hit increases this to a +4 modifier."

Attached is a saved game where this was tested.
[sensor torso.sav.gz](https://github.com/MegaMek/megamek/files/5324310/sensor.torso.sav.gz)

NickAragua:
Fixes #2300 

Mechs with torso mounted cockpits and destroyed heads have their own check for sensor damage and should be mutually exclusive with all other sensor checks.

### DIFF
--- a/megamek/src/megamek/common/actions/WeaponAttackAction.java
+++ b/megamek/src/megamek/common/actions/WeaponAttackAction.java
@@ -1116,9 +1116,8 @@ public class WeaponAttackAction extends AbstractAttackAction implements Serializ
             if (sensorHits > 3) {
                 return Messages.getString("WeaponAttackAction.SensorsDestroyed");
             }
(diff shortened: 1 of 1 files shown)


sixlettervariables (approved):
Looks good! That file was too large to review on my phone.

PR #2391, opened 2020-11-10
2220: Improving Image Null and Invalid Checks

Windchild292:
This is part of fixing https://github.com/MegaMek/mekhq/issues/2220, and during debugging I expanded the null or invalid checks to prevent certain issues like this.

### DIFF
--- a/megamek/src/megamek/common/util/ImageUtil.java
+++ b/megamek/src/megamek/common/util/ImageUtil.java
@@ -211,7 +211,7 @@ public final class ImageUtil {
                 return null;
             }
             final boolean isAnimated = waitUntilLoaded(result);
-            if ((result.getWidth(null) == -1) || (result.getHeight(null) == -1)) {
+            if ((result.getWidth(null) < 0) || (result.getHeight(null) < 0)) {
                 return null;
             }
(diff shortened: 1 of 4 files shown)


megamek/src/megamek/common/util/ImageUtil.java
```diff
@@ -211,10 +207,13 @@ public Image loadImage(String fileName) {
                 return null;
             }
             Image result = Toolkit.getDefaultToolkit().getImage(fileName);
-            if(null == result) {
+            if (result == null) {
+                return null;
+            }
+            final boolean isAnimated = waitUntilLoaded(result);
+            if ((result.getWidth(null) == -1) || (result.getHeight(null) == -1)) {
```
NickAragua, line 214: I'd recommend < 0

NickAragua (changes requested):
Needs a little more protection.

megamek/src/megamek/common/util/ImageUtil.java
Windchild292: The choice was based on the JavaDoc comment in Image::getWidth and Image::getHeight, but it cannot be less than 0 so that's a better way to handle it. Flipped over.

NickAragua (approved):

PR #2395, opened 2020-11-11
Upgrade to Gradle 6.7

sixlettervariables:
This upgrades to Gradle 6.7 and resolves the deprecations noted during builds. This also removes the unused (and likely unfixable) Mac OS X bundle.

When including MegaMek under Gradle 6+ you'll need to give the build a name, like in your `settings_local.gradle`:
```groovy
includeBuild('../megamek') {
    name = 'mm-composite'
}
```
No two `GradleBuild` tasks or includes can share the same name. Under Gradle 6.3+ it uses the directory name by default, leading to these issues.


gradlew
```diff
@@ -28,7 +44,7 @@ APP_NAME="Gradle"
 APP_BASE_NAME=`basename "$0"`
 
 # Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-DEFAULT_JVM_OPTS=""
```
Windchild292, line 31: Why 64m?

gradlew.bat
```diff
@@ -14,7 +30,7 @@ set APP_BASE_NAME=%~n0
 set APP_HOME=%DIRNAME%
 
 @rem Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-set DEFAULT_JVM_OPTS=
+set DEFAULT_JVM_OPTS="-Xmx64m" "-Xms64m"
```
Windchild292, line 33: Same here

Windchild292 (commented):
Two questions first

Windchild292 (approved):
For immediate merge post-stable

PR #2404, opened 2020-11-16
Don't reset position following a failed maneuver

### LINKED ISSUE #2398: Final position of vehicle after skid (and fall) reported in Round Report but not applied to unit
0.47.13 standalone
Windows 10
Java 11.0.8

As seen on the Round Report in the attached screenshot, the J Edgar failed a PSR in hex 2806 and skidded, falling into hex 2706 (and correctly taking damage from the fall). However, the immobilised J Edgar seems to have somehow remained in hex 2806, rather than ending up in hex 2706, into which it should have skidded.

20201112-0226 Fallen J Edgar Location Bug

Before: [autosave_20201112-014448.sav.gz](https://github.com/MegaMek/megamek/files/5530968/autosave_20201112-014448.sav.gz)
After: [20201112-0226.sav.gz](https://github.com/MegaMek/megamek/files/5530952/20201112-0226.sav.gz)
[megameklog 20201112a.txt](https://github.com/MegaMek/megamek/files/5530957/megameklog.20201112a.txt)

Player name: SimonLandmine
Multiple Princess bots, named for their behaviour variant - e.g. 'Sn' = Sniper, etc.

neoancient:
I'm not sure why this line is there. It's possibly a vestige from a copy/paste. But I can't think of any reason it would be correct to reset the Entity's position after processing a failed vehicle maneuver to what it was before.

Fixes #2398

### DIFF
--- a/megamek/src/megamek/server/Server.java
+++ b/megamek/src/megamek/server/Server.java
@@ -7704,7 +7704,7 @@ public class Server implements Runnable {
                             continueTurnFromFishtail = true;
                         }
                         curFacing = entity.getFacing();
-                        entity.setPosition(curPos);
+                        curPos = entity.getPosition();
                         entity.setSecondaryFacing(curFacing);
                         break;
                     }


sixlettervariables (approved):
How odd.

neoancient:
I'm the one that wrote it, and I can't explain it.

PR #2406, opened 2020-11-17
Scroll wheel

### LINKED ISSUE #2399: RFE - 'Mouse wheel zooms map' Client Settings option upgrade
When the 'Mouse wheel zooms map' Client Settings option is unchecked, the mouse wheel scrolls the map, with CTRL+wheel zooming the map.
However, when 'Mouse wheel zooms map' option is checked, the mouse wheel zooms the map ... and so does CTRL+wheel, leaving no mouse-based option to scroll the map.

Would it be possible to configure things so that, when 'Mouse wheel zooms map' is checked, CTRL+wheel scrolls the map?

### LINKED ISSUE #2405: 47.14 Minor scroll-wheel-zoom oddity
### Environment
47.14
Windows 10
Java 11.0.8

### Description
In 47.14, "Mouse wheel zooms map" and "Flip mouse wheel zoom direction" are active by default. In this case, both the minimap and the main map view zoom in when scrolling up, and zoom out when scrolling down. Similarly, if "Flip mouse wheel zoom direction" is deactivated, they both zoom out when scrolling up, and zoom in when scrolling down. This is all as expected.

However, if "Mouse wheel zooms map" is _deactivated_, but CTRL+mouse wheel is used to zoom the map, an inversion occurs. With "Flip mouse wheel zoom direction" still active, the minimap zooms in when scrolling up and zooms out when scrolling down, as before. However, the main map zooms _out_ when CTRL+scrolling up, and _in_ when CTRL+scrolling down. Similarly, if "Flip mouse wheel zoom direction" is also deactivated, the minimap zooms out when scrolling up and zooms in when scrolling down, but the main map zooms _in_ when CTRL+scrolling up, and _out_ when CTRL+scrolling down. 

The net effect of this is that the zoom direction of the main map using CTRL+scroll wheel is always opposite that of the minimap if "Mouse wheel zooms map" is deactivated. It would be nice if this could be flipped to match the behavior of the "Mouse wheel zooms map" option.

SJuliez:
Corrects the mouse wheel behavior on the board and minimap. When the wheel without CTRL zooms, it will scroll with CTRL and vice versa. When scrolling, holding shift scrolls horizontally. Holding shift while zooming has no effect. The direction of the minimap zoom and the board zoom should now always be the same. 
Note that as before, the minimap doesnt care about CTRL and will always zoom. 

Resolves #2405 
Resolves #2399


megamek/src/megamek/client/ui/swing/boardview/BoardView1.java
```diff
@@ -569,57 +569,31 @@ public void mouseWheelMoved(MouseWheelEvent we) {
                 double ihdx = ((double)inhexDelta.x)/((double)HEX_W)/scale;
                 double ihdy = ((double)inhexDelta.y)/((double)HEX_H)/scale;
                 int oldzoomIndex = zoomIndex;
+                
+                boolean ZoomNoCtrl = GUIPreferences.getInstance().getMouseWheelZoom();
+                boolean wheelFlip = GUIPreferences.getInstance().getMouseWheelZoomFlip();
+                boolean zoomIn = (we.getWheelRotation() > 0) ^ wheelFlip; // = XOR
+                boolean doZoom = ZoomNoCtrl ^ we.isControlDown(); // = XOR
+                boolean horizontalScroll = !doZoom && we.isShiftDown();
```
sixlettervariables, line 577: While in testing this does what it says it does, the use of XOR could likely use a small comment/table explaining which modes are active with which key presses. Not required, but would be helpful in the future.

sixlettervariables (approved):

PR #2407, opened 2020-11-18
Issue #1803: Configure view changes Clan unit BV

### LINKED ISSUE #1803: BV Discrepancy

sixlettervariables:
When opening the Configure view for a Clan unit and clicking OK, the BV may change as in #1803. The root cause of this issue is the `JComboBox` supporting the ammo type selection uses strings, and in a number of cases (such as Ultra A/C ammo) there is no difference in the names between IS and Clan ammo. Because of this, no matter what selection logic you use under the hood the JComboBox will take the first matching string when changing the selection.

To fix this I simply switched the JComboBox to use AmmoTypes directly, instead of Strings. The names still appear identical (an issue for another PR), but the selection works. It works because I also removed `AmmoType::equals` and allowed it to fall back to `EquipmentType::equals`. Previously, AmmoType::equals used "loose" equality, ignoring munition types and bomb types. Now we use strict equality (and have an appropriate hash code too).

This fix is not one to take lightly, but one that appears to have no broad affect on MM. I've labeled the areas I wasn't sure of which used the old `AmmoType::equals` and `AmmoType::hashCode` so that I can get eyes on these other spots.

Fixes #1803.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/EquipChoicePanel.java
+++ b/megamek/src/megamek/client/ui/swing/EquipChoicePanel.java
@@ -18,7 +18,6 @@ import java.awt.GridBagConstraints;
 import java.awt.GridBagLayout;
 import java.awt.event.ItemEvent;
 import java.awt.event.ItemListener;
-import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -55,7 +54,7 @@ import megamek.common.weapons.infantry.InfantryWeapon;
  * @author arlith
  * @since 2012-05-20
  */
-public class EquipChoicePanel extends JPanel implements Serializable {
+public class EquipChoicePanel extends JPanel {
     static final long serialVersionUID = 672299770230285567L;
 
     private final Entity entity;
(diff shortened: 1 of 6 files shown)


neoancient (approved):
This at least fixes a known issue and is an overall improvement. If it ends up creating some new bugs, they should be easier to resolve without breaking anything else.

PR #2410, opened 2020-11-19
Allow HarJel II & III in cockpit location

### LINKED ISSUE #2161: 47.9 Harjel in head
I put a similar issue in for MML, but when I hand-edited an MTF with Harjel III in the head, Megamek labels it as invalid. It will still function, but the flag is, as far as I can tell, erroneous. Nothing in the rules or errata for Harjel II or III says that they are restricted from the cockpit location. I can find that for the original Tacops Harjel system but nothing for the II or III.

neoancient:
Per rules clarification, the restriction against HarJel in the same location as the cockpit only applies to the original HarJel system from TacOps. HarJel II and III from IO are permitted in the cockpit location.
https://bg.battletech.com/forums/supplementary-rules/answered-io-harjel-all-types-and-head-locations/

Fixes #2161

### DIFF
--- a/megamek/src/megamek/common/verifier/TestMech.java
+++ b/megamek/src/megamek/common/verifier/TestMech.java
@@ -1595,8 +1595,7 @@ public class TestMech extends TestEntity {
                 }
                 return false;
             }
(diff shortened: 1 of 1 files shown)


HammerGS (approved):
From my understanding of the code, looks good to me.

sixlettervariables (approved):

Issue #2313, opened 2020-10-06
SSW -> MM BV Issues


rjhancock:
Issues noted by the SSW team when comparing their calculations against both MUL and MML:

~~Both MML and SSW both calculate ammo in the front legs of quads as ammo in arms, which affects BV.~~
~~MML isn't counting M-Pods as explosive~~
* iATMs may not count as half heat like streaks in MML

Tested against MML 0.47.9

Fixed (or think we are fixed)
~~Bridge Layer BV may not be in MML~~
~~Stealth STH-5X recordsheet may be wrong in MML~~
~~Turkina U missing coolant pod for at least BV calc purposes~~
~~Wolfhound WLF-2X recordsheet may be wrong in MML~~
~~2KO No-Dachi has an incorrect RS, 2 ERMLs firing to the rear that are firing to the front in the RS~~

WEKarnesky:
Feel free to direct any questions to me, I'll field them to the SSW team if I can't answer.

Maelwys:
For some odd reason in the ON1-H Orion, it has

"Explosive Weapons/Equipment Penalty |   | = -165.0"

60 of this is ammo. I'm not sure where the other 105 comes from. The only thing I can think of is that MM is treating the Primitive Prototype's AC10's crits as explosive ammo for some odd reason.

HammerGS:
Some fixed in https://github.com/MegaMek/megamek/commit/74c3966a520f0b0ca4549e25763a064dde119bad

HammerGS:
> For some odd reason in the ON1-H Orion, it has
> 
> "Explosive Weapons/Equipment Penalty |   | = -165.0"
> 
> 60 of this is ammo. I'm not sure where the other 105 comes from. The only thing I can think of is that MM is treating the Primitive Prototype's AC10's crits as explosive ammo for some odd reason.
> 
> edit
> And according to page 195 of IO, the PPAC10's ammo should be 12 for BV, not 21
> /edit

I've fixed the Ammo BV issue and think I've fixed this part of the issue here -  https://github.com/MegaMek/megamek/commit/b627b4cd8701be2047f9d3cd24c4912fe33def63

HammerGS:
I'm not 100% sure of the code but in mech.java line 3790 looks like it's setting M-Pod to subtract 0 when To:AUE pg 195 shows them as 5 and explosive.

PR #2418, opened 2020-11-21
Fix explosive equipment penalty for mech BV

neoancient:
This fixes two issues with calculating the explosive equipment penalty for mechs:
1. Equipment in legs is being treated like equipment in arms.
2. Weapons split between two adjacent locations is treated like it's all in the primary location. For instance, a heavy gauss rifle with two CT crits and the rest side torso are being treated like all crits are in the CT.

Rules summary:
For each crit of explosive equipment in a vulnerable location, subtract 15 for ammo or 1 for other explosive equipment such as gauss rifle.
Any location with CASE II is not vulnerable. Otherwise, a vulnerable location is any head, CT, or leg, or side torso that has >= 3 engine slots, or any location without CASE that has a vulnerable location next on the transfer diagram. This is transitive, so an arm without CASE is vulnerable if the side torso also doesn't have CASE.

See #2313

### DIFF
--- a/megamek/src/megamek/common/Mech.java
+++ b/megamek/src/megamek/common/Mech.java
@@ -3364,6 +3364,28 @@ public abstract class Mech extends Entity {
     }
 
     /**
+     * Used in BV calculations. Any equipment that will destroy the unit or leg it if it explodes
+     * decreases the defensive battle rating. This is anything in the head, CT, or leg,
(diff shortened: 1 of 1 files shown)


sixlettervariables (approved):
Took a hot second to reason through the booleans, but looks good.

PR #2433, opened 2020-11-25
fix weird building glitch

NickAragua:
I ran into a fairly weird issue where some bot units reduced a building with a turret on top of it to a CF of 0, but the building wouldn't collapse. Instead, after every phase, it would report the building being destroyed by collapse, but not actually collapse it or destroy the turret.

This is an "after the fact" fix. I don't know where the building's CF was set to 0 without collapsing it, there was nothing in the log to indicate any errors.


sixlettervariables (approved):
"Mr. Gorbachev, tear down this building!"

PR #2440, opened 2020-11-26
Fix check for legal move when resolving domino effect

### LINKED ISSUE #2430: NPE in doEntityDisplacement
### Environment
Standalone MegaMek v0.46.1
Java version 1.8.0_271
Platform Windows 10 10.0 (amd64)

### Description
Exception in thread "Packet Pump" java.lang.NullPointerException
	at megamek.server.Server.doEntityDisplacement(Server.java:13546)
	at megamek.server.Server.resolveDfaAttack(Server.java:19918)
	at megamek.server.Server.resolvePhysicalAttack(Server.java:34925)
	at megamek.server.Server.resolvePhysicalAttacks(Server.java:15939)
	at megamek.server.Server.endCurrentPhase(Server.java:3181)
	at megamek.server.Server.changeToNextTurn(Server.java:3494)
	at megamek.server.Server.endCurrentTurn(Server.java:2484)
	at megamek.server.Server.receiveAttack(Server.java:14126)
	at megamek.server.Server.handle(Server.java:32866)
	at megamek.server.Server$PacketPump.run(Server.java:369)

Physical Attack Phase.
SQR-L conducts a DFA against I1A. SQR-L#2 and DJN-1 also declared kick and punch attacks against I1A. I1A has declared a kick against SQR-L#2.

The DFA is successful (I think) and displaces I1A down the hill into J2A. The player controlling J2A is given the option to step out of the domino effect (forward or backward). This puzzled us because stepping forward would step into SQR-L#2. The player decided to step forward and see it it would displace SQR-L#2. The server and client then became unresponsive and the above exception stack trace appeared in the log.

The screenshot is garbled a bit because we saved after the exception and reloaded. On reloading the save, we see that the game has displaced I1A but not J2A, nor resolved any of the physical attacks. The game cannot proceed (neither player can take an action).

### Files
[autosave.sav.gz](https://github.com/MegaMek/megamek/files/5593906/autosave.sav.gz)

[megameklog_npe.txt](https://github.com/MegaMek/megamek/files/5593909/megameklog_npe.txt)

Players: "Josh" (controls the DFA'ing unit), "Mike" (controls the opfor)

neoancient:
Rules for domino effect are found on TW, 152-3.

In the scenario that generated the NPE, the target of a DFA attack was pushed into a hex containing another mech. The other mech was given the choice to move forward or backward to avoid the domino effect, though there was another mech in the forward hex, so that move should have been illegal. When the two MovePaths are compiled on the client side, clipping is left on, so the illegal step is removed from the forward MovePath. isMoveLegal() returns true, since the illegal move is not there. If selected, the empty path is sent to the server, which expects either a null (no action) or a MovePath with a single move in it.

Not clipping the path when compiling it leaves the step in place and correctly returns false to isMoveLegal(). On the server side, the moves are not compiled before checking whether either is legal. This will correctly validate the end position, but will not evaluate such things as changing elevation by 3+ levels.

Fixes #2430

### DIFF
--- a/megamek/src/megamek/client/ui/swing/ClientGUI.java
+++ b/megamek/src/megamek/client/ui/swing/ClientGUI.java
@@ -2053,8 +2053,8 @@ public class ClientGUI extends JPanel implements WindowListener, BoardViewListen
                     MovePath stepBackward = new MovePath(client.getGame(), e);
                     stepForward.addStep(MoveStepType.FORWARDS);
                     stepBackward.addStep(MoveStepType.BACKWARDS);
-                    stepForward.compile(client.getGame(), e);
-                    stepBackward.compile(client.getGame(), e);
+                    stepForward.compile(client.getGame(), e, false);
+                    stepBackward.compile(client.getGame(), e, false);
                     
                     String title = Messages.getString("CFRDomino.Title");
                     String msg = Messages.getString("CFRDomino.Message",
--- a/megamek/src/megamek/server/Server.java
+++ b/megamek/src/megamek/server/Server.java
@@ -12845,6 +12845,8 @@ public class Server implements Runnable {
                                 throw new IllegalStateException();
                             }
                             MovePath mp = (MovePath) rp.packet.getData()[1];
+                            mp.setGame(getGame());
+                            mp.setEntity(violation);
                             // Move based on the feedback
                             if (mp != null) {
                                 // Report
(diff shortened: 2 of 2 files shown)


sixlettervariables (approved):
Interesting failure mode!

neoancient:
LGTM caught an error that will throw an exception if the player chooses not to move out of the way.

sixlettervariables (approved):

PR #2444, opened 2020-11-28
Improve MtfFile parsing performance

sixlettervariables:
Hoist repeated string calls into locals to improve performance of MTF file parsing. I've been running a lot of tests on MHQ and during refit testing I noticed parsing MTF files to be a significant chunk of the time during the tests.

On JDK11.0.7, i5 16G RAM:
- Delete units.cache then load 4MB CPNX
- MechFileParser::parse Total Time before: 7442 ms
- MechFileParser::parse Total Time after: 2929 ms

### DIFF
--- a/megamek/src/megamek/common/loaders/MtfFile.java
+++ b/megamek/src/megamek/common/loaders/MtfFile.java
@@ -178,8 +178,8 @@ public class MtfFile implements IMechLoader {
         int weaponsCount;
         int armorLocation;
         while (r.ready()) {
-            crit = r.readLine();
-            if (crit.trim().isEmpty()) {
+            crit = r.readLine().trim();
+            if (crit.isEmpty()) {
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/loaders/MtfFile.java
```diff
@@ -505,107 +507,103 @@ public Entity getEntity() throws EntityLoadingException {
...
-                        throw new EntityLoadingException("Unsupported tech level: " + rulesLevel.substring(12).trim());
-                }
-                mech.setMixedTech(true);
-                break;
-            case "Mixed":
-                throw new EntityLoadingException("Unsupported tech base: \"Mixed\" is no longer allowed by itself.  You must specify \"Mixed (IS Chassis)\" or \"Mixed (Clan Chassis)\".");
-            default:
-                throw new EntityLoadingException("Unsupported tech base: " + techBase.substring(9).trim());
+        String techBase = this.techBase.substring(9).trim();
+        if (techBase.equalsIgnoreCase("Inner Sphere")) {
```
Windchild292, line 511: We should probably move this into a constant

Windchild292 (approved):
Changes look good to me, a suggestion would be to convert all the uppercase information to lowercase in the first place, do trims when the file is read, put any strings into constants, etc.

neoancient (approved):

megamek/src/megamek/common/loaders/MtfFile.java
sixlettervariables: Noted, just wanted this PR to be 'to-the-point' so as to make the review simple.

PR #2448, opened 2020-11-29
2348: Game summary images

### LINKED ISSUE #2348: RFE - Screenshot Board Overtime
### Environment
MegaMek and HQ MegaMek
Windows
Java 8 update 261

### Description
Would like the option to have Board->Save As Image auto saved like the game auto save every time the board changes.  Or perhaps at the end of each phase. 

### Files
NA

arlith:
The purpose of this work is to create some type of summary for a game.  It does this by saving board or minimap images at the end of every phase.  Two options are added in the Client Settings > Graphics menu (one for saving board images, one for minimap images).

The images are saved in `logs/gameSummaries` (either in a `board` or `minimap` subdirectory).  The PR adds a UUID to the `Game` object, so every saved game can be uniquely identified.  Images are stored in a subdirectory with the UUID so that game summary images can be continued after loading a saved game.

I had originally thought that this branch wasn't finished but really it is.  There are a couple of caveats.  Ultimately, I feel like users want to create some type of animation (gif) or video.  I had thought about doing this automatically but it's not really necessary and really it could quickly get quite complex.  Having the raw images can still be useful (as users can flip through them manually) and by having the images users can also create their own animations or videos (in Linux this is really simple, with just a single call to `convert` or `ffmpeg`).  It's also worth noting that this option can really use up a lot of diskspace: the output image sizes are based upon the size of the board: a 16x17 hex board generates a ~2MB image per phase, which is usually 2-3 (move, fire, physical).  But the options to save these images defaults to off.

Since these changes only really have an effect if the option for them is turned on, I feel like this is pretty safe to merge.

This fixes #2348


arlith:
Here's an example of what this looks like.  Note I scaled this down by 50% to get it under the 10mb filesize limit (original full resolution was ~21mb)

test

arlith:
Here's the same game, just with the minimap summary instead.  This really works best for large games, like company level or larger. (This is not scaled, since the minimap is much smaller.)
test

megamek/src/megamek/client/ui/swing/GUIPreferences.java
```diff
@@ -115,6 +115,9 @@
     public static final String DISPLAY_POS_Y = "DisplayPosY";
     public static final String DISPLAY_SIZE_HEIGHT = "DisplaySizeHeight";
     public static final String DISPLAY_SIZE_WIDTH = "DisplaySizeWidth";
+    public static final String GAME_SUMMARY_BOARD_VIEW = "GameSummaryBoardView";
+    public static final String GAME_SUMMARY_MINI_MAP = "GameSummaryMiniMap";
+    public static final String ENTITY_OWNER_LABEL_COLOR = "EntityOwnerLabelColor";
```
sixlettervariables, line 120: Is this property related? It appears unused.

megamek/src/megamek/client/ui/swing/boardview/BoardView1.java
```diff
@@ -2451,8 +2464,12 @@ public BufferedImage getEntireBoardImage(boolean ignoreUnits) {
                 }
             }
         }
-
         boardGraph.dispose();
+
+        // Restore the zoom setting
+        zoomIndex = oldZoom;
+    	zoom();
```
sixlettervariables, line 2471: Whitespace is off starting here it looks.

megamek/src/megamek/common/Game.java
```diff
@@ -3651,4 +3659,17 @@ private void checkPositionCacheConsistency() {
         }
     }
 
+    /**
+     * Get a string representation of the UUId for this game.
+     *
+     * @return
+     */
+    public String getUUIDString() {
+        if (uuid == null) {
```
sixlettervariables, line 3668: Curious: when would `uuid` be null?

sixlettervariables (commented):
Few questions.

megamek/src/megamek/client/ui/swing/GUIPreferences.java
arlith: `ENTITY_OWNER_LABEL_COLOR` comes from the merge: I assume it was a property that was added at line 118, the same line I added for the two properties related to game summary.  As far as I can tell, all three of those properties are referenced in the code.  I don't see anything in this addition that is unused.

megamek/src/megamek/client/ui/swing/boardview/BoardView1.java
arlith: My code formatter was improperly configured and it was inserting some tabs.  I saw this during the commit and I should've fixed it then.  I've fixed it now (there were a few other places too besides this one).

megamek/src/megamek/common/Game.java
arlith: It wouldn't be.  As the code is written now, it can't be (the constructor initializes it and there's no setter).  I'm pretty sure I wrote this just to be defensive.

Actually, I think I was thinking about saved games from previous versions, but those might just fail to deserialize.

It's easy enough to remove the condition if you want, but I also don't really think it harms anything.  Maybe it would be better to log a message at that point, since it's a condition we'd expect not to happen.

sixlettervariables: I'm okay as is. We also don't guarantee save games across versions so no worries there.

megamek/src/megamek/client/ui/swing/GUIPreferences.java
sixlettervariables: Gotcha, weird that GH marked it as an addition!

sixlettervariables (approved):

SimonLandmine:
Nice one! Thanks for all of your work, @arlith!

PR #2451, opened 2020-11-29
Bug.#2413 fov darken tooltips

### LINKED ISSUE #2413: 47.14 Weird two-tone effect while using "Darken hexes outside field of view"
### Environment
47.14 (first noticed in MekHQ integrated Megamek, reproduces when loading save in standalone MM)
Windows 10
Java 11.0.8

### Description
My preferred client settings should provide a uniform darkening of hexes outside of the field of view of the selected hex. Instead, an odd "two-tone" effect is sometimes observed, with some hexes partially darkened (relative to the specified level) and others fully darkened. 

In a probably-unrelated issue, the slider for "Opaqueness for outside Field of View overlay" seems to decrease slightly over time, without me telling it to do so. 

Screenshot of relevant client settings:
image
(uncertain if other client settings should affect this; can provide them if instructed on how to do so)

Screenshot displaying "two-tone" effect:
image
The more distant regions have the level of darkening I usually expect with the current client settings. 

### Files
[FieldofView.sav.gz](https://github.com/MegaMek/megamek/files/5570855/FieldofView.sav.gz)
Player name: Greyscales

I don't expect the log files to have anything, but am including them anyway. They may include log data from previous scenarios. 
[megameklog.txt](https://github.com/MegaMek/megamek/files/5570857/megameklog.txt)
[mekhqlog.txt](https://github.com/MegaMek/megamek/files/5570858/mekhqlog.txt)

arlith:
Fixes #2413 

Changes include:

1) Made it so the `UnitDisplay` general tab shows snesor information if the sensor option is on, before it was looking for the double blind option

2) Updated the boardview tooltip so that if sensors are on, the tooltip will indicate whether a hex is or is not in sensor range

3) Added a tooltip to the FOV darkening UI widgets in the client settings dialog that indicates that if sensors are on then in-sensor range hexes will be slightly less darkened.

### DIFF
--- a/megamek/i18n/megamek/client/messages.properties
+++ b/megamek/i18n/megamek/client/messages.properties
@@ -2834,7 +2834,7 @@ ProtomechMapSet.Destroyed=Destroyed
 
 #Tactical Overlay - Unit field of view displays
 TacticalOverlaySettingsDialog.FovInsideEnabled=Highlight hexes inside Field of View of selected hex based on distance
-TacticalOverlaySettingsDialog.AlphaTooltip=Alpha values range from 0 (transparent) to 255 (opaque).
+TacticalOverlaySettingsDialog.AlphaTooltip=<HTML>Alpha values range from 0 (transparent) to 255 (opaque).<BR/> If sensors are on, half of this value is used to indicate hexes that are in sensor range but don't have LOS.</HTML>
 TacticalOverlaySettingsDialog.FovHighlightAlpha=Opaqueness Field of View overlay:
 TacticalOverlaySettingsDialog.FovHighlightRingsRadii=Sizes of disks that are to be highlighted: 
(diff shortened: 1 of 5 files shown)


sixlettervariables (approved):
Looks good, would like a double check on the bracket math as I'm new to that area.

arlith:
If you are referring to this:
```
                   int bracket = Compute.getSensorRangeBracket(selectedEntity, null,
                            fovHighlightingAndDarkening.cachedAllECMInfo);
                    int range = Compute.getSensorRangeByBracket(game, selectedEntity, null, los);

                    int maxSensorRange = bracket * range;
                    int minSensorRange = Math.max((bracket - 1) * range, 0);
```

That's copied straight from `FovHighlightingAndDarkening`.  I wouldn't trust myself to write something like that after so long, but since I have a template to work from I'm pretty sure it's right.

PR #2456, opened 2020-11-30
Selectively apply Zweihander crits

### LINKED ISSUE #1449: Zweihander SPA incorrectly applies critical damage
### Environment
.46.2 RC, Windows 10

### Description
Upon successfully striking a unit with a Zweihander attack making use of a hatchet, MegaMek assigns a critical roll to both arms instead of the arm where the weapon is mounted. RAW states that only unarmed Zweihander attacks making use of both hands deliver a critical roll to both arms. See below

The strike and damage resolution
image

The rules
image

Using a tree club, ReBlade, or any other single-handed weapon seems to yield similar results. Otherwise the SPA seems to be working as intended. Doesn't stack with Melee Master which is a blessing in disguise I think. You can also use Spot Welders for Zweihandering which is a delightful meme. Good work y'all

arlith:
They were always applied to both arms, but this should only be true for
punching.  For club attacks, only the arm with the weapon gets the
critical.

Fixes #1449

### DIFF
--- a/megamek/src/megamek/server/Server.java
+++ b/megamek/src/megamek/server/Server.java
@@ -15268,17 +15268,22 @@ public class Server implements Runnable {
     /**
      * Apply damage to mech for zweihandering (melee attack with both hands) as per
      * pg. 82, CamOps
-     * @param ae - the attacking entity
-     * @param missed - did the attack missed. If so PSR is necessary.
+     * 
+     * @param ae           - the attacking entity
+     * @param missed       - did the attack missed. If so PSR is necessary.
+     * @param criticalLocs - the locations for possible criticals, should be one or
+     *                     both arms depending on if it was an unarmed attack (both
+     *                     arms) or a weapon attack (the arm with the weapon).
      */
-    private void applyZweihanderSelfDamage(Entity ae, boolean missed) {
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/server/Server.java
```diff
@@ -15268,17 +15268,22 @@ private void removeDeadAttacks() {
...
-     */
-    private void applyZweihanderSelfDamage(Entity ae, boolean missed) {
+     * 
+     * @param ae           - the attacking entity
+     * @param missed       - did the attack missed. If so PSR is necessary.
+     * @param criticalLocs - the locations for possible criticals, should be one or
+     *                     both arms depending on if it was an unarmed attack (both
+     *                     arms) or a weapon attack (the arm with the weapon).
+     */
+    private void applyZweihanderSelfDamage(Entity ae, boolean missed, List<Integer> criticalLocs) {
```
sixlettervariables, line 15278: I think this would be "easier" to use if `criticalLocs` was varargs, as in `Integer... criticalLocs`

Not at all a requirement.

sixlettervariables (approved):
Looks good, one suggestion for ease of use given the pattern of one or more parameters.

PR #2458, opened 2020-12-01
Combo box to select player names when loading saved game

arlith:
For debugging, we get saved games submitted by players.  Generally, we don't know the names of the players in these saved games so it usually creates an annoying workflow of loading the game, finding the player names, then exiting and reloading.  It would be convenient to have the host dialog populate a list of players in the game to select from, but still function the same as the old load dialog.  This pull request adds this functionality.

I read in the serialized game to scrape the list of players and then `AbstractConnectionDialog` was updated such that it can take a list of player names.  If no player name list is passed (aka, null), then everything functions as before, but if the player name list is supplied then a `JComboBox` is used instead of a `JTextField` and the combo box is populated with the list of player names.  It's still editable, so users can use whatever username they want.

It turns out the action events for `JTextField` and `JComboBox` behave differently.  For the `JTextField` it seems that an edit event is only fired when the user hits enter, and the old behavior was based around this (so when you hit enter in the text box, the dialog closes).  With a `JComboBox` anytime you stop typing it fires a `comboBoxEdited` event, but this event can't distinguish between normal typing and enter being pressed.  I created a work around for this, so the behavior is as expected, but the code is a kludgy.   It could definitely be improved.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/MegaMekGUI.java
+++ b/megamek/src/megamek/client/ui/swing/MegaMekGUI.java
@@ -550,6 +550,9 @@ public class MegaMekGUI  implements IPreferenceChangeListener, IMegaMekGUI {
             newGame = (IGame) xstream.fromXML(gzi);
         } catch (Exception e) {
             MegaMek.getLogger().error("Unable to load file: " + fc.getSelectedFile(), e);
+            JOptionPane.showMessageDialog(frame, Messages.getString("MegaMek.LoadGameAlert.message"),
+            Messages.getString("MegaMek.LoadGameAlert.title"), JOptionPane.ERROR_MESSAGE);
+            return;
         }
 
         Vector<String> playerNames = null;
--- a/megamek/src/megamek/client/ui/swing/MegaMekGUI.java
+++ b/megamek/src/megamek/client/ui/swing/MegaMekGUI.java
@@ -40,13 +40,17 @@ import java.awt.event.WindowEvent;
 import java.awt.image.BufferedImage;
 import java.awt.Window;
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.List;
 import java.util.Locale;
+import java.util.Vector;
+import java.util.zip.GZIPInputStream;
 
 import javax.swing.ImageIcon;
 import javax.swing.JComponent;
@@ -59,6 +63,8 @@ import javax.swing.SwingUtilities;
 import javax.swing.ToolTipManager;
 import javax.swing.UIManager;
 import javax.swing.filechooser.FileFilter;
+
+import com.thoughtworks.xstream.XStream;
 
 import megamek.MegaMek;
 import megamek.client.Client;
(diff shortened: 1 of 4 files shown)


megamek/src/megamek/client/ui/swing/MegaMekGUI.java
```diff
@@ -536,7 +543,24 @@ public String getDescription() {
...
         }
-        HostDialog hd = new HostDialog(frame);
+
+        IGame newGame = null;
+        try (InputStream is = new FileInputStream(fc.getSelectedFile()); InputStream gzi = new GZIPInputStream(is)) {
+            XStream xstream = SerializationHelper.getXStream();
+            newGame = (IGame) xstream.fromXML(gzi);
+        } catch (Exception e) {
+            MegaMek.getLogger().error("Unable to load file: " + fc.getSelectedFile(), e);
+        }
```
SJuliez, line 553: Currently when a game doesn't load, it continues to the host dialog (and notices later, that it cant load). Maybe add 

JOptionPane.showMessageDialog(frame, Messages.getString("MegaMek.LoadGameAlert.message"),
                    Messages.getString("MegaMek.LoadGameAlert.title"), JOptionPane.ERROR_MESSAGE);
return;

to the catch block?

SJuliez (changes requested):
Very cool! I had wanted to do this for quite a while. One small change I'd add and a question: Currently it still sets my last used name as the player name, even if that name is not among the game's players. From a dev perspective at least that is usually not what I want; I'd rather have one of the game's names. Could this be changed?

arlith:
> Very cool! I had wanted to do this for quite a while. One small change I'd add and a question: Currently it still sets my last used name as the player name, even if that name is not among the game's players. From a dev perspective at least that is usually not what I want; I'd rather have one of the game's names. Could this be changed?

From a dev perspective I agree that is probably not what we want but I think what a dev wants is different from what a player wants.  I think from a players perspective this makes the most sense (it's the behavior we have currently).  I think it's still probably best to default to the last used player name rather than try to guess from the list in the saved game.  Alternatively, which option do we assume?  The first one?  That is likely to be right for developers 90% of the time but I maybe not for players.  It's hard to know.  I think for players it's likely that the last used player name is probably the right answer most of the time anyways.

megamek/src/megamek/client/ui/swing/MegaMekGUI.java
arlith: Yeah, that's a fair point.  This briefly crossed my mind.  I chose to not deal with it because I figured the old path would be fine (and it was easier to do nothing about it).  But it does make sense to handle things earlier, before the players enter additional information.  I had assumed that this loading isn't likely to fail, but I guess it is if you load saved games across versions,

NickAragua: I don't think we have any expectation of support for megamek games being loaded across versions. That being said, some error handling won't go amiss.

SJuliez:
Thinking about it again, as even as a dev 98% of games I load are my own and the name is correct anyway. So thats alright.

megamek/src/megamek/client/ui/swing/MegaMekGUI.java
arlith: That's exactly what I meant: we don't have any expectation of supporting games across versions and those are the most likely case of loading errors.

arlith:
Is this a change we want to consider for the release candidate, or wait until after the next release?

NickAragua:
I'm fine with it going into this release as long as we double check that it doesn't break the game connection dialog in MekHQ.

SJuliez (approved):

Windchild292:
I'll test it with MHQ

Windchild292:
Just commited a fix for MekHQ, added a playerName constructor in HostDialog and standardized the constructors into using one with JFrame, playerName, and playerNames

Windchild292 (approved):
Approving changes bar my own

PR #2460, opened 2020-12-01
Fix machine gun array linking and bv

### LINKED ISSUE #2195: Lineholder I/Machine Gun Array Battle Value Issue
So its sort of hard to tell, but there seems to be an issue with the Battle Value of the Machine Gun Arrays in the Lineholder I variant, where there's multiple Arrays in the same location (both arms).

From TM (4th Printing, p 317) the little "F" footnote states "BV is the total of all weapons in a machine gun array, multiplied by 0.67"

p. 317 and 318 show the standard BV for a MG is 5.

So for a 4 MG array, the BV should be (4x5) x0.67 = 13.4
For a 3 MG Array, it should be (3x5) x0.67 = 10.05

But in the Lineholder I, with the two MG arrays in the arms you get the following.

> Weapons with no heat at full BV:
> --
> Machine Gun |   |   | 5.0
> Machine Gun |   |   | 5.0
> Machine Gun |   |   | 5.0
> Machine Gun Array |   |   | 20.1
> Machine Gun |   |   | 5.0
> Machine Gun |   |   | 5.0
> Machine Gun |   |   | 5.0
> Machine Gun Array |   |   | 20.1
> Machine Gun |   |   | 5.0
> Machine Gun |   |   | 5.0
> Machine Gun |   |   | 5.0
> Machine Gun Array |   |   | 20.1
> Machine Gun |   |   | 5.0
> Machine Gun |   |   | 5.0
> Machine Gun |   |   | 5.0
> Machine Gun Array |   |   | 20.1
> Machine Gun |   |   | 5.0
> Machine Gun |   |   | 5.0
> Machine Gun |   |   | 5.0
> Machine Gun |   |   | 5.0
> Machine Gun Array |   |   | 13.4
> Machine Gun |   |   | 5.0
> Machine Gun |   |   | 5.0
> Machine Gun |   |   | 5.0
> Machine Gun |   |   | 5.0
> Machine Gun Array |   |   | 13.4
> Heat Modified Weapons BV: |   |  
> ER Large Pulse Laser |   | 272.0
>   |   | -------------
> Total Weapons BV Adjusted For Heat: |   | 479.20000000000005

As you can see, the 4-gun arrays that are in the torsos are correct at 13.4, but each 3-gun array in the arms is showing 20.1, double the expected result (and higher than the larger 4 gun arrays in the torsos).

I'm not sure if its a calculation era, or if maybe with the two arrays in the arms it may be counting the number of MGs in each array twice (I have no idea how MM assigns MGs to arrays).

[LineholderTest.zip](https://github.com/MegaMek/megamek/files/5159203/LineholderTest.zip)

I tested it in MML (47.9) by removing 3 of the MGs in an arm, along with one of the arrays, and the remaining Array showed 10.05, so it might be that each array is looking in the location and seeing every MG in the location and counting all of them when it comes to its BV calculation.

neoancient:
This fixes some related issues with machine gun arrays. Each array can link up to four machine guns of the same type in the same location. MM uses the bay weapon list in `Mounted` to track which guns are linked. The linking happens in `MechFileParser#postLoadInit`, which is called when the unit is loaded and also called by MML when changes may affect weapon linkings.

Problems:
1. The loop that checks for machine guns to link will link all machine guns of the correct size in the location that have not already been linked until it is full or runs out of guns. This does not allow the unit to split guns between separate arrays. For example, the Linebacker I has two MGAs in each arm, which should each have three MGs. But it always assigns four to the first one and the other two to the remaining one. It is also not possible to have additional unlinked machine guns in the location if the MGA is not full.
2. The BV of a machine gun array is the sum of the linked guns x 0.67. The BV code currently ignores any linking restrictions and includes the BV of all machine guns in the location. So in the case of the Linebacker I, each arm MGA is assigned twice the correct BV.

Fixes:
1. I changed the linking code to look for a contiguous block of qualifying machine guns in the location. Once the first eligible gun is found, any successive slot occupied by anything other than an eligible MG stops the process. So an MGA followed by the linked guns or a block of linked guns followed by the MGA will both work. Placing the MGA in the middle of a block of machine guns will cause the guns preceding the MGA to be linked and those following to be independent. I believe that this fits intuitive behavior.
2. I factored the linking code out of the `postLoadInit` method, which is ~700 lines, which also makes it a distinct unit available for testing.
3. The BV calculations only count the linked guns. This involves functionally identical code in six different places, but I left it to keep the changes minimal since I'm working on a complete rewrite of the BV code. This will fix the bug in the meantime.

Fixes #2195

### DIFF
--- a/megamek/src/megamek/common/Aero.java
+++ b/megamek/src/megamek/common/Aero.java
@@ -1841,14 +1841,14 @@ public class Aero extends Entity implements IAero, IBomber {
             }
             // calc MG Array here:
             if (wtype.hasFlag(WeaponType.F_MGA)) {
-                double mgaBV = 0;
-                for (Mounted possibleMG : getTotalWeaponList()) {
-                    if (possibleMG.getType().hasFlag(WeaponType.F_MG)
-                            && (possibleMG.getLocation() == weapon.getLocation())) {
-                        mgaBV += possibleMG.getType().getBV(this);
+                double mgBV = 0;
+                for (int eqNum : weapon.getBayWeapons()) {
+                    Mounted mg = getEquipment(eqNum);
+                    if ((mg != null) && (!mg.isDestroyed())) {
+                        mgBV += mg.getType().getBV(this);
                     }
                 }
-                dBV = mgaBV * 0.67;
+                dBV = mgBV * 0.67;
             }
             bvText.append(startRow);
             bvText.append(startColumn);
@@ -1994,14 +1994,14 @@ public class Aero extends Entity implements IAero, IBomber {
             }
             // calc MG Array here:
             if (wtype.hasFlag(WeaponType.F_MGA)) {
-                double mgaBV = 0;
-                for (Mounted possibleMG : getTotalWeaponList()) {
-                    if (possibleMG.getType().hasFlag(WeaponType.F_MG)
-                            && (possibleMG.getLocation() == mounted.getLocation())) {
-                        mgaBV += possibleMG.getType().getBV(this);
+                double mgBV = 0;
+                for (int eqNum : mounted.getBayWeapons()) {
+                    Mounted mg = getEquipment(eqNum);
+                    if ((mg != null) && (!mg.isDestroyed())) {
+                        mgBV += mg.getType().getBV(this);
                     }
                 }
-                dBV = mgaBV * 0.67;
+                dBV = mgBV * 0.67;
             }
             // artemis bumps up the value
             // PPC caps do, too
--- a/megamek/src/megamek/common/Dropship.java
+++ b/megamek/src/megamek/common/Dropship.java
@@ -957,14 +957,14 @@ public class Dropship extends SmallCraft {
                 }
                 // calc MG Array here:
                 if (wtype.hasFlag(WeaponType.F_MGA)) {
-                    double mgaBV = 0;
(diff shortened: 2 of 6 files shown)


megamek/src/megamek/common/Aero.java
```diff
@@ -1841,14 +1841,14 @@ public int calculateBattleValue(boolean ignoreC3, boolean ignorePilot) {
             }
             // calc MG Array here:
             if (wtype.hasFlag(WeaponType.F_MGA)) {
-                double mgaBV = 0;
-                for (Mounted possibleMG : getTotalWeaponList()) {
-                    if (possibleMG.getType().hasFlag(WeaponType.F_MG)
-                            && (possibleMG.getLocation() == weapon.getLocation())) {
-                        mgaBV += possibleMG.getType().getBV(this);
+                double mgBV = 0;
```
NickAragua, line 1844: This block of code appears to be repeated something like five times, perhaps we could create a utility function somewhere to reduce duplicated logic?

NickAragua (changes requested):

megamek/src/megamek/common/Aero.java
neoancient: Please read the description, where I addressed this.

NickAragua (approved):

PR #2461, opened 2020-12-01
allow LBX autocannons to swap ammo again

NickAragua:
Noticed during gameplay that a mech equipped with an LBX-10 autocannon was unable to swap from cluster to slug ammo. This PR addresses that issue, and tightens up MML ammo swap restrictions to ensure that both ammo types being considered are of the same general ammo type.


sixlettervariables (approved):
Looks good, pending CI.

Issue #2465, opened 2020-11-30
Cannot fix vehicle engine/motive damage


SuperStucco:
In 47.14, and for several prior releases, if I try to fix engine/motive damage on a vehicle in the Repair Bay it stays immobile (not just the listed 0 MP - actual motive damage).  Even under the GM tools for editing damage, I can unclick all the checked-off boxes for motive damage, click OK, and when checked again the boxes are still ticked.  Sometimes this goes away if I save and close MekHQ before trying repairs but not always.

Please check the attached campaign file.  There is a single vehicle present, with engine/motive damage.  In the Hangar, right-click and GM Mode => Edit damage...  Uncheck one or more boxes under Motive damage and OK the dialog.  Go to the Repair Bay tab and handle the Repair Motive System task (GM complete if necessary).  Go back to the Hangar and again go through GM Mode => Edit damage...  All motive damage boxes are still checked.  Uncheck all boxes, and OK the dialog.  Once again go through GM Mode => Edit damage...  All motive damage boxes are still checked.

[Vehicle engine fix.cpnx.gz](https://github.com/MegaMek/mekhq/files/5618256/Vehicle.engine.fix.cpnx.gz)

sixlettervariables:
Okay this is going to be resolved partially in MHQ when the next dev release comes out. You'll be able to use GM Restore or GM Edit Damage to fix crippled motive systems on vehicles.

NOTE: you will _still_ see Crippled if the vehicle is unmanned. That was not addressed in this fix.

PR #2466, opened 2020-12-02
Issue #2465: Ensure immobility takes into account the crew when applicable

sixlettervariables:
This fixes un-fixable immobility on the MekHQ side for tanks and support tanks, fixing #2465.

Basically we had one~two~ problem~s~:
1. `m_bImmobile` was never cleared or reset when editing a Tank in the `UnitEditorDialog`. I have gone ahead and updated the respective routines to reset this value and apply immobility when instructed by the user.
~2. Even after fixing that issue, if the unit had no crew, `isPermanentlyImmobile` would still return `true` if you explicitly asked to ignore crew effects. This is the more in-depth part of the fix.~

~To fix the second problem, I went ahead and plumbed `ignoreCrew` all the way through your getWalkMP/getRunMP/etc set of methods. I then added `boolean isImmobile(boolean checkCrew)` to Entity, which for Tanks and SupportTanks uses that in its calculation of Walk MP.~

~These fixes together allow you to GM repair motive damage to tanks in MekHQ (or GM Restore Unit).~

A separate issue, not addressed by this PR, is that an uncrewed vehicle will still appear Crippled.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/UnitEditorDialog.java
+++ b/megamek/src/megamek/client/ui/swing/UnitEditorDialog.java
@@ -692,8 +692,8 @@ public class UnitEditorDialog extends JDialog {
         panSystem.add(new JLabel("<html><b>" + Messages.getString("UnitEditorDialog.motiveDamage")
                 + "</b><br></html>"), gridBagConstraints);
         int motiveHits = 0;
-        // FIXME: motive hits arent working quite right
-        if (tank.isImmobile()) {
+        // Do not check the crew when determining if we're immobile here
+        if (tank.isImmobile(false)) {
             motiveHits = 4;
         } else if (tank.hasHeavyMovementDamage()) {
             motiveHits = 3;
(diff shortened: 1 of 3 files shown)


sixlettervariables:
If the plumbing through of `ignoreCrew` is too controversial, I can revert those changes and leave just the fix to `m_bImmobile`. This would leave uncrewed Tanks and Support Tanks as "permanently immobile" without a crew and unfixable in MHQ, but as soon as a crew was assigned, they would automagically appear "permanently mobile".

sixlettervariables:
Okay, after talking this out last night with @neoancient, it was obvious that I had way overengineered this fix.
Insert "There has got to be a better way!" meme: image

This is now targeted to the exact problem facing the `UnitEditorDialog`, but leaves Tanks and Trailers in the position that they will always be crippled without a crew...but that's a problem for another day.

neoancient (approved):

PR #2467, opened 2020-12-02
Remove final bv multiplier for tripod cockpits.

neoancient:
This was in the original rules in FM3145, but not in the latest rules in IO.


HammerGS (approved):
looks right to me

PR #2468, opened 2020-12-02
RISC laser pulse module BV

neoancient:
The RISC laser pulse module increases the linked laser's BV by 15%, not 25%. See IO, p. 196.


HammerGS (approved):
looks right to me

PR #2480, opened 2020-12-05
Loading transported units in MULs

SJuliez:
Transported units in MULs didn't load correctly when their ID was changed by the server. Now the server reconstructs the correct unit transportation when multiple units are added at once, i.e. when loading a MUL.
Things to note:
- I tested this with Jumpship loading, BA clamp loading, infantry loading, from the lobby, late deployment and ingame reinforcement, unload during deployment and it all seemed to work. Im still sure there are quite a few things and situations I haven't tested and might not work. And looking at the changes this may well be better for post-stable (even though MUL loading with transports is quite broken now)
- Fighter squadrons don't get saved. This is done by an explicit check that makes the saver method skip squadrons. When this check is removed, the squadrons do save but simply don't appear when loaded. I have not looked into this further.
- The server doesn't do much checking when reconstructing the load status. I assume that MULs with transports are saved from the lobby, not hand-assembled. In any case, when something is totally off, things should still work, only the units will simply not be transported.
- When a transport is deleted during deployment, all carried units are deleted as well. This was not so yet (I wonder why) but if it isn't done, the carried units don't deploy but throw NPEs. Also, where would they go otherwise. It is possible to unload them before removing a carrier.
- I removed some code that concerns an unofficial rule (BA grab handles for all meks and tanks). Not the whole thing, only part. It would clear out all "transports" (the handles and clamps), removing carried units with them. The comments already marked this as bad and it is. I admit I have not much love for these unoff rules that simply override a limitation that is part of BT. I'll happily spend effort to remove this rule, when asked.
- I also removed a code part that would set the deployment round for carried units as it doesn't work when reinforcing with bay-loaded units. IMO setting a deployment round for carried units should not be necessary and their deployment round shouldn't matter and shouldn't be asked for. If we need to set it, something's wrong anyway. Better fail immediately.
- Sadly had to add further code to the Server but with IDs overwritten on the server side I couldn't think of a way to handle this gracefully and reliably on the client.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/DeploymentDisplay.java
+++ b/megamek/src/megamek/client/ui/swing/DeploymentDisplay.java
@@ -395,6 +395,10 @@ public class DeploymentDisplay extends StatusBarPhaseDisplay {
     private void remove() {
         disableButtons();
         clientgui.getClient().sendDeleteEntity(cen);
+        // Also remove units that are carried by the present unit
+        for (Entity carried: clientgui.getClient().getGame().getEntity(cen).getLoadedUnits()) {
+            clientgui.getClient().sendDeleteEntity(carried.getId());
+        }
         cen = Entity.NONE;
     }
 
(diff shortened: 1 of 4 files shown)


NickAragua (approved):

sixlettervariables:
@SJuliez @NickAragua  Is this a candidate for stable or do we want to hold until after we cut 0.48?

NickAragua:
This can go in the release candidate.

PR #2494, opened 2020-12-09
Fix protomech myomer booster BV calculation

neoancient:
While most units use full running speed (2xwalk) for MASC or supercharger to calculate BV speed factor, protomechs with myomer booster use normal running + 1, along with half jump/umu as with other units.


NickAragua (approved):

neoancient:
Checks failed because my cherry-pick wasn't clean.

NickAragua (approved):

PR #2508, opened 2020-12-13
Colour Selector Button

Windchild292:
This adds a colour selector button to MegaMek for future use relating to colour preferences.


sixlettervariables (approved):
No skin in the 'u' game, but it would avoid discontinuity to match the field/method spelling with the type.

Windchild292:
Since I'm going to be doing the coding spelling it colour speeds up the process.

PR #2515, opened 2020-12-17
Fixing filename access level

Windchild292:
Fixes something I missed during a review, as part of researching the camo issues.


Windchild292:
My local environment didn't show the variable correctly, closing as the access change is proper.

PR #2525, opened 2020-12-21
VGL smoke munitions

### LINKED ISSUE #2522: Vehicular Grenade Launchers
Issue - unable to deploy smoke on any of the hexes within firing arcs.
Potential second issue - Not sure if Incendiary grenades should be able to set woods on fire or cause check for fire, but they do not and they add heat to firing unit if there is another unit in one of the target hexes.
Frags and Chaffs seems to be working fine (I am not familiar with VGL Chaff rules).

All above tested on Tracked Vehicles, Mechs, LAMs in Air Mech mode (same results) and Fighter mode (unable to employ any weapons, seems reasonable).

MM 0.46.1 ,0.47.7  & 0.47.15
Java 1.8.0_261
Windows 10 Home updated

neoancient:
VGL smoke ammo is treated as fragmentation because the ammo has the M_SMOKE flag and the weapon handler checks for M_SMOKEGRENADE. I don't see any reason to have a separate constant for smoke grenades, since the difference between smoke artillery and smoke grenades is in the handler. Smoke grenades are also using the report for heavy smoke instead of light smoke, so I fixed that as well.

Fixes #2522

### DIFF
--- a/megamek/src/megamek/client/bot/princess/FireControl.java
+++ b/megamek/src/megamek/client/bot/princess/FireControl.java
@@ -3041,7 +3041,6 @@ public class FireControl {
                 || AmmoType.M_PAINT_OBSCURANT == ammoType.getMunitionType()
                 || AmmoType.M_SMOKE == ammoType.getMunitionType()
                 || AmmoType.M_SMOKE_WARHEAD == ammoType.getMunitionType()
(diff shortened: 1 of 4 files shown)


sixlettervariables (approved):
Pending CI, but looks good (and frees up a munition type).

NickAragua (approved):

Issue #2537, opened 2020-12-27
NPE in do NuclearExplosion and setTurnIndex.  Plus other stuff in the log.


HammerGS:
Watching the Casual Joker game last night they had a hard crash. From the log.  This is the stream from a couple of minutes before the crash - https://www.twitch.tv/videos/851044772?t=3h21m10s

It would match up to the 23:34 time stamps in the logs.

```
21:06:46,854 INFO [megamek.server.Server] {Connection 2} 
disconnected(), line 1121 : s: removed player Krrast "First Casualty"
Contents of Send Queue: 
s:Contents of Send Queue: 
java.net.SocketException: Software caused connection abort: socket write error
	at java.net.SocketOutputStream.socketWrite0(Native Method)
	at java.net.SocketOutputStream.socketWrite(Unknown Source)
	at java.net.SocketOutputStream.write(Unknown Source)
	at java.io.BufferedOutputStream.flushBuffer(Unknown Source)
	at java.io.BufferedOutputStream.write(Unknown Source)
	at java.io.DataOutputStream.write(Unknown Source)
	at java.io.FilterOutputStream.write(Unknown Source)
	at megamek.common.net.DataStreamConnection.sendNetworkPacket(DataStreamConnection.java:114)
	at megamek.common.net.AbstractConnection.sendNow(AbstractConnection.java:281)
	at megamek.common.net.AbstractConnection.processPacket(AbstractConnection.java:517)
	at megamek.common.net.AbstractConnection.flush(AbstractConnection.java:480)
	at megamek.common.net.DataStreamConnection.flush(DataStreamConnection.java:124)
	at megamek.common.net.AbstractConnection.send(AbstractConnection.java:273)
	at megamek.server.Server.send(Server.java:31178)
	at megamek.server.Server.sendSaveGame(Server.java:1217)
	at megamek.server.commands.LocalSaveGameCommand.run(LocalSaveGameCommand.java:59)
	at megamek.server.Server.processCommand(Server.java:31212)
	at megamek.server.Server.handle(Server.java:31292)
	at megamek.server.Server$1.packetReceived(Server.java:397)
	at megamek.common.net.AbstractConnection.processConnectionEvent(AbstractConnection.java:603)
	at megamek.common.net.AbstractConnection.processPacket(AbstractConnection.java:508)
	at megamek.common.net.AbstractConnection.update(AbstractConnection.java:453)
	at megamek.server.ConnectionHandler.run(ConnectionHandler.java:49)

```

[...]

PR #2538, opened 2020-12-28
Fix nuke and turn npes

NickAragua:
Fixes behaviors described in #2537. 

Two changes:
- defensive coding in "doNuclearExplosion" to avoid referencing non-existent hexes
- when loading a save game, if the "current phase" doesn't have any more turns, end the phase and move on to avoid soft-lock


sixlettervariables (approved):
Are we generating a set of hexes to nuke and sending those along? If so should we ensure the board contains the hexes when we create them (before sending)?

I tried looking for it, but I got lost :)

NickAragua:
Nah, the nuke attack just has an impact site and a radius, and we generate all the coordinates within that radius during the processing method.

PR #2549, opened 2021-01-01
Fix rear/turret equipment loading

### LINKED ISSUE #2544: Rear or turret-mounted weapons failing to load
V.47.15

As title, nothing more to it. Ordering units by Level shows a bunch of new units with the F tag, all due to rear-mounted or turreted weapons.

neoancient:
Modifies case sensitivity support in MtfLoader. Instead of going back and forth between upper case and lower case, all tags are converted to upper case (and will be exported that way in the future). Each tag is stripped from the version of the string that is used for comparison, then after checking all tags the original is truncated and used for the case-sensitive equipment lookup.

While adding tests for the altered code I discovered and fixed a bug that caused armored slots with two mounts (used for superheavies with doubled heat sinks or ammo, or LAM bomb bays) not to be set as armored.

Fixes #2544

### DIFF
--- a/megamek/src/megamek/common/Mech.java
+++ b/megamek/src/megamek/common/Mech.java
@@ -2953,11 +2953,13 @@ public abstract class Mech extends Entity {
     }
 
     public Mounted addEquipment(EquipmentType etype, EquipmentType etype2,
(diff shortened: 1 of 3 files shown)


sixlettervariables (approved):
Good catch.

PR #2570, opened 2021-01-14
AbstractIcon: Full Camouflage Implementation

Windchild292:
This finishes implementing camouflage using AbstractIcon. The standard changes are the implementation of a refresh directory and a better way of handling individual camouflage in the dialog, plus some more dialog cleanup.

This MUST be merged alongside https://github.com/MegaMek/mekhq/pull/2371


megamek/src/megamek/server/ScenarioLoader.java
```diff
@@ -623,100 +618,21 @@ private void parseAutoEject(Entity entity, String eject) {
...
-        }
-        String camoName = getValidCamoName(camoGroup, camoData[1]);
-        if (camoName == null) {
-            throw new ScenarioLoaderException("invalidIndividualCamoName",
-                camoData[1], camoGroup, entity.getDisplayName());
-        }
-
-        entity.setCamoCategory(camoGroup);
-        entity.setCamoFileName(camoName);
+        entity.setCamouflage(new Camouflage(camoData[0], camoData[1]));
```
NickAragua, line 627: How does the camouflage constructor react to being passed bad data for either of those parameters?

Also, let's be a little more defensive here about the size of the array resulting from the split.

NickAragua (changes requested):
Needs a little bulletproofing, should be fine otherwise.

megamek/src/megamek/server/ScenarioLoader.java
Windchild292: If it returns null, it reverts to defaults. Otherwise, that's handled during camouflage creation by a fallback.

Windchild292: Pushed up a better way to handle this, and merged the two methods into a single safer method.

NickAragua (approved):

Issue #2565, opened 2021-01-13
46.16 - Princess becomes stalled on her turn to move units and does not recover


Morkath:
### Environment
0.47.16 RC2 - MekHQ
Windows
Java version 1.8.0_271

### Description
During the second round of a Hide and Seek (Attacker) mission against FRR, Princess became stalled during her turn to move units. and the phase does not progress.

Reloading the save does not fix the issue, she remains stalled.

Did not reproduce when starting the battle from scratch.

### Files
[megameklog.txt](https://github.com/MegaMek/megamek/files/5807790/megameklog.txt)
[TempSave.sav.gz](https://github.com/MegaMek/megamek/files/5807792/TempSave.sav.gz)

Name: My Campaign

BowtieBob:
After reloading the game, you should be able to go to "File/Game/Replace Player.." for each bot to reload them

NickAragua:
It's not a crash, the bot is actually taking a really long time to calculate the move for the Ostscout.

PR #2572, opened 2021-01-14
cache expensive bot computations; forego certain game events

NickAragua:
This addresses #2565.

Two performance improvements:
- cache the results of the isCommander() and isSubCommander() computations which contain multiple method calls, each of which loops through all of an entity's equipment multiple times. These are invariant while the bot is calculating a single unit's move.

- don't fire game events when flipping mech arms during a move calculation

The save game provided in the report went from taking about 30 seconds to evaluate each of the Ostscout's paths to taking about 30 seconds to evaluate the Ostscout's entire path set.

### DIFF
--- a/megamek/src/megamek/client/bot/princess/FireControl.java
+++ b/megamek/src/megamek/client/bot/princess/FireControl.java
@@ -1351,31 +1351,31 @@ public class FireControl {
         return 0;
     }
 
-    private boolean isCommander(final Entity entity) {
+    protected boolean isCommander(final Entity entity) {
         if (owner.getFireControlState().commanderCached(entity)) {
             return owner.getFireControlState().isCommander(entity);
-        } else {
-            owner.getFireControlState().setCommander(entity, 
-                    entity.isCommander() || entity.hasC3M() || entity.hasC3i() || entity.hasC3MM() ||
-                    (owner.getHighestEnemyInitiativeId() == entity.getId()));
+        } 
+
+        owner.getFireControlState().setCommander(entity, 
+                entity.isCommander() || entity.hasC3M() || entity.hasC3i() || entity.hasC3MM() ||
+                (owner.getHighestEnemyInitiativeId() == entity.getId()));
             
-            return owner.getFireControlState().isCommander(entity);
-        }
+        return owner.getFireControlState().isCommander(entity);
     }
 
-    private boolean isSubCommander(final Entity entity) {
+    protected boolean isSubCommander(final Entity entity) {
         if (owner.getFireControlState().subCommanderCached(entity)) {
             return owner.getFireControlState().isSubCommander(entity);
-        } else {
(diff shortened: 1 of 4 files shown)


megamek/src/megamek/client/bot/princess/FireControl.java
```diff
@@ -1352,16 +1352,30 @@ protected double calcCommandUtility(final Targetable target) {
...
     private boolean isSubCommander(final Entity entity) {
-        final int initBonus = entity.getHQIniBonus() + entity.getQuirkIniBonus();  //removed in IO + entity.getMDIniBonus()
-        return entity.hasC3() || entity.hasTAG() || entity.hasBoostedC3() || entity.hasNovaCEWS() ||
-               entity.isUsingSpotlight() || entity.hasBAP() || entity.hasActiveECM() || entity.hasActiveECCM() ||
-               entity.hasQuirk(OptionsConstants.QUIRK_POS_IMPROVED_SENSORS) || entity.hasEiCockpit() ||
-               (0 < initBonus);
+        if (owner.getFireControlState().subCommanderCached(entity)) {
+            return owner.getFireControlState().isSubCommander(entity);
+        } else {
+            final int initBonus = entity.getHQIniBonus() + entity.getQuirkIniBonus();  //removed in IO + entity.getMDIniBonus()
```
Windchild292, line 1370: Do we still need this commented out?

megamek/src/megamek/client/bot/princess/FireControl.java
```diff
@@ -1352,16 +1352,30 @@ protected double calcCommandUtility(final Targetable target) {
...
     }
 
     private boolean isSubCommander(final Entity entity) {
-        final int initBonus = entity.getHQIniBonus() + entity.getQuirkIniBonus();  //removed in IO + entity.getMDIniBonus()
-        return entity.hasC3() || entity.hasTAG() || entity.hasBoostedC3() || entity.hasNovaCEWS() ||
-               entity.isUsingSpotlight() || entity.hasBAP() || entity.hasActiveECM() || entity.hasActiveECCM() ||
-               entity.hasQuirk(OptionsConstants.QUIRK_POS_IMPROVED_SENSORS) || entity.hasEiCockpit() ||
-               (0 < initBonus);
+        if (owner.getFireControlState().subCommanderCached(entity)) {
+            return owner.getFireControlState().isSubCommander(entity);
```
Windchild292, line 1368: This can be pushed outside of the else, so you instead just check if it is not cached and add it if that is the case. Alternatively, setSubCommander could return the boolean.

Windchild292 (commented):
Refactor recommendation

sixlettervariables:
@NickAragua unit test failures in FireControlTest.

sixlettervariables (approved):
Looks good, I presume (for now) we shouldn't reassign commanders if things get broken.

NickAragua:
That's only a transient bot-level indicator, causing it to favor shooting at flagged commanders and units with C3/ECM, etc. The cached info gets cleared out every phase change (and I think during some updates, but I'm not sure).

PR #2574, opened 2021-01-14
Moving to fully use isConventionalInfantry

Windchild292:
This is the MegaMek part of a full repository swapover from using instanceof to determine if a unit is conventional infantry to the isConventionalInfantry method, which returns true for Infantry and all inheritors outside of BattleArmor and inheritors. It does not rely on the other two branches.

To Review:
Whitespace differences off (I found tab spacing in Entity and fixed some other whitespace issues, so this will also heavily drop the number of line changes)

Logic:
(Infantry && !BattleArmor): isConventionalInfantry
(!Infantry || BattleArmor): !isConventionalInfantry

### DIFF
--- a/megamek/src/megamek/client/bot/CEntity.java
+++ b/megamek/src/megamek/client/bot/CEntity.java
@@ -373,7 +373,7 @@ public class CEntity {
         }
 
         // Most units, including BA, are equipped with conventional weapons
-        if (!(entity instanceof Infantry) || (entity instanceof BattleArmor)) {
+        if (!entity.isConventionalInfantry()) {
 
             if (entity instanceof BattleArmor) {
                 number_of_shooters = ((BattleArmor) entity)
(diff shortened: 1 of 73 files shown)


megamek/src/megamek/client/bot/CEntity.java
```diff
@@ -668,7 +668,7 @@ public void characterize() {
         }
 
         if (entity instanceof Infantry) {
-            if (!(entity instanceof BattleArmor)) {
+            if (entity.isConventionalInfantry()) {
```
NickAragua, line 671: Under the hood, this introduces a redundant comparison.

NickAragua (changes requested):
One minor performance improvement, looks fine otherwise.

megamek/src/megamek/client/bot/CEntity.java
Windchild292: You'd prefer
```
if (entity.isConventionalInfantry()) {
} else if (entity instanceof BattleArmor) {
```
?

Because... performance-wise this is an improvement from the current check.

NickAragua: Looked under the hood at isConventionalInfantry(), and turns out it doesn't do what I thought it did. So never mind.

NickAragua (approved):

PR #2590, opened 2021-01-20
SkillLevel Enum

Windchild292:
This creates a SkillLevel enum, with the goal of standardizing how we handle the levels for skills between MegaMek and MekHQ. Currently there are multiple formats spread across the two code bases, and having an enum we can implement from will improve the code base.


megamek/i18n/megamek/common/enums.properties
```diff
@@ -0,0 +1,9 @@
+# This is used to store any common/enums Resources
+## Enums
+# SkillLevel Enum
+SkillLevel.NONE.text=None
+SkillLevel.ULTRA_GREEN.text=Ultra-Green
+SkillLevel.GREEN.text=Green
+SkillLevel.REGULAR.text=Regular
+SkillLevel.VETERAN.text=Veteran
+SkillLevel.ELITE.text=Elite
```
NickAragua, line 9: Might as well add a Heroic while we're here, I've seen it mentioned in a few of the rulebooks.

NickAragua (changes requested):

megamek/i18n/megamek/common/enums.properties
Windchild292: I forgot to close this one... was going to use it to simplify a few open PRs, but because this never merged I used worse workaround instead.

However, I do want to add a better level than current when I finish, so I will add it when I restart working on skills.

Issue #2586, opened 2021-01-19
Null-Signature System Should be Disabled Upon Hit Location Destruction


Windchild292:
TO: AU&E pg 148 states that the null-signature system should be disabled when any crit slot or hit location is hit.

NickAragua:
I can't reproduce this behavior. The system *says* it's on as per #2857, but targeting the unit with weapon attacks shows the proper modifiers.

Windchild292:
This is definitely not working if the mech loses a hit location.
![image](https://user-images.githubusercontent.com/39067288/105503971-3a202600-5c95-11eb-9067-f9abef6d1d32.png)

NickAragua:
Think I'm going to need a save, because I haven't been able to reproduce it in an arbitrary situation.

Windchild292:
Here's a save showing the null signature and chameleon cloak applying on a mech that doesn't have a left arm, and thus the system should be disabled.

[nullSigChameleonPreBug.sav.gz](https://github.com/MegaMek/megamek/files/5857106/nullSigChameleonPreBug.sav.gz)

[nullSigChameleonBug.sav.gz](https://github.com/MegaMek/megamek/files/5857081/nullSigChameleonBug.sav.gz)

Issue #2585, opened 2021-01-19
Chameleon Light Polarization Shield Should Be Destroyed Upon Hit Location Destruction


Windchild292:
TO: AU&E pg 112 states that the chameleon shield should be destroyed when any crit slot or hit location is destroyed.

NickAragua:
Fixed with PR #2596

PR #2596, opened 2021-01-23
more thorough destroyed location equipment destruction

NickAragua:
Fixes behaviors reported in #2585 and #2586, and probably a host of other similar behaviors besides.

The previous approach was to loop through all equipment on a unit, and, if it's mounted in the destroyed location, set hit/missing flag appropriately. However, this approach misses systems that are mounted all across a mech, such as chameleon, null sig, void sig, etc. 

Now, we take advantage of the fact that each critical slot has a piece of equipment associated with it (or two, sometimes for "compact" things?), and hit those pieces of equipment instead.

### DIFF
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -11476,25 +11476,28 @@ public abstract class Entity extends TurnOrdered implements Transporter, Targeta
                     engineHitsThisPhase++;
                 }
                 
+                boolean mountOneIsHittable = (cs.getMount() != null) && cs.getMount().getType().isHittable();
+                boolean mountTwoIsHittable = (cs.getMount2() != null) && cs.getMount2().getType().isHittable();
+                
                 if (blownOff) {
                     cs.setMissing(true);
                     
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/Entity.java
```diff
@@ -11484,13 +11475,35 @@ public void destroyLocation(int loc, boolean blownOff) {
                     && !cs.isDamaged()) {
                     engineHitsThisPhase++;
                 }
+                
+                boolean mountOneIsHittable = (cs.getMount() != null) && cs.getMount().getType().isHittable();
+                boolean mountTwoIsHittable = (cs.getMount2() != null) && cs.getMount2().getType().isHittable();
```
Windchild292, line 11480:
```suggestion
                final boolean mountOneIsHittable = (cs.getMount() != null) && cs.getMount().getType().isHittable();
                final boolean mountTwoIsHittable = (cs.getMount2() != null) && cs.getMount2().getType().isHittable();
```

Windchild292 (approved):
Making them final improves future readability

PR #2606, opened 2021-01-27
avoid building hash collision on maps with lots of buildings

### LINKED ISSUE #2580: 47.16 Building said it took damage during round report, but didn't actually take damage?
### Environment
47.16
Windows 10
Java 11.0.8
MekHQ integrated Megamek

### Description
My Atlas was facing a Warhammer standing on a 1-level tall Hardened Building with 117 CF. I decided to attack the building during both the Weapon Attack Phase and Physical Attack Phase to push its CF low enough for it to collapse. The Round Report recorded the building taking damage as normal, dropping to 75 CF during the Weapon Attack Phase, and then to 33 CF during the Physical Attack Phase - but it didn't actually collapse, and when I checked its CF out of confusion at the start of the next turn, its CF was still at 117? Very strange.

Weapon Attack Phase attacks on the building:
image

Physical Attack Phase attacks on the building:
image

Screenshot of Warhammer still standing on 117 CF building hex 1855 next turn:
image

### Files
Savegame (player name: Coppercoats):
[buildingBug.sav.gz](https://github.com/MegaMek/megamek/files/5825184/buildingBug.sav.gz)

Logs (didn't see anything obvious):
[megameklog.txt](https://github.com/MegaMek/megamek/files/5825182/megameklog.txt)
[mekhqlog.txt](https://github.com/MegaMek/megamek/files/5825183/mekhqlog.txt)

NickAragua:
Fixes #2580 

This revises the algorithm used in the Server.applyBuildingDamage() method, because apparently the hash function for the building results in collisions.

The basic principle is:
loop through all buildings
- for each coordinate within the building
-- if building at coordinate is at CF 0, add the coordinates to list of coordinates to collapse
-- if building CF at coordinate has changed this phase, add it to list of coordinates to update
- add building to list of buildings to update if at least one of its coordinates requires an update
collapse all coordinates in the collapse list
check all coordinates in update list for collapse
send building update for buildings requiring update

This has the drawback of sending unnecessary building update packets across the network when a building is collapsed due to heavy load, but that's generally an infrequent scenario.

Going to sleep on this overnight, as I wonder if this will break the use case where a mech tries to walk over a flimsy building.

### DIFF
--- a/megamek/src/megamek/server/Server.java
+++ b/megamek/src/megamek/server/Server.java
@@ -32509,16 +32509,20 @@ public class Server implements Runnable {
         // Build the collapse and update vectors as you go.
         // N.B. never, NEVER, collapse buildings while you are walking through
         // the Enumeration from megamek.common.Board#getBuildings.
-        Map<Building, Vector<Coords>> collapse = new HashMap<>();
-        Map<Building, Vector<Coords>> update = new HashMap<>();
+        Vector<Coords> collapseCoords = new Vector<>();
+        Vector<Coords> updateCoords = new Vector<>();
+        Vector<Building> buildingsToUpdate = new Vector<>();
+
         Enumeration<Building> buildings = game.getBoard().getBuildings();
         while (buildings.hasMoreElements()) {
             Building bldg = buildings.nextElement();
-            Vector<Coords> collapseCoords = new Vector<>();
-            Vector<Coords> updateCoords = new Vector<>();
             Enumeration<Coords> buildingCoords = bldg.getCoords();
+            
+            boolean buildingUpdated = false;
+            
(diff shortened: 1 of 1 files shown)


sixlettervariables:
For posterity:
- Hash collisions are an eventuality and are not an issue in a Map
- The collision seen was due to two different Buildings having the same ID
- The building ID is set to the hash of their first hex
- In this case, a hash collision would result in an equality collision between two distinct buildings

Future work:
- Buildings should have a unique ID:
    - Monotonically increasing ID
    - Or deterministic: hex.getX() * 10000 + hex.getY()

NickAragua:
Superceded by 2607.

PR #2610, opened 2021-01-28
prevent phantom buildings from crashing bot deployment

### LINKED ISSUE #2609: Princess AI cant deploy or move units in custom map.
[Stand alone megamek dev version 47.16, Windows 64bit ,Java version 1.8.0_271, player name BiggestBoss]

On a custom lunar tileset 40x40 map i have experianced an issue with the princess ai not deploying units except turrets on building and wall tiles.
I then tried connecting as the second player myself and deployed by hand,then disconnected and replaced my second instance of the game with the ai in hopes that it was just a deploy issue but then it turned out that even with the units deployed the ai would not make any moves.I will provide the megameklog document as well as the map and save in hopes that the issue can be found.

[bug log.txt](https://github.com/MegaMek/megamek/files/5888809/bug.log.txt)
[bug.sav.gz](https://github.com/MegaMek/megamek/files/5888813/bug.sav.gz)
[Lunar port base.zip](https://github.com/MegaMek/megamek/files/5888832/Lunar.port.base.zip)

NickAragua:
Fixes #2609

This error was caused by the presence of a building terrain type in a hex, but no CF. Likely this was due to a malformed map definition.

### DIFF
--- a/megamek/src/megamek/common/pathfinder/BoardClusterTracker.java
+++ b/megamek/src/megamek/common/pathfinder/BoardClusterTracker.java
@@ -24,6 +24,7 @@ import java.util.Map;
 import java.util.Set;
 
 import megamek.client.bot.princess.CardinalEdge;
+import megamek.common.Building;
 import megamek.common.BulldozerMovePath;
 import megamek.common.Coords;
 import megamek.common.Entity;
@@ -365,6 +366,12 @@ public class BoardClusterTracker {
         if (!hex.containsTerrain(Terrains.BLDG_CF) && !hex.containsExit(Terrains.FUEL_TANK_CF)) {
             return false;
         } else if (relevantMovementType == MovementType.Walker) {
+            Building building = board.getBuildingAt(coords);
+            
+            if (building == null) {
+                return false;
+            }
+            
             int buildingCF = board.getBuildingAt(coords).getCurrentCF(coords);
             
             return entity.getWeight() > buildingCF;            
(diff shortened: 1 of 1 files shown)


Windchild292 (approved):

megamek/src/megamek/common/pathfinder/BoardClusterTracker.java
```diff
@@ -365,7 +366,13 @@ private boolean buildingPlowThroughRequired(Entity entity, MovementType relevant
         if (!hex.containsTerrain(Terrains.BLDG_CF) && !hex.containsExit(Terrains.FUEL_TANK_CF)) {
             return false;
         } else if (relevantMovementType == MovementType.Walker) {
-            int buildingCF = board.getBuildingAt(coords).getCurrentCF(coords);
+            Building building = board.getBuildingAt(coords);
```
Windchild292, line 369: Could make this final

NickAragua: Does that help performance at all, or is it just a style preference?

Windchild292: The compiler might be able to improve the performance and it notes the variable won't change after being set.

sixlettervariables: The perf benefits only apply to values which can be statically known (e.g. effective constants), as in that case results can be pre-computed, memoized, or omitted. There exists the possibility the JVM could elide a barrier when reading from a _final class member_, but my understanding is that is not implemented. In all other cases it is a nice-to-have to signal intent to other programmers.

Consider for local variables that if the compiler can determine if you violate `final`, it also knows if it is _effectively_ `final`.

PR #2611, opened 2021-01-29
Use armor tech base instead of unit when calculating weight

neoancient:
A mixed tech jumpship/warship/space station calculates armor weight incorrectly for validation when the armor tech base differs from the unit tech base. It should use the tech base of the armor when calculating points per ton instead of the unit tech base.

Fixes MegaMek/megameklab#872

### DIFF
--- a/megamek/src/megamek/common/Jumpship.java
+++ b/megamek/src/megamek/common/Jumpship.java
@@ -2021,21 +2021,13 @@ public class Jumpship extends Aero {
         }
 
         // now I need to determine base armor points by type and weight
-        double baseArmor = 0.8;
-        if (isClan()) {
-            baseArmor = 1.0;
-        }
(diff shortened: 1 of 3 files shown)


megamek/src/megamek/common/Jumpship.java
```diff
@@ -2021,21 +2021,13 @@ public double getArmorWeight(int locCount) {
         }
 
         // now I need to determine base armor points by type and weight
-        double baseArmor = 0.8;
-        if (isClan()) {
-            baseArmor = 1.0;
-        }
+        boolean clan = TechConstants.isClan(getArmorTechLevel(firstArmorIndex()));
```
Windchild292, line 2024: This could be marked final.

Windchild292 (approved):

PR #2617, opened 2021-01-31
Fix elevation comparison when looking for mount target

### LINKED ISSUE #2598: Bug 46.16 Dropship can't mount(load units) if on hexes higher than level 2
### Environment
Stand Alone Mega Mek
Win 10
Java 1.8.0_261

### Description
See attached Board.  
Start Megamek.
Add 2 dropships. I've used the Manatee (Standard) and Seeker (Mech).
Add 2 Mechs. One for each dropship. Any will do.
Right click dropships, configure, Deployment Tab, Starting Altitude = 0.  Bug occurs at any altitude, this just makes testing easier. 
Place one dropship on attached 32x17 board in hex 1708 and another in 1310. 
Place mechs next to dropships.  The mech next to 1708 dropship can mount.  The other cannot. 
image
image
[Bug_01_23_2021.sav.gz](https://github.com/MegaMek/megamek/files/5861545/Bug_01_23_2021.sav.gz)

Attached save game in state that has one mech that is able to load and another that is not just based on underlying hex.

neoancient:
The check for whether the mount button should be enabled passes the elevation of the potential mounting unit relative to the ground, but the method that builds the list of mount targets expects the absolute elevation. The only other place the method is used is when presenting the list of mount targets to the user when the mount button is pressed, and the elevation is calculated correctly there.

Fixes #2598

### DIFF
--- a/megamek/src/megamek/common/Compute.java
+++ b/megamek/src/megamek/common/Compute.java
@@ -6714,10 +6714,8 @@ public class Compute {
      * @param game The game object
      * @return     All adjacent units that can mount the Entity
      */
-    public static List<Entity> getMountableUnits(Entity en, Coords pos,
-                                                      int elev, IGame game) {
-
-        ArrayList<Entity> mountable = new ArrayList<>();
+    public static List<Entity> getMountableUnits(Entity en, Coords pos, int elev, IGame game) {
+        List<Entity> mountable = new ArrayList<>();
         // Expanded to include trains
 
         // the rules don't say that the unit must be facing loader
--- a/megamek/src/megamek/common/Compute.java
+++ b/megamek/src/megamek/common/Compute.java
@@ -6705,10 +6705,19 @@ public class Compute {
         return acceptable;
     }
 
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/common/Compute.java
```diff
@@ -6705,10 +6705,19 @@ public static boolean inDeadZone(IGame game, Entity ae, Targetable target) {
...
+     * Builds a list of all adjacent units that can load the given Entity.
+     * @param en   The entity to load
+     * @param pos  The coordinates of the hex to load from
+     * @param elev The absolute elevation of the unit at the point of loading (surface
+     *             of the hex + elevation over the surface)
+     * @param game The game object
+     * @return     All adjacent units that can mount the Entity
+     */
+    public static List<Entity> getMountableUnits(Entity en, Coords pos,
                                                       int elev, IGame game) {
```
Windchild292, line 6718: This looks odd with the change from arraylist to list

Windchild292 (approved):

PR #2622, opened 2021-02-02
prevent turn skip on game load

NickAragua:
An off-by-one error would result in an entity's turn being skipped on game load if the entity was the only/last entity eligible for that phase.


sixlettervariables (approved):
Not sure how I missed that!

PR #2630, opened 2021-02-06
Fixing Infantry TRO Secondary Weapon Name

Windchild292:
This fixes a bug raised in https://github.com/MegaMek/mekhq/issues/2396


sixlettervariables (approved):
Subtle.

Issue #2597, opened 2021-01-23
Vehicle Numbering


BairdEC:
### Environment
0.47.16 - MM started from MHq
Java build 1.8.0_271-b09
Win10

### Description
MM imported three standard Manticore tanks from MHq, but numbered them 1, 1, and 2.

![Numbering](https://user-images.githubusercontent.com/38081112/105607273-0b6c8300-5d5b-11eb-9190-e2734493f72d.PNG)

### Files
[War Wolves30470519.cpnx.gz](https://github.com/MegaMek/megamek/files/5860495/War.Wolves30470519.cpnx.gz)
[customs.zip](https://github.com/MegaMek/megamek/files/5860498/customs.zip)

SuperStucco:
I've encountered the same thing with different tanks.  If you look at the screen grab, there's also multiple 'Vedette Medium Tank' without any numeric qualifiers.

I tried removing the extra tanks in the MM lobby and manually adding a replacement, each time they came in without the numeric qualifiers.

sixlettervariables:
Are there any logs? This is clearly a race condition in the unit naming logic on the MM side.

BairdEC:
Here you go.  I didn't grab the logs the first time around,  but the game did the same thing when I reloaded the save file.

[mekhqlog.txt](https://github.com/MegaMek/megamek/files/5897263/mekhqlog.txt)
[megameklog.txt](https://github.com/MegaMek/megamek/files/5897264/megameklog.txt)

sixlettervariables:
The logs don't show anything funny, but the code clearly can race between the accesses to the hash table. I'm working on a fix.

sixlettervariables:
This is actually an MHQ bug, interestingly enough.

BullseyeSmith:
https://github.com/MegaMek/mekhq/issues/2412

PR #2632, opened 2021-02-07
Issue #2597: Add thread safety to duplicate name hash

sixlettervariables:
The duplicate name protection code used a `Hashtable` which is thread safe, however, performing multiple operations is not atomic. This switches to use a lock to synchronize access to duplicate name generation. This also adds a lock to synchronize setting the short and display names. This makes both operations atomic and idempotent.

I don't think this is necessary for 0.48.

### DIFF
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -2541,14 +2541,14 @@ public abstract class Entity extends TurnOrdered implements Transporter, Targeta
      * @return A display name for the entity.
      */
     private String createDisplayName(int duplicateMarker) {
-        StringBuilder nbuf = new StringBuilder();
-        nbuf.append(createShortName(duplicateMarker));
+        StringBuilder builder = new StringBuilder();
+        builder.append(createShortName(duplicateMarker));
 
         if (getOwner() != null) {
-            nbuf.append(" (").append(getOwner().getName()).append(")");
+            builder.append(" (").append(getOwner().getName()).append(")");
         }
 
-        return nbuf.toString();
+        return builder.toString();
     }
 
     /**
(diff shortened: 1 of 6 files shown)


sixlettervariables:
I should add I was unable to replicate the original issue.

megamek/src/megamek/common/Entity.java
```diff
@@ -2459,34 +2514,41 @@ public String getDisplayName() {
...
+     }
+
+    /**
+     * Creates a display name for the entity.
+     * @param duplicateMarker A number used to disambiguate two entities with
+     *                        the same name.
+     * @return A display name for the entity.
+     */
+    private String createDisplayName(int duplicateMarker) {
+        StringBuilder nbuf = new StringBuilder();
```
Windchild292, line 2544: nbuf is a strange name for a StringBuilder

megamek/src/megamek/common/Entity.java
```diff
@@ -2507,11 +2569,20 @@ public String getShortName() {
...
+    }
+
+    /**
+     * Creates a short name for the entity.
+     * @param duplicateMarker A number used to disambiguate two entities with
+     *                        the same name.
+     * @return A short name for the entity.
+     */
+    private String createShortName(int duplicateMarker) {
+        StringBuilder nbuf = new StringBuilder();
```
Windchild292, line 2584: Same here

Windchild292 (approved):
Just two comments on variable name oddities

megamek/src/megamek/common/Entity.java
sixlettervariables: I left the original names in the refactor.

Windchild292:
Your discretion on 0.48, especially given https://github.com/MegaMek/mekhq/issues/2412

sixlettervariables:
I have never been able to reproduce this, so it may be worth seeing if one of the bug reporters can test a snapshot of this.

NickAragua (approved):

Windchild292 (approved):

PR #2631, opened 2021-02-07
2625: Preventing Key Bindings when using the Accessibility Window

Windchild292:
This fixes #2625 by preventing key bindings when using the accessibility window. The code change for the fix is commented.

PR #2649, opened 2021-02-11
Issue #2625: Change the default toggle chat to backquote / grave

### LINKED ISSUE #2625: "Enter," Chat Window Shortcut Conflicts with Accessibility SUpport
### Environment
What version of MegaMek does your issue apply to?  Megamek version 0.47.16, stand-alone.
What operating system are you using?  Mac.
What java version are you using? Java/OpenJDK version 1.8.0_282.

### Description
The default shortcut key used to open the chat window, "Enter," conflicts with the Accessibility window used to control the game with a screen reader. 

Steps:
1. Launch Megamek.
2. Open a scenario.
3. Open the accessibility window from the View menu.
4. Try to type a command, such as #help or #tile. Press Enter to submit.

Actual results:
The focus jumps from the Accessibility window, and the command is not submitted. The chat window presumably opens.

Expected Results:
Focus should remain in the command box. The command should be submitted. The chat window should remain closed.

Possible Resolutions:
Change the default chat key shortcut to something less frequently used, such as the grave accent. Long-term, perhaps work on making the chat window accessible and integrating the separate "accessibility window," functionality into it.

### Files
NA

sixlettervariables:
This is an opinionated follow on to #2631 that changes the default toggle chat key to backquote / grave. I'm not a fan of using plain ENTER as the default key as that has meanings all over the place in GUI land (such as "clicking the currently focused button"). Users that wish to have the old behavior can change the default key to ENTER, but going forward our app would be more accessible by default.

Fixes #2625 in another way.

### DIFF
--- a/megamek/docs/Default Key Binds.txt
+++ b/megamek/docs/Default Key Binds.txt
@@ -5,7 +5,7 @@ VIEW->CLIENT SETTINGS->KEY BINDS
 Toggle in-game keyboard shortcuts helper----- Ctrl-K
 
 GENERAL COMMANDS
-Toggle Chat----- Enter
+Toggle Chat----- ` (Backquote/Grave)
 Toggle Chat Command ----- /
 Undo----- Backspace
 Cancel----- Escape
--- a/megamek/mmconf/defaultKeyBinds.xml
+++ b/megamek/mmconf/defaultKeyBinds.xml
@@ -45,8 +45,8 @@
     </KeyBind>
 
     <KeyBind>
-        <command>toggleChat</command> <!-- Enter -->
-        <keyCode>10</keyCode>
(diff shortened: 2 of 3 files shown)


Windchild292 (changes requested):
You also need to update the default key binds text document and (I think, but not certain on this one) defaultKeyBinds.xml

sixlettervariables:
Good catch, my search for VK_ENTER definitely didn't find that!

Windchild292:
Thanks :)
I only know about the two because I looked at doing it myself, but decided on #2631 instead (although wasn't certain it was the best path forwards). There's also the Default Key Binds.txt that needs to be updated @sixlettervariables

Windchild292 (approved):

PR #2661, opened 2021-02-16
2272: Fixing Alt-Click on Building and Mouse Wheel Over Building/Bridge/Fuel Tank Not Clearing Tile

### LINKED ISSUE #2272: Map Editor Bug - ALT+Click on 'building' button acts like SHIFT being held
0.47.10 standalone
Windows 10
Java 11.0.8

When in Map Editor, `ALT+Click` on the 'building' button adds a building to the existing terrain in the terrain 'brush', as if `SHIFT` is also being held, rather than clearing the brush first.

Replication ...
`Click` on 'water' button to add `Water (depth 1)` to the brush.
`ALT+Click` on 'building' button will add a building to the brush, without removing the water.

Windchild292:
This fixes #2272, and additionally fixes having the middle mouse wheel not overwrite the hex setup if it doesn't contain at least one part of that type (i.e. you need to have one of the three types of key building segments or the paintbrush will be cleared).

To Review: first three commits are cleaning up code in the board editor, the last has the fixes for all four issues. The cleanup is mostly if statement formatting, removing NON-NLS comments, adding overrides, whitespace fixes, and fixing a pair of logging calls, and has no functionality changes.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/BoardEditor.java
+++ b/megamek/src/megamek/client/ui/swing/BoardEditor.java
@@ -579,7 +579,9 @@ public class BoardEditor extends JComponent
             button.setToolTipText(tt);
         }
         button.setMargin(new Insets(0,0,0,0));
-        if (bList != null) bList.add(button);
+        if (bList != null) {
+            bList.add(button);
+        }
         return button;
     }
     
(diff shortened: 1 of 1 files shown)


SJuliez (approved):
All good.

PR #2663, opened 2021-02-17
prevent error logging from crashing game

### LINKED ISSUE #2662: (MegaMek 0.47.17) NullPointerException Error Found During a Mission
### Environment
```
MekHQ v0.47.17 RC3
Windows 10
Java 11
```

### Description
Was playing in a Probe mission on the Attacker side, deployed two mech lances (I unfortunately forgot to take screenshots when I ran into the issue). On Round 3 after I collapsed a building a turret was on, Princess immediately entered a state of no action when the physical attack phase came up (I had simul. phys attacks on). Parsed both megameklog and mekhqlog and came across a NPE error in the mekhqlog.

### Files
[mekhqlog_ApocAngelsVSLyranCommon_no1.txt](https://github.com/MegaMek/mekhq/files/5992731/mekhqlog_ApocAngelsVSLyranCommon_no1.txt)
[megameklog.txt](https://github.com/MegaMek/mekhq/files/5992734/megameklog.txt)
[Bug Report Package.zip](https://github.com/MegaMek/mekhq/files/5992749/Bug.Report.Package.zip)

.zip file has the .cpnx file and the customized mechs at the time.

NickAragua:
Probably fixes #2662

getName() in Mounted throws a null reference exception when the Mounted has a null type for some reason, so we avoid using it. Also update the logging mechanism to not print to console.

### DIFF
--- a/megamek/src/megamek/common/Mounted.java
+++ b/megamek/src/megamek/common/Mounted.java
@@ -1261,8 +1261,7 @@ public class Mounted implements Serializable, RoundUpdated, PhaseUpdated {
             return 0;
         }
         // um, otherwise, I'm not sure
-        System.err.println("mounted: unable to determine explosion damage for "
-                + getName());
+        MegaMek.getLogger().error("mounted: unable to determine explosion damage for " + typeName);
         return 0;
     }
 


sixlettervariables (approved):
Odd that type would be null.

PR #2668, opened 2021-02-19
prevent bridge collapse when unit skids under it

NickAragua:
Fixes a behavior where, if a unit skids underneath a bridge, the bridge would collapse. 

I'll leave it up to you guys whether it's stable-worthy.


sixlettervariables (approved):
Now we can't skid into the bridge supports!

PR #2690, opened 2021-03-02
more robust orphan cleanup

NickAragua:
For some reason, the game's entities collection got out of sync with the rest of its entity management collections, so an entity destroyed in the physical phase was showing up in the game (though with 0 CT structure) when loading a save game made in the end phase.

This change makes game more aggressive about keeping the entities collection clear of entities being removed.

### DIFF
--- a/megamek/src/megamek/common/Game.java
+++ b/megamek/src/megamek/common/Game.java
@@ -1392,6 +1392,10 @@ public class Game implements Serializable, IGame {
      * (probably due to double-blind) ignore it.
      */
     public synchronized void removeEntity(int id, int condition) {
+        // always attempt to remove the entity with this ID from the entities collection
+        // as it may have gotten stuck there.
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/Game.java
```diff
@@ -1402,7 +1406,6 @@ public synchronized void removeEntity(int id, int condition) {
             return;
         }
 
-        entities.remove(toRemove);
         entityIds.remove(Integer.valueOf(id));
```
Windchild292, line 1409: Shouldn't we also remove the id before the if statement?

NickAragua: Probably wouldn't hurt, but, unlike entities, we don't use that lookup table to drive "get me all the entities in the game" functionality.

Windchild292 (approved):
Meant to approve with the recommendation.

PR #2725, opened 2021-03-18
Issue #2722: Do not send IP addresses in chat by default

### LINKED ISSUE #2722: RFE: Means of hiding IP in client chat box
From Reddit:

_This may be a bit of an odd question, but has there been any means of preventing Megamek from dumping the IP info of the server and client in the chat box upon start and upon user connection?I have looked through client settings, but found nothing like this.My motivation for this question is one related to privacy during public streams, the Megamek window can technically be cropped in capture software to hide the lower bar/chat, but this also removes the buttons from view (an important learning aspect for viewers), and does nothing in relation to the pop-up chat box that is active on the game screen.(Sidenote: I am aware that most of the IP dump server-side is generally local network data, but this is still something that is private. If someone joins in from an external IP, their IP is shown in chat, which is a concern.) I am ready to look at source code and file a ticket on the Megamek Github regarding this, but I wanted to ask before in case there have been any workarounds related to this._

_My current methodology to protect the IP of someone connecting from an external address is to use netcat (or a similar program) to listen on one port, and forward any connections to the Megamek server's actual port, which makes it appear as if a local client joined. However, this still does nothing to protect the server's own IP, as it is dumped upon startup._

### Environment
What version of MegaMek does your issue apply to?  0.48.0

sixlettervariables:
This PR does two things w.r.t. IP addresses:
1. Ensures they are all logged (server IP and port and client IPs)
2. Ensures they are not sent in chat publicly by default

Resolves #2722 

I've tried to keep this as simple as possible so it is stable backport compatible.

### DIFF
--- a/megamek/src/megamek/common/preference/ClientPreferences.java
+++ b/megamek/src/megamek/common/preference/ClientPreferences.java
@@ -53,6 +53,7 @@ class ClientPreferences extends PreferenceStoreProxy implements
         store.setDefault(MAP_HEIGHT, 1);
         store.setDefault(DEBUG_OUTPUT_ON,false);
         store.setDefault(MEMORY_DUMP_ON,false);
+        store.setDefault(IP_ADDRESSES_IN_CHAT, false);
         setLocale(store.getString(LOCALE));
         setMekHitLocLog();
     }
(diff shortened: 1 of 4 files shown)


NickAragua (approved):

sixlettervariables:
UI will be a separate PR.

sixlettervariables:
New default behavior:
image
image

Updating my preferences to show IPs:
image
image

PR #2732, opened 2021-03-19
Allow offboard artillery to disengage

### LINKED ISSUE #146: Disengage/Remove Unit Button for Offboard Artillery
Would it be possible to add a disengage (or enable the Flee) button for offboard units?  I would be nice to remove them from play when they run out of ammo, so you don't have to click through the targeting and designation phases. Also, eventually I would imagine the Counter-Battery fire rules will be implemented, and then it would be nice to escape when seriously damaged.

neoancient:
Adds a disengage button to the targeting phase display so offboard artillery can leave the fight once their ammunition is expended (or any other reason). This takes effect at the end of the targeting phase instead of the movement phase to keep bookkeeping down, but as the movement phase comes immediately afterward I don't see this as an issue.

Although we don't currently have counter-battery or aerial search and destroy missions (unless someone stuck it in while I wasn't looking) I took the possibility into account. A unit cannot disengage the same turn it fires, leaving it vulnerable to return fire, and the same checks for mobility and active crew are made before disengagement is allowed.

Closes #146

### DIFF
--- a/megamek/i18n/megamek/client/messages.properties
+++ b/megamek/i18n/megamek/client/messages.properties
@@ -2754,6 +2754,7 @@ TargetingPhaseDisplay.fireNext=Next Unit
 TargetingPhaseDisplay.fireSkip=Skip
 TargetingPhaseDisplay.fireTwist=Twist
 TargetingPhaseDisplay.fireSearchlight=Searchlight
+TargetingPhaseDisplay.fireDisengage=Disengage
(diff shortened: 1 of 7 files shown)


NickAragua (approved):
There actually is counter-battery fire - although in practice it gets used pretty rarely. Still, this is super. I'm going to update the bot to be able to disengage off-board artillery as well.

Issue #2694, opened 2021-03-05
47.17 Princess attempting an impossibly difficult PSR


gsparks3:
### Environment
47.17
Windows 10
Java 11.0.10

### Description
I have a savegame where Princess will attempt to stand up a Stalker mech, despite the difficulty of the PSR to stand up being 13 (i.e. impossible). This is obviously incorrect behavior, as the only possible outcome is to damage the unit further while randomly changing its facing. The pilot should either remain prone, or eject (probably the former until the 'mech is further damaged; while the unit in question is missing a leg, it is otherwise relatively intact and could continue fighting while prone). 

![image](https://user-images.githubusercontent.com/59574316/110068949-bf8eff80-7d44-11eb-9b01-85ae2d0ab98a.png)

Reproduction steps:
1. Load the provided savegame as Coppercoats.
2. Replace the MoC "players" with Princess bots on default settings.
3. Hit Done 5x to speed through the movement phase. Observe the movement log of the Stalker attempting to stand with "needs 13" and failing. 

### Files
[getUp.sav.gz](https://github.com/MegaMek/megamek/files/6088184/getUp.sav.gz)

NickAragua:
Fixed, probably, with #2735

PR #2735, opened 2021-03-20
enable the bot to consider sitting still or going to hull down

NickAragua:
Added a very simple "stay on the ground" pathfinder that generates a few possible paths for a mech that's prone - either turn in place or attempt to go 'hull down' then turn in place.

The reason for the latter being that going from prone to hull down is a guaranteed success, and then going from hull down to upright will never result in a fall.

This enables the bot to consider not trying to get up on very high odds of failure, fixing #2694


neoancient (commented):
It looks good except that the new file has tab indents.

Windchild292 (commented):
Still some newly added tab spacing in PathEnumerator, and let's fix the weird formatting for the commented out debug call

Windchild292 (approved):

PR #2742, opened 2021-03-23
enable Princess to find clubs

NickAragua:
Enables the bot to:
- pick up clubs (in the very rare situation where it has absolutely nothing better to do in the firing phase)
- use club attacks (again, in the very rare situation where it has absolutely nothing better to do in the physical phase)
- stop splattering ejected mechwarriors with physical attacks

Also improved defensiveness of some code.

### DIFF
--- a/megamek/src/megamek/client/bot/PhysicalCalculator.java
+++ b/megamek/src/megamek/client/bot/PhysicalCalculator.java
@@ -27,6 +27,7 @@ import megamek.common.IHex;
 import megamek.common.INarcPod;
 import megamek.common.Infantry;
 import megamek.common.Mech;
+import megamek.common.MechWarrior;
 import megamek.common.Mounted;
 import megamek.common.Protomech;
 import megamek.common.Tank;
@@ -281,17 +282,28 @@ public final class PhysicalCalculator {
         }
 
         for (Entity target : game.getEntitiesVector()) {
-
+            // don't consider myself
             if (target.equals(entity)) {
                 continue;
             }
+            
+            // don't consider friendly targets
             if (!target.isEnemyOf(entity)) {
                 continue;
             }
+            
+            // don't consider targets not on the board
             if (target.getPosition() == null) {
                 continue;
             }
+            
+            // don't consider targets beyond melee range
             if (Compute.effectiveDistance(game, entity, target) > 1) {
+                continue;
+            }
+            
+            // don't bother stomping mechwarriors
+            if (target instanceof MechWarrior) {
                 continue;
             }
 
--- a/megamek/src/megamek/client/bot/princess/FireControl.java
+++ b/megamek/src/megamek/client/bot/princess/FireControl.java
@@ -3224,6 +3224,7 @@ public class FireControl {
     /**
      * Return a "Find Club" action, if the unit in question can find a club.
      */
+    @Nullable
     public FindClubAction getFindClubAction(Entity shooter) {
         if (FindClubAction.canMechFindClub(shooter.getGame(), shooter.getId())) {
             FindClubAction findClubAction = new FindClubAction(shooter.getId());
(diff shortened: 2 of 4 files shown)


megamek/src/megamek/client/bot/PhysicalCalculator.java
```diff
@@ -281,19 +282,30 @@ public static PhysicalOption getBestPhysical(Entity entity, IGame game) {
...
                 continue;
             }
+            
+            // don't consider targets beyond melee range
             if (Compute.effectiveDistance(game, entity, target) > 1) {
                 continue;
             }
+            
+            // don't bother stomping mechwarriors
+            if (target instanceof MechWarrior) {
```
Windchild292, line 306: What about abandoned crews? Does this also handle that?

megamek/src/megamek/client/bot/princess/FireControl.java
```diff
@@ -3220,6 +3221,18 @@ Mounted getAntiAirAmmo(final List<Mounted> ammoList,
...
+    /**
+     * Return a "Find Club" action, if the unit in question can find a club.
+     */
+    public FindClubAction getFindClubAction(Entity shooter) {
+        if (FindClubAction.canMechFindClub(shooter.getGame(), shooter.getId())) {
+            FindClubAction findClubAction = new FindClubAction(shooter.getId());
+            return findClubAction;
+        }
+        
+        return null;
```
Windchild292, line 3233:
```suggestion
        return FindClubAction.canMechFindClub(shooter.getGame(), shooter.getId()) ? new FindClubAction(shooter.getId()) : null;
```

megamek/src/megamek/client/bot/princess/FireControl.java
```diff
@@ -3220,6 +3221,18 @@ Mounted getAntiAirAmmo(final List<Mounted> ammoList,
         return unjamVector;
     }
 
+    /**
+     * Return a "Find Club" action, if the unit in question can find a club.
+     */
+    public FindClubAction getFindClubAction(Entity shooter) {
```
Windchild292, line 3227: Please add Nullable Tag if something can be returned as null for ease of future use.

megamek/src/megamek/common/actions/FindClubAction.java
```diff
@@ -54,7 +54,7 @@ public FindClubAction(int entityId) {
      */
     public static boolean canMechFindClub(IGame game, int entityId) {
         final Entity entity = game.getEntity(entityId);
-        if (null == entity.getPosition()) {
+        if (null == entity || null == entity.getPosition()) {
```
Windchild292, line 57:
```suggestion
        if ((null == entity) || (null == entity.getPosition())) {
```

Windchild292 (commented):
In general looks good, just a few comments first.

megamek/src/megamek/client/bot/PhysicalCalculator.java
NickAragua: Probably, MechWarrior inherits from EjectedCrew.

megamek/src/megamek/client/bot/princess/FireControl.java
NickAragua: I don't really find ternary operators more readable in this particular situation.

megamek/src/megamek/client/bot/PhysicalCalculator.java
Windchild292: Wait... it would need to be EjectedCrew if we also wanted the bot to ignore them, instead of just MechWarrior, given MechWarrior extends EjectedCrew.

So, this should be swapped to target instanceof EjectedCrew if we don't want it to attack general ejected crews.

Windchild292: Good to know for future reference

Windchild292 (approved):

Issue #2747, opened 2021-03-25
PAC/8 Ammunition Explosion - Incorrect value


AlphaSled:
v.48.0

A PAC/8 ammunition explosion dealt an amount of damage above what was to be expected.

![Capture](https://user-images.githubusercontent.com/17106678/112405585-f3789780-8ce8-11eb-8237-baf2322676d7.PNG)

Relevant rules on p.47 of the BMM. In short, the ammo explosion should have dealt 56 damage instead of 448.

Save file attached.
[pacexplosion.sav.gz](https://github.com/MegaMek/megamek/files/6201437/pacexplosion.sav.gz)

NickAragua:
How the hell did those techs get 8x the rounds into that autocannon?

NickAragua:
Fixed with PR # 2748

PR #2748, opened 2021-03-25
tone down protomech AC ammo explosion damage

NickAragua:
This fixes the behavior reported in #2747 by bringing protomech AC "damage per shot" in line with all the other autocannon ammo types. The PACs still deal their "rack size" in damage.


HammerGS (approved):
Looks good to me.

PR #2752, opened 2021-03-28
Offboard Artillery Target Not Found NPE

Windchild292:
This fixes a NPE noted on Discord whereby an artillery target may return null. While doing this I also updated a bunch of the logging to use the logger instead of System.err and System.out and added some more nullable annotations.

To Review: Whitespace differences off.
I accidentally merged the first two commits together (cleaning up a few things in BoardView1 and the core bugfix), so I've noted where in BoardView1 the fix is located. View the first two commits separately from the rest.
Commit 1: Core fix and some light cleanup in BoardView1
Commit 2: adds the nullable annotation to Game::getEntity (which may return a null value)
Commit 3: Logging update for Game and BoardView1, fixing a few formatting issues.
Commit 4: Adding nullable annotations to

### DIFF
--- a/megamek/src/megamek/client/ui/swing/boardview/BoardView1.java
+++ b/megamek/src/megamek/client/ui/swing/boardview/BoardView1.java
@@ -1,7 +1,7 @@
 /*
 * MegaMek -
 * Copyright (C) 2000, 2001, 2002, 2003, 2004, 2005, 2006, 2007, 2008 Ben Mazur (bmazur@sev.org)
-* Copyright (C) 2018 The MegaMek Team
+* Copyright (C) 2018-2021 - The MegaMek Team. All Rights Reserved.
 *
 * This program is free software; you can redistribute it and/or modify it under
 * the terms of the GNU General Public License as published by the Free Software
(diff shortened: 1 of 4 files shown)


megamek/src/megamek/client/ui/swing/boardview/BoardView1.java
```diff
@@ -2157,32 +2135,32 @@ private void drawArtilleryHexes(Graphics g) {
...
-            ArtilleryAttackAction a = attacks.nextElement();
-            Coords c = a.getTarget(game).getPosition();
+        for (Enumeration<ArtilleryAttackAction> attacks = game.getArtilleryAttacks();
+             attacks.hasMoreElements(); ) {
+            final ArtilleryAttackAction attack = attacks.nextElement();
+            final Targetable target = attack.getTarget(game);
+            if (target == null) {
+                continue;
+            }
+            final Coords c = target.getPosition();
```
Windchild292, line 2145: This is the bugfix in BoardView

NickAragua (approved):

PR #2754, opened 2021-03-28
Fix scenario bot loading

TenkawaBC:
Fixing the bug of not loading bots properly when a scenario was created.

I added logging, added the connect() call that was missing, and switched to single loop from double-loop to optimize.

Note: The default "Bot" for scenario loading is TestBot. Princess is "Other Bot".  This should be looked at changing the default (ScenarioDialog.java) as Princess is now the Default outside this loader.


megamek/src/megamek/client/ui/swing/MegaMekGUI.java
```diff
@@ -749,19 +749,20 @@ public String getDescription() {
 
         // calculate initial BV
         server.calculatePlayerBVs();
-
+        
         // setup any bots
         for (int x = 0; x < pa.length; x++) {
             if (sd.playerTypes[x] == ScenarioDialog.T_BOT) {
```
Windchild292, line 755: Let us swap this to a switch statement, with an empty default

TenkawaBC: Let's make it an else if. Easy change and there are only three options. Testbot, princess, nothing.

sixlettervariables: Don't worry about TestBot, we're going to get rid of it in the 0.49 line. (Or should, it doesn't work)

TenkawaBC: New commit added to change to else if and switch Princess to be "Bot".  In prep for getting rid of TestBot (Then we can just drop Other Bot from the dropdown in ScenarioDialog and that is the extent of change).

Windchild292 (approved):

PR #2775, opened 2021-04-11
Scenario Additions (Planetary Cond. / Dialog skip)

### LINKED ISSUE #2133: Feature Request: Weather conditions settings in scenario (MMS) files
This is a request for the option to set weather conditions (blowing sand, wind speed, time of day) in scenario (.mms) files.  Or alternatively, if you could set these conditions in a scenario's game options XML file, that would be equally useful.

SJuliez:
- Adds complete planetary conditions settings to .mms scenarios
- Allows settings planetary conditions as fixed, skipping the respective dialog when the scenario starts
- Also allows setting the game options as fixed, skipping that dialog
- Also allows setting player assignment as fixed ("SinglePlayer"), skipping the player assignment/camo dialog and the host game dialog. In this case the first player of the scenario is the local player and the others are connected as Princess. When all is fixed, the game proceeds directly from scenario selection to in-game deployment/first round.
- Updated the example.mms with the new settings and missing settings (camo examples)

Resolves #2133

### DIFF
--- a/megamek/src/megamek/client/ui/swing/MegaMekGUI.java
+++ b/megamek/src/megamek/client/ui/swing/MegaMekGUI.java
@@ -46,6 +46,7 @@ import java.io.InputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Enumeration;
 import java.util.List;
 import java.util.Vector;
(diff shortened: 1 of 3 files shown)


megamek/src/megamek/server/ScenarioLoader.java
```diff
@@ -409,6 +446,69 @@ public IGame createGame() throws Exception {
         return g;
     }
 
+    private void parsePlanetaryConditions(Game g, StringMultiMap p) {
+        if (p.containsKey(PARAM_PLANETCOND_TEMP)) {
```
Windchild292, line 450: Can we separate these if statements so that there is a whitespace between the close `}` and the next if statement?

Windchild292: Also, if we are just going to use a setter immediately after can we just use the same line and combine them. Not much point to separate them out.

megamek/src/megamek/client/ui/swing/MegaMekGUI.java
```diff
@@ -675,71 +676,99 @@ public String getDescription() {
...
-        god.update(g.getOptions());
-        god.setEditable(true);
-        god.setVisible(true);
-        for (IBasicOption opt : god.getOptions()) {
-            IOption orig = g.getOptions().getOption(opt.getName());
-            orig.setValue(opt.getValue());
+        if (!sl.hasFixedGameOptions()) {
+            GameOptionsDialog god = new GameOptionsDialog(frame, g.getOptions(), false);
+            god.update(g.getOptions());
+            god.setEditable(true);
```
Windchild292, line 682: These two are odd... GameOptionsDialog::update nor GameOptionsDialog::setEditable should need to be called in this case.

megamek/src/megamek/client/ui/swing/MegaMekGUI.java
```diff
@@ -675,71 +676,99 @@ public String getDescription() {
...
         // popup planetary conditions dialog
-        PlanetaryConditionsDialog pcd = new PlanetaryConditionsDialog(frame, g.getPlanetaryConditions());
-        pcd.update(g.getPlanetaryConditions());
-        pcd.setVisible(true);
-        g.setPlanetaryConditions(pcd.getConditions());
-
-        // get player types and colors set
+        if (!sl.hasFixedPlanetCond()) {
+            PlanetaryConditionsDialog pcd = new PlanetaryConditionsDialog(frame, g.getPlanetaryConditions());
+            pcd.update(g.getPlanetaryConditions());
```
Windchild292, line 693: Same with this update call here.

Windchild292 (changes requested):
Few things first

megamek/src/megamek/server/ScenarioLoader.java
SJuliez: a) You mean an empty line after }? parsePlanetaryConditions is so uninteresting that I wanted it to be short. No complexity here. 
b) What? You mean like g.getPlanetaryConditions().setTemperature(Integer.parseInt(p.getString(PARAM_PLANETCOND_TEMP)));?
I can do that but it makes lines very long and Id rather have them like this than with a break in the middle.

megamek/src/megamek/client/ui/swing/MegaMekGUI.java
SJuliez: I did not think about these lines one bit, I only placed them in the if {} to make the dialog skippable.

megamek/src/megamek/client/ui/swing/MegaMekGUI.java
SJuliez: ditto.

megamek/src/megamek/server/ScenarioLoader.java
SJuliez: Arlith used to run a code formatter that would cut the lines very short iirc and (I think) sometimes result in awful line breaks. I'd rather break manually. But if no one does this, I can sure use long code lines.

Windchild292: Exactly like this. The extra space is an accessibility bit that assists in separating out the statements to a reader. Otherwise, it can come off as being a single giant chained if statement even though it is not the case

Windchild292 (approved):

PR #2781, opened 2021-04-14
2438: Adding More Detail to Fire Damage Reporting

### LINKED ISSUE #2438: RFE - Record hex where units are damaged by fire in round report/game log
I often use the Game Log post-fight to assign kills in AtB campaigns. However, there is a problem when doing this for units damaged or destroyed by fire. While ignited hexes are noted by coordinates (allowing determination of which unit was responsible for the hex being on fire), any units actually damaged by a fire do not have a hex specified, making it impossible to determine which unit was responsible for the kill.

Adding a note of which hex the unit occupies when it rolls for fire damage would fix this. Example: in the Round Report image below, after "on fire", add "in hex 1913" (the hex the infantry unit was occupying when checking for fire damage). 

image

Secondary request: when noting "Fire spreads to [hex coordinates]" in the round report, also note which hex it spread _from_, as in "Fire spreads to [hex coordinates A] from [hex coordinates B]". This would make it easier to trace back a fire that spread through several hexes before causing damage to a unit.

Windchild292:
This fixes #2438 by adding the origin coordinates for two separate fire messages.


megamek/i18n/megamek/common/report-messages_ru.properties
```diff
@@ -544,7 +543,6 @@
 5143=Ультра Густые Джунгли в <data> выгорают до Густых Джунглей!
 5145=Редкий Лес в <data> выгорает до Пересеченной местности и пожар гаснет!!
 5146=Редкие Джунгли в <data> выгорают до Пересеченной местности и пожар гаснет!!
-5150=Огонь переходит на <data>!
```
NickAragua, line 547: Огонь распространился с [data] на [data]!

NickAragua (approved):
Supplied russian translation. Slight rewording (the previous translation was more like "fire relocates, by means of walking, to [coordinates]", now it's more like "fire spreads from [coords] to [coords]"

PR #2789, opened 2021-04-20
second pass through mounted components

### LINKED ISSUE #2745: ProtoMechs - Arm destruction not disabling weapons
v.48.0

The destruction of a ProtoMech's arm location is not currently destroying/disabling weapons located in said arm when it should.
Relevant rules may be found on page 185 of Total Warfare. Save file where this occurs attached.

[savegame.sav.gz](https://github.com/MegaMek/megamek/files/6193178/savegame.sav.gz)

NickAragua:
Fixes #2745  - it turns out that some equipment may be present in a location but not in a critical slot. In addition to it being possible that a piece of equipment is present in a critical slot but spread across multiple locations (e.g. chameleon).

So, now we go both through all critical slots in the location to mark them and contents as hit, *and* all equipment and mark all equipment with that location as hit.

### DIFF
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -11566,6 +11566,21 @@ public abstract class Entity extends TurnOrdered implements Transporter, Targeta
             }
         }
         
+        // some equipment is not present in critical slots
+        // but is present in the location, so we'll need to look at it as well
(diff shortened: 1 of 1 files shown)


sixlettervariables (approved):
I think I discovered this oddity when trying to repair stuff on the MHQ side, but likely incorrectly assumed it was desired behavior.

Issue #2794, opened 2021-04-23
0.48 NPE Bot Config Dialog when replacing a player from a saved game toggle between Test Bot and Princess


HoneySkull:
### Environment
Starting MegaMek v0.48.0 ...
	Compiled on Fri Apr 16 10:23:22 CDT 2021
	Today is 2021-04-23
	Java vendor AdoptOpenJDK
	Java version 11.0.10
	Platform Windows 10 10.0 (amd64)
	Total memory available to MegaMek: 1,048,576 kB

### Description
When attempting to replace a player with a bot in a load from a saved game, if you click on the Test Bot radio button and then click back to the Princess Bot radio button on the Bot Config dialog, the dialog will remain blank.  (It goes blank when clicking on the Test bot, but does not subsequently show the Princess config controls when selecting back to the Princess settings.
See the Null Pointer exception in the attached logs.

1.  Load a save game that was saved with 2 or more players.
2.  Game->ReplacePlayer
3.  Click on the Test Bot radio button (Bot Config Dialog goes blank as expected)
4.  Then click back on the PrincessBot radio button.  
      **Expected:** PrincessBot controls are shown.  
      **Actual:** Dialog remains blank)

`java.lang.NullPointerException at megamek.client.ui.swing.BotConfigDialog.actionPerformed(BotConfigDialog.java:545)`

### Files
[AssassinSelfDestruct2.sav.gz](https://github.com/MegaMek/megamek/files/6362944/AssassinSelfDestruct2.sav.gz)
[megameklog.txt](https://github.com/MegaMek/megamek/files/6362950/megameklog.txt)

![ConfigureBotError](https://user-images.githubusercontent.com/83041327/115831418-2e004d80-a3d7-11eb-86ef-78ce8a3f5922.png)

HoneySkull:
This issue is easily worked around as that's all you need to do is close the Bot Config Dialog and Game->ReplacePlayer again and the controls are back.  This may not exist after the Lobby GUI changes are integrated, but haven't tested it with any PR versions yet.

NickAragua:
Simple solution: get rid of the testbot button.

HoneySkull:
I killed test bot!   <gulp>  R.I.P. test bot.  you had a good run.

PR #2795, opened 2021-04-24
remove testbot as option from bot config dialog

NickAragua:
This one's pretty straightforward - gets rid of the TestBot option on the add bot dialog, thus addressing #2794. 

Also implements a "better" version of disabling the option to re-name a bot, with an explicit parameter in the constructor. There will be a MekHQ companion for it that *does* allow editing bot names from the briefing screen.


NickAragua:
Never mind, established experimentally that bot name editing works just fine in MekHQ.

sixlettervariables (approved):

PR #2796, opened 2021-04-24
Issue #2704 Indicate that units may still move on facing arrow

### LINKED ISSUE #2704: Showing "has moved" status on icon
[48.0]

Unless I'm missing something there is no way to show whether a unit has moved on icons without viewing its tooltip.
It's an information that is vitally important and I constantly find myself doing the tooltip dance for it (which is not easy because on my setup I have to "nudge" the arrow for some reason for the tooltip to show).
**Would it be possible to show a small visual representation of "has moved" on the icons**, e.g. a colour dot, or better, changing the facing arrow colour? That would be awesome.

SJuliez:
Gives an indication if a unit can still move with the facing arrow. When stacked units with the same facing are present, the arrow will be doubled to give a hint to the player, see screenshot. Ive also greyed out the arrow a little when a unit has moved as the facing is already very obvious for most units from the icon. I think that even with the unit labels on, this is good info for the player.

Resolves #2704 

image


megamek/src/megamek/client/ui/swing/boardview/EntitySprite.java
```diff
@@ -609,7 +609,30 @@ public void prepare() {
                             && !((Infantry) entity).isTakingCover())
                     && !(isAero && ((IAero) entity).isSpheroid() && !board
                             .inSpace())) {
-                graph.draw(bv.facingPolys[entity.getFacing()]);
+                // Indicate a stacked unit with the same facing that can still move
+                if (shouldIndicateNotDone() && bv.game.getPhase() == Phase.PHASE_MOVEMENT) {
```
Windchild292, line 613: brackets

megamek/src/megamek/client/ui/swing/boardview/EntitySprite.java
```diff
@@ -609,7 +609,30 @@ public void prepare() {
...
+                    var tr = graph.getTransform();
+                    // rotate the arrow slightly
+                    graph.scale(1/bv.scale, 1/bv.scale);
+                    graph.rotate(Math.PI / 30, bv.hex_size.width/2, bv.hex_size.height/2);
```
Windchild292, line 617:
```suggestion
                    graph.scale(1 / bv.scale, 1 / bv.scale);
                    graph.rotate(Math.PI / 30, bv.hex_size.width / 2, bv.hex_size.height / 2);
```

megamek/src/megamek/client/ui/swing/boardview/EntitySprite.java
```diff
@@ -609,7 +609,30 @@ public void prepare() {
...
+                    graph.scale(bv.scale, bv.scale);
+                    graph.setColor(GUIPreferences.getInstance().getWarningColor());
+                    graph.fill(bv.facingPolys[entity.getFacing()]);
+                    graph.setColor(Color.LIGHT_GRAY);
+                    graph.draw(bv.facingPolys[entity.getFacing()]);
+                    graph.setTransform(tr);
+                }
+                if (!entity.isDone() && this.bv.game.getPhase() == Phase.PHASE_MOVEMENT) {
```
Windchild292, line 625: Space before, the this is likely pointless, and brackets

megamek/src/megamek/client/ui/swing/boardview/EntitySprite.java
```diff
@@ -622,12 +645,12 @@ public void prepare() {
             }
             // draw red secondary facing arrow if necessary
```
Windchild292, line 646: Should fix this comment

sixlettervariables:
So working clockwise starting with the selected unit:
1. Selected unit facing (red?) Is same as stacked unit facing (gray?)
2. Unit on it's own, cannot move
3. Unit facing different than stacked
4. Unit on its own, can or cannot move?

It's not immediately intuitive given the arrow stacking in case 1.

megamek/src/megamek/client/ui/swing/boardview/EntitySprite.java
```diff
@@ -664,6 +687,21 @@ public void prepare() {
         graph.dispose();
     }
 
+    /** 
+     * Returns true when an indicator should be shown that some of the stacked
+     * units in the hex of this entity's may still move even while the topmost
+     * shown unit may already have moved. 
+     */
+    private boolean shouldIndicateNotDone() {
+        for (var other: bv.game.getEntitiesVector(entity.getPosition())) {
```
Windchild292, line 696:
```suggestion
        for (var other : bv.game.getEntitiesVector(entity.getPosition())) {
```
The spacing is more accessible text, as it makes it into a separate word.

Windchild292: Might also be able to use a stream to check if any fit the check

Windchild292 (commented):
Some comments first

SJuliez:
@sixlettervariables 
I think it is much more obvious when you open a game in MM. In most cases it'll be just 2. and 4.; greyed out arrow means unit has moved; red arrow means unit can still move. When stacked units have different facings, that's no problem, the arrows do the same. When stacked units have the same facing I wanted to give some indication that there is still a unit there that may move even if the topmost unit has already moved, therefore the double arrow. Stacked units = UI hell. The topmost unit is not the selected unit. That would make sense, but it isnt currently.

megamek/src/megamek/client/ui/swing/boardview/EntitySprite.java
```diff
@@ -664,6 +689,19 @@ public void prepare() {
...
+     * Returns true when an indicator should be shown that some of the stacked
+     * units in the hex of this entity's may still move even while the topmost
+     * shown unit may already have moved. 
+     */
+    private boolean shouldIndicateNotDone() {
+        return (bv.game.getEntitiesVector(entity.getPosition()).indexOf(entity) == 1)
+                && bv.game.getEntitiesVector(entity.getPosition()).stream()
+                .filter(e -> e.getId() != entity.getId())
+                .filter(e -> e.getFacing() == entity.getFacing())
+                .anyMatch(e -> !e.isDone());
```
Windchild292, line 702:
```suggestion
        return bv.game.getEntitiesVector(entity.getPosition()).stream()
                .filter(e -> (e.getId() != entity.getId()) && (e.getFacing() == entity.getFacing()))
                .anyMatch(e -> !e.isDone());
```
A further improvement on the provided code would be a combination of the filters, and a removal of the default return.

Windchild292 (commented):
Looks good, just one simplification idea and conflicts need to be fixed.

I'll leave the colour default for a future PR, as I plan on making green defaults into a customizable colour once https://github.com/MegaMek/mekhq/pull/2284 is merged.

megamek/src/megamek/client/ui/swing/boardview/EntitySprite.java
SJuliez: I think I'll do this instead:

/** 
     * Returns true when an indicator should be shown that a unit with the same facing
     * as this unit is stacked below it and can still move.
     */
    private boolean shouldIndicateNotDone() {
        var hexEntities = bv.game.getEntitiesVector(entity.getPosition());
        return hexEntities.stream()
                .filter(e -> hexEntities.indexOf(entity) < hexEntities.indexOf(e))
                .filter(e -> e.getFacing() == entity.getFacing())
                .anyMatch(e -> !e.isDone());
    }

Actually the super special double arrow should be shown not only for the topmost unit but for any unit that is above another with the same facing that may still move, so I'll compare the indexOf instead which makes the id test unnecessary. On combining filters I really like leaving them apart even if that costs a line of code; maybe Im getting old but I find streams super useful but generally harder to read than good ole ifs due to all the additional .filter.stream.anyMatch()()() clutter. But keeping all logical steps apart in separate lines mitigates that. 
Sure gotta test this. Can't be that easy.

Windchild292: You could have the two compares done using an && on different lines.

I've got no proof, but I'd expect it to be slightly better performing

SJuliez: Better performance might be but I'm pretty certain it's irrelevant. I do like separating the filters into lines. Just says "no additional complexity here". (Even with the added clutter of .filter()). I also saw this used like this in all examples so I guess there's merit in it. I'll be obstinate here and keep it like that. Apart from that I think it works now and the merge conflict is gone.

Windchild292 (approved):

Tecmes:
Just one word, which isn't even a word: yay!

PR #2800, opened 2021-04-28
IDE Warning Cleanup (Eclipse) - Resolved 350ish IDE warnings from deprecations from Java 8,9, and 10

HoneySkull:
# Resolved several common categories of IDE warnings.  
This pull request doesn't correspond to a specific issue or request.  Instead is general cleanup of various categories of Eclipse IDE warnings, many resulting from obsolescence of Java 8, 9, and 10 constructs that have been deprecated.  It's somewhat important during development to maintain a short list of compiler warnings as numerous warnings cause background noise and mask potential interesting warnings.  This pull request contains several commits where I attempted to group each type of warning resolution for easier review.

I used @SuppressWarning sparingly and only in cases where I felt that it made sense to keep the original condition causing the warning such as in 'serialVersionUID' for serializable classes that are Swing GUI components as this is the same tactic used by the Java Swing library.

For the most part, these resolutions were automated by the Eclipse IDE.  There are no functional changes to the program from these changes but the end result is resolving all but 12 warning.  The remaining warning require at least some investigation (in my opinion) do resolve safely.  

Here is a high level of the resolutions:

## HexTarget deprecated constructor
Cleaned up Eclipse default IDE warnings for HexTarget.  Deprecated
constructor with unused parameter.  Cleaned up all calling references to
the non-deprecated constructor.

## Default Switch Statement Cases
The IDE's warn against switch statements where all enumeration types are not represented by case statements where no default statement exists.  In the context of resolutions here, the enumeration contained many elements and didn't make sense to include all elements in each switch statement.  In Java, "default" statements are optional.  As such the code here was not incorrect, but generated warnings because not all enum values had cases.  Adding the default case expresses the omission of some elements was intended, but more importantly make the compiler happy. :)

## Unused Java Import Statements
Over the course of modifying files, it's common for class files to retain Java import statement that are no longer used.  Eclipse has a handy Ctrl-Alt-O command that automatically removes and organizes java imports.  I do this automatically and instinctively.  Now that I stated that, you will see future commits where I forget to do that.  (I'm calling it now!)

## KeyEvents Deprecated in Java 9 and 10
Keyboard KeyEvents are more expressive in Java 11 and therefore have deprecated many event enumerations and functions for their 'Ex' variety that differentiates between key down and key up events.

## Removed unsupported @SuppressWarning statements
I removed a few "unsupported" suppress warning statement that seemed to have no negative side effects when removed.  I did not test with IDE's other than Eclipse, I suppose it may be possible that different IDE's utilize these warning types, but assume they are genuinely obsolete.  ( I would hate to @SuppressWarning ('unsupported suppresswarning')  LOL)

## Resolve serialVersionUID for Serializable Java Swing Derived Components
As per the Java Swing library references, I used @SuppressWarning('serial') at-will here.

## Unused Variables
Resolve 'unused variable' IDE warnings.  Only used SuppressWarning
annotation when appropriate in the context.  

## MMLogger Deprecated 'error()' Functions
Resolved and removed references and code for deprecated MMLogger::error() functions and removed deprecated functions from the class.  I did not however, clean up the '::info' or other deprecated functions in this class but should have.  It appears that most deprecated functions in the MMLogger class can now be safely eradicated.


megamek/unittests/megamek/client/bot/princess/NewtonAerospacePathRankerTest.java
```diff
@@ -58,10 +29,14 @@
...
-    private IGame mockGame;
-    private GameOptions mockGameOptions;
+    @SuppressWarnings("unused")
+	private Princess mockPrincess;
+    @SuppressWarnings("unused")
+	private NewtonianAerospacePathRanker mockPathRanker;
+    @SuppressWarnings("unused")
+	private IGame mockGame;
+    @SuppressWarnings("unused")
+	private GameOptions mockGameOptions;
```
HoneySkull, line 39: These private variables are part of a 'broken' jUnit test that had it's test method bodies commented out.   My first instinct beyond just deleting the entire broken test was to comment these variables out also to resolve the unused variable warning, however, I felt if keeping the test, it's better to apply the suppress warning to say, "fix me or delete me."  '@SuppressWarning("unused");' is searchable.

Windchild292:
Did you test this with MekHQ? Most of the logger deprecations were being kept because they are in active use within MekHQ and/or MegaMekLab.

megamek/src/megamek/client/ui/baseComponents/AbstractSplitPane.java
```diff
@@ -18,16 +18,16 @@
  */
 package megamek.client.ui.baseComponents;
 
+import java.awt.Component;
+import java.util.ResourceBundle;
+
+import javax.swing.JFrame;
+import javax.swing.JSplitPane;
```
Windchild292, line 25: We allow wildcard imports, especially for AWT and Swing

megamek/src/megamek/client/ui/baseComponents/AbstractButtonDialog.java
```diff
@@ -42,6 +42,7 @@
  * This is directly tied to MekHQ's AbstractMHQButtonDialog, and any changes here MUST be verified
  * there.
  */
+@SuppressWarnings("serial") // Same-version serialization only (See Swing base classes)
```
Windchild292, line 45: This is completely IDE setup dependent. My IDE (IDEA) gives me the exact opposite warning, mentioning that the suppression is redundant. Further, we do not employ serialization in any of our components.

megamek/src/megamek/client/ui/swing/StatusBarPhaseDisplay.java
```diff
@@ -59,8 +58,6 @@
         implements ActionListener, MouseListener, KeyListener, IPreferenceChangeListener {
     
     //L4J Support
```
Windchild292, line 60: Can remove this comment too.

megamek/src/megamek/common/FixedWingSupport.java
```diff
@@ -417,6 +417,7 @@ public double getPriceMultiplier() {
             case STATION_KEEPING:
                 priceMultiplier = 1 + weight / 75.0;
                 break;
+            default:
```
Windchild292, line 420: Better would be
```
default:
   break; 
```

megamek/src/megamek/server/Server.java
```diff
@@ -27973,7 +28018,7 @@ else if ((null != Compute.stackingViolation(game, other.getId(), curPos))
      * Possible to override 'is explosive' check
      */
     public Vector<Report> explodeEquipment(Entity en, int loc, Mounted mounted, boolean overrideExplosiveCheck) {
-        final String METHOD_NAME = "explodeEquipment(Entity,int,Mounted)";
+        //final String METHOD_NAME = "explodeEquipment(Entity,int,Mounted)";
```
Windchild292, line 28021: Just remove this, don't comment it out.

Windchild292 (commented):
Few comments and queries first.

Also, did you test this with MekHQ? We still use a bunch of deprecated logger functionalities there, and I'm pretty sure this will break that.

Finally, we use four spaces for spacing, not tabs. I noted a few places where this was not the case in the changes.

sixlettervariables (changes requested):
Please break this into separate pull requests based on the class of warning or message being resolved. Otherwise this is merge conflict city for other work.

megamek/src/megamek/client/ui/baseComponents/AbstractSplitPane.java
sixlettervariables: Otherwise they're a nightmare!

megamek/src/megamek/client/ui/baseComponents/AbstractButtonDialog.java
sixlettervariables: VSCode complains about this.

megamek/src/megamek/client/ui/baseComponents/AbstractSplitPane.java
HoneySkull: This was done automatically by Eclipse Ctrl-Shift-O that 'organizes' imports...  I'll revert this and clean it up manually when I split this into it's own pull request.

megamek/src/megamek/client/ui/baseComponents/AbstractButtonDialog.java
HoneySkull: I will revert this set of change.  I had a suspicion that these were IDE specific.  I can change my IDE settings to accommodate. I don't want to simply move the warning from one environment to another.  :)

megamek/src/megamek/common/FixedWingSupport.java
HoneySkull: Yep!  My bad.  Don't want somebody accidentally adding a case statement below the default and have it fall through unintended.

HoneySkull:
@sixlettervariables - I will break this into individual pull requests by warning type with the feedback from this PR.  What should I do with this PR?  Cancel/Close?  Convert to draft?

sixlettervariables:
You can use this branch for a specific type of warning or close it in favor of other PRs. No preference.

HoneySkull:
@Windchild292 - I did not test with MekHQ or MekLab - it didn't occur to me that those projects have a direct dependency on MegaMek.  The intent of this pull request was to resolve the warnings in MegaMek, so I will leave MMLogger as is.  I over reached by removing the methods.   Does it make sense to pull common code into a separate MM project?

Windchild292:
MekHQ relies on a pretty spread out grouping of code originating in MegaMek and MegaMekLab, so it really does not make sense to do so. Utilities like the logger are more likely to be shared between the two, but we also reuse some graphical components and similar.

HoneySkull:
The majority of these have been completed under separate pull requests.

PR #2811, opened 2021-05-01
fix save quirk command

NickAragua:
Apparently, the menu commands need a "NOINFO" between the command and unit IDs.


SJuliez (approved):
Yes that looks right.

PR #2827, opened 2021-05-04
Fix left click targeting issue

### LINKED ISSUE #2819: Left Clicking a Unit no longer targets the unit by default.
### Environment
v.0.49.0 Dev
Java vendor AdoptOpenJDK
Java version 11.0.10
Platform Windows 10 10.0 (amd64)
Total memory available to MegaMek: 1,048,576 kB

### Description
In past MM versions, left clicking a unit during the targeting phase would target the unit by default, enabling the 'Fire' tab. In this latest Dev version, a player must now right click the unit which then deployes the target/select drop down.

### Files

[megameklog.txt](https://github.com/MegaMek/megamek/files/6411924/megameklog.txt)

NickAragua:
Fixes #2819 

The first commit is coalescing the drag and click handling into doing the same thing, as the "drag" functionality in these displays didn't seem to be doing anything at all.

The second commit is cleaning up unused variables and enabling the user to flip arms without resetting torso twist state (technically it's a push/pop, but details).


sixlettervariables (approved):

megamek/src/megamek/client/ui/swing/PhysicalDisplay.java
```diff
@@ -1477,16 +1477,10 @@ public void hexMoused(BoardViewEvent b) {
         if ((b.getModifiers() & InputEvent.CTRL_DOWN_MASK) != 0) {
             return;
         }
-        if (clientgui.getClient().isMyTurn()
```
Windchild292, line 1480: Why remove this check? I did note it wasn't part of the other ones though

megamek/src/megamek/client/ui/swing/TargetingPhaseDisplay.java
```diff
@@ -1524,15 +1514,21 @@ public void actionPerformed(ActionEvent ev) {
     }
 
     private void updateFlipArms(boolean armsFlipped) {
+        if (ce() == null) {
+            return;
+        }
```
Windchild292, line 1519: This can potentially cause issues... we should really just have a local method variable for the entity, which can then be checked once, or better yet use the entity instead of the id for the whole method.

Windchild292 (commented):
Looks mostly good, just two comments first

megamek/src/megamek/client/ui/swing/TargetingPhaseDisplay.java
NickAragua: Think you're going to have to elaborate on that. Sometimes, the user doesn't have an entity selected for whatever reason (perhaps they're not the active player), in which case ce() would come back as null. Which would cause an exception on the line immediately following this.

Windchild292: Going from memory, the return from ce() might change between calls. So using Entity entity = ce(), then using entity for the method, should prevent issues from that.

NickAragua: Theoretically yes. In practice this would mean that the selected entity has been changed or the underlying game object received a "remove entity" update in the middle of this event handler (so, as the user is clicking 'flip arms', they've also somehow managed to simultaneously change the current entity or that entity has become destroyed). It's unlikely to happen during firing/targeting/physical phases, where nothing happens to an entity until the end of a phase. Let's leave it as is for now to keep a lid on scope.

megamek/src/megamek/client/ui/swing/PhysicalDisplay.java
NickAragua: Restored.

Windchild292 (approved):

HoneySkull:
I found the root cause as to why using the BUTTON1_DOWN_MASK behaves differently than the original BUTTON1_MASK.
With the deprecated `getModifiers()` method, the mask is set for button 1 on both the mouse down and mouse release.  With the updated `getModifiersEx()` the BUTTON1_DOWN_MASK is set on the mouse down but is NOT set on the mouse release. 

For example FireDisplay.java and others implement this check to bail if not mouse button 1:
 
        // ignore buttons other than 1
        if (!clientgui.getClient().isMyTurn()
            || ((b.getModifiers() & InputEvent.BUTTON1_DOWN_MASK) == 0)) {
            return;
        }```

This used to work find but on mouseReleased() will always look like the button1 was not used and will therefore exit before doing it's left-click payload such as selecting the hex for target.

To properly fix this, `BoardViewEvent` that is passed into board listeners `hexMouse()` functions needs to carry the mouse event button clicked from the mouse event along with the mask.

I believe, if I fix this properly, the "flip/twist" functionality can be restored correctly if desired.

HoneySkull:
For example, if the user presses button 1 followed by button 2, and then releases them in the same order, the following sequence of events is generated:

MOUSE_PRESSED:  BUTTON1_DOWN_MASK
MOUSE_PRESSED:  BUTTON1_DOWN_MASK | BUTTON2_DOWN_MASK
MOUSE_RELEASED: BUTTON2_DOWN_MASK
MOUSE_CLICKED:  BUTTON2_DOWN_MASK
MOUSE_RELEASED:
MOUSE_CLICKED:

//note the mask modifiers for the releases are 0 for the button being released.

PR #2830, opened 2021-05-04
Destroyed bottom decal issue

SJuliez:
Counting the number of available destroyed decals relied on ImageUtil.loadImageFromFile() returning null. But it no longer returns null when it can't find an image but returns the "fail standard image" red cross on white (which I think is good).

### DIFF
--- a/megamek/src/megamek/client/ui/swing/tileset/TilesetManager.java
+++ b/megamek/src/megamek/client/ui/swing/tileset/TilesetManager.java
@@ -1,18 +1,22 @@
 /*
-* MegaMek -
-* Copyright (C) 2002, 2003, 2004 Ben Mazur (bmazur@sev.org)
-* Copyright (C) 2018, 2020 The MegaMek Team
-*
-* This program is free software; you can redistribute it and/or modify it under
-* the terms of the GNU General Public License as published by the Free Software
-* Foundation; either version 2 of the License, or (at your option) any later
-* version.
-*
-* This program is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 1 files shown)


SJuliez:
image

megamek/src/megamek/client/ui/swing/tileset/TilesetManager.java
```diff
@@ -659,4 +627,29 @@ public synchronized void reset() {
...
+    private int getULightDecalCount() {
+        return getUltraDecalImgCount(FILENAME_SUFFIX_WRECKS_ULTRALIGHT);
+    }
+    
+    /** Returns the number of available ultraheavy destroyed bottom decal images. */
+    private int getUHeavyDecalCount() {
+        return getUltraDecalImgCount(FILENAME_SUFFIX_WRECKS_ASSAULTPLUS);
+    }
+    
+    private int getUltraDecalImgCount(String suffix) {
```
NickAragua, line 641: What's so Ultra about it?

NickAragua (changes requested):
Looks good, one amusing quibble.

megamek/src/megamek/client/ui/swing/tileset/TilesetManager.java
SJuliez: I called it ultra because it looks for the ultralight and ultraheavy decals... or maybe assaultplus and very light... had to use _some_ name

NickAragua: Well, anyway, it's probably a better approach than what I was using.

NickAragua (approved):

PR #2836, opened 2021-05-05
Root fix for Left click issues including movement and LOS issues. #2819, #2800

HoneySkull:
# Proposed fix for several left-click issues introduced with the upgrade to Java 11 Key events
This pull request fixes several left-click issues including movement, and LOS not updating during the movement phase.

One of they key differences between `getModifiersEx()` and the legacy `getModifiers()` is that the button modifiers are not carried with the mouseRelease event as they previously were.  The code here relied on the mouse modifiers to detect if button 1 was pressed on mouse up (BOARD_HEX_CLICK) events.  

The solution here passes a separate value carrying the mouse button that was associated to the event.

With this change it would be possible to restore the torso twist/flip code that was removed in #2819 if desired to cover those cases as they worked before.  The fix previously implemented in #2819 works but is not a full fix because it forces the click and drag functionality together (i.e. the code that used to run only on mouse up now runs on mouse up and mouse down etc)  It also didn't handle cases for movement, deployment, accessibility, etc.

Please consider this set of changes as a more complete fix for the mouse issues.

I've manually tested impacted areas to the best of my app knowledge with the exception of mine deployment.
I am unsure how to test these changes with MekHQ but am open to suggestions.

## Fixes issues during movement phase including updating line of sight during movement.

### DIFF
--- a/megamek/src/megamek/client/event/BoardViewEvent.java
+++ b/megamek/src/megamek/client/event/BoardViewEvent.java
@@ -47,6 +47,7 @@ public class BoardViewEvent extends java.util.EventObject {
     private int type;
     private int modifiers;
     private int entityId;
+    private int mouseButton = 0;
 
     public BoardViewEvent(Object source, Coords c, Entity entity, int type,
             int modifiers) {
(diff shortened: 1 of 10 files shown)


NickAragua:
This looks pretty good, let me take a look at backing out the changes I made tomorrow.

NickAragua:
Thanks for the fix, turns out what I did actually broke charges/DFA and that sort of thing, so this is a lot better.

PR #2849, opened 2021-05-10
Abstract Wreck Sprite NPE

Windchild292:
This fixes a NPE raised in https://github.com/MegaMek/mekhq/issues/2598. Commit 1 is cleanup, commit two fixes the bug

### DIFF
--- a/megamek/src/megamek/client/ui/swing/boardview/AbstractWreckSprite.java
+++ b/megamek/src/megamek/client/ui/swing/boardview/AbstractWreckSprite.java
@@ -79,8 +79,8 @@ public abstract class AbstractWreckSprite extends Sprite {
                 .getAscent());
 
         // create image for buffer
-        final Rectangle bounds = getBounds(); // more efficient than recreation
-        image = ImageUtil.createAcceleratedImage(bounds.width, bounds.height);
+        final Rectangle imageBounds = getBounds(); // more efficient than double recreation
+        image = ImageUtil.createAcceleratedImage(imageBounds.width, imageBounds.height);
         Graphics2D graph = (Graphics2D) image.getGraphics();
         
         // if the entity is underwater or would sink underwater, we want to make the wreckage translucent
--- a/megamek/src/megamek/client/ui/swing/boardview/AbstractWreckSprite.java
+++ b/megamek/src/megamek/client/ui/swing/boardview/AbstractWreckSprite.java
@@ -79,6 +79,7 @@ public abstract class AbstractWreckSprite extends Sprite {
                 .getAscent());
 
         // create image for buffer
+        final Rectangle bounds = getBounds(); // more efficient than recreation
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/client/ui/swing/boardview/AbstractWreckSprite.java
```diff
@@ -70,12 +73,13 @@ public Rectangle getBounds() {
...
         // figure out size
         String shortName = entity.getShortName();
-        Font font = new Font("SansSerif", Font.PLAIN, 10); //$NON-NLS-1$
+        Font font = new Font("SansSerif", Font.PLAIN, 10);
         Rectangle tempRect = new Rectangle(47, 55, bv.getFontMetrics(font)
                 .stringWidth(shortName) + 1, bv.getFontMetrics(font)
                 .getAscent());
 
         // create image for buffer
+        final Rectangle bounds = getBounds(); // more efficient than recreation
```
NickAragua, line 82: First of all, why are we hiding the underlying class member variable?

Second, what this function call does is recreate the bounds from scratch (it invokes the AbstractWreckSprite implementation of getBounds()), so the comment is basically the exact opposite of what the code does.

I'm not saying it's a bad fix, but I think it needs a little elaboration, and the comment needs to be updated.

NickAragua (changes requested):

megamek/src/megamek/client/ui/swing/boardview/AbstractWreckSprite.java
Windchild292: I missed the word double here... because if we called getBounds it would recreate it twice.

I looked at using this getter before two super.getBounds(), this was the easier version to read.

NickAragua: Ok, let's just avoid hiding the member variable then and update the comment.

NickAragua: You could also do that.

NickAragua: I was thinking update the local variable name, but either solution works.

NickAragua: Although, I think your solution is probably better, since it forces an update on the class variable in case it's null or something dumb like that (which apparently it is in some cases).

NickAragua: actually, it forces an update on the class variable either way, so never mind. Up to you.

Windchild292: Updated the variable name and comment

NickAragua (approved):

PR #2852, opened 2021-05-11
Fluff transitions

SJuliez:
Adds gfx to the saxarba tilesets to allow ground texture transitions as in the screenshot. A How To image is included in the docs/boards stuff directory. Also a slight change to the BoardEditor: Adding terrain (shift-clicking) no longer overwrites the theme.
Also, adds the ability to write 
super >4 "ground_fluff:1:05" "" "saxarba/Transitions/toDesert4_100.png"
in a tileset where the ">4" makes it appear for all hex levels of 4 or more.
The code for this is not lovely but I had no better idea that would work without making much bigger changes to Hex and the way the level comparison is done.

image

image


Windchild292 (approved):
For 0.49.2

PR #2880, opened 2021-05-26
Dehexed water and pavement

SJuliez:
For the saxarba tileset, this makes water less hexy by making it use exits like roads do. This happens automatically and for existing maps. The behaviour can be overridden by manually setting the exits, so rivers exiting maps or other stuff can still be edited.

Also adds a version of saxarba's pavement that is also unhexed (pavement:4). It looks like standard pavement but does not have different light/dark versions for hex level (this is possible but would add another few hundred images). It works like the water does but does not get applied automatically to existing maps as it usually covers less area than the hex-y pavement and thus may not touch buildings or roads as intended.

image

image


SJuliez:
I really don't know what the Validate Boards test does there

megamek/src/megamek/client/ui/swing/boardview/BoardView1.java
```diff
@@ -2748,13 +2749,6 @@ private void drawHex(Coords c, Graphics boardGraph,
                 if (!useIsometric()) {
                     g.drawImage(scaledImage, 0, 0, this);
                 }
-                // draw a shadow for bridge hex.
-                if (useIsometric()
```
NickAragua, line 2752: No more bridge shadows? :(

megamek/src/megamek/common/Hex.java
```diff
@@ -181,6 +181,12 @@ public void setExits(IHex other, int direction, boolean roadsAutoExit) {
             }
 
             cTerr.setExit(direction, cTerr.exitsTo(oTerr));
+            
+            // Water gets a special treatment: Water at the board edge
+            // (hex == null) should usually look like ocean and 
+            // therefore always gets connection to outside the board 
+            if ((cTerr.getType() == Terrains.WATER) && (other == null))
+            	cTerr.setExit(direction, true);
```
NickAragua, line 189: one-liner if blocks should still get { }

NickAragua (commented):
One question, one minor formatting quibble.

megamek/src/megamek/client/ui/swing/boardview/BoardView1.java
SJuliez: They're still there! Just a few lines above that.

megamek/src/megamek/common/Hex.java
SJuliez: That code bit is from 4 years ago when I hadnt seen the formatting guide :)

NickAragua (approved):

PR #2885, opened 2021-05-27
KeyCommandBinds maintenance and defaults correction

SJuliez:
I think that keycommands using modifiers all stopped working with the recent replacement of deprecated modifier constants as the new modifier constants use different values. These were not updated in the defaultkeybinds.xml. Correcting this.
Also some code maintenance on the KeyCommandBinds and MegaMekController.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/util/MegaMekController.java
+++ b/megamek/src/megamek/client/ui/swing/util/MegaMekController.java
@@ -1,33 +1,29 @@
 /*
- * MegaMek -
  * Copyright (C) 2000,2001,2002,2003,2004,2005 Ben Mazur (bmazur@sev.org)
- * Copyright © 2013 Nicholas Walczak (walczak@cs.umn.edu)
+ * Copyright (c) 2013 Nicholas Walczak (walczak@cs.umn.edu)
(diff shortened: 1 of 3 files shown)


megamek/src/megamek/client/ui/swing/util/MegaMekController.java
```diff
@@ -105,27 +92,18 @@ public boolean dispatchKeyEvent(KeyEvent evt) {
...
-                modifiers);
 
         // If there's no action associated with this key bind, or the
         // current action is invalid, do not consume this event.
         boolean consumed = false;
 
-        for (KeyCommandBind kcb : kcbs) {
-            // Do we have a binding for this key?
-            if (!keyCmdSet.contains(kcb)) {
+        for (var kcb : KeyCommandBind.getBindByKey(keyCode, modifiers)) {
```
NickAragua, line 100: It doesn't affect functionality, but, in general, I prefer to leave this sort of thing as it was previously (declare and assign variable then use it in a loop), the reason being that I can more easily, while debugging, examine the value that the function call produces.

NickAragua (approved):

PR #2899, opened 2021-06-02
Refactor/test server victory

FreekDS:
In this PR the victory method of the Server (and methods that are used by this method) are refactored to improve overall code health. Tests are added to ensure the refactor actions did not break anything.
The following changes were made:

- Extract duplicate code from ``VictoryResult`` (_getWinningPlayer()_ and _getWinningTeam()_) into a generic private method.
- Add a method _getVictoryResult()_ to ``Game`` to eliminate navigation code in the _victory()_ method (moves the behavior closer to the data)
- Moved method _cancelVictory()_ to ``Game`` as well (moves behavior closer to the data). Redundant method calls are removed from the _victory()_ method in the process
- Move the static _getColorForPlayer()_ of the Server to a member method of Player instead.
- Move the behavior of the _victory()_ method to a new method _processVictory()_ in ``VictoryResult`` (moves behavior closer to the data)
- Remove some code smells from `Player`, `Victory` and `VictoryResult` classes.

Methods that no longer have any use after these refactorings are marked as Deprecated. Each touched part of the code base is covered by new test cases.


megamek/unittests/megamek/common/GameTest.java
```diff
@@ -0,0 +1,59 @@
+package megamek.common;
+
+import junit.framework.TestCase;
```
sixlettervariables, line 3: We would prefer `org.junit.Assert` in line with the existing tests. That is, unless there is a reason to avoid that class in JUnit4.

SJuliez:
Entity is the base class for all game units. Therefore I found "getWinningEntity()" in VictoryResult confusing. I suggest using another name, even getWinningPlayerOrTeam

megamek/unittests/megamek/common/GameTest.java
FreekDS: Thanks for your comment. It is changed to `org.junit.Assert` in 
20827f7

NickAragua:
Let's get the unit tests fixed before merging this.

NickAragua:
Sorry, it took me so long to revisit this. One of the unit tests is still failing. Unfortunately, I have to manually authorize the unit tests to run, could you make a post here when you make the necessary update?

FreekDS:
The problem should be ressolved by now. It did not occur in my local Windows branch, which is the reason it wasn't fixed in the previous commit. 

The problem was that by creating tests for the Server, the static member `serverInstance` is set to the test instance.
In `Compute.d6(int)` it is checked whether this instance is not `null` which it wasn't due to the specific order of executing the tests on the Ubuntu distro.

By mocking `getOptions()` of our mocked game in the server tests, we can workaround the nullpointer exception.

PR #2903, opened 2021-06-03
Camo overlay

SJuliez:
Adds a camo overlay to fake some facing-dependent shadowing and highlighting to the units (except infantry, they're too small for that). It's default on with an advanced setting to turn it off.

Changes this image to this image
In-game:
image


NickAragua (approved):
Looks good.

Windchild292 (approved):

PR #2908, opened 2021-06-04
Lobby PlayerTable oddity

SJuliez:
With the new lobby, the player table remembers now what players were previously selected when the client receives any changes from the server (instead of reverting to having the first player selected as before 0.49). This had the effect that when starting a new game, no player was selected. This PR changes this to select the first player (=self) when no player would otherwise be selected.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/lobby/ChatLounge.java
+++ b/megamek/src/megamek/client/ui/swing/lobby/ChatLounge.java
@@ -1182,6 +1182,7 @@ public class ChatLounge extends AbstractPhaseDisplay implements
                 tablePlayers.addRowSelectionInterval(row, row);
             }
         }
+        // If no one is selected now (and the table isn't empty), select the first player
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/client/ui/swing/lobby/ChatLounge.java
```diff
@@ -1182,6 +1182,10 @@ private void refreshPlayerTable() {
                 tablePlayers.addRowSelectionInterval(row, row);
             }
         }
+        // If no one is selected now (and the table isn't empty), select the first player
+        if (tablePlayers.getSelectedRowCount() == 0 && tablePlayers.getRowCount() > 0) {
```
Windchild292, line 1186: brackets

Windchild292 (approved):
Looks good in general, just one comment before merge

megamek/src/megamek/client/ui/swing/lobby/ChatLounge.java
SJuliez: Right! Done.

PR #2909, opened 2021-06-04
Small AbstractDialog rearrangement

SJuliez:
Small suggestions for AbstractDialog (since I have now been looking at it for quite a bit): 
- Most of the initialization bit can be moved up to the constructor as it needs no dialog content, so that's where I would put it.
- BorderLayout is the default layout of JDialog, so setting it isn't necessary.
- The entity getter seems overkill in the subclassed BVDialog.

### DIFF
--- a/megamek/src/megamek/client/ui/baseComponents/AbstractDialog.java
+++ b/megamek/src/megamek/client/ui/baseComponents/AbstractDialog.java
@@ -75,6 +75,21 @@ public abstract class AbstractDialog extends JDialog implements WindowListener {
         setName(name);
         setFrame(frame);
         this.resources = resources;
+        // Escape keypress
+        final KeyStroke escape = KeyStroke.getKeyStroke(KeyEvent.VK_ESCAPE, 0);
+        getRootPane().getInputMap(JComponent.WHEN_IN_FOCUSED_WINDOW).put(escape, CLOSE_ACTION);
+        getRootPane().getInputMap(JComponent.WHEN_FOCUSED).put(escape, CLOSE_ACTION);
+        getRootPane().getActionMap().put(CLOSE_ACTION, new AbstractAction() {
+            private static final long serialVersionUID = 95171770700983453L;
+
+            @Override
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/client/ui/baseComponents/AbstractDialog.java
```diff
@@ -113,22 +127,6 @@ protected void initialize() {
...
-            private static final long serialVersionUID = 95171770700983453L;
-
-            @Override
-            public void actionPerformed(ActionEvent evt) {
-                cancelActionPerformed(evt);
-            }
-        });
-
-        addWindowListener(this);
-        setDefaultCloseOperation(JFrame.DISPOSE_ON_CLOSE);
```
Windchild292, line 131: I had these here so that they could be overridden if a dev decided they needed to for an individual dialog. Is there are reason why we'd want to override the pairing here without requesting a custom finalization setup?

megamek/src/megamek/client/ui/dialogs/BVDisplayDialog.java
```diff
@@ -42,16 +42,10 @@ public BVDisplayDialog(final JFrame frame, final boolean modal, final Entity ent
...
-        return entity;
-    }
-    //endregion Getters
-
     //region Initialization
     @Override
     protected Container createCenterPane() {
-        final JEditorPane editorPane = new JEditorPane("text/html", getEntity().getBVText());
+        final JEditorPane editorPane = new JEditorPane("text/html", entity.getBVText());
         editorPane.setEditable(false);
```
Windchild292, line 49: Revert. This is a stylistic decision that I tend to use for code I work with, namely using getters/setters instead of the variable itself.

Windchild292 (changes requested):
One comment, one revert request, for the changes

SJuliez:
- Hmmmm, the initialization things can just as well stay where they are if this was intentional.
- I don't follow the stylistic decision to use getters within a class but this is yours so I'll not argue.
This doesn't leave anything worth merging.

PR #2910, opened 2021-06-04
UnitReadoutDialog based on AbstractDlg

SJuliez:
As the title says. No functional change. Separates out the readout dialog from the lobby's "View" menu option into its own class based on AbstractDialog. 
I'll say this was effortless.

### DIFF
--- a/megamek/src/megamek/client/ui/dialogs/EntityReadoutDialog.java
+++ b/megamek/src/megamek/client/ui/dialogs/EntityReadoutDialog.java
@@ -30,6 +30,7 @@ public class EntityReadoutDialog extends AbstractDialog {
 
     private final JFrame parent;
     private final Entity entity;
+    private EntityViewPane entityView;
 
     /** Constructs a non-modal dialog showing the readout (TRO) of the given entity. */ 
     public EntityReadoutDialog(final JFrame frame, final Entity entity) {
@@ -47,7 +48,8 @@ public class EntityReadoutDialog extends AbstractDialog {
 
     @Override
     protected Container createCenterPane() {
-        return new EntityViewPane(parent, entity);
+        entityView = new EntityViewPane(parent, entity);
+        return entityView;
     }
-
+    
 }
--- a/megamek/src/megamek/client/ui/dialogs/EntityReadoutDialog.java
+++ b/megamek/src/megamek/client/ui/dialogs/EntityReadoutDialog.java
@@ -23,12 +23,13 @@ import java.util.Objects;
 import javax.swing.*;
 import megamek.client.ui.baseComponents.AbstractDialog;
 import megamek.client.ui.panes.EntityViewPane;
+import megamek.client.ui.preferences.JTabbedPanePreference;
+import megamek.client.ui.preferences.PreferencesNode;
 import megamek.common.*;
 
 /** A dialog showing the unit readout for a given unit. */
 public class EntityReadoutDialog extends AbstractDialog {
 
-    private final JFrame parent;
     private final Entity entity;
     private EntityViewPane entityView;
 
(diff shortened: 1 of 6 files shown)


SJuliez:
Oh wait - forgot to rename the dialog with its own resource entries!! Grrr.

Windchild292:
Taking a look... other than the above, is there any reason to not just have this dialog wrap around entity view pane? Seems to share the same information but would add the TRO component too.

SJuliez:
Thats the one used in the unit selector, right? I'll look into that. I was just continuing with what the lobby historically did, but good idea that. I wouldnt even bet against there already being such a dialog somewhere...

Windchild292:
It's the one I made that I forgot to replace the unit selector one with...
Made it for MekHQ's unit market based on (as in... nearly a direct copy of) the older unit selector.

SJuliez:
OK, update:
- Based the EntityReadoutDialog on the EntityViewPane
- Used EntityViewPane in the AbstractUnitSelector
- Modernized the MechViewPanel (and gave it a bit of space to the scrollbar, made it scroll even when the mouse wheel is used in the dialog free space or on the fluff image)

It's a bit hacky to set the title of an AbstractDialog to something variable but I don't mind that line.

I would have liked to make it scale but since the whole Unit Selector currently doesn't, this would look strange there.

SJuliez:
Ah I forgot: Although the EntityViewPane->AbstractTabbedPane seems to want to save the index of the selected pane, this doesn't seem to work, neither for the unit selector nor for the separate dialog. I don't know why.

megamek/src/megamek/client/ui/dialogs/EntityReadoutDialog.java
```diff
@@ -0,0 +1,55 @@
...
+    public EntityReadoutDialog(final JFrame frame, final Entity entity) {
+        this(frame, false, entity);
+    }
+
+    /** Constructs a dialog showing the readout (TRO) of the given entity with the given modality. */
+    public EntityReadoutDialog(final JFrame frame, final boolean modal, final Entity entity) {
+        super(frame, modal, "EntityReadoutDialog", "EntityReadoutDialog.title");
+        setTitle(getTitle() + entity.getShortNameRaw());
+        this.entity = Objects.requireNonNull(entity);
+        parent = frame;
```
Windchild292, line 45: Why are you saving this twice? You can just call getFrame() to get the pane this dialog was created with.

megamek/src/megamek/client/ui/dialogs/EntityReadoutDialog.java
```diff
@@ -0,0 +1,55 @@
...
+        initialize();
+    }
+
+    @Override
+    protected Container createCenterPane() {
+        entityView = new EntityViewPane(parent, entity);
+        return entityView;
+    }
+
```
Windchild292, line 54:
```suggestion
    
    @Override
    protected void setCustomPreferences(final PreferencesNode preferences) {
        super.setCustomPreferences(preferences);
        preferences.manage(new JTabbedPanePreference(entityView));
    }
```

Windchild292: I'd leave the preferences off of the AbstractUnitSelectorDialog for now, as any implementation will likely cause issues with MekHQ.

Windchild292 (commented):
In general this looks great, just one oddity and a missing override

megamek/src/megamek/client/ui/dialogs/EntityReadoutDialog.java
SJuliez: Yes, corrected.

megamek/src/megamek/client/ui/dialogs/EntityReadoutDialog.java
SJuliez: I don't know that this does, but I've added it.

Windchild292: So, what this does is it creates the preference that now handles all of the backend of monitoring the tabbed pane. Pretty much a black box that you need to make a one-line call for to ensure the preferences are handled properly.

Windchild292 (approved):

PR #2915, opened 2021-06-06
Improved unit image scaling in the lobby

SJuliez:
In the lobby, scaling gave some unit icons (especially those without a drop shadow, as e.g. the new Warhammers) ugly whitish border pixels. We already have a better scaling filter, so I made use of that.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/lobby/MekForceTreeRenderer.java
+++ b/megamek/src/megamek/client/ui/swing/lobby/MekForceTreeRenderer.java
@@ -35,6 +35,7 @@ import megamek.common.*;
 import megamek.common.force.*;
 import megamek.common.icons.Camouflage;
 import megamek.common.options.OptionsConstants;
+import megamek.common.util.ImageUtil;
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/client/ui/swing/lobby/MekForceTreeRenderer.java
```diff
@@ -124,8 +125,9 @@ public String getToolTipText(MouseEvent event) {
         return null;
     }
 
-    private void setIcon(Image image, int size) {
-        setIcon(new ImageIcon(image.getScaledInstance(-1, size, Image.SCALE_SMOOTH)));
+    private void setIcon(Image image, int height) {
+        int width = height * image.getWidth(null) / image.getHeight(null);
+        setIcon(new ImageIcon(ImageUtil.getScaledImage(image, width, height)));
```
sixlettervariables, line 130: We've seen errors in the past where `getWidth`/`getHeight` returns `0` or `-1`. Lets check for `image.getHeight(null) > 0` before proceeding.

Windchild292 (approved):

PR #2924, opened 2021-06-08
Short building update

NickAragua:
Implements recent errata where mechs in short buildings (i.e. level 1) are no longer entirely covered by the building, but the building is instead treated as partial cover.

Minor re-factoring of weapon handler classes to pull out code that damages buildings.


megamek/src/megamek/common/weapons/ACAPHandler.java
```diff
@@ -96,41 +97,20 @@ protected void handleEntityDamage(Entity entityTarget,
...
-            Vector<Report> buildingReport = server.damageBuilding(bldg, toBldg,
-                    entityTarget.getPosition());
-            for (Report report : buildingReport) {
-                report.subject = subjectId;
-            }
-            vPhaseReport.addAll(buildingReport);
-        }
+        // if the target was in partial cover, then we already handled
+        // damage absorption by the partial cover, if it would have happened
+        IHex targetHex = game.getBoard().getHex(target.getPosition());
```
sixlettervariables, line 102: Does this damage handler (or subsequent handlers) apply to offboard units as well? Said another way: could `targetHex` be null here?

megamek/src/megamek/common/weapons/WeaponHandler.java
```diff
@@ -1479,17 +1479,27 @@ protected void handleEntityDamage(Entity entityTarget,
...
 
         initHit(entityTarget);
+        hit.setLocation(Mech.LOC_LLEG);
         
         boolean isIndirect = wtype.hasModes()
                 && weapon.curMode().equals("Indirect");
         IHex targetHex = game.getBoard().getHex(target.getPosition());
+        // a specific situation where a mech is standing in depth 1 water
+        boolean mechPokingOutOfShallowWater = targetHex.containsTerrain(Terrains.WATER) &&
+                entityTarget.relHeight() <= targetHex.surface();
```
sixlettervariables, line 1489: Why `<=` instead of `>=`?

sixlettervariables (commented):
Two questions.

megamek/src/megamek/common/weapons/WeaponHandler.java
NickAragua: Actually, now that I think about it, it should be == only. We don't want >=, because stuff like superheavy mechs (technically height 3) shouldn't get partial cover benefits from standing in level 1 water.

megamek/src/megamek/common/weapons/ACAPHandler.java
NickAragua: No, this is a handler for autocannons loaded with armor piercing ammo.

NickAragua: Introduced null checking for hex out of an abundance of caution, and needed extra logic in there anyway.

sixlettervariables (approved):
Let the questions roll in!

PR #2927, opened 2021-06-09
Issue #2543 Board images save at current zoom, game summary at zoom 1

### LINKED ISSUE #2543: 47.15 - Appears to have been change to the 'Board > Save As Image' menu option.
MM 47.15 Standalone
Windows 10
OpenJDK 11

In previous versions (47.14 and before), the File > Board > Save As Image option would snapshot the current map at the current display zoom level. Zooming out would thus result in a smaller file size than if the map is zoomed in.

However, in 47.15, it appears that the image is saved at the "mazimum zoom in" resolution, regardless of the currently displayed zoom level.

As an example, a 48x51 map was zoomed out to a level that, with 47.14, produced a 2.88MB image. When zoomed out to the same level with 47.15 and saved in the same way, the image file was 17MB, which was the same size as if the map had been zoomed to the largest size possible (and the image quality was the same as if that had been the case).

This makes it hard to produce an image that can be easily shared on e.g. Discord or other sites with a size cap (as well as causing a certain amount of storage bloat).

Might this be the result of changes required to implement the turn-by-turn snapshot functionality?

(It was suggested that I tag @arlith in this bug report.)

SJuliez:
Arlith changed the board image save method to always use zoom 1 (not only for the game summary). Reverting this for manual saves.

Resolves #2543

### DIFF
--- a/megamek/src/megamek/client/ui/IBoardView.java
+++ b/megamek/src/megamek/client/ui/IBoardView.java
@@ -161,9 +161,11 @@ public interface IBoardView extends MechDisplayListener {
      * mapsheets), draws the hexes onto it, and returns that image.
      *
      * @param ignoreUnits If true, no units are drawn, only the board
+     * @param useBaseZoom If true, save the image at zoom = 1, otherwise at the
+     * current board zoom.
      * @return
      */
-    public abstract BufferedImage getEntireBoardImage(boolean ignoreUnits);
+    public abstract BufferedImage getEntireBoardImage(boolean ignoreUnits, boolean useBaseZoom);
 
     /**
      * Sets the BoardView's currently selected entity.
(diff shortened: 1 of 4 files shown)


Windchild292 (approved):

megamek/src/megamek/client/ui/IBoardView.java
```diff
@@ -161,9 +161,11 @@ public void setMovementEnvelope(Map<Coords,Integer> mvEnvData,
      * mapsheets), draws the hexes onto it, and returns that image.
      *
      * @param ignoreUnits If true, no units are drawn, only the board
+     * @param useBaseZoom If true, save the image at zoom = 1, otherwise at the
+     * current board zoom.
      * @return
```
Windchild292, line 166: Wait, isn't this the exact opposite of what it does?

SJuliez: I don't think so.

```
if (useBaseZoom) {
    zoomIndex = BASE_ZOOM_INDEX;
    zoom();
}
```

SJuliez: Just tested it again and it seems to do what I want it to do, manual image saves work at the current board zoom.

PR #2929, opened 2021-06-10
enable minefield done button at all times

NickAragua:
This changes the behavior of the "done" button on the minefield deployment display. It's now always enabled on "beginMyTurn()" and always disabled on "endMyTurn()" without any consideration for whether the player has deployed any mines.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/DeployMinefieldDisplay.java
+++ b/megamek/src/megamek/client/ui/swing/DeployMinefieldDisplay.java
@@ -140,9 +140,7 @@ public class DeployMinefieldDisplay extends StatusBarPhaseDisplay {
         setActiveEnabled(p.getNbrMFActive());
         setInfernoEnabled(p.getNbrMFInferno());
         setRemoveMineEnabled(true);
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/client/ui/swing/DeployMinefieldDisplay.java
```diff
@@ -140,9 +140,7 @@ private void beginMyTurn() {
         setActiveEnabled(p.getNbrMFActive());
         setInfernoEnabled(p.getNbrMFInferno());
         setRemoveMineEnabled(true);
-        if (!p.hasMinefields()) {
-            butDone.setEnabled(true);
-        }
+        butDone.setEnabled(true);
```
Windchild292, line 143: Do we even need to set it to enabled then? Do we ever disable it?

NickAragua: Yes, when it's not the current player's turn.

Windchild292 (approved):

PR #2934, opened 2021-06-10
Complete Anti-TSM "Green Smoke" warhead implementation

neoancient:
Corrects and updates the anti-TSM munition implementation to the rules in IO, p. 104

1. The results from the cluster hits table is divided by 2 and rounded down, to a minimum of 1 [already implemented, not changed].
2. When a volley hits the target, the hex is filled with green smoke. A mix scatters one hex. The smoke behaves in most ways like heavy smoke [implemented by changing the ATSM handlers to extend the smoke munitions handlers, with modifications as needed, and adding a new smoke type].
3. Any mech with prototype TSM or industrial TSM constructed before 3050 takes an additonal 6 points of damage to every unarmored location when hit by ATSM missiles or enters a hex containing green smoke [implemented in WeaponHandler and Server::processMovement].
4. Any time a mech that meets the criteria outlined above takes a critical hit from being hit by ATSM munitions or exposure to green smoke, it takes an additional critical to the same location [corrected existing implementation to apply one additional critical total instead of applying each resulting critical twice, and apply it to the same location instead of rolling 2d6 and possibly throwing an IndexOutOfRangeException for a non-existent location].
5. Conventional infantry with TSM implants take 2d6 burst damage when hit by ATSM munitions and 1d6 when entering a green smoke hex, unless they have protection against vacuum or toxic atmosphere [new implementation].

I made some other changes to avoid the need to save the state of being hit by ATSM in Entity by changing the boolean `damagedByFire` parameter of `criticalEntity` to `DamageType`, which also involved creating a new type for the inferno missiles that used the boolean. Also renamed the bFrag parameter of `damageEntity` to reflect its current use.

Here are two units with prototype TSM for testing (Locust LCT-1L and Raven RVN-3X, both of which should have prototype TSM based on the intro dates)
[prototype_tsm.zip](https://github.com/MegaMek/megamek/files/6634403/prototype_tsm.zip)

### DIFF
--- a/megamek/i18n/megamek/common/report-messages.properties
+++ b/megamek/i18n/megamek/common/report-messages.properties
@@ -651,7 +651,7 @@
 5181=Light
 5182=Heavy
 5183=<msg:5181,5182> smoke fills <data>!
-5184=<data> smoke fills <data>!
+5184=Green smoke fills <data>!
 5185=Heavy smoke fills <data>!
 5186=Laser Inhibiting smoke fills <data>!
 5187=Chaff fills <data>!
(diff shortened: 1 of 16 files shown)


NickAragua (commented):
Looks good, other than the need to accommodate errata.

Windchild292 (approved):

PR #2939, opened 2021-06-11
Artemis/Apollo validation

neoancient:
Adds checks for rules surrounding Artemis and Apollo fire control systems to entity validation:

- No mixing of Artemis types (IV, V, prototype)
- If any compatible launcher has it, they all have to have it
- The Artemis/Apollo system must be in the same location as the weapon.

In the process I added a F_ARTEMIS_COMPATIBLE flag to WeaponType to make simplify the code to check for compatibility. I added a unit test that checks to make sure the new test has the same results as the old and guard against changes breaking the test. The form of the test is a bit unorthodox but I wrote it to produce meaningful messages on failure.

Fixes MegaMek/megameklab#888

### DIFF
--- a/megamek/src/megamek/common/MechFileParser.java
+++ b/megamek/src/megamek/common/MechFileParser.java
@@ -355,18 +355,8 @@ public class MechFileParser {
 
                 // link up to a weapon in the same location
                 for (Mounted mWeapon : ent.getTotalWeaponList()) {
-                    WeaponType wtype = (WeaponType) mWeapon.getType();
 
-                    // only srm, lrm and mml are valid for artemis
-                    if ((wtype.getAmmoType() != AmmoType.T_LRM)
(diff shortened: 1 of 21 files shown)


Windchild292 (approved):

HuntingTarg:
Hate to do this, however this might be a bug with the above fix and not a new issue; I'm having a problem with Artemis IV on an Omnifighter:
MegaMekLab bug Unit Validation fails Artemis IV when requirement met

[Anishinabe_Chippewa II_Omni ANI-EX1.txt](https://github.com/MegaMek/megamek/files/10693014/Anishinabe_Chippewa.II_Omni.ANI-EX1.txt)

I have done various things with Artemis IV/V and Apollo on Battlemechs and large naval units and it all seems to work fine.  IDK what the underlying issue is here.

HammerGS:
You’ve posted this on a closed PR. This needs to be opened as a new issue. With a link to this.

PR #2946, opened 2021-06-17
prevent prone mechs on top of short buildings from gaining partial cover

NickAragua:
Partial cover only applies to units that are currently tall enough to benefit from it. Also removed an extraneous call to Compute.inBuilding().


megamek/src/megamek/common/actions/WeaponAttackAction.java
```diff
@@ -4582,6 +4582,7 @@ public static boolean targetInShortCoverBuilding(Targetable target) {
         }
         
         return targetHex.containsTerrain(Terrains.BUILDING) &&
+                ((Entity) target).getHeight() > 0 &&
                 ((Entity) target).relHeight() == targetHex.ceiling();
```
Windchild292, line 4586: Brackets (please add for both)

Windchild292 (approved):

PR #2948, opened 2021-06-17
Minimap

### LINKED ISSUE #2762: Minimizing minimap results in empty summary screenshots
### Environment
Standalone MM 0.48.0
Kubuntu 20.04.2
Java 11.0.10

### Description
When minimap is minimized, Save Game Summary Image Of Minimap option generates screenshots 14 px high if minimap was minimized with a green Minimize button at the bottom.
round_1_14_Firing

SJuliez:
A bit of overhaul and cleanup for the minimap.

- The minimap is now used through static calls that either return a dialog (which is the live minimap) or a snapshot image. 
- Externalizes the unit icons to their own class and unifies them a bit (in that they use Path2D drawing). This is in its own package. I didnt want to move the minimap itself in this PR, as merge conflicts with moved classes are really bad.
- Changes some of the standard icons to something more appropriate: vehicles get an upright rectangle (as in tank), mechs are a trapezoid (Schwarzenegger torso), Aero a triangle and so on.
- Hammer suggested making the StratOps icons standard, and I do think they look battle-techy so I did that.
- Game summary minimaps now work regardless of whether the minimap is open or not. They are always saved at a middling zoom level. Only the symbols used depend on the user setting.
- Makes closing the minimap through the window X correctly influence the menu bar setting.
- When the client window is minimized the minimap won't show with changing phases (this happened when letting Princess fight itself) 
- I did cut some code that was supposed to limit the minimap window on small resolutions and big maps but seemed rather bloated for something that any user can handle with applying a half-second mouse wheel operation. The minimap saves its zoom setting anyway. 
- Edit: forgot, I also corrected a map setup XML and added two more.

image
image

Resolves #2762


megamek/src/megamek/client/ui/swing/MiniMap.java
```diff
@@ -389,146 +245,142 @@ public synchronized void update(Graphics g) {
...
+                        } else {
+                            st.nextToken();
+                            red = (int) st.nval;
+                            st.nextToken();
+                            green = (int) st.nval;
+                            st.nextToken();
+                            blue = (int) st.nval;
+
+                            terrainColors[getType(key)] = new Color(red, green, blue);
+                        }
```
sixlettervariables, line 373: This case needs a break.

megamek/src/megamek/client/ui/swing/MiniMap.java
```diff
@@ -731,148 +516,123 @@ protected synchronized void drawMapOrig() {
...
-                                                           Color.yellow);
-                                }
-                            }
-                        }
-                    }
-                }
-                
+            drawDeploymentZone(g);
+            
+            if ((null != game)) {
```
sixlettervariables, line 532:
```suggestion
            if (null != game) {
```

megamek/src/megamek/client/ui/swing/MiniMap.java
```diff
@@ -731,148 +516,123 @@ protected synchronized void drawMapOrig() {
...
+        if (dialog != null) {
+            drawButtons(g);
+            repaint();
         }
+    }
+    
+    /** Indicates the deployment hexes. */
+    private void drawDeploymentZone(Graphics g) {
+        if ((null != client) && (null != game) 
+                && IGame.Phase.PHASE_DEPLOYMENT == game.getPhase() && dialog != null) {
```
sixlettervariables, line 565: When in Rome, I guess:
```suggestion
                && (IGame.Phase.PHASE_DEPLOYMENT == game.getPhase()) && (dialog != null)) {
```

megamek/src/megamek/client/ui/swing/MiniMap.java
```diff
@@ -731,148 +516,123 @@ protected synchronized void drawMapOrig() {
...
+            GameTurn turn = game.getTurn();
+            if ((turn != null) && (turn.getPlayerNum() == client.getLocalPlayer().getId())) {
+                Entity deployingUnit = bv.getDeployingEntity();
+                int dir;
+                // We need to draw the same deployment zone as boardview
+                if (deployingUnit != null && deployingUnit.getOwnerId() == turn.getPlayerNum()) {
```
sixlettervariables, line 571:
```suggestion
                if ((deployingUnit != null) && (deployingUnit.getOwnerId() == turn.getPlayerNum())) {
```

megamek/src/megamek/client/ui/swing/MiniMap.java
```diff
@@ -1111,16 +849,14 @@ private void paintAttack(Graphics g, AttackAction attack) {
         g.drawPolygon(xPoints, yPoints, 4);
 
         // if this is mutual fire, draw a half-and-half line
-        for (Enumeration<EntityAction> iter = m_game.getActions(); iter
-                .hasMoreElements();) {
-            EntityAction action = iter.nextElement();
+        for (var action: game.getActionsVector()) {
```
sixlettervariables, line 852:
```suggestion
        for (var action : game.getActionsVector()) {
```

megamek/src/megamek/client/ui/swing/MiniMap.java
```diff
@@ -1146,148 +882,69 @@ private void paintAttack(Graphics g, AttackAction attack) {
...
+        
+        // Transform for placement and scaling
+        var placement = AffineTransform.getTranslateInstance(baseX, baseY);
+        placement.scale(UNIT_SCALE[zoom]/100.0d, UNIT_SCALE[zoom]/100.0d);
+        g2.transform(placement);
+        
+        // Add a position shift if multiple units are present in this hex
+        Coords p = entity.getPosition();
+        Integer eStack = multiUnits.get(p);
+        if (eStack == null) eStack = 0;
```
sixlettervariables, line 937:
```suggestion
        int eStack = multiUnits.getOrDefault(p, 0);
```

megamek/src/megamek/client/ui/swing/MiniMap.java
```diff
@@ -1344,316 +977,261 @@ private void paintUnit(Graphics g, Entity entity) {
...
                 }
                 if (!s.equals("")) {
-                    FontRenderContext fontContext = new FontRenderContext(null, true, true);
-                    Font font = new Font("SansSerif", Font.BOLD, 100);
-                    FontMetrics currentMetrics = getFontMetrics(font);
-                    int stringWidth = currentMetrics.stringWidth(s);
+                    var fontContext = new FontRenderContext(null, true, true);
+                    var font = new Font("SansSerif", Font.BOLD, 100);
+                    var currentMetrics = getFontMetrics(font);
+                    var stringWidth = currentMetrics.stringWidth(s);
```
sixlettervariables, line 983: Is `var` supported in Java 11 outside of lambda parameters?

megamek/src/megamek/client/ui/swing/MiniMap.java
```diff
@@ -1344,316 +977,261 @@ private void paintUnit(Graphics g, Entity entity) {
...
-                .hasMoreElements(); ) {
-            int[] hex = iter.nextElement();
-            x = hex[0];
-            y = hex[1];
-            baseX = (x * (hexSide[zoom] + hexSideBySin30[zoom])) + leftMargin
-                    + hexSide[zoom];
-            baseY = (((2 * y) + 1 + (x % 2)) * hexSideByCos30[zoom]) + topMargin;
-            exits = hex[2];
+        g.setColor(terrainColors[ROAD]);
+        for (var roadEntry: roadHexes) {
```
sixlettervariables, line 1038:
```suggestion
        for (var roadEntry : roadHexes) {
```

megamek/src/megamek/client/ui/swing/MiniMap.java
sixlettervariables: If it is, my preference for `var` is only when the type is obvious (i.e. the typename itself is on the RHS):
```suggestion
                    var fontContext = new FontRenderContext(null, true, true);
                    var font = new Font("SansSerif", Font.BOLD, 100);
                    FontMetrics currentMetrics = getFontMetrics(font);
                    int stringWidth = currentMetrics.stringWidth(s);
```

megamek/src/megamek/client/ui/swing/MiniMap.java
```diff
@@ -1668,82 +1246,74 @@ private Coords translateCoords(int x, int y) {
...
     }
 
-    void processMouseClick(int x, int y, MouseEvent me) {
+    private void processMouseClick(int x, int y, MouseEvent me) {
         if (y > (getSize().height - 14) && !dragging) {
             if (minimized) {
-                setSize(getSize().width, heightBufer);
-                m_mapImage =  ImageUtil.createAcceleratedImage(getSize().width, heightBufer);
+                setSize(getSize().width, heightBuffer);
+                mapImage =  ImageUtil.createAcceleratedImage(getSize().width, heightBuffer);
```
sixlettervariables, line 1271:
```suggestion
                mapImage = ImageUtil.createAcceleratedImage(getSize().width, heightBuffer);
```

megamek/src/megamek/client/ui/swing/MiniMap.java
```diff
@@ -1777,25 +1339,25 @@ public void gamePhaseChange(GamePhaseChangeEvent e) {
...
+                File dir = new File(Configuration.gameSummaryImagesMMDir(), game.getUUIDString());
                 if (!dir.exists()) {
                     dir.mkdirs();
                 }
-				File imgFile = new File(dir, "round_" + m_game.getRoundCount() + "_" + e.getOldPhase().ordinal() + "_"
+				File imgFile = new File(dir, "round_" + game.getRoundCount() + "_" + e.getOldPhase().ordinal() + "_"
 						+ IGame.Phase.getDisplayableName(e.getOldPhase()) + ".png");
                 try {
-                    ImageIO.write(m_mapImage, "png", imgFile);
+                    ImageIO.write(getMinimapImage(game, bv, 4), "png", imgFile);
```
sixlettervariables, line 1349: Can we make a constant out of `4` so that we have a nice name for it (besides: doesn't show controls).

megamek/src/megamek/common/Hex.java
```diff
@@ -426,6 +426,19 @@ public int terrainLevel(int type) {
     public ITerrain getTerrain(int type) {
         return terrains.get(type);
     }
+    
+    @Override
+    public ITerrain getAnyTerrainOf(int type, int... types) {
+        if (containsTerrain(type)) {
+            return terrains.get(type);
+        }
+        for (int moreTypes: types) {
```
sixlettervariables, line 435:
```suggestion
        for (int moreTypes : types) {
```

megamek/src/megamek/common/Hex.java
```diff
@@ -426,6 +426,19 @@ public int terrainLevel(int type) {
     public ITerrain getTerrain(int type) {
         return terrains.get(type);
     }
+    
+    @Override
+    public ITerrain getAnyTerrainOf(int type, int... types) {
```
sixlettervariables, line 431: Why not just `int... types` ?

sixlettervariables (commented):
Nits here and there, otherwise this is really nice looking.

Thought I was reading C# when I saw `var`, so some of those comments others should chime in on to set some precedence.

megamek/src/megamek/common/Hex.java
SJuliez: To make the call getAnyTerrainOf() invalid. (Enforce giving at least one parameter)
Which I thought int... doesnt do, or does it?

SJuliez: Actually I wanted to comment on this in the PR description but of course forgot. It might seem useless but I find this comes up pretty often; looking in a hex for any terrain of (x,y,z) and then getting whichever of those is present, where x,y,z are usually mutually exclusive. This is usually code-heavy and ugly so I wanted to simplify this.

sixlettervariables: Actually, I really enjoy your usage of the API calling convention itself to enforce At Least One semantics. Elegant and I missed that.

Definite Pit of Success!

megamek/src/megamek/client/ui/swing/MiniMap.java
SJuliez: It's the last case. You mean as a matter of safe style?

sixlettervariables: Yes.

megamek/src/megamek/client/ui/swing/MiniMap.java
SJuliez: > (besides: doesn't show controls)
Sorry I don't get what you mean with this.

sixlettervariables: Ah, like why `4`, is it because it doesn't show controls on the minimap?

SJuliez: Oh, the controls aren't shown because the minimap images are created without a dialog and this
```
        if (dialog != null) {
            drawButtons(g);
            repaint();
        }
```
keeps the buttons from being painted. 
The 4 value is just some middle value; there's no ideal size like for the actual map (where we scale the images, so no zoom other than 1 makes real sense). On the biggest zoom, the biggest maps create minimap images of about 2500 x 4000. Compared to the boardview images that's still small. I just checked: The 160x170 maps make 50MB board images of 8000x10000 (and it takes 6 seconds to write that image). Hmmm. I could set this to full minimap zoom. I'm undecided.

SJuliez:
OK all review changes are in, I changed a few more vars back to explicit, but left the game summary zoom at 4. It's not a bad size and we can change it easily if it comes up.

sixlettervariables (approved):
LGTM!

PR #2950, opened 2021-06-18
don't validate weapon groups for Artemis/Apollo

NickAragua:
Weapon groups are added to aerospace fighters after loading all the weapons; according to the comments it's for the purposes of allowing an aerospace fighter to operate as a squadron. That's fine, but we probably don't need to count those pseudo-weapons when checking whether a fighter has a legal Artemis IV setup.


sixlettervariables (approved):
I think this is correct, but I'd like @neoancient to back me up on this. (I went deep in this code last year tracking down a different issue and believe I came to the same conclusion about weapon groups.)

neoancient (approved):
Yup. I only remember weapon groups in situations where it doesn't matter.

PR #2955, opened 2021-06-20
Board Tagging

### LINKED ISSUE #2760: RFE: Use meta-data tags on maps for filtering
Please add an ability to add search tags to pre-made maps e.g.

- flat
- urban
- forest
- lake

When filtering for maps (including blind random selection), include the ability to have simple 'and/not' combinations of flags to filter by multiple tags in addition to size e.g.

- size: 30 x 30
- AND urban
- AND NOT hills

I'm not a huge fan of the organizational method of "Throw everything in the same bag and search by metadata" over "Use good folder organization".  In this circumstance I think it would be of benefit with the increasing number of large and complex maps included in the program.

SJuliez:
The first two commits have the code changes as well as a minor correction in the fourth commit. The others are boards.

- Adds the "tag" key word for board files so boards can be tagged, e.g.
tag "Forest (Auto)"
tag "Roads (Auto)"
tag "I like this map"
As usual, the tag can be put anywhere in the boardfile but they'll be saved (e.g. by the boardeditor) at the beginning.
- Board tags will be used by the lobby map search filter (in addition to the file name and directory name as before)
- Multiple search strings can now be used in the lobby map search, separated by ";", e.g. ocean;unoff. All search tokens must be fulfilled (they are ANDed)
- Adds a BoardTagger utility program which can be used like the BoardsValidator (I cut the command line stuff from it - if that's needed I can add it back in). This scans all boards and adds automated tags. It also removes them when they're not applicable anymore. Of course it cleans out its tags before adding them (again) so it can be run multiple times without issues. The tags aren't ordered explicitly, so it is possible that board files appear changed when the Tagger is run although only the tag order has changed. The result are the (Auto) tags in the example above. Currently the following auto tags are used:
image
- Note that the (Auto) suffix is used by the Tagger to identify its tags. It isn't necessary to write this ever in the search, as always chunks of search terms are enough, like "urb;wat;for"
- I couldn't think of a good way to identify lakes and rivers so I left it at oceans and water
- Corrects a board (megamek/megamek/data/boards/unofficial/Dervs/Map Pack 2/60x60 Nanking Nanolab.board) that had numerous "\\" entries which stopped the BoardsValidator short. 
- To be clear, manual tags are supported. Mapmakers can write their own tags into the board file. So long as they don't end in (Auto) the Tagger won't change them. The boardeditor will load and save them (although they won't be visible there). A stitched map saved from the lobby will NOT preserve any tags though (but that doesn't affect the original boards).
- And this is the way to get a million code line additions

Resolves #2760


sixlettervariables (approved):
Looks good.

I didn't check the `.board` files out beyond just three or four of the ones I could get to load.

PR #2957, opened 2021-06-20
Board Tagging code only

SJuliez:
A copy of the board tag PR without the board changes :)

### DIFF
--- a/megamek/src/megamek/utils/BoardsTagger.java
+++ b/megamek/src/megamek/utils/BoardsTagger.java
@@ -273,7 +273,7 @@ public class BoardsTagger {
         toRemove.forEach(board::removeTag);
 
         // Give the board any applicable tags
-        matchingTags.keySet().stream().map(Tags::getName).forEach(board::addTag);
+        matchingTags.keySet().stream().filter(t -> matchingTags.get(t)).map(Tags::getName).forEach(board::addTag);
 
         if (DEBUG) {
             System.out.println("----- Board: " + boardFile);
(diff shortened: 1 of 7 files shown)


megamek/src/megamek/utils/BoardsTagger.java
```diff
@@ -0,0 +1,286 @@
...
+        }
+
+        // The board area
+        int area = board.getWidth() * board.getHeight();
+        // The geometric mean of the board sides. Better for some comparisons than the area.
+        int normSide = (int) Math.sqrt(area);
+        int levelExtent = board.getMaxElevation() - board.getMinElevation();
+
+        // Calculate which tags apply
+        Tags.TAG_MEDFOREST.applies = (forest >= normSide * 5) && (forest < normSide * 10) && (forestHU < normSide * 2);
```
sixlettervariables, line 230: This is odd and limits it's usage. Let's find another way, perhaps a map.

sixlettervariables (changes requested):
Still digesting some of this, but one quick issue.

megamek/src/megamek/utils/BoardsTagger.java
SJuliez: What do you mean? 
After some thinking and deleting answers I wrote my conclusion is you mean: Map<Tags, Boolean> to avoid the "applies" field that is without any sense outside of the BoardTagger. Correct? OK I can see that.

sixlettervariables: Yes, also not thread safe and wouldn't be usable from testing if tested in parallel.

SJuliez:
Not finished.

SJuliez:
OK, now it should be okay.

sixlettervariables (approved):

SJuliez:
I'll close this one as it was for review only.

PR #2962, opened 2021-06-22
Connection Watchdog: Renaming File to Match Class

Windchild292:
This PR renames the connection watchdog so the file and class names are the same, and does some light cleanup in the file. The misspell was raised by Akjosch on Discord.


sixlettervariables (approved):
Is this used anywhere? I just realized this is a private class.

Windchild292:
It is used in Server.java... and the spelling might be a cause of Nick and my build issues.

PR #2964, opened 2021-06-23
2199: MUL Parser Game Option-reliant Personnel Load

### LINKED ISSUE #2199: Pilot Abilities are still working even when Pilot Abilities are disabled in Megamek
Even when special abilities are disabled in Megamek, a pilot, pulled from a mul file, that has Pilot abilities will still have those pilot abilities even when they are turned off in the Game options.

### Environment
All versions of MM I suuspect.
What version of MegaMek does your issue apply to?  Note whether it is the stand alone MegaMek or HQ's MegaMek.
What operating system are you using?  Windows, Mac, or Linux?
Windows 10
What java version are you using?  If unsure it will state your java version at the top of the megameklog.txt file.
Recent Java

### Description
Describe the issue and provide screenshots ingame if able.
Provide the rulebook, page number, and text if you feel like this is a rules issue.
And provide the steps you have taken to reproduce the issue in the saved game if applicable.
(Delete the above lines)

### Files
Provide a saved game of when the issue was found.  If possible try to save the game during the phase before the issue occurs so the 
developers can step through the code and find exactly what it is that is breaking in the code.
Provide the megameklog.txt file for any NPE's or errors that MegaMek might be reporting.
Do not paste the entire logfile text into this ticket. Just attach the actual megameklog.txt file
And provide your playername so that they can load the saved game.
(Delete the above lines)

Windchild292:
This fixes #2199 and does a bunch of cleanup for MUL Parser.

The first three commits are all cleanup, then I started to work on the actual bugfix.

This requires https://github.com/MegaMek/megameklab/pull/934 and https://github.com/MegaMek/mekhq/pull/2720.


megamek/src/megamek/common/MULParser.java
```diff
@@ -284,74 +286,65 @@ public MULParser(){
         pilots = new Vector<>();
     }
 
-    public MULParser(InputStream fin){
+    public MULParser(final File file, final @Nullable GameOptions options) throws Exception {
         this();
-        parse(fin);
-    }
 
-    public void parse(InputStream fin){
```
sixlettervariables, line 292: We should still keep an InputStream ctor around.

megamek/src/megamek/common/MULParser.java
```diff
@@ -2390,61 +2386,56 @@ private void parseBAMEA(Element meaTag, Entity entity){
...
         Mounted mountedManip = null;
         int meaMountLoc = Integer.parseInt(meaMountLocString);
         boolean foundMea = false;
-        for (Mounted m : entity.getEquipment()){
-            if (m.getBaMountLoc() != meaMountLoc){
-                continue;
-            }
-            if (m.getType().hasFlag(MiscType.F_BA_MEA)){
+        for (Mounted m : entity.getEquipment()) {
+            if ((m.getBaMountLoc() != meaMountLoc) && m.getType().hasFlag(MiscType.F_BA_MEA)) {
```
sixlettervariables, line 2406: Shouldn't this be `==`?

Windchild292: Whoops, definitely should

megamek/src/megamek/common/MULParser.java
Windchild292: I'd rather keep the single version with a nullable, and just improve the documentation. IMO it makes it less likely someone will miss the options inclusion as that is the standard use case

Windchild292: Wait, misread... We don't utilize an InputStream ctor, so why bother having one?

sixlettervariables: Testing would use an InputStream.

Windchild292: Added an input stream constructor for unit testing, with a note to say it was just for unit testing.

SJuliez (approved):
No complaints

Issue #251, opened 2016-06-14
Three issues in one:  moving the mouse creates black areas;  icons disappear after movement; and the map resists returning to the area of the about to move unit.


robertmcrowe:
[MM19stuff.zip](https://github.com/MegaMek/megamek/files/314599/MM19stuff.zip)

robertmcrowe:
note: while the icon disappears, the text label does not.

arlith:
What appears to be happening here is that there's a super image in the tileset that isn't loading properly.  Given the current logging, I can't tell which, but my plan to fix this will be to add some logging code, so in the future we'll be able to tell if something in a tileset can't be loaded.  I may also write a quick utility for testing too (I think such a tool already exists for the mechsets).

I think I will also add a null-guard to the offending code.  Technically, the way things are written, it *is* possible to get a null image for supers, and the code should reflect this.

arlith:
I actually experienced this bug myself, and it's not because of a bad image in the tileset.  I haven't been able to figure out how exactly, but for some reason I had the road supers be an array list with a null entry.

arlith:
Closing as a duplicate of #284.

PR #2965, opened 2021-06-24
MekHQ #251: Force Generator Unit Table Faction Fallback

Windchild292:
This fixes https://github.com/MegaMek/mekhq/issues/251 by adding parent faction fallbacks (going back 10x, to prevent infinite looping while handling nested forces (like MOC.pm). This also fixes the Magistracy of Canopus Planetary Militia's parent faction as a typo.


neoancient:
What's the issue you're trying to fix? The selected faction should inherit the parent factions' units, so if there aren't any for the selected faction, there shouldn't be any for the parent factions.

HammerGS:
We are consistently getting report of players gettIng Fedcom contracts in AtB and no force being generated. Generally between 3060-3067 during the Civil war era.

Windchild292:
@neoancient 
Exactly as Hammer describes, although I may have misunderstood how this works (although the fix does work properly for MekHQ).

neoancient:
This is a MekHQ problem and should be fixed there. The Federated Commonwealth ceased to exist in 3057, though the Federated Suns portion continued to use the name for another ten years. Accordingly, FC is used to generated a combined Davion and Steiner military. As that didn't exist after 3057, the RAT generator correctly returns an empty table. If you want a Davion unit, the correct faction code is FS. The RAT Generator is not the problem.

Windchild292:
I've thought a bunch on this, and I concur this isn't a proper fix. Will open the MOC.pm fix separately, then think about a pure MekHQ fix.

PR #2980, opened 2021-06-30
Do not use GUIPreferences from Board

sixlettervariables:
Removes coupling of the `Board` with anything GUI related. This likely needs some work on the `BoardView` side to ensure we toggle that value at the correct points. Fixes the issue with the Boards Validator no longer working.

### DIFF
--- a/megamek/src/megamek/common/Board.java
+++ b/megamek/src/megamek/common/Board.java
@@ -46,7 +46,6 @@ import java.util.Set;
 import java.util.Vector;
 
 import megamek.MegaMek;
-import megamek.client.ui.swing.GUIPreferences;
 import megamek.common.Building.BasementType;
 import megamek.common.annotations.Nullable;
 import megamek.common.event.BoardEvent;
@@ -477,7 +476,7 @@ public class Board implements Serializable, IBoard {
         }
         
         // Internally handled terrain (inclines, cliff-bottoms)
-        initializeAutomaticTerrain(x, y);
+        initializeAutomaticTerrain(x, y, /* useInclines: */ true);
         
         // Add woods/jungle elevation where none was saved
         initializeFoliageElev(x, y);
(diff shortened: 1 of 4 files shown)


megamek/src/megamek/common/Board.java
```diff
@@ -477,7 +476,7 @@ private void initializeHex(int x, int y, boolean event) {
         }
         
         // Internally handled terrain (inclines, cliff-bottoms)
-        initializeAutomaticTerrain(x, y);
+        initializeAutomaticTerrain(x, y, /* useInclines: */ true);
```
Windchild292, line 479: Why include a comment here? Decreases readability IMO, especially when using an IDE that shows that automatically.

megamek/src/megamek/common/Building.java
```diff
@@ -266,7 +266,7 @@ public int getDepth() {
         }
 
         public String getDesc() {
-            return desc;
+            return Messages.getString(desc);
```
Windchild292, line 269: It would be better to this in the constructor.

Windchild292 (commented):
Largely looks good, just an improvement suggestion and an odd comment

megamek/src/megamek/common/Building.java
sixlettervariables: That's the problem though, that causes an issue on class init. I may revert this change if I can figure out the real reason the board validator suddenly stopped running.

Windchild292: I'd recommend swapping to a localized resources.getString then, like the ones in the AbstractDialogs. Haven't had any issues with them during unit testing nor initialization in MekHQ

sixlettervariables: It looks like it failed beginning with Java 11, so at least something to dig into.

Windchild292: Not the first oddity with Messages from that point

megamek/src/megamek/common/Board.java
SJuliez: It allows reading this without mouse-over which is better than having to mouse it. I like it when a parameter is otherwise completely meaningless. This call is so short that I can't see any issue with readability.

Windchild292: I guess I've just gotten used to having it constantly showing when there's a constant used.

Windchild292 (approved):
Approved, with #3085 opened to update the enum.

PR #2993, opened 2021-07-05
megamek potato fixes, round 2

NickAragua:
Second round of fixes for exceptions logged while play-testing megamek on a potato-grade computer. I'm not really sure what brought a lot of these about, but they did get logged and the game did do a lot of "jumping" and "freezing", not to mention having to try to start up MegaMek multiple times from MekHQ before it would be able to get through the whole of loading a particular scenario up.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/tileset/TilesetManager.java
+++ b/megamek/src/megamek/client/ui/swing/tileset/TilesetManager.java
@@ -583,7 +583,9 @@ public class TilesetManager implements IPreferenceChangeListener, ITilesetManage
 
         IPlayer player = entity.getOwner();
 
-        Camouflage camouflage = entity.getCamouflageOrElse(player.getCamouflage());
+        Camouflage camouflage = player != null ?
+                entity.getCamouflageOrElse(player.getCamouflage()) :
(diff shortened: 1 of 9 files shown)


megamek/src/megamek/client/ui/swing/tileset/TilesetManager.java
```diff
@@ -583,7 +583,9 @@ public synchronized void loadImage(Entity entity, int secondaryPos) {
 
         IPlayer player = entity.getOwner();
 
-        Camouflage camouflage = entity.getCamouflageOrElse(player.getCamouflage());
+        Camouflage camouflage = player != null ?
+                entity.getCamouflageOrElse(player.getCamouflage()) :
+                new Camouflage();
```
Windchild292, line 588: I'd prefer the following, which adds brackets and removes the need to mentally swap the trinary, especially as part of camouflage.
```suggestion
        Camouflage camouflage = (player == null) ? new Camouflage()
                : entity.getCamouflageOrElse(player.getCamouflage());
```

Windchild292 (approved):
One comment, otherwise looks good

PR #2998, opened 2021-07-10
Advanced Map Generator Panel: Fixing Woods Size Text

Windchild292:
This fixes a bug mentioned on Facebook, and also cleans up a bunch of typos and removes double space usages.


sixlettervariables (approved):
Speeling looks good.

PR #3001, opened 2021-07-11
Corrected Boards

SJuliez:
Repairs two board files.


Windchild292 (approved):
How'd they end up completely empty?

SJuliez:
My BoardsTagger ate them :D Happened when a board had invalid hexes. Bad bad. Reminds me I gotta look at that PR again where I updated the tagger to see if it won't do this again.

PR #3007, opened 2021-07-12
1092/1105/1950/2695: Person/Tech to Unit and Unit to Person/Tech Assignment Menu Standardization

Windchild292:
This is a minor future-proofing implementation, which can be alternatively handled by #3008


Windchild292:
No longer required given #3008 has been merged

Issue #3005, opened 2021-07-12
[49.2] Firing Phase Freeze from Building IllegalArgumentException


K007:
### Environment
49.2 MHQ bundled MM

### Description
Firing phase freeze up's along with a number of other errors that are shown in the .log file

### Files
[megameklog.txt](https://github.com/MegaMek/megamek/files/6803323/megameklog.txt)
[quicksave.sav.gz](https://github.com/MegaMek/megamek/files/6803329/quicksave.sav.gz)
Playername is "June's Raiders"

Windchild292:
We'll also need campaign and customs for the MekHQ-side of things, plus the mekhqlog.txt.

Windchild292:
The three above are the issues found in the mekhqlog.txt. I've not opened megameklog.txt

Windchild292:
And the Compute changes conclude what I'll be handling here. I'm not certain how the buildings stuff works, and it's above my MegaMek knowledge.

Windchild292:
I'm actually going to close this based on the current issues, and let #3020 subsume the building issue.

PR #3009, opened 2021-07-13
Fixing Princess Null Entity Packet state update

Windchild292:
This was found in the MekHQ log in https://github.com/MegaMek/megamek/issues/3005

### DIFF
--- a/megamek/src/megamek/client/bot/princess/Princess.java
+++ b/megamek/src/megamek/client/bot/princess/Princess.java
@@ -2185,7 +2185,6 @@ public class Princess extends BotClient {
     @Override    
     public void receiveEntityUpdate(final Packet packet) {
         super.receiveEntityUpdate(packet);
-        final Entity entity = (Entity) packet.getObject(1);
-        updateEntityState(entity);
+        updateEntityState((Entity) packet.getObject(1));
     }
 }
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/client/bot/princess/Princess.java
```diff
@@ -1626,10 +1626,12 @@ private void checkForBrokenEnemies() {
...
-    /** Update the various state trackers for a specific entity.
-     * Useful to call when receiving an entity update packet */
-    public void updateEntityState(Entity entity) {
-        if(entity.getOwner().isEnemyOf(getLocalPlayer())) {
+    /**
+     * Update the various state trackers for a specific entity.
+     * Useful to call when receiving an entity update packet
+     */
+    public void updateEntityState(final @Nullable Entity entity) {
+        if ((entity != null) && entity.getOwner().isEnemyOf(getLocalPlayer())) {
```
NickAragua, line 1634: My favorite implication here is that it's possible for the server to send an update for a null entity.

NickAragua (approved):

megamek/src/megamek/client/bot/princess/Princess.java
Windchild292: I was surprised to learn this was the case, but it is according to our internals... and the log I'm working with (which also hits some of the potato fixes you made @NickAragua)

PR #3017, opened 2021-07-16
Unit Label display settings

SJuliez:
Adds some more options to unit label display.

Note that the reduced vehicle chassis names ("Ontos") are created by taking the chassis and removing from the end words from a list of generic words like Heavy, Tank, Defense until something non-generic is found. This works reasonably well for many units and there's no pressure for it to find really short names for all units. The first word of the name is never removed.

The label styles are cycled using the keybind and output a description to the chatbox so players know what the style shows even if they don't have a nicknamed unit at the moment. I intentionally didn't implement some client settings for it as there's too many of those already. Cycling through it is easy.

Other suggestions or combinations of features as below, please tell me.

The current style (still available): 
image

Abbreviated style: 
image

Chassis only:
image

Nickname and chassis:
image

Only nicknames:
image

and no labels (as before)

### DIFF
--- a/megamek/src/megamek/MegaMek.java
+++ b/megamek/src/megamek/MegaMek.java
@@ -94,7 +94,7 @@ public class MegaMek {
                 }
             }
 
-//            configureLogging(logFileName);
+            configureLogging(logFileName);
 
             MegaMek.showInfo();
 
--- a/megamek/src/megamek/MegaMek.java
+++ b/megamek/src/megamek/MegaMek.java
@@ -94,7 +94,7 @@ public class MegaMek {
                 }
             }
 
-            configureLogging(logFileName);
+//            configureLogging(logFileName);
 
             MegaMek.showInfo();
 
--- a/megamek/src/megamek/client/ui/swing/GUIPreferences.java
+++ b/megamek/src/megamek/client/ui/swing/GUIPreferences.java
@@ -19,6 +19,7 @@ import java.awt.*;
 import javax.swing.*;
 
 import megamek.client.ui.swing.boardview.BoardView1;
+import megamek.client.ui.swing.boardview.LabelDisplayStyle;
 import megamek.client.ui.swing.util.PlayerColour;
 import megamek.common.Entity;
 import megamek.common.preference.PreferenceManager;
@@ -239,6 +240,7 @@ public class GUIPreferences extends PreferenceStoreProxy {
     public static final String SHOW_KEYBINDS_OVERLAY = "ShowKeybindsOverlay";
     public static final String OPTIONS_SHOW_UNOFFICIAL = "OptionsShowUnofficial";
     public static final String OPTIONS_SHOW_LEGACY = "OptionsShowLegacy";
+    public static final String UNIT_LABEL_STYLE = "UnitLabelStyle";
     
     // RAT dialog preferences
     public static String RAT_TECH_LEVEL = "RATTechLevel";
@@ -347,6 +349,7 @@ public class GUIPreferences extends PreferenceStoreProxy {
         store.setDefault(GAME_SUMMARY_MINI_MAP, false);
         store.setDefault(ENTITY_OWNER_LABEL_COLOR, true);
         store.setDefault(UNIT_LABEL_BORDER, true);
+        store.setDefault(UNIT_LABEL_STYLE, "NICKNAME");
         store.setDefault(GAME_OPTIONS_SIZE_HEIGHT, 400);
         store.setDefault(GAME_OPTIONS_SIZE_WIDTH, 400);
         store.setDefault(FIRING_SOLUTIONS,true);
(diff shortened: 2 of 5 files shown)


megamek/src/megamek/client/ui/swing/GUIPreferences.java
```diff
@@ -347,6 +349,7 @@ protected GUIPreferences() {
         store.setDefault(GAME_SUMMARY_MINI_MAP, false);
         store.setDefault(ENTITY_OWNER_LABEL_COLOR, true);
         store.setDefault(UNIT_LABEL_BORDER, true);
+        store.setDefault(UNIT_LABEL_STYLE, "NICKNAME");
```
Windchild292, line 352: Can we please use the enum .name method instead? Helps prevent issues with name changes (I've made this mistake in MekHQ before)

megamek/src/megamek/client/ui/swing/boardview/EntitySprite.java
```diff
@@ -71,7 +74,14 @@
...
     // Keep track of ECM state, as it's too expensive to compute on the fly.
     private boolean isAffectedByECM = false;
-
+    
+    private static final Set<String> removableNameStrings = new HashSet<String>( 
+            Arrays.asList("Defense", "Platoon", "Heavy", "Medium", "Light", "Artillery", "Tank", 
+                    "Wheeled", "Command", "Standard", "Hover", "Hovercraft", "Mechanized", 
+                    "(Standard)", "Platoon", "Defense", "Transport", "Vehicle", "Air", 
+                    "Assault", "Mobile", "Platform", "Battle Armor", "Vessel", "Infantry",
+                    "Fighting", "Fire", "Suport", "Reconnaissance", "Fast"));
```
Windchild292, line 83: Why not just keep it as an array? Also, new HashSet<String> is completely superfluous in modern Java, as you only need to specify the class (i.e. String) if it is a wildcard on the backend

megamek/src/megamek/client/ui/swing/boardview/EntitySprite.java
```diff
@@ -88,19 +98,96 @@ public EntitySprite(BoardView1 boardView1, final Entity entity,
...
+                if (!pilotNick().isBlank()) {
+                    return "\"" + pilotNick().toUpperCase() + "\"";
+                } else if (!unitNick().isBlank()) {
+                    return "\'" + unitNick() + "\'";
+                } else {
+                    return "";
+                }
+            default: // ONLY_STATUS
+                return "";
             }
```
Windchild292, line 132: Please indent the switch. It's a minor thing, but I've found that it actually really helps with reading.

megamek/src/megamek/client/ui/swing/boardview/EntitySprite.java
```diff
@@ -88,19 +98,96 @@ public EntitySprite(BoardView1 boardView1, final Entity entity,
...
-                        .toUpperCase();
+            switch (GUIPreferences.getInstance().getUnitLabelStyle()) {
+            case FULL:
+                return standardLabelName();
+            case ABBREV:
+                if (entity instanceof Mech) {
+                    return entity.getModel();
+                } else {
+                    return abbreviateUnitName(standardLabelName());
+                }
```
Windchild292, line 111: Ternary candidate

megamek/src/megamek/client/ui/swing/boardview/EntitySprite.java
```diff
@@ -88,19 +98,96 @@ public EntitySprite(BoardView1 boardView1, final Entity entity,
...
         }
     }
+    
+    /** 
+     * Returns a shortened unit name string, mostly for vehicles. Words contained
+     * in the removableNameStrings list are taken away from the end of the name
+     * until something is encountered that is not contained in that list. 
+     * On Mech names this will typically have no effect.
+     */
+    private static String reduceVehName(String unitName) {
```
Windchild292, line 142: Can we name this reduceVehicleName? Helps with future-proofing

Windchild292 (commented):
Few things first

megamek/src/megamek/client/ui/swing/boardview/EntitySprite.java
SJuliez: I'll change the initialization to Set.of(), much simpler. The reduceVehicleName() must find if a term is contained in that word list, which is simplest and fastest in a HashSet (although that millisecond won't be important)

sixlettervariables:
@SJuliez did not expect that, [but you're correct](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Set.html#of(E...))! The `IllegalArgumentException` thrown in the static initializer causes that ClassDefNotFound exception. What awful behavior...sets should just deduplicate:
> **Throws:**
> `IllegalArgumentException` - if there are any duplicate elements

megamek/src/megamek/MegaMek.java
```diff
@@ -94,7 +94,7 @@ public static void main(String[] args) {
                 }
             }
 
-            configureLogging(logFileName);
+//            configureLogging(logFileName);
```
NickAragua, line 97: Is this commented out code something that needs to be retained?

NickAragua (commented):
One question, looks fine otherwise.

megamek/src/megamek/MegaMek.java
SJuliez: Ouch! Good you saw that. I always do that to get errors in the console. Corrected it.

PR #3035, opened 2021-07-30
update board tag parsing

NickAragua:
The "Tags" enum wasn't parsing tags properly, so I introduced a lookup table.


Windchild292 (approved):
Looked at it... and I'm not a fan of the backend decisions made in saving this, but this should handle the problem for now

PR #3041, opened 2021-08-02
Princess configuration

### LINKED ISSUE #2853: [0.49] Text on bot behaviour control panel is too small
When I go to add a bot to a megamek match it opens a window that lets you set the behaviour using sliders. The text for the sliders is too small and the letters become almost unreadable (left side of the pic). I'm using the 0.49 development snapshot of mekhq and the recommended adoptopen JDK 11 java distro. I remember having issues with garbled text in meklab (the right side of the pic) and I solved them by updating to the adoptopen JDK 16 release. However the new 0.49 release doesn't recognize JDK 16 and the issues returned when I installed 11 instead.

Since we'll be stuck with JDK 11 for the foreseeable future you might be able to fix the problem by using a different font. Only certain bits have problems, most of the text is fine.

megamek bot text problem

### LINKED ISSUE #2958: RFE - Bot 'player' configuration saved as part of savegame for reloading
Suggestion - save bot configuration along with savegame, so that when game is reloaded, the 'ghost player' need only be confirmed as 'resume bot', and the saved configuration will be resumed. (Possibly with a 'resume all bots' option.)

This will be useful for long games featuring multiple bots, where the game may need to be saved and resumed several times, as it will avoid having to reconfigure the bots every time (and possibly remember things like retreat edges, target unit#s, etc)
(e.g. when using the 'role' settings for bots, so that Princess behaves 'appropriately' for the units in question)

I understand that, at present, the server doesn't know which players are bots and which are full clients. This RFE would require that to be tracked. It is suggested that, when the game is reloaded, missing 'players' would continue to be marked as 'ghost player', with the option to replace them with a newly-created bot configuration. Meanwhile, missing 'bots' could be replaced by 1) a player joining the game with their name, 2) a newly-configured bot, or 3) resuming the bot settings saved against that player.

### LINKED ISSUE #2986: Princess setting tooltips and labels are incorrect
The Princess setting dialog has incorrect labels and ranges for Bravery and Self Preservation.  They do not perform as indicated in the tooltips and range labels, and conflict with the available help documentation (which appears to be correct).

Tooltips as they stands now:
Bravery - How quickly will I try to escape when damaged?
Range - Run Away!  -----  I fear Nothing!

Self Preservation - How worried about enemy damage am I?
Range - It's only a flesh wound  -----  It hurts! It really hurts!

Instead, they should look like this (matching the help documentation):
Bravery - How worried about enemy damage am I?
Range - I fear Nothing!  -----  Run Away!

Self Preservation - How quickly will I try to escape when damaged?
Range - It's only a flesh wound  -----  It hurts! It really hurts!

SJuliez:
- Modernizes the Bot Config Dialog
- Princess Presets now only care about the 5 behavior sliders (technically, the rest is saved too, but when selecting a Preset only the five slider values are pulled). The other settings are scenario specific and don't feel like they belong to a "Behavior Preset"
- When adding a bot, the Configuration (<- that means all settings, including prio targets) of already existing local bots (= those added through the Add Bot button in the lobby) can be copied; they are offered at the bottom of the presets list (see screenshots)
- Princess Configurations are saved in savegames; they can be recalled by the new Replace Players Dialog; When this is opened all stored configs are used to preselect and preconfigure Princess bots. If no changes are wanted, it's basically sufficient to press OK after calling up the dialog and all previously existing bots will be restored
- When reinstating bots in the lobby, they are local bots and fully configurable
- Hex and unit targets are now shown with some more info to them. Units can be added in the lobby from the main unit list. For both there's a dialog to enter them somewhat conveniently. Duplicate entries are avoided.

Resolves #2986 
Resolves #2958 
Resolves #2853 

image
image
image


megamek/src/megamek/client/ui/dialogs/BotConfigTargetHexDialog.java
```diff
@@ -0,0 +1,170 @@
...
+        listLabel.setLabelFor(coordsList);
+        listLabel.setDisplayedMnemonic(KeyEvent.VK_S);
+        listLabelPanel.add(listLabel);
+        
+        coordsList.setSelectionMode(ListSelectionModel.MULTIPLE_INTERVAL_SELECTION);
+        coordsList.setLayoutOrientation(JList.VERTICAL);
+        coordsList.setCellRenderer(new BuildingHexRenderer());
+        coordsList.setFont(UIUtil.getScaledFont());
+        coordsList.setVisibleRowCount(6);
+        coordsList.setPrototypeCellValue(new Coords(-21, 22));
```
NickAragua, line 107: This seems a little specific.

megamek/src/megamek/client/ui/swing/lobby/LobbyMekPopup.java
```diff
@@ -372,6 +374,23 @@ private static JMenu squadronMenu(ClientGUI cg, boolean enabled, ActionListener
...
     }
     
+    /**
+     * Returns the "Priority Target" submenu, allowing to assign units as a strategic
+     * target to a local Princess bot.
+     */
+    private static JMenu prioTargetMenu(ClientGUI cg, boolean enabled, ActionListener listener,
+            Collection<Entity> entities) {
+
+        JMenu menu = new JMenu("Set Prio Target for");
```
NickAragua, line 384: I'm sure that we have enough bits that we can spell "Priority" fully.

NickAragua (changes requested):
Two minor nitpicks, looks pretty good otherwise.

HammerGS:
Just a comment from the forums that might be applicable here. 

https://bg.battletech.com/forums/megamek-games/princess-settings-popup-tooltip-vs-help/

megamek/src/megamek/client/ui/dialogs/BotConfigTargetHexDialog.java
SJuliez: If you mean the prototype cell value, that's only passed to the renderer to get the cell size when the list is empty. This prevents it from changing size when it goes from empty to containing items (the changed size doesn't appear directly, only when resizing the dialog, but then it looks a little silly)

megamek/src/megamek/client/ui/swing/lobby/LobbyMekPopup.java
SJuliez: OK, done

SJuliez: OK, done

NickAragua (approved):

PR #3050, opened 2021-08-04
Send dedicated server round reports by email

mjog:
This adds the ability to send round reports via email from the dedicated server, so that players get a record of a game as it progresses, but also to notify players in long running games (e.g. very large games, double blind games, those with simultaneous phases, etc) when something has happened, so they know to re-connect and take their turn.

When starting a dedicated server, if `-mail path/to/email.properties` is specified, an email service is initialised when the server starts up. The given file contains standard JavaMail properties to configure the SMTP connection details. Then, for any players who have set an email address during game setup, a copy of every round report will be emailed to them when generated.

The emails sent are set up so that all email for the same turn are threaded together, to give the same effect as the GUI round report in email clients:

mm-email-example

### DIFF
--- a/megamek/src/megamek/client/bot/BotClient.java
+++ b/megamek/src/megamek/client/bot/BotClient.java
@@ -224,6 +224,10 @@ public abstract class BotClient extends Client {
         });
     }
 
+    public boolean isBot() {
+        return true;
+    }
+
     BotConfiguration config = new BotConfiguration();
 
     protected BoardClusterTracker boardClusterTracker;
--- a/megamek/src/megamek/client/ui/swing/lobby/PlayerSettingsDialog.java
+++ b/megamek/src/megamek/client/ui/swing/lobby/PlayerSettingsDialog.java
@@ -119,7 +119,12 @@ public class PlayerSettingsDialog extends ClientDialog {
     public boolean getForceGP() {
         return butForceGP.isSelected();
     }
-    
+
+    /** Returns the player's email address. */
+    public String getEmail() {
+        return fldEmail.getText().trim();
+    }
+
     // PRIVATE
 
     private final Client client;
@@ -150,7 +155,11 @@ public class PlayerSettingsDialog extends ClientDialog {
     private JComboBox<String> cmbPilot = new JComboBox<String>();
     private JComboBox<String> cmbXP = new JComboBox<String>();
     private MMToggleButton butForceGP = new MMToggleButton(getString(PSD + "butForceGP"));
-    
+
+    // Email section
+    private JLabel labEmail = new JLabel(getString(PSD + "labEmail"), SwingConstants.RIGHT);
+    private JTextField fldEmail = new JTextField(20);
+
     private JPanel panStartButtons = new JPanel();
     private TipButton[] butStartPos = new TipButton[11];
     private JButton butBotSettings = new JButton(Messages.getString(PSD + "botSettings"));
@@ -179,6 +188,9 @@ public class PlayerSettingsDialog extends ClientDialog {
             mainPanel.add(mineSection());
         }
         mainPanel.add(skillsSection());
+        if (!(client instanceof BotClient)) {
+            mainPanel.add(emailSection());
+        }
         mainPanel.add(Box.createVerticalGlue());
     }
     
@@ -251,7 +263,16 @@ public class PlayerSettingsDialog extends ClientDialog {
         panContent.add(butForceGP);
         return result;
     }
-    
+
+    private JPanel emailSection() {
+        JPanel result = new OptionPanel(PSD + "header.email");
+        Content panContent = new Content(new GridLayout(1, 2, 10, 5));
+        result.add(panContent);
+        panContent.add(labEmail);
+        panContent.add(fldEmail);
+        return result;
+    }
+
     private JPanel buttonPanel() {
         JPanel result = new JPanel(new FlowLayout());
         butOkay.addActionListener(listener);
@@ -282,8 +303,9 @@ public class PlayerSettingsDialog extends ClientDialog {
         cmbXP.setSelectedIndex(client.getRandomSkillsGenerator().getLevel());
         butForceGP.setSelected(client.getRandomSkillsGenerator().isClose());
         cmbMethod.addItemListener(e -> adjustSkillMethodTip());
+        fldEmail.setText(player.getEmail());
     }
-    
+
     private void setupStartGrid() {
         panStartButtons.setAlignmentX(Component.LEFT_ALIGNMENT);
         for (int i = 0; i < 11; i++) {
--- a/megamek/src/megamek/server/Server.java
+++ b/megamek/src/megamek/server/Server.java
@@ -80,6 +80,7 @@ import megamek.common.options.IOption;
 import megamek.common.options.OptionsConstants;
 import megamek.common.preference.PreferenceManager;
 import megamek.common.util.BoardUtilities;
+import megamek.common.util.EmailService;
 import megamek.common.util.fileUtils.MegaMekFile;
 import megamek.common.util.SerializationHelper;
 import megamek.common.util.StringUtil;
@@ -150,6 +151,9 @@ public class Server implements Runnable {
     private ServerSocket serverSocket;
 
     private String motd;
+
+    private EmailService mailer;
+
 
     private static class ReceivedPacket {
         public int connId;
@@ -341,7 +345,12 @@ public class Server implements Runnable {
     private final Object serverLock = new Object();
 
     public Server(String password, int port) throws IOException {
-        this(password, port, false, "");
+        this(password, port, false, "", null);
+    }
+
+    public Server(String password, int port, boolean registerWithServerBrowser,
+                  String metaServerUrl) throws IOException {
+        this(password, port, registerWithServerBrowser, metaServerUrl, null);
     }
 
     /**
@@ -352,11 +361,14 @@ public class Server implements Runnable {
      *                                  used
      * @param registerWithServerBrowser a <code>boolean</code> indicating whether we should register
      *                                  with the master server browser on megamek.info
+     * @param mailer an email service instance to use for sending round reports.
      */
     public Server(String password, int port, boolean registerWithServerBrowser,
-                  String metaServerUrl) throws IOException {
+                  String metaServerUrl, EmailService mailer) throws IOException {
         this.metaServerUrl = metaServerUrl;
         this.password = password.length() > 0 ? password : null;
+        this.mailer = mailer;
+
         // initialize server socket
         serverSocket = new ServerSocket(port);
 
--- a/megamek/src/megamek/server/Server.java
+++ b/megamek/src/megamek/server/Server.java
@@ -620,6 +620,10 @@ public class Server implements Runnable {
             conn.close();
         }
 
+        if (mailer != null) {
+            mailer.shutdown();
+        }
+
         connections.removeAllElements();
         connectionIds.clear();
         if (serverBrowserUpdateTimer != null) {
--- a/megamek/src/megamek/server/Server.java
+++ b/megamek/src/megamek/server/Server.java
@@ -841,7 +841,7 @@ public class Server implements Runnable {
         sendServerChat(connId, motd);
 
         // send info that the player has connected
-        send(createPlayerConnectPacket(connId));
+        transmitPlayerConnect(player);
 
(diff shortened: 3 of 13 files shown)


mjog:
This is marked as WIP since I could use some advice about where to actually ask players for their email addresses.

My initial thought was in `ConnectionDialog`, but the code paths to communicate that to the server along with the player name aren't really  appropriate, and since most people won't want to enter their email address, it's poor usability.

The alternative would be to add it to the Player Settings dialog, which seems to be more useful, but may not have any way to easily save the email address (like the connection dialog does for player names), but will look into it.

mjog:
Latest updates fixes a few issues and implements setting email address from PlayerSettingsDialog, which is I think all I'm going to set out to do here.

I'd like to be able to persist the last used email address in the client prefs (like the last used player name), but there's no place to do that in the current client/server model that really fits, since it's the server that effectively creates the player objects and sends them to the client, rather than the other way around. That model makes a lot of sense for re-connecting to existing games, but less so when connecting to a new game. This a problem not specific to this work however (e.g. it also affects other possible remembered player prefs like choice of camo), so I'm not going to try to fix it here.

As such I've unmarking this a WIP. Feedback welcome!

megamek/src/megamek/client/ui/swing/lobby/PlayerSettingsDialog.java
```diff
@@ -179,6 +188,7 @@ private void setupDialog() {
             mainPanel.add(mineSection());
         }
         mainPanel.add(skillsSection());
+        mainPanel.add(emailSection());
```
NickAragua, line 191: Can we make adding this element conditional on the player not being a bot?

megamek/src/megamek/server/Server.java
```diff
@@ -31119,6 +31156,22 @@ private void sendReport() {
      * Send the round report to all connected clients.
      */
     private void sendReport(boolean tacticalGeniusReport) {
+        if (mailer != null) {
+            for (var player: game.getPlayersVector()) {
+                if (!StringUtil.isNullOrEmpty(player.getEmail())) {
```
NickAragua, line 31161: Let's set this to send one email per unique email address in the email address collection. I don't think there's any value in sending multiple report emails to one person.

Additionally, this seems like something that's best spun off into a separate thread, as I don't think the server should be waiting for a potentially lengthy email sending process to complete before moving on.

NickAragua (changes requested):
Can we make configuration and sending of email conditional on the player not being a bot (both the player starting out as a bot and the player disconnecting and being replaced by a bot)?

I'm also mildly concerned about the possibility of this feature being used to spam people. I've suggested some changes to reduce that, and also to prevent email sending (in case of slow email server) from blocking the Server from moving on during turn processing.

megamek/src/megamek/server/Server.java
sixlettervariables: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/Executors.html#newSingleThreadExecutor()

megamek/src/megamek/client/ui/swing/lobby/PlayerSettingsDialog.java
mjog: Fixed - this is now only shown for non-bots.

megamek/src/megamek/server/Server.java
mjog: > Let's set this to send one email per unique email address in the email address collection. I don't think there's any value in sending multiple report emails to the same address.

This is a little bit difficult to do since for double blind situations, each player needs to get their own copy of the report. In the end, since the email address isn't a remembered preference, if people don't want to receive multiple copies, they simply shouldn't enter their email address more than once.

> Additionally, this seems like something that's best spun off into a separate thread, as I don't think the server should be waiting for a potentially lengthy email sending process to complete before moving on.

Yep, email sending is already handled on a separate thread in EmailService (see commit 7be6ed1c). The executor framework was overkill for this, and wouldn't easily allow sending multiple queued messages over the same transport, so I just went with a minimal solution instead.

mjog:
Thanks for the comments, I hear the feedback about not spamming people, but as-is, it works by opt-in that is required for every game. So there's no chance of people getting multiple unwanted copies, unless they (or someone else) deliberately configured multiple players with the same email address.

This could be abused of course, an adversary could enter someone else's email address when they connect as a player, so that they get sent unsolicited round reports. Of course there are literally an unlimited number of other ways that could be achieved, and most of those are much easier than setting up and using MegaMek dedicated server games, so I don't think the risk of that is very hight at all. :)

In the end, since it requires opt-in, and it only available for dedicated server games, and only for dedicated server games that are explicitly and correctly configured to send email, it's highly unlikely to be a source of spam for people.

megamek/src/megamek/client/ui/swing/lobby/PlayerSettingsDialog.java
mjog: Oh, I missed the part about the case where a player disconnects and is replaced by a bot (didn't even realise that was possible!) - will look into that tonight.

mjog:
Thinking some more though about possible attack vectors, there is a risk that adversaries could join a game as an observer with someone else's email address, and it would be pretty trivial do that multiple times, leading the recipient potentially receiving multiple copies of the round reports. This is perhaps not unlikely risk for public and high-profile games. To mitigate that, observers shouldn't be included when sending round reports either.

So there still is some work to do here: Reports should only be sent to non-bot, non-observer players.

rjhancock:
I'm concerned with the potential for both spamming and privacy issues related to sharing of email addresses. In the grand scheme of things, the fears are most likely overblown, but they are there.

I think a good compromise would be to either configure a single email address server side to send them to (still opens it up to abuse but much less so) or to adjust this to send to a configurable collection server for others to view.

This is just my opinion however.

NickAragua:
It's not a particularly major "privacy breach", but we do send a copy of the Player class out to the client during updates and when the client does a save. This potentially exposes other people's email address to parties other than themselves and the server. So, I think either encrypting the email or not sending it during client updates/client saves is probably worthwhile as well.

Granted, I'm pretty sure the value of isolated emails linked to a throwaway user name and a list of mechs is pretty low, so I'm not really sure how much effort we want to put into it. But, on the other hand, we're not in a hurry and it's good to get it right.

rjhancock:
Wasn't thinking so much a breach as that wouldn't apply but a potential leak is there, however small it is (thinking massively isolated here).

I do think something like this has promise just thinking of how to best do it without having privacy/spam issues (or reduce them to barest minimal)

mjog:
> I'm concerned with the potential for both spamming and privacy issues related to sharing of email addresses. In the grand scheme of things, the fears are most likely overblown, but they are there.

I hear that, so let me run through some scenarios to hopefully reassure you it's unlikely to be an issue either way :)

1. Playing a standard game (i.e. without a dedicated server)
    * Spam: Not a problem, since a dedicated server is not in use, and hence no email will ever be sent
    * Privacy:
      * Players that do nothing (i.e. do not enter their email address) are at zero risk
      * Players that do enter their email address will have that transmitted to the server and saved in saved game files there
        * Solution: Don't enter your email address unless you trust the host
      * Players that do enter their email address will have that transmitted to other players
        * TODO: The server should not transmit email addresses to other players - I'll look into implementing this

2. Public/private dedicated server game /without/ email service enabled & configured
    * Spam: Same as above - no email can be sent if the email service is not enabled
    * Privacy: Same as above

3. Private dedicated server game /with/ email without email service enabled & configured
    * Spam: Possible problem
      * Solution: Host shouldn't allow players to join that they don't trust, or kick those that are abusive
    * Privacy: Same as above

4. Public dedicated server game /with/ email without email service enabled & configured
    * Spam: Possible problem
      * Solution: Host shouldn't allow players to join that they don't trust, or kick those that are abusive
      * TODO: Server should only email players who are actually players, not bots or observers
    * Privacy: Same as above

So in the end, the situation is basically: Don't enter your email address unless you trust the other players and the dedicated server host, and people running dedicated servers with the email service enabled & configured have a responsibility to moderate the players that do connect by kicking kick those that are abusive.
 
> I think a good compromise would be to either configure a single email address server side to send them to (still opens it up to abuse but much less so) or to adjust this to send to a configurable collection server for others to view.

This more or less defeats the stated purposes above for implementing this in the first place, and particularly for double-blind games this would leak information, so I don't think this is an effective solution, unfortunately.

rjhancock:
There are additional avenues of possible spam/privacy issues as well as possible legal issues (GDPR/CPPA both MIGHT come into play as dealing with PII).

Server itself could get hacked and have it's settings changed without the "owning" player knowing about it.

Regardless, there are still things that need to be considered before it could be merged. Let's get this done right and reduce the issues closer to 0.

NickAragua:
Let's be fair though, this is a pretty low-risk low-value situation. This isn't a situation where the information is being retained over a long term or any amount of money is exchanging hands. At the end of the day, MegaMek is provided "as is" so while we should take some reasonable effort to prevent information leakage, this isn't an ISO 9000 type situation. I'm ok with the proposed todos.

rjhancock:
Oh I agree and even stated above the risk is insanely low. I just prefer getting as close to 0 as possible. :)

I'm just thinking of odd ball situations that could happen that I'd like to see be made harder to happen is all. To be fair, I think both solutions can even work together and let users flip between them or have both if desired.

mjog:
Latest two updates implements two things:

1. Filters out private player data (e.g. the email address) when sending to other connections. Player's see their own email address when (re)connecting, but other players every even get sent it.
2. Sends round report email only to actual players, not observers.

The only thing outstanding now is handling the situation where a human player is replaced by a bot. There's no way I can tell to determine if a player is a bot on the server, so I'm not exactly sure how to deal with that, tbh. Any suggestions?

mjog:
Some possibilities include:

 * Having the bot send an update after connecting to null out its email address (has same problem as setting other player defaults as I outlined above though)
 * Adding `IPlayer.isBot` method or equivalent, and having the bot transmit that state (probably requires changing COMMAND_CLIENT_NAME packets, which is maybe out of scope for this)

NickAragua:
I like the idea of a connecting bot sending a "don't send emails" flag to the server via a separate message.

megamek/src/megamek/client/ui/swing/lobby/PlayerSettingsDialog.java
mjog: latest pish ensures that bots don't get emailed either

mjog:
Okay, latest changes ensure that bots don't get emailed either.

I ended up a simple `isBot` to `IPlayer` and passed that through when players first connect, since it may well be useful to know whether a player is a bot in the future for other things, and a "don't send email" flag is at the moment redundant since that can be handled by just not sending an email address.

There' some cleanup around the packets that get sent when a client first connects that could be made here, I'll submit a MR for that separately once this lands since it's otherwise unrelated.

I don't think there's anything outstanding left to do here, is this good to merge?

NickAragua:
I'll take a look tonight.

mjog:
One thing I released I haven't done is documented is the email.properties file. There's no specific docs for the dedicated server aside from a brief mention on the readme, so maybe this should be added to the wiki page instead?

A complete example would look like:

```
megamek.smtp.from = MyMegaMekGame <megamek@example.com>
megamek.smtp.login = megamek
megamek.smtp.password = myl33tp4ssw0r3
mail.smtp.auth = true
mail.smtp.starttls.enable = true
mail.smtp.host = smtp.example.com
mail.smtp.port = 587
```

megamek/src/megamek/client/bot/BotClient.java
```diff
@@ -113,7 +113,7 @@ public void run() {
     }
 
     public BotClient(String playerName, String host, int port) {
-        super(playerName, host, port);
+        super(playerName, true, host, port);
```
NickAragua, line 116: The other way you could implement the isBot() property is just having the BotClient class override it and return a fixed true, while the regular Client returns a fixed false. That way you don't have to modify the constructor and store the value.

NickAragua (commented):
One more suggestion for consideration.

megamek/src/megamek/client/bot/BotClient.java
mjog: Good point, ta!

Fixed.

mjog:
Thanks for merging this!

rshddb40 (approved):

PR #3055, opened 2021-08-06
attacks on hull down tanks now go to appropriate location

### LINKED ISSUE #3048: Potential Bug (0.48): Hulldown Turretless Vehicles (observed with LRM carrier) takes hits to 'BD' location
### Environment
Version 0.48 MekHQ packaged MegaMek
OS: Windows 10
Java version: OpenJDK 11

### Description
Hull down turretless vehicle (LRM carrier in this case) was fired at by other units whilst in a 'prepared position' terrain type. This vehicle was fired on and the shots hit the 'BD' location.

In case this changes things, the vehicle was fired at using Swarm LRMs.

Screenshot: 
image

### Files
Apologies - no MegaMekLog available in this case as I forgot to save this when we stopped playing for the night.

NickAragua:
What it says on the tin. ToHitData.SIDE_X didn't have the same numbers as Tank.LOC_X. Fixes #3048

### DIFF
--- a/megamek/src/megamek/common/LargeSupportTank.java
+++ b/megamek/src/megamek/common/LargeSupportTank.java
@@ -14,6 +14,7 @@
 package megamek.common;
 
 import java.util.ArrayList;
+import java.util.Map;
 
 import megamek.MegaMek;
 import megamek.common.options.OptionsConstants;
@@ -49,6 +50,15 @@ public class LargeSupportTank extends SupportTank {
     // tanks have no critical slot limitations
     private static final int[] NUM_OF_SLOTS = { 25, 25, 25, 25, 25, 25, 25, 25 };
 
+    // maps ToHitData - SIDE_X constants to LOC_X constants here for hull down fixed side hit locations
+    protected static final Map<Integer, Integer> SIDE_LOC_MAPPING = 
+        Map.of(ToHitData.SIDE_FRONT, LOC_FRONT,
+                ToHitData.SIDE_FRONTLEFT, LOC_FRONTLEFT,
+                ToHitData.SIDE_FRONTRIGHT, LOC_FRONTRIGHT,
+                ToHitData.SIDE_REARLEFT, LOC_REARLEFT,
+                ToHitData.SIDE_REARRIGHT, LOC_REARRIGHT,
+                ToHitData.SIDE_REAR, LOC_REAR);
+    
     @Override
     public String[] getLocationAbbrs() {
         return LOCATION_ABBRS;
--- a/megamek/src/megamek/common/LargeSupportTank.java
+++ b/megamek/src/megamek/common/LargeSupportTank.java
@@ -14,7 +14,6 @@
 package megamek.common;
 
 import java.util.ArrayList;
-import java.util.Map;
 
 import megamek.MegaMek;
 import megamek.common.options.OptionsConstants;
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/common/Tank.java
```diff
@@ -98,6 +98,9 @@
 
     private static String[] LOCATION_NAMES_DUAL_TURRET = { "Body", "Front",
             "Right", "Left", "Rear", "Rear Turret", "Front Turret" };
+    
+    // maps ToHitData - SIDE_X constants to LOC_X constants here for hull down fixed side hit locations
+    private static final int[] SIDE_LOC_MAPPING = { 1, 4, 3, 2 };
```
neoancient, line 103: Can we use the named constants here?

megamek/src/megamek/common/Tank.java
```diff
@@ -958,7 +960,7 @@ public HitData rollHitLocation(int table, int side, int aimedLocation,
                 // the opposite direction to which it entered the hex it went
                 // Hull Down in hit side they come in from
                 else {
-                    nArmorLoc = side;
+                    nArmorLoc = SIDE_LOC_MAPPING[side];
```
neoancient, line 963: What about superheavy tanks/large support vehicles, which have two additional side locations?

neoancient (changes requested):

megamek/src/megamek/common/Tank.java
NickAragua: Good question, I'll have to do some research on that one.

NickAragua: As it turns out, superheavy tanks and large support vehicles can't go "hull down", but are effectively subject to partial cover rules (TacOps 2018 reprint). That's a little beyond the scope of this change, but I did introduce a mapping in the relevant class for future use.

megamek/src/megamek/common/LargeSupportTank.java
```diff
@@ -49,6 +50,15 @@
     // tanks have no critical slot limitations
     private static final int[] NUM_OF_SLOTS = { 25, 25, 25, 25, 25, 25, 25, 25 };
 
+    // maps ToHitData - SIDE_X constants to LOC_X constants here for hull down fixed side hit locations
```
neoancient, line 53: This should be in SuperHeavyTank, which is the parent class of LargeSupportTank, so it will cover both.

neoancient: Actually, nevermind about that. SuperHeavyTank overrides `rollHitLocation, so this never gets used anyway.

NickAragua: To be honest, given SuperHeavyTank's "rollHitLocation" implementation, I think this discussion is moot - a hit on a particular side always directly translates to a single hit location with no random variation (unlike a standard-sized tank, for example, a side hit can't hit the front or the back). So even if a superheavy tank is considered hull down, it doesn't affect which location gets hit with an incoming attack. I'm going to remove the specialized location mapping I introduced, because it's already carried out within that method.

NickAragua: That being said, I don't think the "hull down"/"partial cover" rules are being applied correctly to large vehicles currently anyway, I'm not seeing a partial cover/hull down modifier, and, as written, attacks over the partial cover won't actually hit the turret. Might as well see if I can get that implemented while I'm in there.

NickAragua: I took a look at the relevant code and I'm not optimistic that I won't break something else while implementing that, so maybe I'll focus on just fixing the existing bug and hold off on dealing with superheavy tank hulldown/partial cover rules until after the dev release.

neoancient (approved):

PR #3058, opened 2021-08-07
stop bot walking into water with leg armor breaches

### LINKED ISSUE #3031: [0.48] Princess could avoid water but doesn't. Ends up with breached legs.
### Environment
0.48, both Megamek and MekHQ
Window 10
OpenJDK 11

### Description
image
Locust located at 2439 and 2335. 
They have both have one leg that has no armor.  
Using Replace Player, set princess to all 5s, auto flee to the south, she will not avoid the water on 2440 & 2441. 
Without armor a location always breaches. 

### Files
[autosave_2021-07-27_23-43-53.sav.gz](https://github.com/MegaMek/megamek/files/6897169/autosave_2021-07-27_23-43-53.sav.gz)

NickAragua:
Pretty simple improvement to bot behavior which prevents it from stepping into water with breached leg armor. I could probably make it more accurate for tripods and quads, but it seems to me like stepping into water with a guaranteed fall over is generally a bad idea, so we'll avoid it entirely unless feedback indicates otherwise (the only situation I can think of where I'd even be willing to consider intentionally stepping into water with a breached leg is if I had a quad mech with only one breached leg).

Fixes #3031

### DIFF
--- a/megamek/src/megamek/common/pathfinder/CachedEntityState.java
+++ b/megamek/src/megamek/common/pathfinder/CachedEntityState.java
@@ -155,8 +155,8 @@ public class CachedEntityState {
                         ((backingEntity.getArmor(TripodMech.LOC_RLEG) > 0) ? 0 : 1);
             } else if (backingEntity instanceof Mech) {
                 numBreachedLegs = 
-                        ((backingEntity.getArmor(TripodMech.LOC_LLEG) > 0) ? 0 : 1) +
-                        ((backingEntity.getArmor(TripodMech.LOC_RLEG) > 0) ? 0 : 1);
+                        ((backingEntity.getArmor(Mech.LOC_LLEG) > 0) ? 0 : 1) +
+                        ((backingEntity.getArmor(Mech.LOC_RLEG) > 0) ? 0 : 1);
             } else {
                 numBreachedLegs = 0;
             }
--- a/megamek/src/megamek/common/pathfinder/CachedEntityState.java
+++ b/megamek/src/megamek/common/pathfinder/CachedEntityState.java
@@ -7,6 +7,8 @@ import java.util.Map;
 import megamek.common.Entity;
 import megamek.common.Mech;
 import megamek.common.MiscType;
+import megamek.common.QuadMech;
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/common/pathfinder/CachedEntityState.java
```diff
@@ -132,4 +135,33 @@ public boolean isAmphibious() {
...
+                        ((backingEntity.getArmor(QuadMech.LOC_RARM) > 0) ? 0 : 1);
+            } else if (backingEntity instanceof TripodMech) {
+                numBreachedLegs = 
+                        ((backingEntity.getArmor(TripodMech.LOC_LLEG) > 0) ? 0 : 1) +
+                        ((backingEntity.getArmor(TripodMech.LOC_CLEG) > 0) ? 0 : 1) +
+                        ((backingEntity.getArmor(TripodMech.LOC_RLEG) > 0) ? 0 : 1);
+            } else if (backingEntity instanceof Mech) {
+                numBreachedLegs = 
+                        ((backingEntity.getArmor(TripodMech.LOC_LLEG) > 0) ? 0 : 1) +
+                        ((backingEntity.getArmor(TripodMech.LOC_RLEG) > 0) ? 0 : 1);
```
SJuliez, line 159: For what it's worth, lines 158 and 159 could refer to Mech.LOC_ instead of TripodMech.LOC_

megamek/src/megamek/common/pathfinder/DestructionAwareDestinationPathfinder.java
```diff
@@ -282,6 +289,17 @@ private boolean friendlyFireCheck(Entity shooter, IGame game, Coords position, b
         return false;
     }
     
+    /**
+     * Simplified logic for whether a mech going into the given hex will flood its legs.
+     */
+    private boolean waterCheck(BulldozerMovePath path) {
```
SJuliez, line 295: Maybe call this willFloodLeg() or something more expressive than waterCheck()?

SJuliez (approved):
If I may, this looks good to me with two minor suggestions.

Issue #2951, opened 2021-06-18
Landed Drop ship missing Firing at Grounded Dropship Penalty


HammerGS:
Reported from Discord
"Hey, so I don't know if this is Megamek thing, an Aero Sanity Mod thing, or a StratCon thing for balance, but:

Where is the -4 grounded dropship modifier? From what Total Warfare says, that modifier is supposed to be there whether the grounded dropship is immobile or not"

Poked around in Weaponattackactions.java line 3727 and couldn't find reference to a -4 per the rules on Total Warfare pg 110

### Environment
What version of MegaMek does your issue apply to?  0.49.1
What operating system are you using?  Windows

### Description
![image](https://user-images.githubusercontent.com/13752452/122615250-7342a380-d045-11eb-94c0-0e78597951b5.png)

### Files
No save available

vulcanmagus:
I'm the player in from the screenshot
here's my game save
[Epoch.sav.gz](https://github.com/MegaMek/megamek/files/6679372/Epoch.sav.gz)
[megameklog.txt](https://github.com/MegaMek/megamek/files/6679376/megameklog.txt)

Player Name: Epoch

PR #3060, opened 2021-08-08
update target to-hit modifiers for large objects

NickAragua:
Looking at issue #2951, it turns out from TW that even non-spheroid dropships that are technically mobile get the "immobile" modifier when being shot at. While looking at the relevant section I also noticed some modifiers that get applied to large support tanks and small craft.


megamek/src/megamek/common/actions/WeaponAttackAction.java
```diff
@@ -4274,8 +4277,7 @@ else if ((atype != null)
             ToHitData immobileMod = Compute.getImmobileMod(target, aimingAt, aimingMode);
             // grounded dropships are treated as immobile as well for purpose of
             // the mods
-            if ((null != te) && !te.isAirborne() && !te.isSpaceborne() && (te instanceof Dropship)
-                    && ((Aero) te).isSpheroid()) {
+            if ((null != te) && !te.isAirborne() && !te.isSpaceborne() && (te instanceof Dropship)) {
                 immobileMod = new ToHitData(-4, Messages.getString("WeaponAttackAction.ImmobileDs"));
             }
```
Windchild292, line 4282: This code is odd to me... why are we just overriding when it could instead be an else/if statement that assigns a value to the immobileMod. That would decrease the processing required

megamek/src/megamek/common/actions/WeaponAttackAction.java
```diff
@@ -4287,10 +4289,20 @@ else if ((atype != null)
...
         // -1 to hit a SuperHeavy mech
-        if (te != null && (te instanceof Mech) && ((Mech) te).isSuperHeavy()) {
+        if ((te != null) && (te instanceof Mech) && ((Mech) te).isSuperHeavy()) {
             toHit.addModifier(-1, Messages.getString("WeaponAttackAction.TeSuperheavyMech"));
         }
         
+        // large support tanks get a -1 per TW
+        if (te instanceof LargeSupportTank) {
+            toHit.addModifier(-1, Messages.getString("WeaponAttackAction.TeLargeSupportTank"));
+        }
```
Windchild292, line 4299: What about large Fixed-Wing Craft? They don't use LargeSupportTank. I'm also not sure about Support VTOLs being handled by this either

megamek/src/megamek/common/actions/PhysicalAttackAction.java
```diff
@@ -241,8 +242,12 @@ protected static void setCommonModifiers(ToHitData toHit, IGame game,
                 toHit.addModifier(-2, "target is large support tank");
             }
```
Windchild292, line 243: What about large Fixed-Wing Craft? From what I can tell they don't use LargeSupportTank. I'm also not sure about Support VTOLs being handled by this either (I know this code hasn't changed, but it's the same bit as added code below)

Windchild292 (commented):
A few rule/functionality questions first.

NickAragua:
Made some changes to better catch "large support vehicles" unless they're airborne, localized some of the messages in Physical Attack Action, introduced if-else block to reduce immobile computations.

megamek/src/megamek/common/actions/WeaponAttackAction.java
```diff
@@ -4274,12 +4274,15 @@ else if ((atype != null)
         boolean mekMortarMunitionsIgnoreImmobile = wtype != null && wtype.hasFlag(WeaponType.F_MEK_MORTAR) 
                 && (atype != null) && (munition == AmmoType.M_AIRBURST);
         if (wtype != null && !(wtype instanceof ArtilleryCannonWeapon) && !mekMortarMunitionsIgnoreImmobile) {
-            ToHitData immobileMod = Compute.getImmobileMod(target, aimingAt, aimingMode);
+            ToHitData immobileMod = null;
```
Windchild292, line 4277: The null assignment isn't required given it will be initialized immediately after.

Windchild292 (commented):
Just one more thing, otherwise this looks good

Windchild292 (approved):

PR #3077, opened 2021-08-13
fix always-retreat bot behavior

NickAragua:
The semantics behind one of the behavior settings got changed, so Princess is getting updated to account for it.


sixlettervariables (approved):
Brave Sir Robin mode huh.

Windchild292 (approved):

PR #3096, opened 2021-08-20
Russian names cleanup

HeckfyEx:
I had removed from the pool names from the different slavs, patronymics, and hypocorisms and sprinkled instead some minority's names, namely: Adyghe, Bashkyr, Buryat, Erzya, Moksha, Yakut, Chechen, Tartar and some common names of Turcic origin. Addes some weights, mostly by feel though.


Windchild292:
Will review likely in the morrow

Windchild292 (commented):
I'm largely concerned withe the reduction of (what seem to be) variant options for female Russian given names.

HeckfyEx:
The culled names were either family names, patronymics, hypocorisms (which do not work in any official capacity), used aberrant/ambiguous spelling(like 4 or 5 progressively more weird ways to write Anastasiya) or were Western Slav names (one of the biggest offenders, IIRC) and the special mention goes to hypocorisms for male names like Kostya and Kolya.

HeckfyEx:
1) Added and/or weighted slightly higher surnames from the list of family names common to every region.
2) Tried to correct and unify  as much ambiguous spelling as possible.
3) Removed or changed to proper Russian form surnames ending in "off". My reasoning is that those not Russian surnames, but of Russian origin and should be seeded in pools of "western" countries if they are not there already.
4)Added some family names of notable Afro-russians.

Windchild292:
Can you restore the aberrant/ambiguous spelling and hypocorisms to the female name list? These are both valid ways of names changing over the years (same as the male names, but... let us just not do that for base confusion). Sam, Rob and Bob are three hypocorisms off the top of my head that became regular names over centuries.

Further, the Russian list is also Russian origin, so can you restore the "off"-ending surnames because of that.

HeckfyEx:
1) The thing with aberrant spelling is that in this case it isn't result of names changing over years( that would be vowel shifts and changes mostly) but of combination of 2 or 3 ways of transliterating Cyrillic alphabet into Latin one e.g. armenian surname Аладж(а)ян became transliterated as Alajajan which doesn't read as clearly as Alazhayan or the one that written as Jarov when both Zharov and Yarov exist.
2) I'm somewhat iffy on hypocorisms mostly because a) we historically operate on names whitelist - first it was Menologium, then menologium with governmentally approved dictionary and b) adding every hypocorysm to every name (or even more common ones) will be mind-blowing amount of work and will bloat the list to high heaven and beyond. I can copy over the ones that were in the old list instead if you ok with that.
3) Not so much as restore it, but duplicate every surname that ends on "v" and change it to "ff"

Windchild292:
1) Ah, differing translation spellings are fine to leave off
2) Just the old ones is fine, leaves a bit more variance.
3) Just the old ones is fine

HeckfyEx:
Done.

Windchild292 (approved):

PR #3099, opened 2021-08-22
Update Anubis ABS-3MC.mtf

IanBellomy:
Anubis ABS-3MC: Right torso only has 11 crits. Appears to be missing '-Empty-' in slot 12.


HammerGS (approved):

HammerGS:
Not entering a history text entry for this.

HammerGS:
and thanks!

PR #3100, opened 2021-08-22
Fix Quad Location Keys

IanBellomy:
Appears that some quad mech files have quad-specific crit-table location keys and others have biped keys.

It would be nice for other mtf consumers if keys were consistent. 

AFAIK there's no _need_ for quad mech files with biped keys to have biped keys instead of quad keys, but someone may want to chime in on that.


HammerGS:
I would want on of the coders to confirm changing the name isn't going to cause issues. I'm good with it from a data POV.

Windchild292:
Just tested the blue flame in MekHQ, no issues with the change.

Windchild292 (approved):

PR #3112, opened 2021-08-26
Update Atlas III AS7-D3.mtf

IanBellomy:
99.5% certain 'Clan Streak SRM6 Ammo' should be 'IS Streak SRM6 Ammo'.
The D3 has an IS Streak 6 (not a clan one) and the earlier D2 has IS Streak 6 ammo.


HammerGS (approved):

HammerGS:
Thanks for the fix.

PR #3130, opened 2021-09-02
MegaMek XML Util: Fixing Two Escaping Issues

Windchild292:


sixlettervariables (approved):
I don't recall why I left it at 1.0 XML escaping, but if it works I don't see why we can't move to 1.1.

PR #3131, opened 2021-09-03
3089: Fixing messages implementation to handle localization

### LINKED ISSUE #3089: 0.49.3-Snapshot: Game Options Russian Display
Currently Game Options displays in Russian for Russian language computers, not following the client option.

Windchild292:
This should properly fix the localization issues that have been popping up since 0.49.0 and the update to Java 11. It thus fixes #3089.


SJuliez (approved):
Nice! Seems to work. And I no longer have a mix of German and English option texts.

PR #3138, opened 2021-09-06
fix NPE resulting in inability to close 'add combat unit' dialog

NickAragua:
Just NPE protection, and also added description for posterity of nested ternary operator logic I couldn't decipher.


Windchild292 (changes requested):
getGame is rarely null protected, although not required to create an Entity (apparently...)

I'd recommend just tagging Entity::getGame as nullable, but would prefer preventing an Entity from having a null Game ever (as it should be non-nullable).

Windchild292 (approved):

PR #3139, opened 2021-09-06
restore 'rotate board' option

NickAragua:
What it says on the tin. When right-clicking to "set a board as", there's now an option in the popup menu to set it as a 180 degree rotated version.


megamek/src/megamek/client/ui/swing/lobby/MapListPopup.java
```diff
@@ -35,23 +35,24 @@ static ScalingPopup mapListPopup(List<String> boards, int numButtons, ActionList
         boolean oneSelected = boards.size() == 1;
 
         ScalingPopup popup = new ScalingPopup();
-        popup.add(singleBoardMenu(oneSelected, listener, numButtons, boards));
+        popup.add(singleBoardMenu(oneSelected, false, listener, numButtons, boards));
+        popup.add(singleBoardMenu(oneSelected, true, listener, numButtons, boards));
```
Windchild292, line 39: Can we prevent rotation for non-even boards here instead of where we handle the rotate?

Windchild292 (commented):
In general looks good, just one improvement idea

megamek/src/megamek/client/ui/swing/lobby/MapListPopup.java
NickAragua: I'll take a look. The board collection here is just a list of strings so I don't know if we have immediate access to any of that metadata at this point.

HoneySkull:
Sincere "thank you" for this one!

megamek/src/megamek/client/ui/swing/lobby/MapListPopup.java
NickAragua: Turns out it wasn't too much trouble to omit the 'rotate board' option when displaying non-even maps, since the map filter tells us the width of the maps being looked at. I'm going to leave the safety checks in the other places to avoid mangled maps if we come in there from other workflows.

Windchild292 (approved):

PR #3144, opened 2021-09-07
protection vs npes when retrieving force camo

NickAragua:
Fixes another NPE when retrieving camo, this time for forces, addressing a behavior that prevents loading a game.

### DIFF
--- a/megamek/src/megamek/client/generator/skillGenerators/AbstractSkillGenerator.java
+++ b/megamek/src/megamek/client/generator/skillGenerators/AbstractSkillGenerator.java
@@ -52,6 +52,10 @@ public abstract class AbstractSkillGenerator implements Serializable {
     }
 
     public SkillLevel getLevel() {
+        if (level == null) {
+            level = SkillLevel.REGULAR;
+        }
+        
         return level;
     }
 
--- a/megamek/src/megamek/common/force/Force.java
+++ b/megamek/src/megamek/common/force/Force.java
@@ -117,6 +117,10 @@ public final class Force implements Serializable {
     }
 
     public Camouflage getCamouflageOrElse(final IGame game, final Camouflage camouflage) {
+        if (getCamouflage() == null) {
+            setCamouflage(new Camouflage());
+        }
+        
         return getCamouflage().hasDefaultCategory()
                 ? ((getParentId() == NO_FORCE) ? camouflage : getParent(game).getCamouflageOrElse(game, camouflage))
                 : getCamouflage();


Windchild292 (changes requested):
This is not a proper fix. AbstractIcon is non-nullable by design and, unless I'm missing something, cannot possibly be uninitialized.

NickAragua:
Honestly, I'm getting a little tired of this "by design is not nullable" stuff. It obviously is nullable, no matter how much you wish for it to not be nullable.

Windchild292:
You're treating the symptom, not the problem. Finding the actual problem will prevent further issues, which go deeper than the basis. I always nullable annotate, which makes it easy to tell what is permissible.

As for the second exception, the actual error is in MekHQ, not MegaMek. Namely, AtBDynamicScenario's effectiveOpforSkill isn't saved to file and thus causes the exception. If that value should be transient, then it needs reinitializing. It also meant that previously the OpFor skill wasn't persisting between saves. Same issue with AtBDynamicScenario's effectiveOpforQuality.

Windchild292:
Skill Level is only serialized in MekHQ, and is null because of the explanation above.

Camouflage... I think you are might be correct, provided it is a null camouflage instead of an unfound force (if you still have the exception I should be able to tell). That could be fixed by setting a default value for a force's camouflage, thus preventing the base problem.

NickAragua:
Don't have the logs any more from the first exception, but it happened when loading this file from yesterday's nightly build
[midmission.sav.gz](https://github.com/MegaMek/megamek/files/7122624/midmission.sav.gz)

Windchild292:
Yup, serialization issue. For some unknown reason the force camouflage failed to serialize and thus is properly loading as null.

Windchild292:
I can only recreate this by loading an pre-force camouflage save file... like the provided save file. It dates from 0.49.3-SNAPSHOT, not 0.49.4-SNAPSHOT... and thus loading is not guaranteed.

Windchild292:
The second exception is properly fixed by https://github.com/MegaMek/mekhq/pull/2869, the first is related to migrating MegaMek save games, which we do not support. I believe this can be closed.

NickAragua:
I don't buy the argument that we don't need to null proof in this situation. Having encountered the error, no amount of "this should theoretically not happen" is going to convince me otherwise.

Windchild292:
It's not theoretical. The force camouflage is user error, the skill level is missing serialization.

If you have this occur with a valid save file, then we can add null protection. If not, I'm never going to approve this.

PR #3148, opened 2021-09-09
Remove MegaMekFile

mjog:
There's no reason for MegaMekFile to exist, this removes it.


Windchild292 (changes requested):
MegaMekFile is not a wrapper class for file but instead a way to handle user data. This removes functionality without replacing it.

mjog:
> MegaMekFile is not a wrapper class for file but instead a way to handle user data. This removes functionality without replacing it.

Not at all. See 0ccada6

NickAragua:
If you're going to do this, I think you'll need to make a companion pull request for MekHQ and MegaMekLab, because those use MegaMekFile as well. Additionally, if we do this, we'll probably need to coordinate with the people who code MekWars, because I bet that uses this data structure also.

Windchild292:
It doesn't maintain current functionality, and actually exasperates the current directory issues.

The code in question will return the userdata directory if it exists regardless of it being empty or containing the specified file. MegaMekFile returns the path for the specified file, which allows overwriting a single image without overwriting the whole directory. The current problem is that we don't properly merge directories, so if you have a directory that is used for userdata it needs to be the whole directory instead of just the individual file.

Just try running this with the /userdata/data/ directory created, but empty... or just using a single file (I used the default format for MekHQ, which has a user ranks file automatically).

Windchild292:
I'm closing this PR as all it does it break any uses of the userdata folder that aren't a complete data replacement.

PR #3159, opened 2021-09-22
nerf mausers

NickAragua:
Per recent errata, infantry weapons have been nerfed to do a maximum of .6 damage per trooper when used as primary weapons. They do, however, get the "infantry burst fire" modifier.

### DIFF
--- a/megamek/src/megamek/common/Infantry.java
+++ b/megamek/src/megamek/common/Infantry.java
@@ -73,6 +73,8 @@ public class Infantry extends Entity {
             | FIRE_ENGINEERS | MINE_ENGINEERS | SENSOR_ENGINEERS
             | TRENCH_ENGINEERS;
 
+    public static final double PRIMARY_WEAPON_DAMAGE_CAP = 0.6;
+    
     /**
      * squad size and number
      */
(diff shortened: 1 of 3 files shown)


neoancient (approved):
Math.min() would be much more concise.

megamek/src/megamek/common/Infantry.java
```diff
@@ -73,6 +73,8 @@
             | FIRE_ENGINEERS | MINE_ENGINEERS | SENSOR_ENGINEERS
             | TRENCH_ENGINEERS;
 
+    public static final double PRIMARY_WEAPON_DAMAGE_CAP = 0.6;
```
Windchild292, line 76: Can we have this in MegaMekConstants instead? Would be nice to start grouping our constants together there.

megamek/src/megamek/common/Infantry.java
```diff
@@ -2275,12 +2277,27 @@ public double getDamagePerTrooper() {
             return 0;
         }
 
-        double damage = primaryW.getInfantryDamage() * (squadsize - secondn);
+        // per 09/2021 errata, primary infantry weapon damage caps out at .6
+        double adjustedDamage = primaryW.getInfantryDamage() > PRIMARY_WEAPON_DAMAGE_CAP ? 
+                PRIMARY_WEAPON_DAMAGE_CAP : primaryW.getInfantryDamage();
```
Windchild292, line 2282: Math.min would be better

Windchild292 (approved):

megamek/src/megamek/common/Infantry.java
neoancient: This constant has no use outside this class, and this is the logical place for it.

Windchild292: It is isolated here, but I'm thinking about the far future. ​If we have all of our constants isolated to a single file, we could eventually do a constant definition file for significantly increased user customizability.

PR #3160, opened 2021-09-22
Correct MUL Force Loading

SJuliez:
Corrects loading force info saved with MULs.

@Windchild292 I currently have no way of creating a MUL with camo info. I assume it's stored as 
Force_Name|ID|CamoPart1|CamoPart2\Subforce|ID|CamoPart1|CamoPart2 
In that case it should now work as well. If you have such a MUL I'll test it.


Windchild292:
This PR will break MekHQ hard, as it relies on this format. We can't use "/" as the escape as that is required for pathing, hence the swap to using single and double bars. I missed updating Forces::forceStringFor in the force camouflage PR, my apologies for the inconvenience.

Windchild292:
Subsumed by #3161

PR #3161, opened 2021-09-22
Corrected MUL Force Saving

Windchild292:
This subsumes #3160 by moving MegaMek force strings to the same format at MekHQ force strings.

Plus a zip where this is the case: [The Killing Kind.zip](https://github.com/MegaMek/megamek/files/7213652/The.Killing.Kind.zip)

### DIFF
--- a/megamek/src/megamek/common/force/Forces.java
+++ b/megamek/src/megamek/common/force/Forces.java
@@ -359,17 +359,16 @@ public final class Forces implements Serializable {
         return (owner != null) && !owner.isEnemyOf(player);
     }
     
-    public String forceStringFor(Entity entity) {
-        List<Force> ancestors = forceChain(entity);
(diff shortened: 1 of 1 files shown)


SJuliez:
Ah now I get it, it uses a ||. OK from my side.
But I do think ForceDescriptor.getForceString() (l.850) must be updated to this format too, in l.863 replace the \\ with a || and in 867 the substr(1) with a substr(2) I think.

SJuliez:
ForceDescriptor is the force generator

Windchild292:
I had thought ForceDescriptor was how it worked across MegaMek, so it has already been updated.

SJuliez:
Ah yes, now that I look, it actually has :)

SJuliez (approved):

PR #3165, opened 2021-09-24
remove unnecessary logging in CommonMenuBar

NickAragua:
This method is called pretty much everywhere (MovementPhaseDisplay, FiringPhaseDisplay etc), every time the user takes an action that would change the state of one or more buttons. This tends to flood the log file with unnecessary messages, as the "CommonMenuBar" doesn't have a lot of the buttons that the phase-specific bars have.. It seems to me like we can just skip this for performance reasons as it doesn't really do anything.


SJuliez (approved):
Ah yes. I intended this as an error message but since the menu bar is no longer a somewhat-copy of the game bars, it doesn't work that way.

PR #3168, opened 2021-09-25
Array bounds protecting Entity::getCritical

Windchild292:
This prevents it from breaking when checking locations that may or may not exist, like the Center Leg of a Tripod 'Mech.

### DIFF
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -4626,8 +4626,8 @@ public abstract class Entity extends TurnOrdered implements Transporter, Targeta
     /**
      * Returns a critical hit slot
      */
-    public CriticalSlot getCritical(int loc, int slot) {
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/Entity.java
```diff
@@ -4626,8 +4626,8 @@ public int getHeatCapacity() {
     /**
      * Returns a critical hit slot
      */
-    public CriticalSlot getCritical(int loc, int slot) {
-        return crits[loc][slot];
+    public @Nullable CriticalSlot getCritical(int loc, int slot) {
+        return ((loc < crits.length) && (slot < crits[loc].length)) ? crits[loc][slot] : null;
```
NickAragua, line 4630: Not the most readable expression ever, but whatever.

NickAragua (approved):

PR #3192, opened 2021-10-09
Fixing Resource Bundles with Missing Locale

Windchild292:
Found this while working on another PR.


SJuliez (approved):
Makes it actually use MM's language pref, right?

Windchild292:
Yeah, did the code for this before learning I needed to add it

PR #3200, opened 2021-10-11
naval vessels passing over bottom mud no longer take MP penalty

### LINKED ISSUE #2217: 47.9 (and possibly earlier versions) - Mud on seabed affects Naval (Displacement) vessels sailing over it
Standalone MM 47.9
Windows 10
Java 1.8.0_261

A water (depth 2) hex with 'Mud' also applied to the hex.
A Naval (Displacement) vessel moving across it ... takes a movement penalty from the Mud, and also has to check for bogging down.

NickAragua:
fixes #2217

Took the easy way out since naval vessels never actually go over the mud. Don't give amphibious vehicles the penalty either if they're above the bottom of the lake.


NickAragua:
Had to make some additional changes to prevent bog down nagging when passing over submerged mud hexes.

PR #3222, opened 2021-10-29
New AbstractDialog children appear centered; modernized LOSDialog

SJuliez:
When AbstractDialog children appear for the very first time (no stored prefs), they appear centered on screen and - if necessary - reduced to 80% width/height of the screen - this will happen mostly for dialogs that don't pack() well such as HTML texts and the icon choosers as these would otherwise extend to full screen height which is annoying. 

To see it in use with a small dialog that isn't likely to be resized by the player I also updated the LOSDialog (Ctrl-Alt-L in game) to be an AbstractButtonDialog child. The screens show the placement and size for fresh Client Settings and LOSDialog. 

A weakness might be multiple monitors but I currently have only one monitor and have no way to even test this.

image
image

### DIFF
--- a/megamek/i18n/megamek/client/messages.properties
+++ b/megamek/i18n/megamek/client/messages.properties
@@ -1554,8 +1554,8 @@ KeyBindingsDisplay.heading=#FFFFFFKeyboard Shortcuts - press {0} to toggle this 
 
 
 #LOS / Ruler tool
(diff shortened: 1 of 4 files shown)


SJuliez:
OK all done.

Windchild292 (approved):
Code approved, just need to test locally to ensure it works with two monitors

PR #3226, opened 2021-10-31
don't prevent attempts to hack ourselves using our own code

NickAragua:
When we pass a MekHQ data structure through megamek to serialize/deserialize it, don't block the deserialization.

Fixes #2934


Windchild292 (commented):
What about MegaMekLab?

NickAragua:
We usually don't invoke megamek directly from MML.

Windchild292:
I know, but... best to just approve for this exact case (cause we don't tend to from MHQ either)

Windchild292 (approved):

PR #3229, opened 2021-11-02
AbstractIcon: Force Icon Rework

Windchild292:
This PR contains changes required for MekHQ AbstractIcon Force Icons.

I've been merging subsets over the past year (this isn't the first branch for this rework, just the last one), which is both why this isn't a huge PR and why there's some changes in files that should be untouched.


megamek/src/megamek/client/ui/dialogs/AbstractIconChooserDialog.java
```diff
@@ -105,21 +116,18 @@ protected Container createCenterPane() {
 
     @Override
     protected JPanel createButtonPanel() {
-        JPanel panel = new JPanel(new FlowLayout(FlowLayout.CENTER, 20, 2));
-        panel.setBorder(new EmptyBorder(5, 5, 5, 5));
+        final JPanel panel = new JPanel(new GridLayout(1, 3));
+        panel.setName("buttonPanel");
```
SJuliez, line 120: Gridlayout is likely to stretch the buttons. I notice you seem to like that or not care; Modern UIs including IntelliJ itself or this gituhub don't seem to do that. On small windows, the buttons float on the bottom right, on bigger ones bottom left or center.

SJuliez (approved):
All ok. Only one comment but it's "only" GUI and this would be something to decide to do one way or the other in general.

megamek/src/megamek/client/ui/dialogs/AbstractIconChooserDialog.java
Windchild292: I think it works well for our style of application, and that's why I used it in the base components and swapped it over here to follow their formatting.

PR #3230, opened 2021-11-03
Fix #2917 - missing water tile hexes on build for nightlies.

### LINKED ISSUE #2917: 0.49.2 Windows Java version 11.0.10 water tiles not loading in MM
0.49.2 Windows Java version 11.0.10

Tiles for water are not loading and giving a bland hex with red cross
image

HoneySkull:
I'm adding a possible quick fix for the nightly missing water tiles based on my understanding that source files should not be atlased and deleted as part of the build process as they reported in the Gradle createImageAtlases task.  The existing code was also attempting to ignore directories containing 'hexes/largeTexture' but is failing because the file path uses back slashes instead of forward slashes.  This change should prevent both the largeTexture and Structured_Water directories from image atlas processing.


HoneySkull:
#2917

HoneySkull:
The file path for the linux build uses forward slashes... the windows nightly build uses back slashes.
image

image

I will need to change the code to ignore both forward slash and back slash versions.

HoneySkull:
After downloading the CI_CD build for my second commit, all necessary water files appear to be present in the build.

HoneySkull:
No problem.  My thought was this would help with people testing with the nightly build for the quick fix.  I completely understand wanting to minimize change for the quick fix.  If this isn't part of the quick fix, 0.49.5 megamek-only release will have missing water tiles in the distributed zip files.  (same as 0.49.3 and 0.49.4 megamek-only release)

NickAragua:
I see. Do we know if this code is used anywhere outside of image atlas creation?

Windchild292:
It is not, but it's also a longer-term issue so fine to leave for now.

Also, fixes #3230, fixes #2917.

HoneySkull:
Okay - that will give me a little more time to look a little closer.  I have a feeling this should still be able to work with a partial atlas image.  My assumption that the *.tileinc files are specifying filenames and cannot be found because they are atlasized may not be correct.  This change corrects by avoiding the issue instead of solving it.

HoneySkull:
Fixing the case of the file paths in the "tileinc" files in the Structured_Water directory resolved this issue for the cases where the images in that directory were added to the atlas and removed from the directory.   Scanning for other non-displayable tiles, I also found a *.tileinc file in the 'Structured_Pavement' directory that also had a case mismatch and included that fix here as well.  

I downloaded the ci-cd build triggered from my commit on GitHub and installed in a clean directory and confirmed the files could be found in the atlas file and no blank tiles were found.  

I removed the 'ignore' check for "Structured_Water" as it was not necessary.  

I left my change to the "largeTextures" line so that it will also be effective on Windows builds as well as the Linux/Mac builds.

Windchild292 (approved):

Windchild292:
(Approved for 0.49.6)

Issue #3231, opened 2021-11-03
Feedback on Dockable unit display 


HammerGS:
From Lisbon
So, a couple of thoughts about the dockable unit display (these are the very first impressions after the battle started):

It's a bit big, as in, bigger than it needs to. All the info is the same, but the space is almost double on my monitor compared to the little window I usually have on the corner just outside the main battle window
When using the display as that mentioned little window, the elements in the various tabs don't update very well, specially the armor and the "extras" tab, because in the undocked version they keep the format of the original docked one, so unless we use a large height for the window, a couple of elements get hidden
Minor stuff, but I figured I would share"
and

A bit of extra feedback: the game no longer saves unit display window position, so every time you load a game, the unit display is docked. It also doesn't save size (but it does save position) within the same game session

@mjog

HoneySkull:
I noticed when the window is undocked and you re-size the unit display, especially with the Armor tab, the positioning of the content isn't relative to the window as Hammer stated above.  As a result, the window is required to be larger than in has to be in order to see the full content.

Might be kind of nice to have an over/under tab configuration so you could see two tabs at once.  (i.e. The armor tab is always on top, and then have the remaining tabs on the lower half of the docked window.  Having the armor and weapons tabs always visible would be sweet.

JonhyBurner:
The possibility to have a double display, of our unit and the target, is also interesting.

I also think that the docked display could be slimmer or the user be able to choose a width, because on a 16:9 monitor it occupies a lot of space that could (should) be otherwise occupied with more map.

mjog:
NB: fixes for this stuff going up over in #3233

mjog:
@HoneySkull 

> I noticed when the window is undocked and you re-size the unit display, especially with the Armor tab, the positioning of the content isn't relative to the window as Hammer stated above. As a result, the window is required to be larger than in has to be in order to see the full content.

Will see what I can do about that - might be a layout issue? Can you provide screenshot(s) that illustrate the issue though?

> Might be kind of nice to have an over/under tab configuration so you could see two tabs at once. (i.e. The armor tab is always on top, and then have the remaining tabs on the lower half of the docked window. Having the armor and weapons tabs always visible would be sweet

Yeah that sounds like a good approach to explore - but tbh the unit display code is all custom rendering and themed images, it's a bit of a mess, so probably outside the scope of this for the moment.

mjog:
> The possibility to have a double display, of our unit and the target, is also interesting.

[...]

PR #3233, opened 2021-11-03
Detachable unit detail tweaks

mjog:
Addressing issues reported in #3231

- [x] Save detached pane's size and location
- [x] Save detached state (attached or detached)
- [ ] ~~Content positioning in the window when detached~~
- [x] Reduce the space taken by the pane


mjog:
Nothing else to do here - per discussion in the bug, the issues with content positioning are caused by pre-existing issues with UnitDisplay, and hence should get fixed elsewhere.

Can I get a code review and/or get this merged?

Windchild292 (approved):

PR #3239, opened 2021-11-08
Moves undo back to backspace

BlueNexus:
Backspace is a much more intuitive button for it, in my opinion, and is also consistent with how it worked in 0.48.0.

This is subjective, note, so feel free to close this if you feel it's a bad idea.


SJuliez:
Ctrl-Z is widely used for Undo across applications of all kinds. It was suggested in another Issue to implement that key setting. I see that is would be consistent with older versions but sometimes things need to change.

PR #3261, opened 2021-11-18
Implement TacOps active probe rules for hidden units

NickAragua:
Previously, active probes could only find hidden units at the end of a move; now they can do so in the middle of the move if the "tacops active probes" rule is turned on. Doing so interrupts the unit's movement.

### DIFF
--- a/megamek/src/megamek/server/ServerHelper.java
+++ b/megamek/src/megamek/server/ServerHelper.java
@@ -20,7 +20,6 @@ package megamek.server;
 
 import java.util.*;
 import megamek.common.*;
-import megamek.common.enums.GamePhase;
 import megamek.common.options.OptionsConstants;
 import megamek.common.weapons.other.TSEMPWeapon;
 
@@ -543,8 +542,7 @@ public class ServerHelper {
         
         // Units without a position won't be able to detect
         // check for this before calculating BAP range, as that's expensive
-        if ((detector.getPosition() == null) ||
-                (detectorCoords == null)) {
+        if ((detector.getPosition() == null) || (detectorCoords == null)) {
             return false;
         }
        
@@ -578,23 +576,24 @@ public class ServerHelper {
         for (Entity detected : hiddenUnits) {            
             // Can only detect units within the probes range
             int dist = detector.getPosition().distance(detected.getPosition());
+            boolean beyondPointBlankRange = dist > 1;
 
             // Check for Void/Null Sig - only detected by Bloodhound probes
-            if (dist > 1 && (detected instanceof Mech)) {
-                Mech m = (Mech)detected;
+            if (beyondPointBlankRange && (detected instanceof Mech)) {
+                Mech m = (Mech) detected;
                 if ((m.isVoidSigActive() || m.isNullSigActive()) && !detectorHasBloodhound) {
                     continue;
                 }
             }
 
             // Check for Infantry stealth armor
-            if (dist > 1 && (detected instanceof BattleArmor)) {
+            if (beyondPointBlankRange && (detected instanceof BattleArmor)) {
                 BattleArmor ba = (BattleArmor) detected;
                 // Need Bloodhound to detect BA stealth armor
                 if (ba.isStealthy() && !detectorHasBloodhound) {
                     continue;
                 }
-            } else if (dist > 1 && (detected instanceof Infantry)) {
+            } else if (beyondPointBlankRange && (detected instanceof Infantry)) {
                 Infantry inf = (Infantry) detected;
                 // Can't detect sneaky infantry
                 if (inf.isStealthy()) {
@@ -607,7 +606,7 @@ public class ServerHelper {
             }
 
             LosEffects los = LosEffects.calculateLOS(game, detector, detected, detectorCoords, detected.getPosition(), false);
-            if (los.canSee() || dist <= 1) {
+            if (los.canSee() || !beyondPointBlankRange) {
                 detected.setHidden(false);
                 server.entityUpdate(detected.getId());
                 Report r = new Report(9960);
@@ -623,7 +622,7 @@ public class ServerHelper {
             }
         }
 
-        if (vPhaseReport.size() > 0 && game.getPhase() == GamePhase.MOVEMENT
+        if (!vPhaseReport.isEmpty() && game.getPhase().isMovement()
                 && (game.getTurnIndex() + 1) < game.getTurnVector().size()) {
             for (Integer playerId : reportPlayers) {
                 server.send(playerId, server.createSpecialReportPacket());
--- a/megamek/src/megamek/server/ServerHelper.java
+++ b/megamek/src/megamek/server/ServerHelper.java
@@ -550,7 +550,7 @@ public class ServerHelper {
             return false;
         }
        
-        int probeRange = 100;//detector.getBAPRange();
+        int probeRange = detector.getBAPRange();
         
         // if no probe, save ourselves a few loops
         if (probeRange <= 0) {
--- a/megamek/src/megamek/server/ServerHelper.java
+++ b/megamek/src/megamek/server/ServerHelper.java
@@ -20,6 +20,7 @@ package megamek.server;
 
 import java.util.*;
 import megamek.common.*;
+import megamek.common.enums.GamePhase;
 import megamek.common.options.OptionsConstants;
 import megamek.common.weapons.other.TSEMPWeapon;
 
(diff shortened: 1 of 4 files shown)


megamek/src/megamek/server/ServerHelper.java
```diff
@@ -530,6 +531,108 @@ public static boolean detectMinefields(Game game, Entity entity, Coords coords,
...
+                
+        // Get all hidden units in probe range
+        List<Entity> hiddenUnits = new ArrayList<>();
+        for (Coords coords : detectorCoords.allAtDistance(probeRange)) {
+            for (Entity entity : game.getEntitiesVector(coords, true)) {
+                if (entity.isHidden() && entity.isEnemyOf(detector)) {
+                    hiddenUnits.add(entity);
+                }
+            }
+        }
```
Windchild292, line 566:
```suggestion
        List<Entity> hiddenUnits = detectorCoords.allAtDistance(probeRange).stream()
                .flatMap(coords -> game.getEntitiesVector(coords, true).stream()
                        .filter(entity -> entity.isHidden() && entity.isEnemyOf(detector)))
                .collect(Collectors.toList());
```

megamek/src/megamek/server/ServerHelper.java
```diff
@@ -530,6 +531,108 @@ public static boolean detectMinefields(Game game, Entity entity, Coords coords,
...
+
+        boolean detectorHasBloodhound = detector.hasWorkingMisc(MiscType.F_BLOODHOUND);
+        boolean hiddenUnitFound = false;
+        
+        for (Entity detected : hiddenUnits) {            
+            // Can only detect units within the probes range
+            int dist = detector.getPosition().distance(detected.getPosition());
+
+            // Check for Void/Null Sig - only detected by Bloodhound probes
+            if (dist > 1 && (detected instanceof Mech)) {
```
Windchild292, line 583:
```suggestion
            if ((dist > 1) && (detected instanceof Mech)) {
```

megamek/src/megamek/server/ServerHelper.java
```diff
@@ -530,6 +531,108 @@ public static boolean detectMinefields(Game game, Entity entity, Coords coords,
...
+                Mech m = (Mech)detected;
```
Windchild292, line 584:
```suggestion
                Mech m = (Mech) detected;
```

megamek/src/megamek/server/ServerHelper.java
```diff
@@ -530,6 +531,108 @@ public static boolean detectMinefields(Game game, Entity entity, Coords coords,
...
+                if ((m.isVoidSigActive() || m.isNullSigActive()) && !detectorHasBloodhound) {
+                    continue;
+                }
+            }
+
+            // Check for Infantry stealth armor
+            if (dist > 1 && (detected instanceof BattleArmor)) {
```
Windchild292, line 591:
```suggestion
            if ((dist > 1) && (detected instanceof BattleArmor)) {
```

megamek/src/megamek/server/ServerHelper.java
```diff
@@ -530,6 +531,108 @@ public static boolean detectMinefields(Game game, Entity entity, Coords coords,
...
+                BattleArmor ba = (BattleArmor) detected;
+                // Need Bloodhound to detect BA stealth armor
+                if (ba.isStealthy() && !detectorHasBloodhound) {
+                    continue;
+                }
+            } else if (dist > 1 && (detected instanceof Infantry)) {
```
Windchild292, line 597:
```suggestion
            } else if ((dist > 1) && (detected instanceof Infantry)) {
```

megamek/src/megamek/server/ServerHelper.java
```diff
@@ -530,6 +531,108 @@ public static boolean detectMinefields(Game game, Entity entity, Coords coords,
...
+                    continue;
+                }
+                // Need bloodhound to detect non-sneaky inf
+                if (!detectorHasBloodhound) {
+                    continue;
+                }
+            }
+
+            LosEffects los = LosEffects.calculateLOS(game, detector, detected, detectorCoords, detected.getPosition(), false);
+            if (los.canSee() || dist <= 1) {
```
Windchild292, line 610:
```suggestion
            if (los.canSee() || (dist <= 1)) {
```

megamek/src/megamek/server/ServerHelper.java
```diff
@@ -530,6 +531,108 @@ public static boolean detectMinefields(Game game, Entity entity, Coords coords,
...
+                Report.addNewline(vPhaseReport);
+                reportPlayers.add(detector.getOwnerId());
+                reportPlayers.add(detected.getOwnerId());
+                
+                hiddenUnitFound = true;
+            }
+        }
+
+        if (vPhaseReport.size() > 0 && game.getPhase() == GamePhase.MOVEMENT
+                && (game.getTurnIndex() + 1) < game.getTurnVector().size()) {
```
Windchild292, line 627:
```suggestion
        if (!vPhaseReport.isEmpty() && game.getPhase().isMovement()
                && ((game.getTurnIndex() + 1) < game.getTurnVector().size())) {
```

Windchild292 (commented):
Few things first

megamek/src/megamek/server/ServerHelper.java
NickAragua: Not sure that this makes it easier to read or debug or has any performance benefits.

Windchild292: Readability is arguably better, it is far more modern, and it scales far better (given player proclivities this is key).

NickAragua: Modern != better necessarily, and I'm not really sure how it scales better. Do the "stream" functions not have to loop through each element in the collection?

megamek/src/megamek/server/ServerHelper.java
```diff
@@ -530,6 +530,108 @@ public static boolean detectMinefields(Game game, Entity entity, Coords coords,
...
+                Report.addNewline(vPhaseReport);
+                reportPlayers.add(detector.getOwnerId());
+                reportPlayers.add(detected.getOwnerId());
+                
+                hiddenUnitFound = true;
+            }
+        }
+
+        if (!vPhaseReport.isEmpty() && game.getPhase().isMovement()
+                && (game.getTurnIndex() + 1) < game.getTurnVector().size()) {
```
Windchild292, line 626:
```suggestion
                && ((game.getTurnIndex() + 1) < game.getTurnVector().size())) {
```

Windchild292 (commented):
Formatting miss

megamek/src/megamek/server/ServerHelper.java
Windchild292: This is discussed in the Stream JavaDoc, and can be better explained by either reading that or searching google for why to use streams over loops.

https://docs.oracle.com/javase/8/docs/api/java/util/stream/package-summary.html

NickAragua: I read the document, and while it promises "greater parallelism", that does not happen in reality (I figured it was the case because similar technology in the .NET environment is also slower than just basic loops). I ran some tests on a 100-something x 100-something map, with a BAP-equipped Loki on the "detector" end and ~40 hidden Lokis. I set the BAP range to 100 artificially to get meaningful numbers. The raw data indicates that the "collection" method you propose is actually slower by about 2ms per run of the code in question, and much worse if I discard the high outliers.

So, I will not be implementing your suggestion. 

Here's the raw data, including a save. The logging is "entityID: time in ms to run detect hidden unit routine".

[collectdata.txt](https://github.com/MegaMek/megamek/files/7599899/collectdata.txt)
[loopdata.txt](https://github.com/MegaMek/megamek/files/7599900/loopdata.txt)
[quicksave.sav.gz](https://github.com/MegaMek/megamek/files/7599902/quicksave.sav.gz)

The code I used for time measurement is as follows (comment/uncomment the relevant implementation you want to test).

// Get all hidden units in probe range
        long currentMillis = System.currentTimeMillis();
        
        List<Entity> hiddenUnits = new ArrayList<>();
        for (Coords coords : detectorCoords.allAtDistanceOrLess(probeRange)) {
            for (Entity entity : game.getEntitiesVector(coords, true)) {
                if (entity.isHidden() && entity.isEnemyOf(detector)) {
                    hiddenUnits.add(entity);
                }
            }
        }
        
        /*List<Entity> hiddenUnits = detectorCoords.allAtDistanceOrLess(probeRange).stream()
                .flatMap(coords -> game.getEntitiesVector(coords, true).stream()
                        .filter(entity -> entity.isHidden() && entity.isEnemyOf(detector)))
                .collect(Collectors.toList());*/
        
        long endMillis = System.currentTimeMillis();
        long result = endMillis - currentMillis;
        MegaMek.getLogger().error(String.format("%d: %d", detector.getId(), result));

NickAragua: However, this did have the side effect of uncovering a bug in my implemetation, which is that I was only checking the outer edge of the BAP radius and nothing inside it.

Windchild292: That's interesting, and... I agree on the performance bit now. I do think it is more readable, but that's (especially in this case) very much a personal viewpoint.

Sorry about the delayed response, missed the reply messages on Discord.

Windchild292 (approved):

Issue #3238, opened 2021-11-07
Water tiles disappear when zoomed out


mjog:
Water tiles disappear when the board is zoomed out to step 9.

I assume this is related to the animated water hexes? It's been a problem since at least ~0.49.3 and is still a problem under current development mainline.

To reproduce:

1. Load a map with water
2. Zoom in as far as possible (optional, but makes it easier to determine what zoom level you are on)
3. Zoom out to at least step 9
4. Notice water tiles have disappeared and been replaced by a land tile

### Environment
What version of MegaMek does your issue apply to?  >= ~0.49.3
What operating system are you using?  Windows, Linux
What java version are you using? OpenJDK 11

mjog:
Actually, only an issue under linux - maybe a file name issue then?

HammerGS:
Maybe. I thought all the images had to be 84x72 but its part of the data side I don't know well.

mjog:
Doing some more testing:

* Happens when run from source, not when run from a built zip
* Happens when run from source and `saxarba_hq` tileset is selected, doesn't happen `classic` or `saxarba`

mjog:
#3266 doesn't help, unfortunately.

mjog:
Don't think it's a linux filesystem case sensitivity issue, since all the file names in `megamek/data/images/hexes/hq_saxarba.tileset` check out.

Might actually still be a problem on Windows since when it was being tested there, that was being done via a build rather from source.

PR #3266, opened 2021-11-21
Fix incorrect saxarba structured water tile sizes

mjog:
When running `./gradlew clean build` on linux, a whole bunch of errors show up like:

```
Skipping image data/images/hexes/Structured_Water/SaxarbaWater_D0_Structure_15.png because dimensions don't match.  Image is 86 x 74
```

and

```
Skipping image data/images/hexes/Ahne/iLarge_Textures/BigGrass.jpg because dimensions don't match.  Image is 400 x 300
```

and 

```
Skipping image data/images/hexes/saxarba/SMV_Buildings/BuildingSystem-03-Advance-04-ReinforcedLow-I-099.png because dimensions don't match.  Image is 210 x 180
```

Etc.

This fixes the saxarba structured water hex errors, at least. I just ran the following to convert the problematic ones automatically, so there may be some alignment issues in there:

```
for file in megamek/data/images/hexes/Structured_Water/*.png; do echo $file; identify $file | grep -qv 84x72 && mogrify -crop 84x72+0+0 -gravity NorthWest $file ; done
```

Does not help with #3238 unfortunately.


Windchild292:
I think this would be better handled by making changes to how the image atlases work instead of changing the images and causing alignment issues. 

I've got three ideas for this, but I've never looked into how atlases work:
1) Using a mixture of the atlas and isolated images, with the latter being used for any images that aren't consistent with the baseline.
2) We get the data files redone properly to ensure length/width equals
3) Just using images instead of any uses of atlases

HammerGS:
Re #3 the Atlas got implemented to reduce releases size, and reduce the number of unzipping issue due to the sheer number of files. If the images are the wrong size that's the logical fix to me.

SJuliez (approved):
This is exactly what needed to be done. Apologies that I never saw that the image sizes were incorrect. Please merge.

PR #3270, opened 2021-11-23
Updating Flat Look And Feel from 0.26 to 1.6.4

Windchild292:


SJuliez (approved):
Checked it out, works.

PR #3276, opened 2021-11-25
ITerrainFactory and ITerrain: Removing Interfaces

Windchild292:
This removes the ITerrainFactory and ITerrain interfaces, and the completely useless Terrain Factory setup (we just use new Terrain... for all of them)


megamek/src/megamek/server/Server.java
```diff
@@ -9832,26 +9820,24 @@ public int processTeleguidedMissileCFR(int playerId, List<Integer> targetIds, Li
...
-                    if (rp.connId != playerId) {
-                        MegaMek.getLogger().error("Expected a " 
-                                + "COMMAND_CFR_TELEGUIDED_TARGET CFR packet " + "from player  " + playerId
-                                + " but instead it came from player " + rp.connId);
-                        continue;
-                    }
-                    return (int)rp.packet.getData()[1];
-                } // If no packets, wait again
+                ReceivedPacket rp = cfrPacketQueue.poll();
+                int cfrType = rp.packet.getIntValue(0);
```
SJuliez, line 9826: Won't this lead to an exception when the queue is empty in contrast with the old implementation?

megamek/src/megamek/server/Server.java
```diff
@@ -9869,24 +9855,22 @@ public int processTAGTargetCFR(int playerId, List<Integer> targetIds, List<Integ
...
-                    // Check packet came from right ID
-                    if (rp.connId != playerId) {
-                        MegaMek.getLogger().error("Expected a " + "COMMAND_CFR_TAG_TARGET CFR packet "
-                                        + "from player  " + playerId
-                                        + " but instead it came from player " + rp.connId);
-                        continue;
-                    }
-                    return (int) rp.packet.getData()[1];
+                rp = cfrPacketQueue.poll();
+                int cfrType = rp.packet.getIntValue(0);
```
SJuliez, line 9859: as above

SJuliez (commented):
I commented on one change in two places that Im doubtful about but otherwise all good.

megamek/src/megamek/server/Server.java
Windchild292: Nope... we already have an isEmpty check just a few lines before, and immediately before this assignment.
```
try {
    while (cfrPacketQueue.isEmpty()) {
        cfrPacketQueue.wait();
    }
} catch (InterruptedException e) {
    return false;
}

ReceivedPacket rp;
if (cfrPacketQueue.size() > 0) {
    rp = cfrPacketQueue.poll();
```

SJuliez: Ah ok didn't see that in the diff view.

SJuliez (approved):
All good.

PR #3278, opened 2021-11-25
revert 2786.xml to before space stations

NickAragua:
There's some kind of bug in the rat generator editor that clobbers everything when saving the rat generator state; reverted the affected file for now.


NickAragua:
This is not at all what I expected it to come out as.

PR #3282, opened 2021-11-29
CommonConstants: Interface Removal

Windchild292:
There was only a single constant, a newline character setup or \r\n. I've implemented it across the suite as \n instead, to prevent issues in Linux-origin systems while Windows handles \n fine.

I also cleaned up some useless conversions, removed some useless try-catches, swapped to StringBuilder when possible, did some chaining fixes (but not all... too many to actually handle them all), plus fixed any spacing issues I noticed.

Requires https://github.com/MegaMek/mekhq/pull/2979


Windchild292:
I went with \n instead of using System.lineSeparator is because I want it to export the same file for all systems, not one for Linux/Mac and one for Windows. It's nicer when we have data changes to keep the same format.

PR #3283, opened 2021-11-30
Logging: Reducing Throwable::printStackTrace Uses

Windchild292:
This PR reduces the number of throwable::printStackTrace uses because of how they tend to extend the debugging process in MekHQ, and swaps a few other logged calls that are often seen in the MekHQ log so they are properly in the MegaMek log.

To review:
Whitespace differences off, as I was hitting a lot of old files in need of formatting fixes.
MapMenu is the largest changed file, as it not just had the most calls of printStackTrace but I also removed the unnecessary ActionListener uses and fixed a bunch of other formatting issues in that file.

### DIFF
--- a/megamek/src/megamek/client/Client.java
+++ b/megamek/src/megamek/client/Client.java
@@ -270,11 +270,11 @@ public class Client implements IClientCommandHandler {
             try {
                 log.close();
             } catch (IOException e) {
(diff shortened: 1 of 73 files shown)


Windchild292:
The Log4j2 swapover and other changes have made this PR unusable, so I'm closing it.

PR #3287, opened 2021-12-02
IMoralUtil: Removing Interface and Fixing Method Names

Windchild292:
This removes the IMoralUtil interface and fixes the Morale typo... all names should be Morale not Moral.


SJuliez (approved):
Looks good to me.

PR #3289, opened 2021-12-04
IPlayer: Removing Interface

Windchild292:
This removes the IPlayer interface and updates Mockito to 4.1.0 (required to fix a Mockito bug that came up in testing). Requires https://github.com/MegaMek/mekhq/pull/2991


SJuliez (approved):
All good.

PR #3291, opened 2021-12-04
IOptions: Interface Removal

Windchild292:
This removes the IOptions interface, and just the interface with that name and not the system.


SJuliez (approved):
OK

PR #3292, opened 2021-12-04
Enum All The Things: AimingMode Enum Replacing IAimingModes Interface

Windchild292:
This removes the IAimingModes interface and replaces it with the AimingMode enum.
While working on this I found that quad mechs swap their cover and aiming mode values when doing a head hit edge reroll.


SJuliez (approved):
All good.

PR #3295, opened 2021-12-06
IBoard: Interface Removal

Windchild292:
This is the last of our easily collapsible interfaces, and the one with the most junk in the main file. I've removed about a half dozen completely unused methods there, and a few other commented out ones.

During personal review I found a bunch of formatting issues and IDE issues in FireProcessor, which I largely cleaned up.


SJuliez (approved):
Thumbs up.

PR #3297, opened 2021-12-08
Rewriting Telemissile AMS to handle LGTM, Performance, and Code Issues

Windchild292:
I found this while looking into LGTM issues, and... this rewrite improves the performance by removing useless checks and unused bulk processing.


megamek/src/megamek/common/Entity.java
```diff
@@ -6616,38 +6616,18 @@ public void reloadEmptyWeapons() {
...
-                }
-            }
-        }
+     * Assign AMS systems to an incoming telemissile attack. This allows AMS bays to work against
+     * these modified physical attacks.
+     */
+    public void assignTMAMS(final TeleMissileAttackAction telemissileAttack) {
+        // AMS Bays can fire at all incoming attacks each round
+        // Point defense bays are added too, provided they haven't first at something else already.
+        getActiveAMS().stream()
```
SJuliez, line 6625: // Point defense bays are added too, provided they haven't >>>first<<< at something else already.

"fired" I guess

SJuliez (approved):
A small comment mistake

megamek/src/megamek/common/Entity.java
Windchild292: Nice catch, fixed

PR #3298, opened 2021-12-08
Entity::assignAMS Performance Refactor

Windchild292:
There were a bunch of formatting oddities in the method, plus minor performance issues raised from #3297.


SJuliez (commented):
Logic is good.

I'm all for streams as long as they're readable but I'd prefer not having a longish double loop inside nested foreach statements. Also, Im not sure if the inner foreach couldnt be reformulated as a filter and collect at which point it would be much better.

Windchild292:
That's... a significant improvement on my first version. Coded the loop into a filter followed by a map, and it's a lot better than the previous.

SJuliez (approved):
OK. At least only one big foreach ;)

Issue #3273, opened 2021-11-24
0.49.5 Offboard disengage causes crash


EvilAtlas:
Fill out the fields below with as much pertinent information as possible to help the developers solve the issue.
Search for your issue inside open and closed issues first before reporting so as not to make duplicates.
(Delete the above lines)

### Environment
I'm using 0.49.5. I'm playing a MekHQ campaign, and this issue came up repeatedly during one of the battles.
I'm using Windows.
The Java version is 11.0.13.

### Description
I am in a battle where I have a single offboard field artillery platoon supporting an attack force and the AI enemy has two Thumper artillery turrets (among other things). I saw the Thumpers were doing counter-battery fire, so when my turn came during the indirect fire step, I pressed the disengage button in hopes of saving my troops. The platoon disappeared from the set of unit boxes in the upper right and all the buttons were greyed out, but the game didn't progress, still saying it was my turn for indirect fire. The unit info box also still had the artillery platoon selected.

I repeatedly tried closing the MegaMek session and loading the autosave from the beginning of the turn. The first attempt or two successfully restarted the game, but disengaging still caused the game to get stuck. Subsequent attempts had trouble loading the match at all, with the main GUI glitching out and either failing to produce the map or just giving a white screen, and then the game froze.

I used task manager to kill the Java session, waited a bit, and then reopened MekHQ. The game loaded in successfully and I tried again, to the same increasingly broken results. After rebooting MekHQ once again, I fired a round instead of disengaging, and the game progressed as normal. At the beginning of the next turn, I tried disengaging again, and it broke once more.

### Files
[disengagebugfiles.zip](https://github.com/MegaMek/megamek/files/7593061/disengagebugfiles.zip)
My player name is Iron Titans.

PR #3301, opened 2021-12-09
3273: Quirk NPE and Code Cleanup

Windchild292:
This fixes an NPE found in the logs in #3273, fixes how VTOLs aren't handled properly in the Quirk Tool, and cleans up the code in the region.

For Review: The first commit is a refactor, the second commit is the bugfix, the third commit are review fixes.


SJuliez (approved):
OK

PR #3305, opened 2021-12-10
CodeQL GitHub Action

Windchild292:
We use this one MekHQ to detect potential issues, and I thought it would be wise to add it to MegaMek too.


Windchild292:
GitHub action has been tested (finding a lot of new alerts, hence the failure there) by this PR, and I'm merging so it can be used.

PR #3307, opened 2021-12-12
Log4j 2 Swapover

Windchild292:
This removes the MMLogger setup, which uses Log4j 1, and replaces it with a full Log4j 2 setup. Log4j 1 is extremely outdated, with multiple major security vulnerabilities.

MegaMek::showInfo has been made public and standardized with MML and MHQ.

Logging Levels work different, and do not nicely allow the internal verbosity changes we use for princess... but a user should be able to customize all logging through the resources file instead.

To review: Whitespace differences off (I fixed a bunch in the copyright headers as I was opening the files anyway)

Requires https://github.com/MegaMek/megameklab/pull/977 and https://github.com/MegaMek/mekhq/pull/3002


SJuliez (approved):

SJuliez:
I thought the internal logger config (DefaultMMLogger) allowed the different loglevels. Is it necessary to remove that?

Windchild292:
I wanted to go the easily changeable and maintainable route, which does not allow for internal customization of a logger's level.

The external file does provide sufficient customization for our use though, although it's definitely not a perfect solution.

PR #3348, opened 2022-01-02
Internal Code Cleanup: Wave 3

Windchild292:
This wave fixes the following:
- [x] Lambda can be replaced with method reference
- [x] Unnecessary Semicolon Use
- [x] Missing spacing after a comma - Lowercase bar some ",f", and numbers bar ",2" are done
- [x] Unnecessary Boxing
- [x] Unnecessary to String Conversion
- [x] Unnecessary toString Calls
- [x] Unnecessary String.format Calls 
- [x] Removing Commented Out System.err and System.out Calls

Plus a lot of issues around copyright headers and the "created on" additional comments.

Next wave will finish the lowercase ",f" and ",2" missing spaces, 

For Review:
Whitespace differences off.


SJuliez (approved):
OK!
My only comments, fwiw, would be that I believe String.isBlank() is the most modern version and I'm not sure if XYZTest.java outputs must really go to the Logger but if this makes sense then Im ok with it.

Windchild292:
Fixed, and the loggers will work with utilities.

PR #3356, opened 2022-01-04
Suite Localization: Creating a temporary locale redirect in MMOptions

Windchild292:
This is the first MegaMek step towards suite standardized locales, which changes direct preference manager calls to instead be done through MMOptions, using its inheritance from SuiteOptions.


SJuliez (approved):
OK looks good to me.

PR #3373, opened 2022-01-14
Add WrapLayout for use by the whole suite

SJuliez:
MHQ and now also MML are using WrapLayout and I wrote something similar once for MM which WrapLayout could replace. So, best to have it in MM and we can subsequently remove it from MHQ and MML.

### DIFF
--- a/megamek/src/megamek/client/ui/WrapLayout.java
+++ b/megamek/src/megamek/client/ui/WrapLayout.java
@@ -1,0 +1,194 @@
+/*
+ * This file is part of MegaMek.
+ *
+ * MegaMek is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * MegaMek is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/client/ui/WrapLayout.java
```diff
@@ -0,0 +1,194 @@
...
+    private Dimension layoutSize(Container target, boolean preferred) {
+        synchronized (target.getTreeLock()) {
+            // Each row must fit with the width allocated to the container.
+            // When the container width = 0, the preferred width of the container
+            // has not yet been calculated so lets ask for the maximum.
+
+            int targetWidth;
+            Container container = target;
+
+            while ((container.getSize().width) == 0 && (container.getParent() != null)) {
```
Windchild292, line 108: Just noticed the first brackets are around the first parameter and not the comparison 😂

Windchild292 (approved):
Sounds good, just a single bracket oddity

megamek/src/megamek/client/ui/WrapLayout.java
SJuliez: Hehe I had to look at it for a while to see it :D. Corrected now. Hopefully...

Windchild292 (approved):

PR #3374, opened 2022-01-15
(DRAFT) Updating to Gradle 7.3.3

Windchild292:
This is still extremely WIP, this just provides another testing vector


Windchild292:
The update can't occur until the GR Git plugin handles Gradle 7.3, and has been partially subsumed by other changes I've done with our build scripts. Closing as I'll handle the update when it is possible.

PR #3380, opened 2022-01-17
3371: Fixing TurnOrdered NPEs for null Entity game and null Player team

### LINKED ISSUE #3371: [49.7] Isometric View Not Applying to Map and Leads to Game-ending NPE
### Environment
What version of MegaMek does your issue apply to?  MegaMek launched from MekHQ. Specifically, the latest successful Gradle run of the Company Generator branch that had an artifact. ( https://github.com/MegaMek/mekhq/actions/runs/1685479719 )
What operating system are you using?  Windows 10
What java version are you using?  Adoptium 11.0.13

### Description
Unfortunately I wasn't able to take a lot of screenshots other than the error message telling me to submit a bug report. It is highly reproducible, though. (Apologies ahead of time for the childish campaign name...it was just supposed to be a test!) Here are the steps:

1) In MekHQ, open latest non-autosave campaign save "30691110"
2) Start Battle for pending Diversion Scenario.
3) Hit Done in MegaMek.
4) Place mines wherever to get through this phase. The strangeness is better demonstrated next phase.
5) In Deployment Phase, first, hit the T key to toggle Isometric View on and off, and you'll notice no changes other than 'Mechs at elevation moving up and down. It also shaves off many of the pixels that make up the deployment zone outline when toggled.
6) I'm not sure if the NPE can be forced before Round 1, but it seems to happen when Isometric View is left newly toggled on and a whole new Round begins. Play advances normally if Isometric View is never touched, in my attempts so far. (Screenshot inside attached ZIP file.)

### Files
(NOTE: I don't know if the autosave being run through the actual MegaMek.exe will produce the same errors. When opening MegaMek from the same artifact package, the Isometric View, deployment zone appearance, and turn advancement worked without any issue all the way through the victory screen. This appears to only happen specifically when MM is launched via MekHQ from the one Pending Scenario in the campaign.)
[Huh.zip](https://github.com/MegaMek/megamek/files/7864211/Huh.zip)

Windchild292:
This fixes #3371

### DIFF
--- a/megamek/src/megamek/common/TurnOrdered.java
+++ b/megamek/src/megamek/common/TurnOrdered.java
@@ -361,9 +361,13 @@ public abstract class TurnOrdered implements ITurnOrdered {
             }
 
             if (item instanceof Entity) {
-                Entity e = (Entity) item;
-                bonus = e.getGame().getTeamForPlayer(e.getOwner()).getTotalInitBonus(false)
-                        + e.getCrew().getInitBonus();
+                final Entity entity = (Entity) item;
+                if (entity.getGame() != null) {
+                    final Team team = entity.getGame().getTeamForPlayer(entity.getOwner());
+                    if (team != null) {
+                        bonus = team.getTotalInitBonus(false) + entity.getCrew().getInitBonus();
+                    }
+                }
             }
 
             if (rerollRequests == null) { // normal init roll


NickAragua (approved):
How in the hell would an entity not have a game while a game is in progress?

Well, whatever. It's just a routine for calculating initiative bonus. Better not get the bonus than crash the game.

Windchild292:
@NickAragua
It's the exact same with the team, and I've got no clue for either... but I've seen it enough times to know there's some path by which it is possible.

PR #3388, opened 2022-01-19
Weapon list sort order

SJuliez:
Adds the method getSortingName() to EquipmentType (and therefore also WeaponType, AmmoType, MiscType) that returns a string similar to the name but optimized for sorting equipment. Using it will sort ACs in the order AC2 / 5 / 10 / 20 instead of alphabetic AC 10 / 2 / 20 / 5 and Lasers in the order Small Medium Large instead of having them all over the place by their first letter. Other equipment like the Gauss variations are grouped together somewhat. Std, OS and IOS versions of missiles are listed apart. I worked over all equipment (CL/IS up to experimental) except Infantry weapons.

Internally it adds a string class field sortingName for this. When it is left at null, getSortingName() returns the name of the equipment, otherwise that sortingName. Some intermediate classes like LRMWeapon override getSortingName() to provide the sorting name for all the multitudes of LRMs and SRMs programmatically.

For review: this is really repetitive, EquipmentType has the only interesting changes. Other than the sortingName addition or sometimes the getSortingName override, I removed many useless "this." and some useless "non-javadoc" and empty comments and sometimes moved a package or import line from over the copyright to below. If you've seen LRMWeapon and AdvancedSRMWeapon, you've seen it all.

Before: 
image

After:
image

This doesn't depend on anything but I'll have a small update for MML that requires this PR.

### DIFF
--- a/megamek/src/megamek/common/EquipmentType.java
+++ b/megamek/src/megamek/common/EquipmentType.java
@@ -1519,7 +1519,7 @@ public class EquipmentType implements ITechnology {
      * order instead of the order AC10/2/20/5 and S/M/L Lasers will be grouped together.
      * @return A String similar to getName() but modified to support a better sorting
      */
-    public String getNaturalNameSortingString() {
+    public String getSortingName() {
         return (sortingName != null) ? sortingName : name;
     }
 }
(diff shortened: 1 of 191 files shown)


Windchild292:
I've gone over this using spot checking, and it looks good to me.

Windchild292 (approved):

PR #3404, opened 2022-01-25
MHQ 3072: Preventing the RAT Generator from writing empty parent faction tags

Windchild292:
The second commit prevents https://github.com/MegaMek/mekhq/issues/3072 from occurring, while the first cleans up some of the RAT Generator file I/O code.

### DIFF
--- a/megamek/src/megamek/client/ratgenerator/RATGenerator.java
+++ b/megamek/src/megamek/client/ratgenerator/RATGenerator.java
@@ -835,7 +835,7 @@ public class RATGenerator {
             dispose = false;
         }
     }
-    
+
     /**
      * If the year is equal to one of the era marks, it loads that era. If it is between two, it
      * loads eras on both sides. Otherwise, just load the closest era.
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/client/ratgenerator/RATGenerator.java
```diff
@@ -1045,7 +1044,7 @@ private void parseChassisNode(int era, Node wn) {
     }
     
     private void parseModelNode(int era, ChassisRecord cr, Node wn) {
-        String modelKey = (cr.getChassis() + " " + wn.getAttributes().getNamedItem("name").getTextContent()).trim();
+        String modelKey = (cr.getChassis() + ' ' + wn.getAttributes().getNamedItem("name").getTextContent()).trim();
```
NickAragua, line 1047: Switching these out is very unlikely to impact performance (since we are dealing with an activity that involves reading and writing data from physical storage) and only contributes negatively to the signal:noise ratio.

NickAragua (approved):
Not really a fan of pointlessly swapping double quotes for single quotes, but since you've already done the work, there's no point to spending the additional time to roll it back.

PR #2558, opened 2021-01-08
Potential fix for this issue by allowing isEligibleFor to proceed if …

TenkawaBC:
…the advanced artillery pre-plotted option is selected and artillery is present. This allows for the artillery to be plotted even with on-board artillery.

PR #3414, opened 2022-01-28
3319: Fixing Unofficial On Map Predesignate Invalid Turn Processing NPEs

### LINKED ISSUE #3319: AI is being called to take a turn before they have any units to deploy. Kicking them allows skip but trying to skip disconnects user.
### Environment
Windows 10
49.5
jdk-11.0.11.9-hots

### Description
Pirate Aircraft Opfor in one specific map is taking its turn on the first round despite not deploying until round 6. This causes it to not do any movement and hold up the game indefinitely until kicked. Once kicked, using /skip to skip their turn disconnects the user from the session.

### Files
[Helena's Valkyries30270424.cpnx.gz](https://github.com/MegaMek/megamek/files/7723827/Helena.s.Valkyries30270424.cpnx.gz)
[megameklog.txt](https://github.com/MegaMek/megamek/files/7723829/megameklog.txt)
[Helena's Valkyries.zip](https://github.com/MegaMek/megamek/files/7723842/Helena.s.Valkyries.zip)
[BuggedScenarioAI.sav.gz](https://github.com/MegaMek/megamek/files/7723853/BuggedScenarioAI.sav.gz)

Windchild292:
This fixes #3319.

The first three commits are cleanups, the fourth is the bugfix.


megamek/src/megamek/client/bot/princess/PathRanker.java
```diff
@@ -53,28 +52,14 @@ public PathRanker(Princess princess) {
         owner = princess;
     }
 
-    /**
-     * Gives the "utility" of a path; a number representing how good it is.
-     * Rankers that extend this class should override this function
-     */
-    RankedPath rankPath(MovePath path, Game game) {
```
NickAragua, line 60: Let's not remove this, in case we want to override the PathRanker class in the future without overriding the rankPath method.

megamek/src/megamek/client/bot/princess/Princess.java
```diff
@@ -1981,8 +1981,13 @@ private void unloadTransportedInfantry(MovePath path) {
             return;
         }
         
-        Entity movingEntity = path.getEntity();
-        Coords pathEndpoint = path.getFinalCoords();
+        final Entity movingEntity = path.getEntity();
+        final Coords pathEndpoint = path.getFinalCoords();
+        if (pathEndpoint == null) {
```
NickAragua, line 1986: This approach is incorrect. A null path endpoint means the entity isn't moving, so you can simply use the entity's current position instead. You still probably want an "is this point on the board" check though; technically off-board units don't get to participate in the movement phase, but with the movement phase code the way it is, you never know.

NickAragua (changes requested):
This needs a revision to the way it's approached.

Windchild292:
I did a deeper dive, and this was introduced by the changes in #2558. I've got a proper cause fix now, which also improves the performance of said unofficial option.

megamek/src/megamek/client/bot/princess/PathRanker.java
Windchild292: It's completely unused, so there's no reason to keep it. It'll still be in our history if that ends up being needed at some point in the future.

megamek/src/megamek/client/bot/princess/Princess.java
Windchild292: It's also just a bandaid. I've reverted the null handling here and instead fixed the core issue, which would also be the cause of a bunch of other issues across the project (unofficial rule implementation issue)

NickAragua (approved):

PR #3416, opened 2022-01-29
New Util: programmatically filterable unit list

SJuliez:
Not sure if we already have something like this somewhere...

This is a small tool that writes out a filtered list of all the units. The filter is entered in line 42 as a boolean expression that can be anything with an Entity as input and all units for which it is true will be printed. This allows filters that would likely not be implemented in the advanced search even if someone upgraded it. I added a few example filters as comments.

@HammerGS Useful?

### DIFF
--- a/megamek/src/megamek/utils/FilteredUnitListTool.java
+++ b/megamek/src/megamek/utils/FilteredUnitListTool.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) 2022 - The MegaMek Team. All Rights Reserved.
+ *
+ * This file is part of MegaMek.
+ *
+ * MegaMek is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * MegaMek is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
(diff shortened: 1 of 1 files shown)


Windchild292 (commented):
The code is solid, but I'm not sure how useful a purely programmatical util is.

gsparks3:
Trying to think of what the minimum would be for something that's useful to someone who doesn't compile Megamek themselves (i.e. me). Being able to set up custom filters like this would be great in terms of searching for things that would never reasonably get added to the advanced unit search, but it needs to actually be usable to be useful. 

Working up a full GUI is best for the end-user, but that's a lot of extra dev work because GUIs are a pain. Not ideal.

Maybe something crude like a command-line executable that ingests a user-editable text file for the filter definition, then spits out another text file with the results? That might wind up being useful to a pretty narrow set of users, though, since you'd have to have enough technical knowledge to look up / learn the various possible entity checks to design your desired filter. A list of valid entity checks somewhere would make it easier. 

Any other ideas?

pakfront:
A command line utility (or an added command line option to MegaMek) might be useful, but it should be able to consume a common searching syntax  like SQL and be a able to write out a MUL as well as a simple list.

HammerGS:
We have a CSV export tool that works well for almost everything. But as I post a couple of weeks ago could use an improvement in the unit types. I think this could be useful in some situations. 

Ultimately the advanced search could be perfect with some tweaks.

SJuliez:
This is intended as a dev-tool and I had no intention of spending a lot of coding on it, therefore no command-line and no parsing. On usefulness, I used it to find SV with fusion engines of tech level C which it did.

Windchild292 (approved):

PR #3428, opened 2022-02-03
FontComboBox

Windchild292:
This adds a custom ComboBox that handles Font selection. It's an accessibility change as this makes it possible for users to customize fonts, which is crucial for Dyslexia support. I'll start implementing this by handling https://github.com/MegaMek/mekhq/issues/3097.

### DIFF
--- a/megamek/src/megamek/client/ui/comboBoxes/FontComboBox.java
+++ b/megamek/src/megamek/client/ui/comboBoxes/FontComboBox.java
@@ -37,9 +37,9 @@ public class FontComboBox extends MMComboBox<FontDisplay> {
             public Component getListCellRendererComponent(final JList<?> list, final Object value,
                                                           final int index, final boolean isSelected,
                                                           final boolean cellHasFocus) {
-                final JLabel label = (JLabel) super.getListCellRendererComponent(list, value, index, isSelected, cellHasFocus);
+                super.getListCellRendererComponent(list, value, index, isSelected, cellHasFocus);
                 if (value instanceof FontDisplay) {
-                    label.setFont(((FontDisplay) value).getFont());
+                    setFont(((FontDisplay) value).getFont());
                     list.setToolTipText(value.toString());
                 }
                 return this;
--- a/megamek/src/megamek/client/ui/comboBoxes/FontComboBox.java
+++ b/megamek/src/megamek/client/ui/comboBoxes/FontComboBox.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) 2022 - The MegaMek Team. All Rights Reserved.
+ *
+ * This file is part of MegaMek.
+ *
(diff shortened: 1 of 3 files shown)


megamek/src/megamek/client/ui/comboBoxes/FontComboBox.java
```diff
@@ -0,0 +1,56 @@
...
+            @Override
+            public Component getListCellRendererComponent(final JList<?> list, final Object value,
+                                                          final int index, final boolean isSelected,
+                                                          final boolean cellHasFocus) {
+                final JLabel label = (JLabel) super.getListCellRendererComponent(list, value, index, isSelected, cellHasFocus);
+                if (value instanceof FontDisplay) {
+                    label.setFont(((FontDisplay) value).getFont());
+                    list.setToolTipText(value.toString());
+                }
+                return this;
```
SJuliez, line 45: This looks strange because you work on "label" but return "this". I think this does work but it looks off. Maybe return label here. Although I'd even prefer leaving away the "label = " in line 40 and the "label." in l. 42 and keeping the return this.

megamek/src/megamek/client/ui/comboBoxes/FontComboBox.java
```diff
@@ -0,0 +1,56 @@
...
+            }
+        });
+        addActionListener(evt -> {
+            final FontDisplay fontDisplay = getSelectedItem();
+            if (fontDisplay != null) {
+                setFont(fontDisplay.getFont());
+            }
+        });
+    }
```
SJuliez, line 54: Puh maybe I get this wrong from looking at code but the FontComboBox looks like it should write each font name in its own font which is good so you dont accidentally choose Symbol. But Im not sure what the actionlistener is doing here cause the font writing wouldnt change... or would it?

SJuliez (approved):
Im very ok with changing fonts! One comment, one question.

megamek/src/megamek/client/ui/comboBoxes/FontComboBox.java
Windchild292: This changes the font for the actual ComboBox, while the list cell renderer handles it for selection.
image
instead of
image

megamek/src/megamek/client/ui/comboBoxes/FontComboBox.java
Windchild292: Removed the label and just used setFont, which works perfectly.

megamek/src/megamek/client/ui/comboBoxes/FontComboBox.java
SJuliez: Ahh, ok! Then all good.

Windchild292:
I'm merging this enhancement as it is completely unused at this time, and thus is fine for 0.49.7.

PR #3431, opened 2022-02-04
MUL ID

SJuliez:
Adds the MUL ID to Entity and BLK/MTF files.
Also adds a hyperlink to our MechViews leading to the MUL page of that unit, when a MUL ID is available.
Best look at the commits separately as some contain several thousand unit files.


megamek/src/megamek/common/MechView.java
```diff
@@ -1336,6 +1345,30 @@ public String toHTML() {
             return value + "<br/>\n";
         }
     }
+
+    /**
+     * Displays a hyperlink. Does not add a line break after itself.
+     */
+    private static class HyperLinkElement implements ViewElement {
+
+        private final String adress;
```
Windchild292, line 1356: I think you mean address 😉

megamek/src/megamek/common/MechView.java
```diff
@@ -252,6 +252,8 @@ public MechView(final Entity entity, final boolean showDetail, final boolean use
                     Messages.getString("MechView.Source")));
             if (!entity.getSource().isEmpty()) {
                 sHead.add(new LabeledElement("", entity.getSource()));
+            } else {
+                sHead.add(new LabeledElement("", "(unkown)"));
```
Windchild292, line 256: It should be "unknown", not "unkown"

megamek/src/megamek/common/MechView.java
```diff
@@ -245,11 +245,20 @@ public MechView(final Entity entity, final boolean showDetail, final boolean use
         if (useAlternateCost && entity.getAlternateCost() > 0) {
             cost = entity.getAlternateCost();
         }
-        sHead.add(new LabeledElement(Messages.getString("MechView.Cost"),//
+        sHead.add(new LabeledElement(Messages.getString("MechView.Cost"),
                 dFormatter.format(cost) + " C-bills"));
-        if (!entity.getSource().isEmpty()) {
-            sHead.add(new LabeledElement(Messages.getString("MechView.Source"), entity.getSource()));//
+        if (entity.hasMulId()) {
+            sHead.add(new HyperLinkElement("http://www.masterunitlist.info/Unit/Details/" + entity.getMulId(),
```
Windchild292, line 251: The actual link should probably be a MMConstant.

Windchild292 (commented):
A pair of spelling mistakes and a suggestion for the location of a constant. Other than that, I'm good with the code for this PR.

megamek/src/megamek/common/MechView.java
SJuliez: indeed :)

megamek/src/megamek/common/MechView.java
SJuliez: I must have been really tired and impatient...

megamek/src/megamek/common/MechSummary.java
```diff
@@ -118,6 +119,8 @@ public String getModel() {
         return (m_sModel);
     }
 
+    public int getMulId() { return mulId; }
```
Windchild292, line 122: We don't do single line getters...

Windchild292 (commented):
Just one style issue to fix

HammerGS:
I've sent Juliez an excel sheet with another 450 odd matches to the MUL. Testing the branch I haven't seen any issues and think this should be considered for 49.7 depending on those new matches.

megamek/src/megamek/common/MechSummary.java
SJuliez: lol this was an IntelliJ auto-add

SJuliez:
Very nice! Of the 7016 units that can possibly have a MUL ID 6626 now have one, leaving 390 units without. 245 of those are Infantry that may not have a corresponding MUL entry. Of the remaining 145, 62 are Aero, 40 are Meks. So there's really only a handful of entries missing.

Concerning merging this, I'd prefer making informed decisions rather than following some rule. This PR changes thousands of unit files of which I've only really tested or looked at a few. It also changes the unit loading and saving code and makes the units not-backward-compatible (this is something that I could actually reverse, thinking about it now. I was doing what had been done in the past with other changes to mtf but it's not actually necessary). I'd consider this a very dangerous PR to merge because if it breaks something it's going to break it real hard.

SJuliez:
Next round! Down to 226 of 7016 units (not unofficial, not Gun Emplacement) without MUL ID (27 Meks, 13 Aero, 167 Infantry, no PM, and 19 Tanks)

HammerGS:
We've got unit files conflicting as I fixed them in master.

PR #3446, opened 2022-02-13
Adding the Legendary SkillLevel

Windchild292:
This is part 1 of handling https://github.com/MegaMek/mekhq/issues/3111, and was a tier I missed when adding the enum.

This requires and is required by https://github.com/MegaMek/mekhq/pull/3116


SJuliez (approved):
Legen... wait for it

PR #3451, opened 2022-02-17
Properly Treating Blank Passwords as not being Passwords

Windchild292:
This fixes a bug I found while working on #3450


SJuliez (approved):
This kinda precludes using three blanks as a password but I guess that's not a bad idea.

PR #3463, opened 2022-02-23
Standardized Code Utilities

Windchild292:
This creates a suite setup for Java code simplification utilities.

It requires https://github.com/MegaMek/mekhq/pull/3134 because it moves some current functionality.


megamek/src/megamek/common/util/StringUtil.java
```diff
@@ -1,33 +1,26 @@
...
+import megamek.common.preference.PreferenceManager;
+
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
-import java.util.Comparator;
 import java.util.Vector;
 
-import megamek.common.preference.PreferenceManager;
-
 public class StringUtil {
```
SJuliez, line 23: Surprise, I had expected for this to be deleted and integrated into StringUtility. Not good?

SJuliez (approved):
Quick question, but otherwise all good.

megamek/src/megamek/common/util/StringUtil.java
Windchild292: I took the most useful suite-wide methods from the two utilities and moved them into the new utilities setup. I'm going to investigate what's left here and MekHQ's Utilities to see if their functionality can be moved or removed, but felt that would be best left for its own PR.

PR #3466, opened 2022-02-24
FontDisplay: Fixing missed equals and hashcode overrides

Windchild292:
Missed these in the initial commit. I also simplified the constructor, which helps with setting the value nicely.


megamek/src/megamek/client/ui/displayWrappers/FontDisplay.java
```diff
@@ -35,6 +37,10 @@
     //endregion Variable Declarations
 
     //region Constructors
+    public FontDisplay(final String familyName) {
+        this(Font.decode(familyName).deriveFont(Font.PLAIN, 16f), familyName);
```
SJuliez, line 41: If you're interested, you could try getting the UIManager default font for comboboxes and derive from it. The size would probably be 12 then.
Like:
Font font = UIManager.getFont("ComboBox.font");
if (font != null) etc.

image

megamek/src/megamek/client/ui/displayWrappers/FontDisplay.java
```diff
@@ -49,12 +55,34 @@ public Font getFont() {
...
+    @Override
+    public boolean equals(final @Nullable Object other) {
+        if (other == null) {
+            return false;
+        } else if (this == other) {
+            return true;
+        } else if (other instanceof FontDisplay) {
+            return getFont().equals(((FontDisplay) other).getFont());
+        } else if (other instanceof Font) {
+            return getFont().equals(other);
```
SJuliez, line 76: As someone who did not write this class I would never write fontDisplay.equals("Verdana") or fontDisplay.equals(someFont) and expect it to return true without taking a deep look at it first. Far be it from me to create unnecessary work but in this case I'd say isFont(String fontName) and isFont(Font font) would be easier on other users of this class such as me.

SJuliez (changes requested):
Sorry, small PR only and still I'd like to request a change to the equals...

megamek/src/megamek/client/ui/displayWrappers/FontDisplay.java
Windchild292: Both were testing setups that don't compare properly (which lead to the constructor change), so I've removed them. I was trying to simplify selected item assignment by making them possible.

megamek/src/megamek/client/ui/displayWrappers/FontDisplay.java
Windchild292: It's in a constructor, so it needs to be inline... which means using the UIManager isn't the most stable way to code. However, I've swapped to 12 as that's the default.

Windchild292 (commented):
I concur on both, review changes applied

SJuliez (approved):

PR #3469, opened 2022-02-25
Add CalculationReport tools

SJuliez:
I noticed that I need to write a conversion report for AlphaStrike conversion. Looking at the BV or cost reports, these are really done per pedes with mountains of stringbuilder.append(ROW_START). I wanted a tool for this, so this is the result and I'd rather add it now than keep it buried in the AS conversion branch.

This consists of an interface "CalculationReport" that writes up ths standard interaction with any of the report classes, such as addLine(String, String, String), addHeader() and so on.
The interface is implemented by four classes. These are:
- HTMLCalculationReport: This creates a report much like it's done now, only without having to use any HTML manually
- SwingCalculationReport: This creates a report using Swing Labels, ready to deploy in a panel or wherever; this is a bit superior to HTML as it knows how wide it wants to be and it is easier to format if any formatting is required. (see screen below)
- TextCalculationReport: This is a pure text report, formatted for display in a monospace font. Good for export (see screen below)
- In addition, there is FlexibleCalculationReport, which internally builds all three of the above and has any of them ready, for example for export to clipboard.

Adding a line to any of the four is the same regardless of which Report type is used (and they can be typed as CalculationReport instead of their class): addLine("Damage:", "2+4+6", " = 12"); Adding any of them to a dialog is simple as they all have a toJComponent() method.

For the examples below, I changed the bv calculation of Mech to the new way (which cuts around 700 lines of code and, I believe, a few bugs along the way). This is not part of the PR as I would need to convert all unit types and thats a lot. Something for later.
The code using it starts with:
bvReport.addHeader("Battle Value Calculations For " + getChassis() + " " + getModel());
bvReport.addSubHeader("Defensive Battle Rating Calculation:");

To switch between these two screenshots I changed only one line:
CalculationReport bvReport = new TextCalculationReport(); <-> CalculationReport bvReport = new SwingCalculationReport();
(and set the font to monospaced for the text)
image
image

(There is a bug in the screenshots that has been corrected in the code)

### DIFF
--- a/megamek/src/megamek/client/ui/swing/calculationReport/CalculationReport.java
+++ b/megamek/src/megamek/client/ui/swing/calculationReport/CalculationReport.java
@@ -188,7 +188,7 @@ public interface CalculationReport {
     CalculationReport addResultLine(@Nullable String type, @Nullable String calculation, @Nullable String result);
 
     /**
-     * Returns the CalucaltionReport as a JComponent that can be added to a dialog or other Swing component.
+     * Returns the CalculationReport as a JComponent that can be added to a dialog or other Swing component.
      *
      * @return The CalculationReport wrapped in a JComponent form.
      */
--- a/megamek/src/megamek/client/ui/swing/calculationReport/CalculationReport.java
+++ b/megamek/src/megamek/client/ui/swing/calculationReport/CalculationReport.java
@@ -1,0 +1,205 @@
+/*
+ * Copyright (c) 2022 - The MegaMek Team. All Rights Reserved.
+ *
+ * This file is part of MegaMek.
+ *
+ * MegaMek is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * MegaMek is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with MegaMek. If not, see <http://www.gnu.org/licenses/>.
+ */
+package megamek.client.ui.swing.calculationReport;
+
+import megamek.common.annotations.Nullable;
+
+import javax.swing.*;
+import java.util.Locale;
+/**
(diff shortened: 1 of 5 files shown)


megamek/src/megamek/client/ui/swing/calculationReport/CalculationReport.java
```diff
@@ -0,0 +1,205 @@
...
+     *
+     * @param type The first element of this line, such as "Damage: "
+     * @param calculation A calculation or other info, displayed after the type
+     * @param result A result or other info, displayed on the right side
+     * @return The CalculationReport itself. Enables multiple stringed addLine calls
+     */
+    CalculationReport addResultLine(@Nullable String type, @Nullable String calculation, @Nullable String result);
+
+    /**
+     * Returns the CalucaltionReport as a JComponent that can be added to a dialog or other Swing component.
```
Windchild292, line 191:
```suggestion
     * Returns the CalculationReport as a JComponent that can be added to a dialog or other Swing component.
```

megamek/src/megamek/client/ui/swing/calculationReport/HTMLCalculationReport.java
```diff
@@ -0,0 +1,104 @@
...
+
+    @Override
+    public JComponent toJComponent() {
+        final JEditorPane editorPane = new JEditorPane("text/html", this.toString());
+        editorPane.setEditable(false);
+        editorPane.setCaretPosition(0);
+        return editorPane;
+    }
+
+    public CalculationReport addResultLine(String type, String calculation, String result) {
```
Windchild292, line 58: Missing override annotation

megamek/src/megamek/client/ui/swing/calculationReport/SwingCalculationReport.java
```diff
@@ -0,0 +1,108 @@
...
+    private final GridBagConstraints gbc = new GridBagConstraints();
+    private final String COL_SPACER = "   ";
+    private final static int PADDING = 15;
+
+    @Override
+    public JComponent toJComponent() {
+        return report;
+    }
+
+    public CalculationReport addResultLine(String type, String calculation, String result) {
```
Windchild292, line 39: Missing override annotation

Windchild292 (commented):
Looks good, just a typo and a few missing overrides

SJuliez:
OK corrected.

Windchild292 (approved):

PR #3474, opened 2022-03-02
MML Issue 1045: EquipmentType/MiscType updates, constants, data

SJuliez:
Provides the basis for a fixing a few equipment items not showing up in MML.
- Adds internalName constants in EquipmentTypeLookup for industrial equipment
- By way of exception to the rule, I moved the annotations in EquipmentTypeLookup to the same line as the constant to make this less of a waste of space
- Adds the constants to the MiscType definitions of industrial eqp (the internalNames arent changed, just made a constant) as well as the Nail/Rivet Gun which is also an industrial eqp.
- In MiscType, corrects the ISCVDuneBuggy chassis mod to be TANK_EQUIPMENT instead of SUPPORT_TANK and removes the comment on both the CV Dune buggy and SV dune buggy that there is only one eqp for SV and CV and there should be two, because there are, in fact, two.
- and... I accidentally left in the change to the FilteredUnitListTool but I did use it to find out if there is an SV Dune Buggy and in fact there is, the Dromedary Water Transport.

@HammerGS I'd like to note that the rules say that Dune Buggy is only available to CV. Even though there is an official SV... Has this been errata'd? And, I could not find any mention in the rules of a Buzzsaw, yet we have one that points to TM.

### DIFF
--- a/megamek/src/megamek/common/EquipmentType.java
+++ b/megamek/src/megamek/common/EquipmentType.java
@@ -1502,4 +1502,16 @@ public class EquipmentType implements ITechnology {
     public String getSortingName() {
         return (sortingName != null) ? sortingName : name;
     }
+
+    /**
+     * Returns true if this equipment is any of those identified by the given type Strings.
+     * Best use the constants defined in EquipmentTypeLookup.
+     *
+     * @param eType An equipment type to check
+     * @param eTypes More equipment types to check
+     * @return true if the internalName of this equipment matches any of the given types
+     */
+    public boolean isAnyOf(String eType, String... eTypes) {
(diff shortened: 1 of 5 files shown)


HammerGS:
Hi, Regarding the Dune Buggy - SV's get them on the table pg 122 of TM as well as a combat chassis. But looks like different rules for each one. 

Buzzsaw is a legacy equipment was originally in Unbound Module, I thought it was in a very early Tech Manual or TacOps but can't find it. It's been replaced with the Dual Saw or Chainsaw officially.

Windchild292 (approved):

Issue #3400, opened 2022-01-24
Hidden units in movement phase may not have the opportunity to unhide


pakfront:
### Environment
MegaMek Version 0.49.6
Timestamp Sat Jan 08 11:45:41 PST 2022
Java Vendor AdoptOpenJDK
Java Version 11.0.10
Windows 10

### Description
**Summary**
It is quite possible for a player with hidden units to miss their opportunity to unhide them in the move phase.  Hidden units are not in the initiative list and therefore do not block progression to the next phase. So you have to make sure you get all your unhiding done before the last move in the initiative order is finished. Otherwise MM will move to the next phase without you changing all your upcoming hide states.

**Details**
MegaMek does not do hidden units exactly as described in Total War, but it is a reasonable approximation.
IN TW, hidden units are to be unhidden at the beginning of the Movement or Firing phase. They are immediately visible on the map, and can fire and be targeted. They cannot move that turn, at least according to my rulebook. 
"Unless the player plans to move a hidden unit during the Movement Phase, hidden units are not counted for the purposes of determining movement order during that phase. If the player plans to activate a hidden unit during a turn, he must reveal that unit and place it on the map at the start of then turn's Movement Phase. If the player plans to attack using a hidden unit, he must reveal  that unit and place it on the map at the beginning of of the Weapons Attack phase. Hidden units revealed during the Movement Phase cannot move during that phase" (Total Warfare "Hidden Units" p. 259)

In MM units are not unhidden at the beginning of the phase, rather they are marked to be unhidden in a previous phase. Units must be marked to unhide in a previous phase and will reveal themselves in an upcoming phase. This means a hidden unit may not be revealed during the current phase. So, if you want to have a hidden unit fire, you must mark it to reveal during the preceding movement phase, rather than at the beginning of the fire phase.
You do so be right-clicking on the unit and "Viewing" it, then going to the Extras tab of the unit display, changing the activation dropdown to "firing" and pressing  the Activate Hidden Unit button. Note: You cannot select it in the Unit Overview as it is not in the initiative list. 
This is clunky, but would work out fine except for one issue:

[...]

PR #3483, opened 2022-03-04
Hiddenunits

pakfront:
**DRAFT** - but comments welcome
This addresses https://github.com/MegaMek/megamek/issues/3400
It uses the pattern established by UnloadStrandedGameTurn to insert an UnhideHiddenGameTurn turn at the beginning of the turn order vector for TargetingPhase (for reveal during for Movement) and OffboardPhase (for reveal during FiringPhase.) This pops up a dialog where the user can unhide any units before regular play continues.

**Minor Issues:**
Needs more testing overall.
Need to make sure does not conflict with the old way of revealing by using the unit inspector, including resetting the reveal phase property. 
Client displays do not show unhidden units until one more turn has happened (need to figure out how to push the updated info to clients.)
Need to test with bots to make sure it doesn't hang them up. Not expecting them to use it correctly. ;)
Popup dialog is ugly.

### DIFF
--- a/megamek/i18n/megamek/client/messages.properties
+++ b/megamek/i18n/megamek/client/messages.properties
@@ -3883,6 +3883,11 @@ ReplacePlayersDialog.config=Bot Settings...
 #Point Blank Shot Dialog
 StatusBarPhaseDisplay.pointblankShot=Pointblank shot from a hidden unit!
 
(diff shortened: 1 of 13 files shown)


HammerGS:
Is there anything needed to move this forward?

PR #3486, opened 2022-03-05
Shadow Hawk LAM SHD-X2: Bomb Bays instead of Cargo

SJuliez:
Was looking for additional fuel on any LAM and found that this one had Cargo instead of Bomb Bays. No idea if there is some necessity behind that. The bomb bays let it actually load bombs in MM.


HammerGS (approved):

HammerGS:
Legacy from when the units where made I think.

PR #3490, opened 2022-03-06
RATGeneratorEditor 

SJuliez:
Some more changes to the RatGeneratorEditor requested by Hammer.
- sort factions alphabetically
- add weight column to unit table
- color model factions missing in chassis red
- also made the faction table headers multiline and the columns stretchable which makes the whole table fit in much less space
image


SJuliez:
@HammerGS Can now copy between model and faction. Please give it a test if this really does what it looks it does because with these tables there's always a chance they're showing something that isnt really present in the data.

HammerGS:
Gonna test this and confirm no issues.

SJuliez:
OK got a new version. This writes back the copied rows to the actual data. Please test.

SJuliez:
According to Hammer, this works. Therefore I'd like to merge it.

PR #3506, opened 2022-03-09
3502: Disable Jump Button when prone

### LINKED ISSUE #3502: Disable Jump Button when Entity is Prone
I found this when playing an exhibition match for the Solaris tournament.
image

Windchild292:
This fixes #3502 while cleaning up some related code and fixing some related NPEs. These are separated nicely into individual commits.


Windchild292:
This is now ready for review.

SJuliez (approved):

PR #3508, opened 2022-03-10
Fix landing/takeoff length for small craft and dropships

pavelbraginskiy:
@NickAragua  Additional fix to the aerodyne hovering fix from earlier. Dropships already override their landing length and will inherit the takeoff length from the small craft class.

### DIFF
--- a/megamek/src/megamek/common/SmallCraft.java
+++ b/megamek/src/megamek/common/SmallCraft.java
@@ -1012,4 +1012,14 @@ public class SmallCraft extends Aero {
     protected int calculateWalk() {
         return walkMP;
     }
+
+    @Override
+    public int getTakeOffLength() {
+        return 20;
+    }
+
+    @Override
+    public int getLandingLength() {
+        return 8;
+    }
 }


pavelbraginskiy:
I'm still a little unsure on how to interpret the rules on landing/takeoff distance. I've talked to some people and depending on how you interpret it, the short landing might just be limited to fighters but the short takeoff might be usable by small craft/dropships since those count as VSTOL. But it wasn't that way in megamek before, so I'm unsure.

pavelbraginskiy:
image
image
Some of the ongoing discussion on the subject

NickAragua:
@HammerGS may be able to weigh in on the rules aspect of this.

SJuliez:
From the rules, I think I agree with the landing length of 8 for SC but not with the takeoff length of 20. According to TM p188 aerodyne SC have VSTOL. According to TW p88 VSTOL-equipped units lift off in 10.

SJuliez (changes requested):

SJuliez:
Replaced by #4077

PR #3514, opened 2022-03-19
Updating to Launch4j 2.5.2

Windchild292:


Windchild292:
This update can't be done yet, will handle it later

PR #3540, opened 2022-03-29
DockingCollar Refactor

SJuliez:
This is a refactoring centered on DockingCollar with a bit of fallout around it. This also gives Docking Collars a hard restriction to a maximum of one dropship per collar which used to be ensured by always passing 1 as the size of transport.
This PR requires an adjustment in MML.

### DIFF
--- a/megamek/src/megamek/common/DockingCollar.java
+++ b/megamek/src/megamek/common/DockingCollar.java
@@ -1,295 +1,145 @@
 /*
-* MegaMek -
-* Copyright (C) 2003, 2004 Ben Mazur (bmazur@sev.org)
-* Copyright (C) 2018 The MegaMek Team
-*
-* This program is free software; you can redistribute it and/or modify it under
-* the terms of the GNU General Public License as published by the Free Software
-* Foundation; either version 2 of the License, or (at your option) any later
-* version.
-*
-* This program is distributed in the hope that it will be useful, but WITHOUT
-* ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
-* FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
-* details.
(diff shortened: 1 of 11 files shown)


megamek/src/megamek/common/DockingCollar.java
```diff
@@ -1,295 +1,145 @@
...
  */
-
 public class DockingCollar implements Transporter {
 
-    // Private attributes and helper functions.
-
-    /**
-     *
-     */
-    private static final long serialVersionUID = -4699786673513410716L;
```
Windchild292, line 35: We can't remove these from anything that isn't graphical yet in MegaMek, or it will break saving.

Windchild292 (changes requested):
One breaking bug, otherwise this looks good.

megamek/src/megamek/common/DockingCollar.java
SJuliez: How so? I can save & load with it.

SJuliez: or "without it" one might say

Windchild292: Interesting... I've been having constant save failures whenever I try to remove them in MM, despite that the XStream docs say there's no relationship. For safety, can we please keep it in for now and I'll remove it instead with the XStream removal?

SJuliez: Sure, can do that. I read a bit more about it and while it's not necessary to add one, it will be auto-calculated from the object signature (methods, fields etc.) when missing, that calculation is seemingly very sensitive and so save games could break even when there is no reason to. So I'll keep them in savegame-relevant classes from now on.

Windchild292 (approved):

PR #3320, opened 2021-12-16
Internal Code Cleanup: Wave 2

Windchild292:
This follows up from #3296 to further cleanup our internals.

The focus of this cleanup is as followed:
1) Complete $NON-NLS-1$ / $NON-NLS-2$ Cleanup
2) Adding all missing Overrides
3) Fixing All C-style Arrays
4) Removing all unnecessary Collection<> declarations
5) StandardCharsets.UTF_8 Swapover
6) String.equals("") Swapover

Plus a bunch of other fixes, mainly formatting, logging, and boxing. These changes handle 2 LGTM issues and a difference of 951 Code QL issues (982 fixes, 31 new alerts... a 50% reduction).

To review:
Whitespace differences off

PR #3541, opened 2022-03-29
Suite Cleanup 1: MegaMek Changes

Windchild292:
This is the first suite cleanup, which is primarily targeted at handling accessibility whitespace issues, some Java 2-8 migrations (the List::isEmpty change namely being an issue since Java 2... which was released on December 8th, 1998), and at removing some useless comments.
Do note that MegaMek has already seen a lot of the changes fixed or heavily reduced from the four previous cleanups (#3296, #3320, #3348, #3440), and thus is significantly smaller than it would otherwise be... while still being more changes than the other two combined.

Specifically, this does:
1. Fixes most accessibility spacing problems, especially uses of `if(`, `for(`, `while(`, `try{`, `catch(`, `}catch`, and `){`.
2. Fixes all `if`, `for`, `while`, and `try` statements without braces
3. Removes remaining useless debugTime Code
4. Modernize String::trim::length and String::trim::isEmpty uses to String::isBlank
5. Fix a ton of useless empty Javadoc comment setups and related spacing issues
6. Removing useless `/* package */` comments
7. Removing old comments that were probably meant to region the code
8. Modernizing Anonymous  Types to Lambdas
9. Renaming `arg0` to `evt` for events
10. Modernizing Anonymous Types to Method References
11. Fixing missing `@Override` annotations
12. Removing useless `GEN-FIRST` comments
13. Modernizing Explicit Type declarations to <>
14. Modernizing Lambdas to Method References
15. Modernizing List::size to List::isEmpty
16. Fixing Unnecessary Semicolon Use
17. Removing Unnecessary Boxing
18. Fixing case spacing issues
19. Fixing some  `=  ` spacing issues
20. Removing any unused imports

There's also some resource leak fixes, a bunch of spacing and whilespace fixes, and some other simple IDE-recommended fixes I found while working on the above.

To review: Whitespace differences off, and I'd recommend doing a final compare instead of using the individual commits (although they are separated) given the number of fixes included.

This was coded alongside with https://github.com/MegaMek/megameklab/pull/1080 and https://github.com/MegaMek/mekhq/pull/3197, but doesn't require nor is required by them. Once the trio are merged I'll start working on the next wave, which will be fixing the hundreds of typos I've found while working on this PR.


SJuliez (approved):

SJuliez:
This is almost too much even for a PC browser. One thing I noticed is that it removes something around 3000 lines of code. I couldn't directly see where so I'll assume that this is as it should be (?)

PR #3560, opened 2022-04-09
Correct Screen Launchers to be Capital weapons

SJuliez:
Screen Launchers weren't appearing as Large Craft equipment in MML (taken from a comment on discord). According to SO, p.103 and TO AUE p. 228 Screen launchers are Capital Weapons. According to TM p.237 they are only available to DS which agrees with the former. TM p.342 lists them as available to SC which contradicts TM p.237. Overall I assume they are indeed Capital and not available to SC and TM 342 is wrong.

With this PR they are Capital and appear to all Large Craft as Capital Weapons.

@HammerGS: I couldnt find this in any errata but I'm not sure what the best way is to search for forum entries (the forum search engine feels like no big help)


HammerGS (approved):

HammerGS:
They are available to Capital Ships.

PR #3584, opened 2022-04-20
Cost calculations

SJuliez:
This PR 
- moves the bulk of cost calculations out of Entity and the subclasses into their own package and classes and makes full use of the CalculationReport classes. 
- adds a separate cost calculation dialog with the ability to switch between dry (unloaded) cost and full cost and export of the report. 
- a lobby context menu item to show the cost calculation; I used this for testing but I guess there's no harm in keeping it
- displays weapons & equipment in "stacked" form, i.e. "5 Medium Laser     200000" rather than listing each Medium Laser separately
- makes the cost calculation available for all units except Gun Emplacements that have no cost

For review: The first commit moves out all the stuff and is big. The other commits are all small and self-contained and do mostly what the title says.

image


Windchild292 (approved):
I'm approving this, but I'm not sure about the design choice for using statics instead of creating objects, as it will cause issues when we start adding unit testing.

PR #3596, opened 2022-04-25
Connection Rework Phase One

Windchild292:
This is the first phase of the connection rework, split off of #3450.  This PR splits off the ConnectionEventType and PacketCommand Enums, migrates the packet setup to varargs, and creates the events setup.

Finally, this likely has bugs that I've missed in testing. I've tried to test it, but there's a huge chance I've missed something... but we need these changes to fix bugs in the core connection setup without making it too hard to work with.


megamek/src/megamek/common/net/enums/PacketCommand.java
```diff
@@ -0,0 +1,162 @@
...
+        return this == CFR_DOMINO_EFFECT;
+    }
+
+    public boolean isCFRTagTarget() {
+        return this == CFR_TAG_TARGET;
+    }
+
+    public boolean isCFR() {
+        return isSendingBoard() || isCFRHiddenPBS() || isCFRTeleguidedTarget() || isCFRAMSAssign()
+                || isCFRAPDSAssign() || isCFRDominoEffect() || isCFRTagTarget();
```
SJuliez, line 159: Sure that sendingBoard is a CFR?

megamek/src/megamek/server/Server.java
```diff
@@ -30428,9 +30433,13 @@ private void receiveGameOptionsAux(Packet packet, int connId) {
...
-        for (AbstractConnection conn : connections) {
-            send(conn.getId(), new Packet(Packet.COMMAND_GAME_VICTORY_EVENT));
+        if (connections == null) {
+            return;
         }
+
+        connections.stream()
+                .filter(Objects::nonNull)
+                .forEach(connection -> connection.send(new Packet(PacketCommand.GAME_VICTORY_EVENT)));
     }
```
SJuliez, line 30443: Just to make sure: It is intentional that this is no longer Server::send(...), instead it is now Connection::send? I looked at it quickly and it looked to me like it ends up being the same. But just making sure.

SJuliez (approved):
OK for what it's worth, this looks good to me. Two places where I left a question.

megamek/src/megamek/common/net/enums/PacketCommand.java
Windchild292: Nice catch, fixed.

megamek/src/megamek/server/Server.java
Windchild292: It's identical, and easier to read this way. I've further changed it to use Server::send(Packet), which handles null checks and send to all available connections.

PR #3606, opened 2022-05-01
Logging Modernization: Updating a Bunch More System.out, System.err, and printStackTraces

Windchild292:
To Review: Whitespace differences off

The LGTM report is pre-existing, and will thus be handled in a LGTM cleanup instead.

### DIFF
--- a/megamek/src/megamek/client/bot/princess/BotGeometry.java
+++ b/megamek/src/megamek/client/bot/princess/BotGeometry.java
@@ -158,7 +158,8 @@ public class BotGeometry {
             boolean flip = getDirection() > 2;
             HexLine[] edges = a.getEdges();
             if ((edges[getDirection()] == null) || (edges[(getDirection() + 3) % 6] == null)) {
-                LogManager.getLogger().error(new IllegalStateException("Detection of NULL edges in ConvexBoardArea: " + a));
+                LogManager.getLogger().error("Detection of NULL edges in ConvexBoardArea: " + a,
+                        new IllegalStateException());
                 return 0;
             }
             if (edges[getDirection()].getIntercept() == getIntercept()) {
--- a/megamek/src/megamek/client/bot/princess/BotGeometry.java
+++ b/megamek/src/megamek/client/bot/princess/BotGeometry.java
@@ -158,8 +158,7 @@ public class BotGeometry {
             boolean flip = getDirection() > 2;
             HexLine[] edges = a.getEdges();
             if ((edges[getDirection()] == null) || (edges[(getDirection() + 3) % 6] == null)) {
-                LogManager.getLogger().error("Detection of NULL edges in ConvexBoardArea: " + a,
-                        new IllegalStateException());
+                LogManager.getLogger().error("Detection of NULL edges in ConvexBoardArea: " + a);
                 return 0;
             }
             if (edges[getDirection()].getIntercept() == getIntercept()) {
--- a/megamek/src/megamek/client/ui/swing/RandomMapPanelAdvanced.java
+++ b/megamek/src/megamek/client/ui/swing/RandomMapPanelAdvanced.java
@@ -20,6 +20,7 @@ import megamek.client.ui.swing.util.VerifyIsPositiveInteger;
 import megamek.client.ui.swing.widget.VerifiableTextField;
 import megamek.common.MapSettings;
 import megamek.common.util.BoardUtilities;
+import org.apache.logging.log4j.LogManager;
 
 import javax.swing.*;
 import javax.swing.border.LineBorder;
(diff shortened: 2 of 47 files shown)


Windchild292:
The LGTM report is pre-existing, and will thus be handled in a LGTM cleanup instead.

megamek/src/megamek/common/pathfinder/LongestPathFinder.java
```diff
@@ -166,16 +167,16 @@ public Deque<MovePath> doRelax(Deque<MovePath> v, MovePath mpCandidate, Comparat
                      * Current implementation of doRelax() assumes that v is
                      * sorted in such way that this situation is impossible.
                      */
-                    System.err.println(new IllegalStateException(
-                            "Top Move Path uses more MPs than Move Path Candidate."));
+                    LogManager.getLogger().error("",
```
NickAragua, line 170: I gotta wonder why we're bothering creating an exception here; it's not like we're throwing it - seems like logging just the string will serve the same purpose.

megamek/src/megamek/common/weapons/CapitalMissileBearingsOnlyHandler.java
```diff
@@ -43,41 +43,25 @@ public class CapitalMissileBearingsOnlyHandler extends AmmoBayWeaponHandler {
...
             || weapon.curMode().equals(Weapon.MODE_CAP_MISSILE_WAYPOINT_BEARING_EXT));
 
+    // Defined here so we can use it in multiple methods
+    Mounted bayWAmmo;
+    int range;
+    Coords targetCoords;
+
     /**
-     * This constructor may only be used for deserialization.
+     * This constructor can only be used for deserialization.
```
NickAragua, line 52: Really don't get the beef with "may". It's a perfectly valid usage of the word.

megamek/src/megamek/client/ui/swing/RandomMapPanelAdvanced.java
```diff
@@ -1048,14 +1049,14 @@ private boolean isFieldVerified(VerifiableTextField field) {
         String result = field.verifyTextS();
         if (result != null) {
             result = field.getName() + ": " + result;
-            new RuntimeException(result).printStackTrace();
+            LogManager.getLogger().error("", new RuntimeException(result));
```
NickAragua, line 1052: Not sure why we're bothering to initialize a new exception here when just logging the string contents of it will do.

NickAragua (changes requested):
A couple of places where we're unnecessarily initializing exception objects that proceed to get thrown away immediately. Looks fine otherwise.

megamek/src/megamek/client/ui/swing/RandomMapPanelAdvanced.java
Windchild292: This is done to log the full stacktrace as part of the error. It should be logging the results as the error message instead of blank though, which I'll fix.

megamek/src/megamek/client/bot/princess/BotGeometry.java
```diff
@@ -158,7 +158,8 @@ public int judgeArea(ConvexBoardArea a) {
             boolean flip = getDirection() > 2;
             HexLine[] edges = a.getEdges();
             if ((edges[getDirection()] == null) || (edges[(getDirection() + 3) % 6] == null)) {
-                LogManager.getLogger().error(new IllegalStateException("Detection of NULL edges in ConvexBoardArea: " + a));
+                LogManager.getLogger().error("Detection of NULL edges in ConvexBoardArea: " + a,
+                        new IllegalStateException());
```
NickAragua, line 162: Because of how often this function gets called (with a big O of #paths for moving unit * #enemies for each moving unit), I'm really against initializing unnecessary throwaway objects here, even if it only happens in a fraction of the calls to the function. I get that the logger derives some information from the exception, but in this case, performance trumps pretty much every other consideration.

NickAragua (changes requested):
One place where I feel this will negatively impact bot performance, everywhere else is fine.

megamek/src/megamek/client/bot/princess/BotGeometry.java
Windchild292: Done

megamek/src/megamek/common/weapons/CapitalMissileBearingsOnlyHandler.java
Windchild292: With further review it is completely valid, but can lead to misreads (as I did before making the naming change). Using "can" prevents those misreads.

NickAragua (approved):

PR #3615, opened 2022-05-03
Fixing all Assert Usages

Windchild292:
Asserts do not work properly outside of isolated builds, so this fixes all assert uses outside of the net setup (which will be in the network rework instead). This largely removes the unused calls, only moving them to actual error handling when the call should be required.

The main changes are in the Preference setup, which uses asserts as part of the core design. Otherwise, there's a lot of cleanup in otherwise untouched files because of the change from just swapping to throwing exceptions to removing the unnecessary calls during the dev process.

This requires https://github.com/MegaMek/megameklab/pull/1096 and https://github.com/MegaMek/mekhq/pull/3261.

To review: Whitespace differences off.


megamek/src/megamek/client/ui/preferences/PreferencesNode.java
```diff
@@ -107,26 +107,28 @@ public void manage(final PreferenceElement... elements) {
...
-    public void initialize(final Map<String, String> initialValues) {
-        assert initialValues != null;
-        assert !isInitialized();
+    public void initialize(final Map<String, String> initialValues) throws Exception {
+        if (!isInitialized()) {
+            throw new Exception();
+        }
         setInitialized(true);
-        setInitialValues(initialValues);
+        setInitialValues(Objects.requireNonNull(initialValues));
```
Windchild292, line 115: Double check this

Windchild292:
This is ready for review

rjhancock (approved):

megamek/src/megamek/client/ui/swing/skinEditor/SkinSpecEditor.java
```diff
@@ -253,11 +253,16 @@ private void setupEditPanel() {
...
-            editPanel.add(skinEditPanel);
+            // TODO : Tear out broken Skin Editor
+            try {
+                SkinSpecification skinSpec = SkinXMLHandler.getSkin(selectedComp.getComp());
+                enableBorders.setSelected(!skinSpec.noBorder);
+                enableBorders.setEnabled(true);
+                skinEditPanel.setupSkinEditPanel(skinSpec);
+                editPanel.add(skinEditPanel);
+            } catch (Exception ignored) {
+
```
NickAragua, line 264: ?

megamek/src/megamek/client/ui/swing/skinEditor/SkinSpecPanel.java
```diff
@@ -648,9 +641,6 @@ public void setupSkinEditPanel(SkinSpecification skinSpec) {
 
     @Override
     public void valueChanged(ListSelectionEvent e) {
-        if (e.getValueIsAdjusting()) {
```
NickAragua, line 651: ?

megamek/src/megamek/common/AmmoType.java
```diff
@@ -10587,38 +10587,6 @@ private static AmmoType createISNailRivetGunAmmoHalf() {
         return ammo;
     }
 
-    /*
-     * private static AmmoType createCLNailRivetGunAmmo() { AmmoType ammo = new
```
NickAragua, line 10591: I think we keep these around in case they get implemented. Unless there's no such thing as clan nail rivet gun ammo or clan anil rivet gun half ammo, in which case, never mind.

megamek/src/megamek/common/AmmoType.java
```diff
@@ -8585,7 +8585,7 @@ private static AmmoType createISLRT20Ammo() {
         ammo.bv = 23;
         ammo.cost = 30000;
         ammo.flags = ammo.flags.or(F_HOTLOAD);
-        ammo.setModes(new String[] { "", "HotLoad" });
+        ammo.setModes("", "HotLoad");
```
NickAragua, line 8588: Does setModes have an appropriate overload?

NickAragua: I guess it does, but the other question is, is "setModes" used in any other location in MM/MHQ?

megamek/src/megamek/common/EquipmentType.java
```diff
@@ -682,7 +679,6 @@ public boolean isNextTurnModeSwitch(String mode) {
      * @see #hasModes()
      */
     public EquipmentMode getMode(int modeNum) {
-        assert ((modes != null) && (modeNum >= 0) && (modeNum < modes.size())) : "Invalid Mode";
```
NickAragua, line 685: I feel like these two methods are used enough and critical enough that "manual" validation is appropriate (obviously the asserts don't work, but...)

megamek/src/megamek/common/WeaponOrderHandler.java
```diff
@@ -233,8 +233,8 @@ public synchronized static void saveWeaponOrderFile() throws IOException {
                 weapOrderMap.put(unitId, weapOrder);
             }
             return weapOrderMap;
-        } catch (Exception e) {
-            throw new IOException(e);
+        } catch (Exception ex) {
+            throw new IOException(ex);
```
NickAragua, line 237: I gotta say that the idea of re-throwing a generic exception as an IOException is a little weird, but we probably better leave it as is for now.

megamek/src/megamek/common/weapons/CapitalMissileBearingsOnlyHandler.java
```diff
@@ -432,16 +431,17 @@ public void convertHexTargetToEntityTarget(Vector<Report> vPhaseReport) {
                 inArc.add(a);
             }
         }
+
         if (inArc.isEmpty()) {
-            newTarget = aaa.getTarget(game);
```
NickAragua, line 436: This seems like a weird change to make. Is newTarget initialized somewhere earlier in the method?

NickAragua (commented):
A couple of questions.

megamek/src/megamek/common/AmmoType.java
Windchild292: It's a varargs method, so the string array creation isn't necessary.

megamek/src/megamek/common/WeaponOrderHandler.java
Windchild292: I do too, but wanted to keep the rethrows consistent for now (I am planning on looking at them once everything is migrated to proper logging methods).

megamek/src/megamek/common/EquipmentType.java
Windchild292: Given it's worked without the checks thus far I think it's fine to skip... and is a definite slowdown to add that level of check to.

megamek/src/megamek/common/weapons/CapitalMissileBearingsOnlyHandler.java
Windchild292: It's setting a method variable immediately before a return, so it's setting a variable then immediately clearing that state.

megamek/src/megamek/client/ui/swing/skinEditor/SkinSpecEditor.java
Windchild292: If you mean the tear out note, it was a code misread.
Otherwise... there's a potential exception that can be thrown, but in all likelihood won't. I'll be doing a full look at catches in the future, and will probably change it then.

megamek/src/megamek/client/ui/swing/skinEditor/SkinSpecPanel.java
Windchild292: Both cases just immediately return, so it's just bad handling

megamek/src/megamek/common/AmmoType.java
Windchild292: It hasn't been used since MM was on GitHub, and the text predates that.

It's also just not useful to retain ancient code just in case...

PR #3625, opened 2022-05-07
Enum All The Things: Separate Entity Enums

Windchild292:


megamek/src/megamek/client/bot/CEntity.java
```diff
@@ -264,15 +265,15 @@ public void characterize() {
...
 
         // Make a guess as to whether MASC should be turned on or off
-        // TODO: Link this to a Bot configuration file
+        // TODO : Link this to a Bot configuration file
         runMP = entity.getRunMP();
         if (entity instanceof Mech) {
-            Entity.MPBoosters mpBoosters = ((Mech) entity).getMPBoosters();
-            if (mpBoosters.hasMASCAndOrSupercharger()) {
+            MPBoosters mpBoosters = entity.getMPBoosters();
+            if (!mpBoosters.isNone()) {
```
rjhancock, line 272: Is there an opposite version of this method so negation doesn't need to be done? For example a `isAvailable()` method?

megamek/src/megamek/common/Mech.java
```diff
@@ -972,29 +973,11 @@ public int getWalkHeat() {
...
-     * @see megamek.common.Entity#getRunMP(boolean, boolean, boolean)
-     */
     @Override
-    public int getRunMP(boolean gravity, boolean ignoreHeat,
-            boolean ignoreModularArmor) {
+    public int getRunMP(boolean gravity, boolean ignoreHeat, boolean ignoreModularArmor) {
         MPBoosters mpBoosters = getArmedMPBoosters();
-        if (mpBoosters.hasMASCAndOrSupercharger()) {
-            return mpBoosters.calcRunMP(
+        if (!mpBoosters.isNone()) {
```
rjhancock, line 979: Same here. a `isAvailable()` method would help here instead of negating a `isNone()` one. The call can be the same internally but it shows intent better than negation.

megamek/src/megamek/common/Mech.java
```diff
@@ -1164,10 +1146,9 @@ public int getOriginalSprintMPwithoutMASC() {
     @Override
     public String getSprintMPasString() {
         MPBoosters armedBoosters = getArmedMPBoosters();
-        if (armedBoosters.hasMASCAndOrSupercharger()) {
-            return getRunMPwithoutMASC() + "(" + getSprintMP() + ")";
-        }
-        return Integer.toString(getSprintMP());
+        return !armedBoosters.isNone()
```
rjhancock, line 1149: Same concept here.

megamek/src/megamek/client/bot/CEntity.java
Windchild292: There is not.

megamek/src/megamek/common/Mech.java
Windchild292: I disagree, and find redirect methods to cause more confusion than basic boolean logic

megamek/src/megamek/client/bot/CEntity.java
rjhancock: Skimming through this review I see there is a way to add one. From a readability stand point (I skipped over the `!` about half the time is why I'm asking this) it would be good to add.

megamek/src/megamek/common/Mech.java
rjhancock: I've found them to be helpful regardless of language. It's still Boolean logic it just reads in a way that makes it easier for newer people (and even experienced once) to read.

rjhancock (approved):

PR #3627, opened 2022-05-08
(DRAFT) Detachable Unit Detail Removal

Windchild292:
This invalidates #3223, #3272, #3309, #3372, #3579, #3616

PR #3629, opened 2022-05-09
Detachable Unit Display Removal

### LINKED ISSUE #3223: [nightly 49.4] Un-docking dockable unit display UI issues
Basically, when you undock the "dockable" unit display, the armor part is a way down the unit display area, and there's no way to see the bottom part without expanding the window downwards. Which sometimes is not possible. This needs to be addressed somehow - in some cases, it can be simply by moving the displayed stuff upwards, but in some cases we need a scrollbar or something.

image

Additionally, when trying to cycle through targets and check to-hit numbers, the un-dockable version "disappears". Which is annoying when you are on a single monitor and don't want half the play area taken up by the dockable version.

### LINKED ISSUE #3272: 0.49.5 can't set equipment mode
With the new dockable unit display I can no longer set equipment modes at any time during the game.  With the old pane there was a dropdown menu to allow for mode changes for equipment like ECM and targeting computers.  That dropdown menu is no longer there, and the "Mode" button only changes weapon modes on the weapon list.

### LINKED ISSUE #3309: [0.49.5] Unit 'Crippled' (etc) status not visible when unit window docked
Standalone MegaMek 0.49.5
Windows 10
Java AdoptOpenJDK 11.0.8

When a unit is crippled, the [CRIPPLED] status is displayed in the window title. However, when the unit window is docked, the window title is no longer visible, so it is no longer immediately obvious when a selected unit is crippled or not.

Undocked
image

Docked
image

### LINKED ISSUE #3372: [49.6] & [0.49.7] [Bug] Docked unit display getting larger, no way to resize
MM 49.6 Standalone
Windows 10
Java 11

For some reason it appears that, on reloading a saved game, the docked unit display is now taking up about a third of the screen width, and there seems to be no way to change the width. When undocked, the unit display floats on top of any other window unless minimised, even when MM is not in focus. The MiniMap does not do this.

This could be resolved, as well as adding usability, by making it possible to change the width of the docked unit display.

[autosave_2022-01-14_03-59-51.sav.gz](https://github.com/MegaMek/megamek/files/7872336/autosave_2022-01-14_03-59-51.sav.gz)
 (Player name: Landmine, no custom units)

### LINKED ISSUE #3579: Issues with Unit Overview window
### Environment
Ver 0.49.7
Windows 10.

### Description
First, it is not a bug report. It is actually the request.

For now it can be attached to the main window, but it also adds some annoyance. 

First, it is always on top. Yes it is fine - if we ignore the fact that it is not optional. Yes I do need it to be always on top sometimes, but not always. Why it can't be an option? It makes me too annoyed especially when I use the other programs as well - for wait a while until the AI make its move. Because of this usually I have to set aside the window THEN bring back it manually each time when I did alt-tab. Each time, I mean. Perhaps you want it to be convenient, but for me it cause more annoyance than feel comfortable actually.

Second, it now needs too much length.

r01
r02
My resolution is 1600x900. Perhaps some of you says that it is too small? Maybe. But I think that 1600x900 is too large to actually see the details and it must be smaller than that. Anyway, the window needs too much length in height, and it is hard to check if the additional details, and it almost hides change mode as well. Why it must be soooooo long? 

Can't we back to the old style? The lobby and unit cache window are already a mess, even for my large resolution(again, 1600x900 is already large enough). And am I have to suffer the problem with unit overview, that must be get along with me all the times, as well?

### LINKED ISSUE #3616: [49.8 Test Branch] New MM GUI Issues with Unit Window
23:00:13,261 INFO  [megamek.MegaMek] {main}
megamek.MegaMek.initializeLogging(MegaMek.java:115) - Starting MegaMek v0.49.8-SNAPSHOT
	Build Date: 2022-05-03T01:13:29.469020
	Release Date: 2022-02-18
	Today: 2022-05-02
	Origin Project: MekHQ
	Java Vendor: AdoptOpenJDK
	Java Version: 11.0.10
	System Locale: en_US

Very strange.  Im on the test branch.  Upgraded from a very recent nightly build.  Screenshots below.  

Edit2:  I guess this has turned into a "general spacing issues" post that is highlighted by the one-third screen docked window.  Ill just add any new edits in comments.

Edit:  Added another screenshot to bring up a spacing issue on armor tab that existed before, but is exacerbated by using the window undocked.  Also, I cannot get the floating button issue to replicate.  Ive tried to duplicate it numerous ways.

3 issues:

1. The Unit screen now defaults to takes up almost a third of the screen.  Before it took up about a quarter.  I preferred the old way.  ;)  Doesnt seem draggable to the right to reduce its size.  
2. When I detached the window, because it was too big, when I changed tabs on the detached window, got a strange appearance where the buttons went about an inch below the top of the detached window, and they covered the pilots face.  This happened on all tabs.  (Closing and relaunching MM did make the lower button issue go away).  EDIT: Added a close up.
3. Also with the detached window - it now does not minimize when you switch to another program, and always stays on top unless manually minimized.  I dont think this is correct.  Its been a while since I played with it detached, but I recall that it used to minimize with the rest of the program?  You can see its on the front screen with me as I type this.  EDIT:  Now that I think about it, the floating mini-map pop up correctly minimizes with the program, and does not show up as its on separate window when hovering over the game icon on the taskbar.  So perhaps the unit window needs the same setting as the minimap?

And thank you for the turret pop up!

[mekhq.log](https://github.com/MegaMek/megamek/files/8608495/mekhq.log)
[The Bluecoats30621106A.cpnx.gz](https://github.com/MegaMek/megamek/files/8608496/The.Bluecoats30621106A.cpnx.gz)
[megamek.log](https://github.com/MegaMek/megamek/files/8608500/megamek.log)

Screenshot (413)_LI

Screenshot 2022-05-03 001642_LI

Screenshot 2022-05-03 003226
Screenshot (422)
Edit:
Close up of split.
Screenshot 2022-05-03 011039

Windchild292:
This tears out the Detachable Unit Display and replaces it with (largely) the original code.
It replaces #3627, which kept having odd phase issues. By invalidation, this closes #3223, closes #3272, closes #3309, closes #3372, closes #3579, and closes #3616.

I will be following this up with an AbstractDialog rework (some baseline changes included) and Skin Editor dialog update immediately after 0.49.8.


NickAragua (approved):
This one, we'll definitely need people to bang on before we hit the release.

Issue #2738, opened 2021-03-21
48.0 Initiative bonus stacking issues


gsparks3:
While discussing the Mobile Headquarters optional rules, page 195 of the old TacOps rulebook states that "A side may only benefit from one Mobile HQ at a time; this includes a ’Mech mounting a cockpit command console cockpit (see p. 300)." Please verify that this remains true in more recent TacOps printings; however, I did not see any errata for this section.

In Megamek, the bonuses from a Mobile Headquarters and Command Console currently stack, which goes against this description. 

Reproduction steps:
1. Open a Megamek game. Make sure the "TacOps Mobile HQs" Game Option is active. 
2. Add a Mobile Headquarters (Standard) to your force, as well as any Command Console mech (I use the Orion ON1-V-DC)
3. During deployment, note the +2 initiative bonus (from the Mobile Headquarters; the CC bonus only applies post-deployment)
4. Following deployment, note the total +4 initiative bonus (from both the MHQ and CC). 

The MHQ/CC initiative bonus also currently stacks with the Command Mech or Battle Computer Quirks if those are present. However, I did not see any rules indicating that they should _not_ stack. There are also several canon CC mechs having said Quirks (Command Console Atlas, Battlemaster, Cyclops, etc), combining the effects into a single unit, so it's plausible that this is expected to stack. 

To make things weird, there is also the canon Awesome AWS-9Ma, which combines both a Command Console and 3 effective tons of Communications Equipment (enough for a +1 initiative bonus from the MHQ rules) into the same unit. However, there are other things that can be done with Communications Equipment, so not being able to stack those on that unit is probably fine. 

**UPDATE:** Much more serious issue discovered - initiative bonuses from _multiple players on the same team_ will stack when they shouldn't. If I deploy by myself, no matter how many Wolverines I bring (with Quirks active for Command Mech), I only ever get +1 initiative. However, if I split multiple Wolverines across multiple players (e.g. deploy with myself and N Princess bots on the same team, with one Wolverine each), the team gets +1 initiative _per player with a command mech_. This causes massively unbalanced initiative bonuses in multiplayer games. This seems to apply for the Command Console / Mobile HQ effects as well. All of these effects (Command Mech, Battle Computer, Command Console, Mobile HQ) should be team-wide and apply once each at most, on top of CM not stacking with BC and CC not stacking with MHQ.

gsparks3:
[...]

PR #3634, opened 2022-05-11
Adjust initiative bonus processing

NickAragua:
The way it works now is that, for each player, only the highest initiative bonus applies. Same goes for each team. Addresses #2738.


Windchild292:
This also has CI issues, especially relating to negative initiative modifiers

Windchild292 (approved):

PR #3644, opened 2022-05-18
update randomInt comment to be more meaningful

NickAragua:
Adds a meaningful comment to Compute.randomInt(), as I keep forgetting exactly how it works every time which leads to a large number of off-by-one errors.


Windchild292 (approved):
Approved for 0.49.8

PR #3672, opened 2022-06-01
Excised No Torso Twist & Replaced Quirks

pokefan548:
Per BattleMech Manual Errata v6.0 PRE.


pokefan548:
Correct. Some 'Mechs had no other quirks, and so were given quirk(s) as replacements for No Torso Twist. The Daboku also got an added quirk at the same time.

PR #3687, opened 2022-06-02
special "get damaged crits" logic for gun emplacement; minor bugfix

NickAragua:
Gun emplacements don't technically have critical slots, and a piece of equipment on a gun can only take one hit; so we just return 0 or 1. Among other things, this allows gun emplacements with damaged systems to show up in the repair bay, and for damage to those systems to carry over from MegaMek to MekHQ.

Also fix a null pointer exception when clicking on systems on gun emplacements in the unit display.

### DIFF
--- a/megamek/src/megamek/common/GunEmplacement.java
+++ b/megamek/src/megamek/common/GunEmplacement.java
@@ -15,7 +15,6 @@
 package megamek.common;
 
 import megamek.client.ui.swing.calculationReport.CalculationReport;
-import megamek.client.ui.swing.calculationReport.DummyCalculationReport;
 import megamek.common.battlevalue.GunEmplacementBVCalculator;
 import megamek.common.cost.CostCalculator;
 import megamek.common.enums.AimingMode;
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/common/GunEmplacement.java
```diff
@@ -443,11 +442,15 @@ public boolean hasStealth() {
...
             
-            initialBuildingCF = occupiedStructure.getCurrentCF(getPosition());
-            initialBuildingArmor = occupiedStructure.getArmor(getPosition());
+            if (occupiedStructure != null) {
+                initialBuildingCF = occupiedStructure.getCurrentCF(getPosition());
+                initialBuildingArmor = occupiedStructure.getArmor(getPosition());
+            }
         } else {
             initialBuildingCF = initialBuildingArmor = 0;
         }
```
Windchild292, line 456: This doesn't handle the previous failure case when the occupied structure is null.

Windchild292 (approved):

Issue #3722, opened 2022-06-16
Princess should respect ignoreTarget list with artillery


SuperStucco:
Currently in 49.7, when Princess has artillery units they will still target units that have been added to the ignoreTarget list.  In my current game, I've added the IDs of turreted buildings, but Princess is still targeting them with its artillery.

PR #3726, opened 2022-06-22
Check for ignored units while princess is building a target list for indirect fire

Lu9us:
Per https://github.com/MegaMek/megamek/issues/3722, Princess is currently able to target ignored units with indirect fire, this change simply checks the ignore list while generating the list of targets for indirect fire.

This does not stop Princess from targeting units that might be stacked on top of ignore listed units that would require a more significant change, though I think that might be possible.


NickAragua (approved):
Looks good, merging once auto-build completes.

Windchild292 (approved):

PR #3732, opened 2022-06-24
(DRAFT) Include MapSettings in savegame from LOUNGE phase

pakfront:
DRAFT
Moving MapSettings from ChatLounge to Game so that is serializes into the savegame.
~~Note: MapSettings is a singleton,~~ so coordinating the marshalling and unmarshalling is the main task here.


pakfront:
Found a better path to this fix.

Issue #3744, opened 2022-06-29
Bug - Mast Mount not in Support Vee equipment for VTOL


pheonixstorm:
I originally posted this under MML but the temporary fix is actually for MiscType.

When creating a support vee VTOL, Mast Mount does not show up in the equipment list for any rules level or date. It does properly display in the Combat Vee equipment list.

pheonixstorm:
Closing as resolved by #3745 and #3766

PR #3745, opened 2022-06-29
Fixes the Mast Mount equipment issue for Support VTOLs

pheonixstorm:
This is a band-aid fix for issue #3744 

There is a greater bug somewhere in the code that causes support VTOLs to not classify them as VTOLs. This issue affects any equipment only allowed on VTOLs such as the Mast Mount and VTOL Jet Booster. These may be the only items that are VTOL only so let me know if the underlying bug is worth a separate ticket.

### DIFF
--- a/megamek/src/megamek/common/MiscType.java
+++ b/megamek/src/megamek/common/MiscType.java
@@ -8584,7 +8584,7 @@ public class MiscType extends EquipmentType {
         misc.tonnage = 0.5;
         misc.tankslots = 0;
         misc.cost = 50000;
-        misc.flags = misc.flags.or(F_MAST_MOUNT).or(F_VTOL_EQUIPMENT);
+        misc.flags = misc.flags.or(F_MAST_MOUNT).or(F_VTOL_EQUIPMENT).or(F_SUPPORT_TANK_EQUIPMENT);
         misc.bv = BV_VARIABLE;
         misc.rulesRefs = "350, TO";
         misc.techAdvancement.setTechBase(TECH_BASE_ALL).setIntroLevel(false).setUnofficial(false)


NickAragua (approved):

NickAragua:
Although, doesn't this mean that non-VTOL support tanks will be allowed to have mast mounts now?

pheonixstorm:
Yes, just like they can currently add VTOL Jet Boosters to a tracked SV. I spent hours digging and looking. I tried instanceof VTOL,  Support VTOL, I even tried replacing it with ETYPE and I couldn't get it to recognize Mast Mounts when constructing a support VTOL w/o the tag in the PR.

SJuliez:
@pheonixstorm maybe you're interested to add to this PR: The support tank flag for Mast mounts is correct. The unit filtering is "hosted" in the shouldShow() method of MML's AbstractEquipmentDatabaseView. This method checks equipment types, filter buttons, text filter and equipment availability. Additional filtering within the "Other" category (MiscTypes and stuff) is in EquipmentDatabaseCategory beginning with

OTHER ("Other",
            (eq, en) -> ((eq instanceof MiscType)
                    && !UnitUtil.isPhysicalWeapon(eq)


In this block, adding the lines:

&& (!eq.getInternalName().equals("ISVTOLJetBooster") || (en instanceof VTOL))
&& (!eq.getInternalName().equals("ISMastMount") || (en instanceof VTOL))

will hide these two equipments from any CV/SV that isn't a VTOL. (I think specific equipment should be addressed by internal name rather than some flag that may or may not continue to be just that equipment only. Ideally a constant should be added for such internal names to EquipmentTypeLookup)

It could be said that even this is sort of a band-aid, as both Mast Mounts and VTOL Jets should actually be filtered out in TestTank.legalForMotiveType(), at least for CV VTOLs. I don't know if this method can apply to Support VTOLs but probably not which means that the additions in OTHER would still be necessary.

SJuliez:
Nvm I'll make the MML PR

pheonixstorm:
Neither part should require the TANK or SUPPORT tags. If I removed the flag F_TANK_EQUIPMENT for the Jet Booster it no longer shows up in the Combat Vehicle equipment list except for VTOLs. Mast Mounts already worked properly for combat vehicles.

In the MML src we already have `if (eq.hasFlag(MiscType.F_VTOL_EQUIPMENT) && (tank instanceof VTOL))` in UnitUtil under isTankEquipment (line 2912).

Adding yours does work, but it only hides the problem. That line above allows for only the VTOL_Equipment tag to be used for Combat Vehicles and should also work for Support Vehicles. The problem is why it doesn't work for Support VTOLs which are supposed to extend VTOL.

This is why I went with a fix that leaves part of the problem that has been around for awhile. 

Anyway, long night and I'm tired.

PR #3763, opened 2022-07-04
Don't overwrite lrm and srm ranges for non-deadfire.

### LINKED ISSUE #3757: [0.49.Snapshot] Minimum Range for Clan LRM
### Environment
V: 0.49 commit 85db651a8a49b20eb10d4381116b4889d9ee8934
O: Linux
J: openjdk11

### Description
I have a mech with clan LRM15 and in the overview (picture attached) it shows correct Min: ---
But in the overlay in the map it shows the red area "min" and in the to hit calculation it states "+1 (minimum range)"

### Files
Screenshot_2022-07-02_23-19-17

neoancient:
The range adjustment code does not need to assign ranges for missile launchers in the case of non-deadfire missiles, as these ranges have already been assigned using the weapon range brackets. This only causes a problem for LRMs, which apply the IS minimum range to Clan launchers as well, but I also removed the else block from the SRM section as it is superfluous.

Fixes #3757

### DIFF
--- a/megamek/src/megamek/common/WeaponType.java
+++ b/megamek/src/megamek/common/WeaponType.java
@@ -470,12 +470,6 @@ public class WeaponType extends EquipmentType {
                 mRange = 10;
                 lRange = 15;
                 eRange = 20;
-            } else {
-                minRange = 6;
-                sRange = 7;
-                mRange = 14;
-                lRange = 21;
-                eRange = 28;
             }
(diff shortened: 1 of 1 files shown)


HammerGS (approved):
Thanks for cleaning up my clumsy attempt. :)

Windchild292 (approved):

PR #3767, opened 2022-07-05
Server collections

neoancient:
1. Replaces the connections and pendingConnections Vectors with CopyOnWriteArrayList, which is also thread-safe but offers better performance when reads are much more common than writes or when the list is small, both of which are the case. It also fails safe, unlike Vector.
2. No longer exposes connections outside Server.
3. Replaces Hashtable with the more performant ConcurrentHashMap.

This fixes the problems with #3691.

### DIFF
--- a/megamek/src/megamek/server/Server.java
+++ b/megamek/src/megamek/server/Server.java
@@ -134,7 +134,7 @@ public class Server implements Runnable {
 
     private final boolean dedicated;
 
-    private Vector<AbstractConnection> connectionsPending = new Vector<>(4);
+    private final List<AbstractConnection> connectionsPending = new CopyOnWriteArrayList<>();
 
     private Hashtable<Integer, AbstractConnection> connectionIds = new Hashtable<>();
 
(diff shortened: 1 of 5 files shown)


NickAragua (approved):

Windchild292 (changes requested):
The use of copied lists causes resource leaks, and thus this does not work as intended.

neoancient:
What's holding the reference?

neoancient:
It's been two weeks since my request for explanation, with none forthcoming. `CopyOnWriteArrayList` has been part of the Java concurrency framework since 1.5 and is intended for use in just this sort of situation. I don't what your claim about resource leaks is based on, but the class holds the data in a private array that is eligible for garbage collection after the array is copied during modification. The only time a reference is held is when the internal iterator class is created, but once the iterator passes out of scope the reference to the old copy is released and it is then eligible for garbage collection. My rewrite of `getConnections` avoids exposing an iterator outside the Server class, so that is not an issue.

HammerGS:
Lets just get this merged. We are in Dev cycle and can make adjustments based of what the QA players encounter.

NickAragua:
The only argument I could see for resource leaks is if a particular AbstractConnection or other object within one of those CopyOnWrite collections doesn't properly dispose of itself (i.e. close external resources such as network connections) when it is garbage collected. But probably "socket" objects (which are used under the hood in AbstractConnection) are self-disposing as well when they go out of scope. 

So it's a pretty thin argument. Let's get it merged and then we'll have people playtest it by repeatedly connecting and disconnecting clients and see if any problems develop. Especially this would be useful to check for dedicated servers.

PR #3774, opened 2022-07-08
fixes to deployment zone GUI

NickAragua:
Two changes:
- change NumberFormatter settings so that user can empty the textbox out
- institute minimum 0 value
- adjust layout slightly


sldfgunslinger2766:
🤘Thanks for adding this feature!

Issue #3749, opened 2022-06-29
RFE - Make /Seeall work for non-observers or have option to force Princess as an observer.


Thom293:
Pursuant to a discussion on discord, one or both of the following options:

Option 1, suggested by Look-Busy Loop, Pakfront and Illiani:

- remove requirement from seeall that person must be an observer
- add consent requirement to seeall, copying code from /defeat
- if consent is given seeall fires, else it throws up a chat box error saying it failed due to lack of consent

Option 2:  EDIT: I DONT THINK THIS ONE WILL WORK.  I tried to set her as an observer while she had units, and she could not deploy.

Make princess have a tick box to give her a flag so she can be set as an observer by the game manager, or force-set her as an observer-with-units in the lobby.  

Then the current /Seeall command might work to give her vision of the whole map.

Perhaps making Princess an observer-with-units and then using the /seeall command on her would give the single-blind effect.  Or, making her an observer-with-units MIGHT break a bunch of stuff.  Dont know.

IllianiCBT:
Related to: https://github.com/MegaMek/megamek/issues/203

Thom293:
The first /Seeall option appears to roughly work.  

Look-Busy Loop was kind enough to point me to a salient section of code.  I deleted some things.  Tested twice and appears to work.  

Tested vs 2 princess bots on same team.  A and B.  I waited 10 turns, and they never approached.  On turn 10 gave PrincessA the /seeall 1 command, and a few turns later she pops over the hill coming for me.  Princess B did not immediately approach, but eventually she did.  I assumed if I gave seeall to A and the share vision option is enabled, B would have seen me immediately too?  

Anyway - here is the code Look-Busy Loop showed me, and the 2nd shows what I deleted.  No idea if  I over-deleted or if anything else was broken as a result.  Id have no idea how or where to add any consent code.

From megamek\common\Player.java file.

![Screenshot 2022-06-29 221108](https://user-images.githubusercontent.com/23645569/176586425-6d99f185-74d1-451a-9d43-e97d652e1d2d.png)

![Screenshot 2022-06-29 215150](https://user-images.githubusercontent.com/23645569/176586450-41b469cd-9365-4077-98d0-c3c93a22715f.png)

IllianiCBT:
This definitely looks like it has potential. My biggest, and only, concern is whether there is interactions with other systems that aren’t immediately obvious.

Without looking at the code myself, I won’t even guess.

Thom293:
[...]

Issue #203, opened 2016-05-18
Single Blind Option for Princess


scJazz2:
Reference... https://sourceforge.net/p/megamek/feature-requests/1354/

after chatting with Arlith, Akjosch, and others on Esper #battletech...

I would like to RFE a Single Blind option for Princess games. Princess does a /seeall for each Princess player? Human plays as Blind.

pheonixstorm:
I would rather see princess fixed to allow for double blind myself. It makes more sense than to handicap the player and allow the bot to cheat.

scJazz2:
low hanging fruit dude... low hanging. It is at least a start

pheonixstorm:
Yeah, but it would still be a lot of work compared to just fixing princess which would probably just be harder but take less time. If you do single blind then later on fix princess you have to still remove the old code. But, that's just me looking at it from a programmers perspective rather than a players.

BLOODWOLF333:
Single blind already kinda works.  /seeall 1 (if princess is #1 under /who) and that commands MM to give her full vision while you are still under the blind rules.  AJ had mentioned something in passing about thats how AI's in other games do it as well where the bot always knows where the player is the whole time IIRC.

pheonixstorm:
Yeah, it is one of the AI cheats for almost ALL RTS/FPS games as a means to keep the AI coding down since AI has never been a high priority for games.

dericpage:
Single blind would not only be easier to code but it also would not need to be intrusive.  What Princess can or cannot see is not housed within the Princess code at all.  That's handled by the game engine itself.  So, if Princess gets updated to "properly" handle double-blind play, there's no need to remove any of the single-blind code.  The option would simply need to be turned off.

BLOODWOLF333:
Was thinking about this and wondered if to simulate double-blind with princess that at the beginning of the game princess would have the locations of all the players units after they all deploy.  Then unless she has LOS to the players units she would go off the assumption that all the units are still where the player deployed them(last known location) and would receive an update on all the player unit locations every six turns (1 min game time).  That way she would at least go in the right direction even if going after ghost targets until the one minute update occurs but, she wouldn't always know each player unit's exact location 100% of the time.

pheonixstorm:
Interesting idea, worth looking into I think.

scJazz2:
Hmmm,.. that does of course mean fiddling with Princess instead of the Visibility code. Still interesting idea.

dericpage:
[...]

PR #3780, opened 2022-07-11
Single blind v2 - Makes requested changes and fixes a vanilla bug.

Thom293:
This is the same as the other PR with 2 exceptions:  It fixes an already existing bug from vanilla so now it will give a proper failure message, and this also makes Windchild's suggested changes.  I will close the other PR.  Original text below:

Added /singleblind toggle command that is only assignable to bots. Its gives them vision of the entire map, just like /seeall. It cannot be assigned to a human player.

Bots given vision with /singleblind do not share /singleblind vision with anyone on their team - so observers, other bots with units, or humans with units on their team do not see anything even if the bot can, therefore I don't think a vote mechanic is needed. One could certainly be added by someone with more skill than me later if desired.

Usage is just like see all. "/singleblind 1" to give to a bot in the player 1 spot.

Anyone is welcome to edit or improve this.

If approved, this will resolve https://github.com/MegaMek/megamek/issues/203 and moot or partially resolve https://github.com/MegaMek/megamek/issues/3749.

This should not change the current /seeall command or observer code in any way. This was created by basically just copying the /seeall command, renaming all copied references, and swapping the observer requirement for a bot requirement.

I believe all of the changes requested by Windchild have been made, and work.  But if not please let me know.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/lobby/ChatLounge.java
+++ b/megamek/src/megamek/client/ui/swing/lobby/ChatLounge.java
@@ -2335,7 +2335,27 @@ public class ChatLounge extends AbstractPhaseDisplay implements
         }
         return true;
     }
-    
+
+    // New SingleBlind - I dont think this is used.
+    /**
+     * Returns true if the local player can see all the given entities.
+     * This is true except when a blind drop option is active and one or more
+     * of the entities are not on his team.
+     */
+    boolean canSingleBlind(Collection<Entity> entities) {
+        if (!game().getOptions().booleanOption(OptionsConstants.BASE_BLIND_DROP)
+                && !game().getOptions().booleanOption(OptionsConstants.BASE_REAL_BLIND_DROP)) {
+            return true;
+        }
+        for (Entity entity: entities) {
+            if (!entityInLocalTeam(entity)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+
     /**
      * Returns true if the local player can see the given entity.
      * This is true except when a blind drop option is active and one or more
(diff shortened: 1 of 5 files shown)


NickAragua (approved):

Thom293:
@NickAragua Thank you!  I hope its useful.

@HammerGS Usage description.:  

This is a chat command toggle - /singleblind.  It is intended to give vision of the entire map to a bot player as a partial working option for the bot in Double Blind games.  You should get a failure message if it is attempted to be assigned to a non-bot player.

To use, type:

/who - this will give you player numbers.  In this example, Princess (a bot) is player 1.
/singleblind 1 - this will give a bot in the Player 1 spot vision of the entire map at the start of the next phase (after any player hits done).  It is a Toggle, so if you type /singleblind 1 again, it will remove the vision at the start of the next phase.

Notes:  

If it is turned off, unless Princess has direct LOS on an enemy unit, she will forget where enemy units were, and at least in my tests, she will always return to her home edge and mill about, even if she has travelled quite far away from it.

While it is active, she does not share the /singleblind sight with her teammates, even if "teammates share vision" option is enabled.  So she will see enemy units and move toward them, but other humans and bots on her team will not see them.  So if you want different bots on the same team to have singleblind vision, you need to assign it to each of them. However, once she gets direct LOS on an enemy unit, because of normal vision rules, she will share their location with teammates and remain engaged, even if you turn /singleblind off.

You should get a failure message and it should not work if you attempt to assign it to a human.  If you somehow manage to assign it or otherwise give it to a human player, please report it so it can be fixed.

HammerGS:
I'll add this as a text file in the bots folder tomorrow.

Issue #3786, opened 2022-07-14
[Nightly] Bug: If a bot is assigned to a slot in MM, game perpetually thinks that slot is a bot, even if replaced by a human.


Thom293:
This vanilla bug came to light when I was testing the singleblind command.  

SB should be assignable only to bot players, and it works correctly in that regard.  But if a bot player is kicked and replaced by a human of the same name, you can still assign SB to that player slot, even though it checks for isBot first.  So, I think this means the game still thinks that player slot is a bot, even though it is now occupied by a human.  The bot flag appears to never get reset to false upon player change.  After discussion with others in discord, we believe this to be a bug behavior in the vanilla game.  It seems it only sets the isbot flag once, and then never checks for it again, even if there is a disconnect/reconnect.  

So, I think what it should do is somehow the bot status of a player position gets checked/reset when a new player connects, and/or gets cleared when a player disconnects.  (This should also happen for the singleblind flag in the same location).

This is beyond my current ability to solve correctly or efficiently.  This is more difficult than a copy/rename job, so Im essentially just guessing on how to fix it and we dont want that.  

I did guess anyway, and maybe my failure will help someone succeed.  I did attempt a few different fixes, and my best attempt only mostly works.  Upon disconnect, It clears the singleblind flag, clears the bot flag, and will allow a human to connect to the slot in question and not be able to have SB assigned to it.  That is the desired and expected behavior.  However, if you kick a bot, and then replace it with a bot, the new bot cant be assigned SB, so it appears a newly joining bot does not reset its flag to bot, or my attempted change prevents it from doing so.

I also tried to fix it in the receivePlayerName section, but I could not get it to work.  But I probably just did it wrong.

For reference, here is my attempted-but-only-partially-successful fix from the server.java file.  I am quite sure there are better/cleaner ways to fix it.  I just dont know  what they are.  

![Screenshot 2022-07-13 225902](https://user-images.githubusercontent.com/23645569/179065157-d130792d-2033-4d69-8588-dd54fc7ce9c4.png)

Issue #3784, opened 2022-07-14
Artillery may not be using double and single blind correctly in some cases


pakfront:
MegaMek v0.49.9-SNAPSHOT
In GameManager,createArtilleryPacket there is a test against getSingleBlind and getSeeAll that probably should be against isSingleBlind and canSeeAll. The former (get) does not check if the modes are valid, only if they are set, the latter does the check for validity.

```
                if ((aaa.getPlayerId() == p.getId())
                        || ((team != Player.TEAM_NONE)
                        && (team == game.getPlayer(aaa.getPlayerId()).getTeam()))
                        || p.getSingleBlind() || p.getSeeAll()) {
                    v.addElement(aaa);
                }
```

Thom293:
I trust you completely on the matter.  I simply tried to copy the seeall format exactly, so if it had underlying oddities, then singleblind probably does too.

PR #3787, opened 2022-07-14
Game Master flag, preliminary work

pakfront:
This add a new 'gameMaster' (GM) flag to Player. It allows for privileged actions to facilitate refereeing and game mastering of games. At the moment this is limited to allowing /seeall on a GM, even if it is not an observer. More advanced features will be added in the future.
- player using /gm command will start voting for them to be a Game Master
-- other players can vote for this with /allowGM
- player who is already a GM can remove GM mode by using /gm again. No vote is required to disable.
-  tries to rationalize some of the voting code. Probably more can be done, but it does fix the problem where a solo human player cannot trigger a vote to be counted.
- Resolves https://github.com/MegaMek/megamek/issues/3786
- adds more detail to Player List UI (this could use some prettying up later)
- properly copies some flags when Player is cloned
- Resolves https://github.com/MegaMek/megamek/issues/3784
- Consolidates double blind checks against seeAll and singleblind


Thom293:
This looks great!  I had no idea you were so close.  You are welcome  to manipulate singleblind to add a vote or otherwise work with GM mode, however you wish.

Just a note: in the vanilla game - I could not get /seeall to work with a password, but I just may not know what I am doing.  Or it may not have worked because I was running it through the IDE.  Not sure.  I just could not get it to work.

pakfront:
Thanks. Singleblind functionality looks good as is. I don't think a vote is needed. I will probably clean up a few things here and there and I will test to see if I can get a password to work with seeall

megamek/src/megamek/common/Player.java
```diff
@@ -84,12 +83,10 @@ public final class Player extends TurnOrdered {
...
     private boolean admitsDefeat = false;
-    
-    /**
-     * Boolean that keeps track of whether a player has accepted another 
-     * player's request to change teams.
-     */
-    private boolean allowingTeamChange = false;
+
+    //TODO Voting should not be stored in savegame. mark transient?
+    private boolean votedToAllowTeamChange = false;
```
NickAragua, line 88: Yeah, transient is the way to go here.

megamek/src/megamek/server/commands/AllowGameMasterCommand.java
```diff
@@ -0,0 +1,94 @@
...
+package megamek.server.commands;
+
+import megamek.common.Player;
+import megamek.server.GameManager;
+import megamek.server.Server;
+
+/**
+ * This command allows a player to allow another player to switch teams.
+ *
+ * @author arlith
```
NickAragua, line 28: Give yourself a little credit. ;)

megamek/src/megamek/server/commands/AllowGameMasterCommand.java
```diff
@@ -0,0 +1,94 @@
...
+    protected static void voteYes(Server server, Player player ) {
+        player.setVotedToAllowGameMaster(true);
+
+        // Tally votes
+        boolean allowGameMaster = true;
+        int voteCount = 0;
+        int eligiblePlayerCount = 0;
+        for (Player p : server.getGame().getPlayersVector()) {
+            if (p.getTeam() != Player.TEAM_UNASSIGNED) {
+                allowGameMaster &= p.getVotedToAllowGameMaster();
```
NickAragua, line 67: Do we want to shortcut this to break out of the loop whenever we hit a "no" vote?

NickAragua (changes requested):
A little documentation and one minor optimization.

megamek/src/megamek/server/commands/AllowGameMasterCommand.java
pakfront: Currently there is no way to vote "no." False just mean you haven't voted yet. So the vote hangs unresolved until someone restarts the vote count with /gm (/joinTeam uses the same logic)

 Adding a "no" option would be a new feature, one worth having. A some generic voting code would be good.

NickAragua: Let's limit scope and leave it for another pull request, then.

NickAragua (approved):

Thom293:
Nice.   Thank you both.

Issue #3800, opened 2022-07-21
Bug: 49.8 MASC OR Supercharger allows Sprinting reguarless of checkbox


MarcJC:
![image](https://user-images.githubusercontent.com/20606833/180117721-08a49f6d-4bba-4ad0-8ead-99f09ba16acc.png)
When running a test game I found that my mechs were able to sprint on specific conditions.  I doubled checked and found that the check box for sprinting is off.  

The issue only occurs when either MASC or a Supercharger is activated.  If both systems or no acceleration systems are active the bug allowing sprinting will not occur.  

It seems that the mech does not have to mount both systems like shown in the picture.  Another mech was able to sprint and it is only equipped with MASC.

PR #3801, opened 2022-07-21
properly handle sprint option for long moves (masc or supercharger)

pakfront:
fixes https://github.com/MegaMek/megamek/issues/3800

### DIFF
--- a/megamek/src/megamek/common/MoveStep.java
+++ b/megamek/src/megamek/common/MoveStep.java
@@ -2403,47 +2403,41 @@ public class MoveStep implements Serializable {
                 } else {
                     movementType = EntityMovementType.MOVE_RUN;
                 }
-            } else if (canUseSprint(game)
-                    && ((getMpUsed() <= sprintMPNoMASC)
-                            || ((getMpUsed() <= sprintMP) && (isMASCUsedEntity || isSuperchargerUsedEntity)))
-                    && !isRunProhibited() && !isEvading()) {
-                if (entity.getMovementMode() == EntityMovementMode.VTOL
-                        || (entity.getMovementMode() == EntityMovementMode.WIGE
-                                && getClearance() > 0)) {
-                    movementType = EntityMovementType.MOVE_VTOL_SPRINT;
-                } else {
-                    movementType = EntityMovementType.MOVE_SPRINT;
-                }
-            } else if ((getMpUsed() <= sprintMPOneMASC) && !isRunProhibited()
-                    && !isEvading()) {
-                // decide if using MASC or Supercharger if need only one
-                // choose Super if even
-                MPBoosters mpBoosters = entity.getArmedMPBoosters();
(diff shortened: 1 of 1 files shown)


HammerGS:
Looks like we might have a couple of files.

pakfront:
> Looks like we might have a couple of files.

yup, removed them from the request.

megamek/src/megamek/common/MoveStep.java
```diff
@@ -2403,47 +2403,41 @@ && getClearance() > 0) {
...
-            } else if ((getMpUsed() <= sprintMPOneMASC) && !isRunProhibited()
-                    && !isEvading()) {
-                // decide if using MASC or Supercharger if need only one
-                // choose Super if even
-                MPBoosters mpBoosters = entity.getArmedMPBoosters();
-                int scTarget = mpBoosters.hasSupercharger() ? entity.getSuperchargerTarget() : 2000;
-                int mascTarget = mpBoosters.hasMASC() ? entity.getMASCTarget() : 2000;
-                if (mascTarget < scTarget) {
+            } else if (canUseSprint(game)) {
+                if (((getMpUsed() <= sprintMPNoMASC) || ((getMpUsed() <= sprintMP) && (isMASCUsedEntity || isSuperchargerUsedEntity))) && !isRunProhibited() && !isEvading()) {
```
NickAragua, line 2407: I feel like this if statement series is complicated enough that it could use a comment explaining the "real life" scenarios we're dealing with.

NickAragua (changes requested):

pakfront:
I'm going through the existing code now and trying to simplify it and document it better. A few rules questions arise.
Does a unit with 'Poor performance' quirk get to ignore the penalty if it is using MASC or Supercharger?
In the current code the answer is 'sometimes' which seems wrong either way.

pakfront:
This turned into a deeper dive, I cleaned up Sprint and Run cases and think i fixed a few edge cases as well

pakfront:
> I'm going through the existing code now and trying to simplify it and document it better. A few rules questions arise. Does a unit with 'Poor performance' quirk get to ignore the penalty if it is using MASC or Supercharger? In the current code the answer is 'sometimes' which seems wrong either way.

I can find no reference that says MASC or SC allows ignoring 'Poor performance' so I have changed the code to not allow any running or sprinting if you have 'Poor performance', even if you have MASC or SC.

NickAragua:
@HammerGS Want to weigh in on the question?

HammerGS:
Nothing in the quirk prevents the use of MASC or SC. Given MASC/SC is based on Walking poor performance but I'll confirm with the rules team.

pakfront:
Just to be clear:
In the old code, a unit with MASC or SC could sometimes ignore the limitations of 'Poor Performance'
In the new code,  a unit with MASC or SC can never ignore the limitations of 'Poor Performance'

HammerGS:
Catching up on this. Its good to go rules wise.

NickAragua (approved):

PR #3807, opened 2022-07-22
Ratgen mul

SJuliez:
Requested by @HammerGS. Adds a button to the RATGenEditor that opens the MUL entry of the currently selected unit in a browser window.

image


NickAragua (approved):
Looks pretty straightforward.

PR #3823, opened 2022-07-29
Consolidate MASC/SC Target Number lookup

pakfront:
fixes an issue where only MASC was respecting the (unofficial)  alternate options that reduce initial Target Number.  Supercharger now uses them too, which matches the behavior describe by the options tooltip.

### DIFF
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -13968,13 +13968,7 @@ public abstract class Entity extends TurnOrdered implements Transporter, Targeta
     }
 
     public int getMASCTarget() {
-        if ((game != null) && game.getOptions().booleanOption(OptionsConstants.ADVANCED_ALTERNATE_MASC_ENHANCED)) {
-            return ALTERNATE_MASC_FAILURE_ENHANCED[nMASCLevel];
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/Entity.java
```diff
@@ -13991,7 +13985,18 @@ public void setSuperchargerUsed(boolean used) {
...
         // uses same TNs as MASC
-        return MASC_FAILURE[nSuperchargerLevel];
+        return getMASCorSuperchargerTarget(nSuperchargerLevel);
+    }
+
+    /** @return Target number taking into account game options */
+    private int getMASCorSuperchargerTarget(int nLevel) {
+        if ((game != null) && game.getOptions().booleanOption(OptionsConstants.ADVANCED_ALTERNATE_MASC_ENHANCED)) {
+            return ALTERNATE_MASC_FAILURE_ENHANCED[nLevel];
+        } else if (game != null && game.getOptions().booleanOption(OptionsConstants.ADVANCED_ALTERNATE_MASC)) {
```
Windchild292, line 13995:
```suggestion
        } else if ((game != null) && game.getOptions().booleanOption(OptionsConstants.ADVANCED_ALTERNATE_MASC)) {
```

Windchild292 (commented):
Looks good, bar one legacy code issue

Issue #3806, opened 2022-07-22
Ebony Magistrate Assault Commando Bug


MoonPepp:
Fill out the fields below with as much pertinent information as possible to help the developers solve the issue.
Search for your issue inside open and closed issues first before reporting so as not to make duplicates.
(Delete the above lines)

### Environment
Bug recreated in 0.48.0 and v0.49.8
Windows 10
Java 11

### Description
While adjusting the anti-mech value of the Assault Commandos, if you adjust the mech value downwards (Base 8, down to 7) it declares the unit invalid and displays the BV as going down (36 to 32) There is also a discrepancy of the listed BV between Mega Mek Labs (52 BV2) and Mega Mek (BV 32) 
Opened MM in .48.0 and .49.8 and replicated the bug.

### Files
No saved game. 
![Commandos with adjustment](https://user-images.githubusercontent.com/109823504/180506363-238266d2-89fc-437c-83ae-02b9712fa72b.png)
![Commandos without Adjustment1](https://user-images.githubusercontent.com/109823504/180506424-6a694067-19ec-47e8-b91d-9fde9430008c.png)

gsparks3:
Note: this is a generic bug for any and all CI. With an anti-mek skill of 8, their calculated BV is displayed as if the anti-mek skill is 5. At values of 6 or 7 the correct (lower) value is displayed.

SJuliez:
The Ebony Magistrate Assault Commando shows as invalid after a change when Manei Domini options are not activated because they use Dermal Armor which allows them to carry their 3 support guns instead of only 2.

PR #3828, opened 2022-07-30
Remove special treatment for Anti-Mek 8 BV calculation

### LINKED ISSUE #3748: [49.7] Infantry BV calculation for encumbering armour
Bundled Megamek with MekHQ 0.49.7

Megamek currently calculates infantry BV incorrectly for infantry wearing encumbering armour. 

Infantry incapable of making anti mek attacks don't pay the BV for the anti mek step in the calculation.
https://bg.battletech.com/forums/techmanual/anti-mech-bv-questions/msg1845225/#msg1845225

None of the infantry wearing ballistic plate on the MUL match the BV in megamek, for example:
http://www.masterunitlist.info/Unit/Details/1449/heavy-support-infantry-irian-technologies-swat-team
e bv

SJuliez:
This is a fix for #3806 as far as BV calculation goes. 
Fixes #3748 

The code used to treat Anti-Mek 8 as if it was 5 as a special infantry treatment. TM p.309 seems clear that No Anti-Mek gear -> Anti-Mek = 8 for the purpose of BV calculation. MM assumes no Anti-Mek gear when Anti-Mek = 8 and lowers the weight of the unit; it adds weight for the gear when Anti-Mek < 8. Therefore Anti-Mek = 8 should not be treated as 5, but as 8. That is, 8 should be treated as 8. This makes the whole special treatment unnecessary. So this PR removes this special infantry treatment for BV.

This leads to values of BV that are lowest at Anti-Mek = 8 and rise with every improvement of the skill.

In LAMPilot the usePiloting parameter is also unnecessary and this method was never called with usePiloting=false for anything but Infantry.

image

### DIFF
--- a/megamek/src/megamek/common/Crew.java
+++ b/megamek/src/megamek/common/Crew.java
@@ -129,18 +129,6 @@ public class Crew implements Serializable {
     public static final String SPECIAL_BALLISTIC = "Ballistic";
     public static final String SPECIAL_MISSILE = "Missile";
 
-    private static final double[][] bvMod = new double[][]{
-            {2.42, 2.31, 2.21, 2.10, 1.93, 1.75, 1.68, 1.59, 1.50},
-            {2.21, 2.11, 2.02, 1.92, 1.76, 1.60, 1.54, 1.46, 1.38},
-            {1.93, 1.85, 1.76, 1.68, 1.54, 1.40, 1.35, 1.28, 1.21},
-            {1.66, 1.58, 1.51, 1.44, 1.32, 1.20, 1.16, 1.10, 1.04},
-            {1.38, 1.32, 1.26, 1.20, 1.10, 1.00, 0.95, 0.90, 0.85},
-            {1.31, 1.19, 1.13, 1.08, 0.99, 0.90, 0.86, 0.81, 0.77},
-            {1.24, 1.12, 1.07, 1.02, 0.94, 0.85, 0.81, 0.77, 0.72},
-            {1.17, 1.06, 1.01, 0.96, 0.88, 0.80, 0.76, 0.72, 0.68},
-            {1.10, 0.99, 0.95, 0.90, 0.83, 0.75, 0.71, 0.68, 0.64},
-    };
-
     //region extraData inner map keys
     public static final String MAP_GIVEN_NAME = "givenName";
     public static final String MAP_SURNAME = "surname";
@@ -985,13 +973,6 @@ public class Crew implements Serializable {
         return (getGunnery() != 4) || (getPiloting() != 5);
     }
 
-    /**
-     * @return the BV multiplier for this pilot's gunnery/piloting
-     */
-    public double getBVSkillMultiplier() {
-        return getBVImplantMultiplier() * getBVSkillMultiplier(getGunnery(), getPiloting());
-    }
-
     public double getBVImplantMultiplier() {
         int level = 1;
         if (options.booleanOption(OptionsConstants.MD_PAIN_SHUNT)) {
@@ -1004,19 +985,6 @@ public class Crew implements Serializable {
             level = 5;
         }
         return level / 4.0 + 0.75;
-    }
-
-    /**
-     * Returns the BV multiplier for a pilots gunnery/piloting. This function is
-     * static to evaluate the BV of a unit, even when they have not yet been
-     * assigned a pilot.
-     *
-     * @param gunnery  the gunnery skill of the pilot
-     * @param piloting the piloting skill of the pilot
-     * @return a multiplier to the BV of whatever unit the pilot is piloting.
-     */
-    public static double getBVSkillMultiplier(int gunnery, int piloting) {
-        return bvMod[MathUtility.clamp(gunnery, 0, 8)][MathUtility.clamp(piloting, 0, 8)];
     }
 
     public boolean hasEdgeRemaining() {
--- a/megamek/src/megamek/common/Crew.java
+++ b/megamek/src/megamek/common/Crew.java
@@ -16,7 +16,6 @@
 package megamek.common;
 
 import megamek.client.generator.RandomGenderGenerator;
-import megamek.codeUtilities.MathUtility;
 import megamek.common.enums.Gender;
 import megamek.common.icons.Portrait;
 import megamek.common.options.IOption;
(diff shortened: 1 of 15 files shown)


NickAragua (approved):

SJuliez:
Re-reading the rules and another open issue I think I'll have to look at this some more.

HammerGS:
I wonder if it would be best to set up the use of the equipment along with the skills.

SJuliez:
@HammerGS What do you mean? I was also looking at #3748 (and the linked forum post there) and coming to the conclusion that Mech Inf (as the rules say) but also Inf with encumbering armor treat their AM skill as 5 (neutral) but they don't get the AM weapon BV bonus (TM 309 and 315). All other Inf get the AM weapon BV bonus but those that don't have the gear have their AM skill fixed at 8. I have this coded with the result that for Mech Inf/encumbered Inf the BV doesnt care what the AM skill is (it should be hidden but that's for another day). Still, almost none of the infantry BV match the MUL.

SJuliez:
In fact the example BV for Infantry given in TM p. 309 is matched by MM but the MUL also lists another value.

HammerGS:
> @HammerGS What do you mean? I was also looking at #3748 (and the linked forum post there) and coming to the conclusion that Mech Inf (as the rules say) but also Inf with encumbering armor treat their AM skill as 5 (neutral) but they don't get the AM weapon BV bonus (TM 309 and 315). All other Inf get the AM weapon BV bonus but those that don't have the gear have their AM skill fixed at 8. I have this coded with the result that for Mech Inf/encumbered Inf the BV doesnt care what the AM skill is (it should be hidden but that's for another day). Still, almost none of the infantry BV match the MUL.

Sorry I should have expanded my comment and provided better details. :) my comment is more about the construction and play stuff than the BV but it would impact BV depending on the skill. 

I was referencing TM pg 155 - _This kit includes the necessary climbing gear and a satchel charge; with it, the unit can possess a
better Anti-’Mech Skill Rating than 8 in game play. Without it, the unit receives a fixed Anti-’Mech Skill Rating of 8 during
construction (see p. 314 to alter a unit’s Anti-’Mech Skill Rating)._

TM pg 314 - 
_Anti-’Mech Default Skill: Unless the unit has an Anti-’Mech Infantry Kit (see p. 155), all non-mechanized conventional
infantry have a default Anti-’Mech Skill Rating of 8, which cannot be improved._

I'll use the Foot Rifle Platoon
image

We have two versions in MM - one AM one not. But the in game both units can still swarm and leg attack. But I can add the lighter unit (lower one with no AM kit) and set it's AM skill to 2 but it doesn't have the equipment that per TM allows it.

MML associates the equipment as training
image

To me, we should change MML "Anti-Mek Training" to "Anti-Mek Equipped." as this adds the kit and required weight. 

In MegaMek units with the equipment would be allowed to change their Anti-Mek skill, but those without the equipment would be fixed at 8. One potential downside would be a player wanting to have a better AM skill in MM, we'd need to have a way to add it but that would like create an invalid unit.

gsparks3:
Note: if you take an explicitly _non_-anti-mek-equipped platoon and improve its Anti-Mek skill in the lobby, the weight of the platoon will increase - because it seems the MM lobby is smart enough to automatically add the weight for the AM gear if not initially present.

SJuliez:
- There is no anti-mek gear in MM, the weight is calculated on-the-fly by looking at the AM skill, adding the 15kg if it is lower than 8.
- Both the normal and the AM version of a certain infantry behave exactly the same when set to the same AM skill.
- @HammerGS: My understanding of the rules is that regardless of AM gear and AM skill a non-mech (and non-encumbered, see TO) infantry may make an AM attack. At skill 8 (=without the "gear") it'll have a low chance of success. (Otherwise, why force a skill of 8 at all as in TM 314 as opposed to mech inf that are barred from AM attacks?)

So for me the questions are:
- Does skill 8 mean having no AM gear? (the rules state No AM gear => skill 8. This is the other direction. Can the troops with AM gear just be bad at it?). 
- If the answer to the above question is yes (then MM is doing it correctly), can we remove the AM versions of infantry?
- If the answer is no, then we'd have to separate the gear from the skill which would mean more UI or at least keeping the AM versions of units.
- Is my understanding of the rules correct w.r.t. being able to make AM attacks regardless of gear and skill? (Then MM is doing it correctly).

HammerGS:
> Does skill 8 mean having no AM gear? (the rules state No AM gear => skill 8. This is the other direction. Can the troops with AM gear just be bad at it?).

Any skill level can make AM attacks but without the equipment the best you're going to get is Skill 8. 

> If the answer to the above question is yes (then MM is doing it correctly), can we remove the AM versions of infantry?

I seem to recall we added the AM infantry to make it clear some infantry could attack with a better skill. But its not an elegant solution as it leaves us with  a confusing mix of units that imply one can and one can't make AM attacks. I've always felt the best solution is to find a way to show - _This unit is equipped with AM stuff so it's skill can be better than 8. If we have that I'd fully support removing the extra units._ Maybe a help tip? 

I also think the check box in MML should be renamed. 

The other part I don't like is the sudden change of weight in the lobby when the skill goes below 8. Those sudden changes can throw off the units used to transport them. I feel like we should be clear to the player. This is unit is equipped to make AM attacks as such its skill can be better than 8, but weighs X tons. This unit isn't equipped to make AM attacks so the best skill is 8 but it weighs X tons.

pheonixstorm:
One other thing not mentioned is how the cost is calculated for infantry. AM infantry cost 5X more as well due to training. This cost is for a full squad/platoon with troops, which MekHQ does not add when you purchase said squad/platoon. Those you have to hire separate. 

MML uses proper TW cost calculations for AM and non-AM infantry based on that checkbox.

However this goes, there are likely other areas this will affect in possibly weird ways.

SJuliez:
OK this should now be in a working state. The changes only affect BV calculation. I opened issues for the construction stuff.

I realized that the special "treat AM skill as 5" for BV must be used for all infantry that can't make AM attacks which are mechanized, encumbered and field gunners and those don't receive the AM weapon BV bonus. So I had to correct what I had previously done. Crew is written without an entity reference and can't decide what skill to use for infantry. There are several solutions but I opted to remove the BV calculation from Crew as it's not directly TW-gameplay-related, our entity classes can use some slimming and BV is in its own package anyway. So now, infantry that can't make AM attacks still show an AM skill but it has no influence on BV.

NickAragua (approved):
Looks fine; we should probably flag this for some QA attention due to the widespread nature of the changes.

PR #3840, opened 2022-08-07
BV calculation report for BA

SJuliez:
Adds a BV calculation report for BattleArmor. The calculation is done per Trooper (I guess to be able to account for damage), so the report lists every trooper even if there's no damage and every trooper is the same.

image
image

### DIFF
--- a/megamek/src/megamek/client/ui/swing/calculationReport/CalculationReport.java
+++ b/megamek/src/megamek/client/ui/swing/calculationReport/CalculationReport.java
@@ -203,4 +203,21 @@ public interface CalculationReport {
         return ((resultPrefix != null) ? resultPrefix : "")
                 + String.format(Locale.US, "%1$,.1f", result);
     }
+
+    /**
+     * Formats the given double with only the necessary digits and at most three digits. Uses the fixed Locale.US
+     * as the Java way of converting "" + value seems to use Locale.US by default as well.
+     */
+    static String fmt(double d) {
+        long timesThousand = Math.round(1000 * d);
+        if (timesThousand % 1000 == 0) {
+            return String.format(Locale.US, "%1$.0f", d);
+        } else if (timesThousand % 100 == 0) {
+            return String.format(Locale.US, "%1$.1f", d);
+        } else if (timesThousand % 10 == 0) {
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/client/ui/swing/calculationReport/CalculationReport.java
```diff
@@ -203,4 +203,21 @@ private String getRoundedResultString(String resultPrefix, double result) {
         return ((resultPrefix != null) ? resultPrefix : "")
                 + String.format(Locale.US, "%1$,.1f", result);
     }
+
+    /**
+     * Formats the given double with only the necessary digits and at most three digits. Uses the fixed Locale.US
+     * as the Java way of converting "" + value seems to use Locale.US by default as well.
+     */
+    static String fmt(double d) {
```
NickAragua, line 211: Can we name this a little more descriptively? "fmt" doesn't really tell me anything about what the method does.

NickAragua (changes requested):
Just one request.

Also, we really need to de-couple the BV report and the actual calculation - having all this string concatenation in there can't possibly be good for performance. But that's a task for another day.

megamek/src/megamek/client/ui/swing/calculationReport/CalculationReport.java
SJuliez: Invited that change request with such a name :). There's these lines:
bvReport.addLine("Offensive Battle Rating:",
                    fmt(oBV) + " x " + fmt(speedFactor), "= " + fmt(oBV * speedFactor));
where the method is used three times. Making it "formatDoubleForReport" or "formatReportThreeDigits" or something like that would make the report calls very long and I guess rather less readable. I'm all for descriptive names but had hoped to keep this one short and let people mouse hover it to see what it does. (I couldn't think of a good way to have the report do the formatting by itself). Call it format instead?

SJuliez: OK changed the name and made a few small corrections. I couldn't think of any name that was more descriptive and not ridiculously long.

NickAragua (approved):
Sorry that took so long, completely spaced out on it.

PR #3882, opened 2022-09-07
Base Components: DefaultScrollablePanel

Windchild292:
This creates a new base component that I've found very useful in MHQ.


SJuliez (approved):
I can't say that I would know how to use this.
(Just kidding, I would look in MHQ if I needed an example :-) )

PR #3888, opened 2022-09-09
Update BSK .mtfs and Quirks

McStarley:
A small proposed correction to the current Bull Shark implementations. This moves the Bull Shark(s) out of MWO Canon and into the Unofficial folder, as they do not appear in MWO and are apocryphal, only appearing in the HBS video game. The designs are changed to more closely match those originally seen in-game, along with adding some fluff for both. Also gives the chassis the Distracting quirk.

In addition, cleaned the canonUnitQuirks list slightly, moving a small number of units to their proper sections and alphabetising them.

Please review and advise any corrections needed or steps otherwise missed.

### DIFF
--- a/megamek/docs/history.txt
+++ b/megamek/docs/history.txt
@@ -3,7 +3,9 @@ VERSION HISTORY:
 0.49.10-SNAPSHOT
 + Data: Shorten RAT names for FM3145
 + PR #3887 - Excised Rumble Seat and move to Legacy quirk.
-+ PR #3886 - Simplify handling of submunition names 
++ PR #3886 - Simplify handling of submunition names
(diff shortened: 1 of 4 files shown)


HammerGS:
Thank for the PR but a couple of things around the Bullshark. We will move soon making the unofficial and unsupported folders and optional download. The Bullshark is popular unit in MegaMek and I would like to keep it in the Mechs folder. 

But the folder should be renamed to accurately represent its source. Something I can handle on my end. 

I would like to retain the fluff you've done for it. Can you split this into two PR's one for the Bullshark, leaving it in the folder but with the fluff. Then a separate PR for the quirk fixes.

HammerGS (changes requested):
See comments in top of thread.

Issue #3776, opened 2022-07-08
[RFE] Add 0 (zero) as a selectable weapon/equipment quantity in Open unit, advanced search


rickremer:
MegaMek & MegaMekLab v 0.48.0
(not going thru MMHQ)
Windows 10 Home
Java v 8 upd 311

![Screenshot 2022-07-08 083024](https://user-images.githubusercontent.com/47697445/178001974-36952838-507d-4a44-9390-5bc69f30259c.png)
[screencap shows 'Mech Mortar 4 with no 'zero' at top of quantity pulldown]

Please add 0 (zero) as a selectable weapon/equipment quantity in Open unit, advanced search.  For example, one might wish to look for units with no SRM type weapons, criteria might be "0 SRM 6 And 0 SRM 4 And 0 SRM 2".

Seosaidh:
I would like to second this feature.

Seosaidh:
PR #3901 opened.

SJuliez:
Implemented. Therefore closing this.

PR #3901, opened 2022-09-14
Issue #3776 [RFE] Add 0 (zero) as a selectable weapon/equipment quantity in Open unit, advanced search

Seosaidh:
The evaluate function (MechSearchFilter.java) was changed to handle 0 as a valid quantity. Additionally, the gui was modified to allow 0 as a selection in the drop-down (AdvancedSearchDialog.java).

Specifically, 0 was added to the available options for the weapon and unit drop-downs.
In the unit filter. If the name of the weapon/equipment in the mech summary matches that of a node with a requested quantity of 0, then the unit is not matched. If the name matches and the requested quantity is greater than 0 and the unit quantity is greater than or equal to the requested quantity, the unit matches.
If no equipment is found where the name matches the requested weapon/equipment, and the requested quantity is 0, then the unit matches. Otherwise it does not.

### DIFF
--- a/megamek/docs/history.txt
+++ b/megamek/docs/history.txt
@@ -1,6 +1,5 @@
 VERSION HISTORY:
 ----------------
-+ Issue #3776 - Add ability to filter out mechs with specified equipment.
 0.49.10 (2022-09-12 1500 UTC)
 + Data: Shorten RAT names for FM3145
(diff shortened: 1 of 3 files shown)


HammerGS:
Thanks for the PR. Can you remove History.txt as we manage that if the PR is accepted.

Issue #3903, opened 2022-09-15
Ahab AHB-X missing rear-mounted medium lasers 0.49.9


BLR-IIC:
I was just modifying an Ahab AHB-X in MML and noticed that it was 2 tons underweight.  It's 2 tons underweight because it's missing its rear-mounted medium lasers.

HammerGS:
The problem is we have to build of the RS, the RS doesn't have them. The TRO entry doesn't list them and Sarna has a cites the old TRO 2750. I've had it underweight with an [unresolved errata ][(url](https://bg.battletech.com/forums/errata/record-sheets-3050-upgrade-unabridged/msg1771278/#msg1771278))

If you have a current canon reference to the 2 laser please let me know.

BLR-IIC:
I've got nothing then.  I was just going off Sarna not owning the actual record sheets or TROs this variant was featured in.  Looks like it's going to be awhile before any of that errata gets resolved!

HammerGS:
This matches my research. Lots of these unit things that come I’ve caught and asked about over the years. If it’s something in the canon folders it’s good to ask before putting it tons of work.

PR #3904, opened 2022-09-15
Fix AHB-X missing weapons

McStarley:
Regarding issue #3903 
Add two missing medium lasers to aft weapons bay, as described in TRO 3050U.

(Is pull-requesting for this too excessive? It's such a minuscule change...)


HammerGS:
For stuff like this it's better to open an issue, attach the file and I'll add it.

PR #3925, opened 2022-10-09
Correctly load damaged CI from MUL

SJuliez:
Applies a reduced trooper count from a .MUL to both the "men" and "shootingMen" fields so that the CI really knows how strong it is (when loading a damaged CI and then choosing "edit damage" in the lobby, this would show the wrong value.)

### DIFF
--- a/megamek/src/megamek/common/MULParser.java
+++ b/megamek/src/megamek/common/MULParser.java
@@ -1449,15 +1449,14 @@ public class MULParser {
                 }
             } else if (type.equals(INTERNAL)) {
                 if (entity.getOInternal(loc) < pointsVal) {
-                    warning.append("The entity, ")
-                            .append(entity.getShortName())
-                            .append(" does not start with ")
-                            .append(pointsVal)
-                            .append(" points of internal structure for " +
-                                    "location: ")
+                    warning.append("The entity, ").append(entity.getShortName()).append(" does not start with ")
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/common/MULParser.java
```diff
@@ -1449,15 +1449,14 @@ private void parseArmor(Element armorTag, Entity entity, int loc) {
...
-                            .append(" does not start with ")
-                            .append(pointsVal)
-                            .append(" points of internal structure for " +
-                                    "location: ")
+                    warning.append("The entity, ").append(entity.getShortName()).append(" does not start with ")
+                            .append(pointsVal).append(" points of internal structure for location: ")
                             .append(loc).append(".\n");
                 } else {
                     entity.setInternal(pointsVal, loc);
+                    if (entity.isConventionalInfantry()) {
```
neoancient, line 1457: This needs to happen for BattleArmor also.

neoancient (changes requested):

SJuliez:
OK for BA internal damaged = destroyed and the applydamage needs to happen at another place in the MUL Parser. But I've used the instanceof check in all places to make it happen for both BA and CI. I also added a change to the lobby damage editor dialog so that lobby damage arrives correctly.

neoancient (approved):

PR #3935, opened 2022-10-13
Destroy Infantry field guns and artillery with trooper damage

SJuliez:
TO:AUE p.123 notes that field guns count as destroyed when they can no longer be crewed by the platoon. Despite the different rules text, I assume this extends  to field arty. Per discussion with Hammer, the rules team says they're dead even for purposes of MHQ. Therefore with this PR field guns and arty get destroyed with trooper damage. As a result, some code that dealt with trooper damage and crew requirements is removed (FireControl and AttackAction).


pheonixstorm:
While this is open I have a question. How does this affect the unit after it is imported back into HQ? MM sends back the crew with injuries and destroyed field guns. Will those guns need to be repaired, replaced, or just have the dead crew replaced to be functional?

NickAragua (approved):

Issue #3938, opened 2022-10-14
RFE: Add Ability to Filter Via Weapon Groups/Types/Classes (i.e. All Mechs With Any Type of LRM)


Seosaidh:
### Environment
N/A

### Description
It would be nice to be able to filter out units based on entire classes of weapons in the advanced filter screen instead of having to select every single instance of a weapon. For example, if I wanted to get all units that mount LRMs, and I'm playing advanced tech in 3080, there are a _lot_ of weapons from the weapons table I have to add to the filter. It would be nice to just have to add one item.

Seosaidh:
This was attempted in PR #3932. However, the approach taken there mistakenly used the Aero weapon class, which doesn't correspond to what the average person would understand for weapon types in non-aero units. The main example being that gauss weapons are classified as auto-cannons, while the only "gauss weapon" is the gauss bay.

SJuliez:
I believe this function is present.

PR #3942, opened 2022-10-18
#3938 Implement Weapon Type/Group/Class Filter

Seosaidh:
Add table of weapon types/classes such as "laser", "energy", "rac", etc. and allow filtering on these groups.

The previous attempt (PR #3932) mistakenly used the aero weapon class table. This implementation uses the names of the weapons with a contains comparison instead of an equals comparison. For example. If you choose to filter on all mechs with at least one laser, then a match occurs if "laser" is found in the equipment name, even if there isn't an exact match.

The allowed weapon "classes" are laid out in a table on the advanced search screen, and add to the general filter tokens similar to the weapons and equipment tables. One major difference is that they weapon class table doesn't allow filtering the table entries by tech/year, as with a handful of exceptions all weapon classes are available at all tech levels/years.


NickAragua (approved):
Looks good code-wise. Perhaps someone can pull down this build artifact and test it?

HammerGS:
Been testing this and haven't found any more oddities. I'd say it's ready for final review and merging.

megamek/src/megamek/client/ui/swing/unitSelector/TWAdvancedSearchPanel.java
```diff
@@ -1563,6 +1566,7 @@ public boolean matches(String name) {
                     WeaponClass.MACHINE_GUN.matches(name);
             } else if (this == RAC) {
                 if (name.toLowerCase().contains("rotary")) {
+                    System.err.println(name);
```
NickAragua, line 1569: We try to avoid System.err.println logging - I would use LogManager.getLogger() if you really need to log this. Otherwise, let's just get rid of it.

NickAragua (changes requested):
One minor adjustment.

megamek/src/megamek/client/ui/swing/unitSelector/TWAdvancedSearchPanel.java
Seosaidh: *facepalm* I meant to remove that.

Seosaidh: I removed in the latest commit.

NickAragua (approved):
Good to go.

PR #3944, opened 2022-10-21
Infantry refactor

SJuliez:
When there, why not improve. 

Removes unnecessary comments (for overrides mostly), improves a few method and field names, makes use of methods and improves the way some things are represented in the code.

The real changes are in Infantry, the changes in the other files adapt to changed method names.
This should be merged together with the adaptations for MML and MHQ.


megamek/src/megamek/common/Infantry.java
```diff
@@ -114,26 +90,19 @@ public class Infantry extends Entity {
...
-     */
     public static final int LOC_INFANTRY = 0;
     public static final int LOC_FIELD_GUNS = 1;
 
-    /**
-     * Infantry only have critical slots for field gun ammo
-     */
+    // Infantry only have critical slots for field gun ammo
     private static final int[] NUM_OF_SLOTS = { 20, 20 };
     private static final String[] LOCATION_ABBRS = { "MEN", "FGUN" };
```
NickAragua, line 107: Since we're polishing this, can we change the "MEN" location to "troopers" or "soldiers" or something?

NickAragua (commented):
I think I'm ok with this for the most part. We'll definitely want to give infantry a good beating in terms of testing after we merge it.

megamek/src/megamek/common/Infantry.java
HammerGS: I like troopers myself. But agree with the men part.

SJuliez: Sure, done. I was afraid this would break things but at least loading a damaged CI from a MUL is compatible across this change.

NickAragua (approved):
Yeah, I think locations might be determined by index rather than name. Although I could be wrong, it's been more than five minutes since I last looked at a MUL file.

Issue #3945, opened 2022-10-21
RFE: update Unit Display GUI to allow a tabbed and non tabbed versions


kuronekochomusuke:
Description

update the Unit Display GUI to allow it to be switchable between the current version with tabs and a non tabbed version that shows all the tabs at once.   

this will allow users with small monitors to continue to use the current tabbed display.  users with large or dual monitors can expand the display to see all the tabs at once.

tabbed version
![image](https://user-images.githubusercontent.com/116095479/197251446-9d12b5c4-04e9-4b4d-afce-b893bd8fc052.png)

non tabbed version
![image](https://user-images.githubusercontent.com/116095479/197251586-8f1fc8ae-f644-43ee-b021-46f046908d26.png)

the "switch view" button at bottom to toggle between the tabbed and non tabbed view of the Unit Display

PR #3946, opened 2022-10-21
#3945 RFE: update Unit Display GUI to allow a tabbed and non tabbed v…

kuronekochomusuke:
…ersions

implements RFE #3945

-  added a button that switches the panel layout type when clicked
-  when switching to the CardLayout remove the child panels and readd so that the names are set correctly.  show the tabStrip
-  when switching to the GridLayout set the columns, rows and make all the tabs visible.  hide the tabStrip used by the CardLayout
-  added logic to do show panel when the layout is CardLayout

### DIFF
--- a/megamek/src/megamek/client/ui/swing/GUIPreferences.java
+++ b/megamek/src/megamek/client/ui/swing/GUIPreferences.java
@@ -606,23 +606,23 @@ public class GUIPreferences extends PreferenceStoreProxy {
         return store.getBoolean(DISPLAY_START_TABBED);
     }
 
-    public int getDisplaSplitABCLoc() {
+    public int getDisplaySplitABCLoc() {
         return store.getInt(DISPLAY_SPLIT_ABC_LOC);
     }
 
-    public int getDisplaSplitBCLoc() {
+    public int getDisplaySplitBCLoc() {
         return store.getInt(DISPLAY_SPLIT_BC_LOC);
     }
 
-    public int getDisplaSplitA1Loc() {
+    public int getDisplaySplitA1Loc() {
         return store.getInt(DISPLAY_SPLIT_A1_LOC);
     }
 
-    public int getDisplaSplitB1Loc() {
+    public int getDisplaySplitB1Loc() {
         return store.getInt(DISPLAY_SPLIT_B1_LOC);
     }
 
-    public int getDisplaSplitC1Loc() {
+    public int getDisplaySplitC1Loc() {
         return store.getInt(DISPLAY_SPLIT_C1_LOC);
     }
 
--- a/megamek/src/megamek/client/ui/swing/GUIPreferences.java
+++ b/megamek/src/megamek/client/ui/swing/GUIPreferences.java
@@ -125,8 +125,8 @@ public class GUIPreferences extends PreferenceStoreProxy {
     public static final String DISPLAY_START_TABBED = "DisplayStartTabbed";
     public static final String DISPLAY_SPLIT_ABC_LOC = "DisplaySplitABCLoc";
     public static final String DISPLAY_SPLIT_BC_LOC = "DisplaySplitBCLoc";
+    public static final String DISPLAY_SPLIT_A1_LOC = "DisplaySplitA1Loc";
     public static final String DISPLAY_SPLIT_B1_LOC = "DisplaySplitB1Loc";
-    public static final String DISPLAY_SPLIT_B2_LOC = "DisplaySplitB2Loc";
     public static final String DISPLAY_SPLIT_C1_LOC = "DisplaySplitC2Loc";
     public static final String DISPLAY_SIZE_HEIGHT = "DisplaySizeHeight";
     public static final String DISPLAY_SIZE_WIDTH = "DisplaySizeWidth";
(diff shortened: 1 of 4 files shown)


HammerGS:
Gave it a test and everything seems to work. But one suggestion. 

image

Would it be possible to be able to drop and adjust the red lines? I was finding that I needed to drag the window larger to see the slot information rather than just drag the bar over. 

Also wonder how the panel would work if the top left (General) has lots of information to display.

kuronekochomusuke:
i can look into adding dividers between the panels, to see how hard it would be.

some of the sub panels have issues displaying a lot of information and they probably need to be updated to correct some of the issues.

kuronekochomusuke:
how about like this?  the splitters make it a little clunky, but it allows you to adjust the sizes to fit the area you have.

image

kuronekochomusuke:
- updated to use split panes and adjusted the location of the panels in the non tabbed version
- removed code in PilotPanel that was causing it to be very wide and causing issues with the split panes
- added a preference that remembers if you were using tabbed or non tabbed, then starts up with the same version

HammerGS:
Been playing with this today and I really like it. Almost gives a control panel feel to things.

HammerGS:
On more thought. The window positions persist between switching, but could we get them to persist between games (specifically restarting a new game) like lots of our other window's options do?

Great job with this!

kuronekochomusuke:
- save splitter locations to preferences and use them on load

HammerGS:
Thanks for the save locations. The more I use the more I'm wondering about the Pilot panel. 

image

With the exception of Command and toughness the date is duplicated on the General tab.  This might be beyond the scope of what you want to do. But I wonder if the Command and Toughness got added to the General tab, then when you go to full mode. You could not show the Pilot tab, giving more space to spread the other tabs. 

Or even just not show the pilot tab.

Any thoughts or opinions on that?

kuronekochomusuke:
the Pilot tab is also used switch crew around when you have more than one crew, so i don't think we can skip it   

on Pilot tab there is also also a lot of hidden labels on it, that get shown when they have the attribute.   i had to add a scroll pane to it in order it get it to work okay.

I was wondering about the pilot tab, if all those labels on it should be converted to display the info more like the summary  tab.   the way it is now does not play well with the splitter panes.   

also I think i have seen an issue with the summary tab, when it has a lot if info to display.  so they both could use some work.

I think updating the pilot tab and the summary tab should be new RFEs or issues, if they are going to be updated.

gsparks3:
Yeah, doesn't the Pilot tab also display e.g. SPAs, Edge settings, etc?

pakfront:
Yes. The recent changes I made duplicated a lot of the usefulpilot info the
general(on purpose) I think you could get rid of pilot - or use it to show
all crew members for multi-crew

kuronekochomusuke:
does the Swap roles button on the pilot panel do anything?  looks like it just sets the text on a button

do you need to know the commanders or crews info while in a battle?  the pilot panel looks like the only place to see the commanders or crews info.

here is the what the battlemaster  looks like with the command console.  
image

just trying to see if the pilot panel is adding any value or not.

what is a unit that has multiple crew, but does not have a command console?  that is a case i have not looked at.

HammerGS:
I might do something for Tripods. But I think someone would need to double check.

kuronekochomusuke:
this is the code for the Swap roles button.  would some other code be checking the text value on the button and then doing something based on it?

    private void updateSwapButtonText() {
        if (btnSwapRoles.isSelected()) {
            btnSwapRoles.setText(Messages.getString("PilotMapSet.keepRoles.text"));
        } else {
            btnSwapRoles.setText(Messages.getString("PilotMapSet.swapRoles.text"));
        }        
    }

HammerGS:
This was the PR that implemented the Multi-Crew units https://github.com/MegaMek/megamek/pull/555/

I think some testing is needed to determine the impacts.

kuronekochomusuke:
- rearranged the panels by size,
- moved pilot panel to first column, think the resizing works better with a 2x2x2  layout vs the 1x3x2 layout
- turned on one touch expand, so can easily hide the pilot panel or other ones

image

HammerGS:
This latest iteration is really good. I'm going to tease it to the community to get some player feedback.

megamek/src/megamek/client/ui/swing/unitDisplay/UnitDisplay.java
```diff
@@ -68,41 +83,197 @@ public UnitDisplay(@Nullable ClientGUI clientgui,
...
+        displayP.add(splitABC);
+
+        GUIPreferences.getInstance().setDisplayStartTabbed(false);
+    }
+
+    /*
+    * Save splitter locations to preferences
+    *
+    */
+    public void saveSpliterLoc() {
```
NickAragua, line 271: Hate to be that guy, but can we correct the method's spelling to "saveSplitterLoc"?

megamek/src/megamek/client/ui/swing/GUIPreferences.java
```diff
@@ -590,6 +602,30 @@ public int getDisplayPosY() {
         return store.getInt(DISPLAY_POS_Y);
     }
 
+    public boolean getDisplayStartTabbed() {
+        return store.getBoolean(DISPLAY_START_TABBED);
+    }
+
+    public int getDisplaSplitABCLoc() {
```
NickAragua, line 609: Think this and the other "Displa" should probably be "Display".

NickAragua (changes requested):
I'm happy with it code-wise, just needs a little polishing.

NickAragua (approved):

HammerGS:
I’ve been using these and really like them. But I have 43” Widescreen. Haven’t had any major issues with the recent improve.

kuronekochomusuke:
the only other thing i can think of adding is a Client Settings for the panel order, like the Button Order.   just not sure how much it would be used.

image

yall think this would be helpful?

HammerGS:
> the only other thing i can think of adding is a Client Settings for the panel order, like the Button Order. just not sure how much it would be used.
> 
> image
> 
> yall think this would be helpful?

I think this would be a good option, and if possible, include the option to not show a panel.

PR #3954, opened 2022-11-02
AlphaStrike conversion changes

SJuliez:
* Proto gliders no longer get a 2" base movement (only their 8"g e.g.) (CGL requested to-be-errata change)
* Fixes a bug where the OV value was not computed using S damage when M damage is 0


NickAragua (approved):
Looks good.

Issue #426, opened 2016-11-25
New Drop Down Option: Display acitve weather conditions and game effects


VampireSeraphin:
Please add a new menu similar to "show round report" that displays active weather conditions and their in-game effects. This will be especially pressing as Phoenix finishes up his weather rework.

HammerGS:
I would recommend @pheonixstorm look into this as part of his weather project stuff.

pheonixstorm:
I think I can add that in. Best to have @arlith decide on how best to do this though. I can see it done with the round report or as a separate weather dialog. I had planned on looking into adding the temperature to the round report already as it is not displayed and has a greater impact on the game in extreme cases.

Once I get a little guidance from arlith on how best to implement this I will get it added (I hope, still a bit confudled about how to build extra bits into the round reports)

VampireSeraphin:
I'd prefer to see it as a separate drop down, which you can fill with temperature, visibility, planetary conditions like gravity, and the current weather + the values of the weather modifiers. This would keep the round report from getting more cluttered and make it easier to upgrade later if need be.

arlith:
The round reports are all created in server, having creating new `Report` objects and adding them to the current `Vector<Report>`.  Can't you already see most of the weather conditions from the init report, which you can go back and view with the round report menu?

Regardless, you could potentially do something like the player list, where there's a menu option that pops up a small dialog.  I'm not a big fan of the player this though.  Can't really think of much else.  I feel like a complete report in the init phase is probably sufficient.  Otherwise, there's not much of a great way to communicate info outside of tooltips and the unitDisplay, unless you add a new dialog.

kuronekochomusuke:
implemented in #3969

PR #3969, opened 2022-11-07
RFE 426 overlay for Planetary Conditions

kuronekochomusuke:
RFE #426 overlay for Planetary Conditions

cloned the KeyBindingsOverlay to display Planetary Conditions

Are there any other conditions that would be good to add to this?

image


SJuliez:
Also: I remember that people had difficulty with this initially because the keybind didn't work properly. iirc you'd have to add the keybind default to the keybinds.xml file that is somewhere in the config folder.

SJuliez:
Thirdly :) you might wanna make sure that it doesnt show up in the board editor.

kuronekochomusuke:
- colors can be set in advance options
- added condition indicators
- change temperature color based on extreme temperature ranges
- align to right side

image

SJuliez:
> Also: I remember that people had difficulty with this initially because the keybind didn't work properly. iirc you'd have to add the keybind default to the keybinds.xml file that is somewhere in the config folder.

Sorry if I was not making sense: I was speaking of the key bind overlay. But the same is true for the new overlay.

kuronekochomusuke:
I tried updating the keybinds.xml for  this and added in null checks for the  board editor

kuronekochomusuke:
- add advance option to hide default conditions

kuronekochomusuke:
- scale overlay based on the GUI scale

image

image

HammerGS:
Looking good!

kuronekochomusuke:
- add advance setting to show the header, labels, values, and indicators

image

indicatorScreenshot 2022-11-11 101335

kuronekochomusuke:
This is now ready for review

megamek/src/megamek/client/ui/swing/boardview/PlanetaryConditionsOverlay.java
```diff
@@ -0,0 +1,405 @@
...
+        @Override
+        public void gameTurnChange(GameTurnChangeEvent e) {
+            // The active player has changed
+            changed = true;
+        }
+    };
+
+    @Override
+    public void preferenceChange(PreferenceChangeEvent e) {
+        if (e.getName().equals(KeyBindParser.KEYBINDS_CHANGED)) {
```
SJuliez, line 397: I believe it's not hurtful to just set changed to true here without doing any checks (PreferenceChangeEvents don't happen all the time); this way it'll react to changes to its advanced settings.

SJuliez (changes requested):
I like this a lot! Just one small request then I'll merge it.

SJuliez (approved):

PR #3991, opened 2022-11-18
Fixes #3727 Magma crust breakthrough and liquid magma damage not working as expected

### LINKED ISSUE #3727: [49.7] Magma crust breakthrough and liquid magma damage not working as expected
Bundled Megamek with MekHQ 0.49.7

[save1.sav.gz](https://github.com/MegaMek/megamek/files/8959816/save1.sav.gz)
image

image

image

**A)** MM seems to check for breaking through a magma crust when turning in that hex, when it should only check upon entering the hex.

[save2.sav.gz](https://github.com/MegaMek/megamek/files/8959874/save2.sav.gz)

image

**B)** No damage is being applied for breaking through the crust until leaving the hex the following turn. If the crust breaks and converts to liquid magma the exposed locations should take damage immediately in the movement phase where that occurs. 

The wording of the rule (TacOps advanced ground movement) is that damage should be caused when starting a movement phase in liquid magma or if entering the hex. I guess this might be an issue with the order of movement and determining when the hex becomes magma - strictly the unit enters a crust hex first, which then changes. 

**C)** Also the additional second damage amount is not being applied for starting and ending the movement phase in liquid magma.

pheonixstorm:
From the original report, A is fixed. Magma crust breaking will only get checked per hex, not per moveStep. The popup showed the correct number but processMovement did not know this rule. In fact in testing this, this has been a bug since at least 0.47 and possibly before.

Per rules on p34/35 TO:AR damage should be applied if a mech starts its turn in magma, walks into a magma hex, or breaks a crust hex. Additionally, if a unit starts and ends its turn in a magma hex it should take additional damage.

TODO: units starting and ending their movement in a magma hex should take additional damage
units walking into a magma hex should not make a bog down check. RAW only mention jumping into the hex needing a bog down roll.

### DIFF
--- a/megamek/src/megamek/server/ServerHelper.java
+++ b/megamek/src/megamek/server/ServerHelper.java
@@ -445,9 +445,21 @@ public class ServerHelper {
                 hex.addTerrain(new Terrain(Terrains.MAGMA, 2));
                 gameManager.sendChangedHex(curPos);
                 for (Entity en : entity.getGame().getEntitiesVector(curPos)) {
-                    gameManager.doMagmaDamage(en, false);
+                    if (en != entity) {
+                        gameManager.doMagmaDamage(en, false);
+                    }
                 }
             }
+        }
+    }
+
+    /**
+     * Check for movement into magma hex and apply damage. TODO: If first and last step are in magma double it.
+     */
+    public static void checkEnteringMagma(Hex hex, int elevation, Entity entity, Coords curPos, GameManager gameManager) {
+
+        if ((hex.terrainLevel(Terrains.MAGMA) == 2) && (elevation == 0) && (entity.getMovementMode() != EntityMovementMode.HOVER)) {
+            gameManager.doMagmaDamage(entity, false);
         }
     }
 
--- a/megamek/src/megamek/server/ServerHelper.java
+++ b/megamek/src/megamek/server/ServerHelper.java
@@ -454,7 +454,7 @@ public class ServerHelper {
     }
 
     /**
-     * Check for movement into magma hex and apply damage. TODO: If first and last step are in magma double it.
+     * Check for movement into magma hex and apply damage.
      */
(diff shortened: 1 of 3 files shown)


megamek/src/megamek/server/ServerHelper.java
```diff
@@ -445,12 +445,24 @@
...
+                    }
                 }
             }
         }
     }
 
+    /**
+     * Check for movement into magma hex and apply damage. TODO: If first and last step are in magma double it.
+     */
+    public static void checkEnteringMagma(Hex hex, int elevation, Entity entity, Coords curPos, GameManager gameManager) {
```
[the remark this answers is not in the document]
pheonixstorm: removed

pheonixstorm:
The bugs have been fixed. I will split the Bog down changes into its own PR once some of the other coders are around to help me understand the code better.

This can be reviewed and approved at any time now.

megamek/src/megamek/server/ServerHelper.java
```diff
@@ -445,12 +445,24 @@ public static void checkAndApplyMagmaCrust(Hex hex, int elevation, Entity entity
...
             }
         }
     }
 
+    /**
+     * Check for movement into magma hex and apply damage. TODO: If first and last step are in magma double it.
+     */
+    public static void checkEnteringMagma(Hex hex, int elevation, Entity entity, GameManager gameManager) {
+
+        if ((hex.terrainLevel(Terrains.MAGMA) == 2) && (elevation == 0) && (entity.getMovementMode() != EntityMovementMode.HOVER)) {
```
NickAragua, line 461: Does this cause airborne aerospace units to take magma damage?

NickAragua (commented):
Looks fine, one question.

megamek/src/megamek/server/ServerHelper.java
pheonixstorm: No, the code is a near copy of the crust check code at line 431. I also tested it.

NickAragua (approved):

PR #4000, opened 2022-11-23
Fixes #3818 Airborne Aerospace Units Unaffected By Extreme Temperatures

### LINKED ISSUE #3818: Airborne Aerospace Units Unaffected By Extreme Temperatures
### Environment
MegaMek v0.49.8

### Description
Airborne aerospace units do not have their heat modified by extreme temperatures (less than 0C or more than 50C).

pheonixstorm:
Added code to resolveAeroHeat to account for extreme temps. Tested and works with ASF, SC, and DS. Would probably work with JS and WS if they could survive entering an atmosphere...

### DIFF
--- a/megamek/src/megamek/server/ServerHelper.java
+++ b/megamek/src/megamek/server/ServerHelper.java
@@ -95,26 +95,9 @@ public class ServerHelper {
             r.subject = entity.getId();
             vPhaseReport.add(r);
         }
-        
+
         // Add or subtract heat due to extreme temperatures TO:AR p60
-        if ((game.getPlanetaryConditions().getTemperatureDifference(50, -30) != 0) && !game.getBoard().inSpace()) {
-            if (game.getPlanetaryConditions().getTemperature() > 50) {
-                int heatToAdd = game.getPlanetaryConditions()
-                        .getTemperatureDifference(50, -30);
-                entity.heatFromExternal += heatToAdd;
-                r = new Report(5020);
-                r.subject = entity.getId();
-                r.add(heatToAdd);
-                vPhaseReport.add(r);
-            } else {
-                entity.heatFromExternal -= game.getPlanetaryConditions()
-                        .getTemperatureDifference(50, -30);
-                r = new Report(5025);
-                r.subject = entity.getId();
-                r.add(game.getPlanetaryConditions().getTemperatureDifference(50, -30));
-                vPhaseReport.add(r);
-            }
-        }
+        adjustHeatExtremeTemp(game, entity, vPhaseReport);
 
         // Combat computers help manage heat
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/server/ServerHelper.java
```diff
@@ -95,6 +95,26 @@ public static void resolveAeroHeat(Game game, Entity entity, Vector<Report> vPha
             r.subject = entity.getId();
             vPhaseReport.add(r);
         }
+        
+        // Add or subtract heat due to extreme temperatures TO:AR p60
+        if (game.getPlanetaryConditions().getTemperatureDifference(50, -30) != 0) {
```
NickAragua, line 100: I'd suggest not doing this if the game board is in space.

NickAragua: Also, I feel like we can just call getTemperatureDifference once instead of four times. Internally, that function is one giant, inefficient WTF, so I'd rather minimize the number of times it's called.

NickAragua: Finally, we need to make sure we're not doing any of this if the temp is within "normal" range. Currently, it looks like it'll issue a report even in "normal" temp range, which we want to avoid so that we don't clutter things up any more than we already do.

NickAragua (changes requested):

megamek/src/megamek/server/ServerHelper.java
pheonixstorm: Blame whoever wrote the original code for the mess please lol, this section was a modified copy paste from GameManager line 18481 (Mech heat in extreme temps)

As to questions in order
1. That shouldn't be a problem, have any idea where the variable is set if the map is in space or not? Would help speed up that fix
2. As with my opening statement, this is the original code slightly modified. I can _try_ to clean it up a little.
3. If it doesn't for mechs it shouldn't for aero. I did test to be sure and it does not generate a report in the Heat Phase.

pheonixstorm: Nevermind on one, it was easier to find than I thought. That one is now fixed.

pheonixstorm:
Forgot to push the space map fix.

megamek/src/megamek/server/ServerHelper.java
NickAragua: I'm not asking you to clean up the getTemperatureDifference function, I'm asking you to call it once and store the results. Right now, you're calling it between three and four times each time through the logic block that you added, whereas you can get away with just once.

pheonixstorm: Ok, misunderstood what you were asking for then. Will have it done tonight, both aero and original mech code.

NickAragua (approved):

Issue #3054, opened 2021-08-06
RFE - TO Infantry movement on pavement


gsparks3:
Total Warfare p. 61 describes a MP bonus for ground vehicles moving on pavement. This is explicitly noted as not affecting motorized or mechanized infantry. However, Tactical Operations: Advanced Rules p. 25 describes an optional rule that allows this bonus MP effect to apply to motorized and mechanized infantry.

Megamek does not currently implement this optional rule; this RFE requests that it be added.

Windchild292:
Added in #4005

PR #4005, opened 2022-11-24
Completes #3054 TO Infantry movement on pavement

pheonixstorm:
Added optional TacOps rule Movement on Pavement p25
Infantry will now gain move bonus on pavement same as vehicles if this option is checked.


megamek/src/megamek/common/Infantry.java
```diff
@@ -1497,6 +1497,16 @@ public boolean isSquad() {
         return squadCount == 1;
     }
 
+    @Override
+    public boolean isEligibleForPavementBonus() {
+        if ((game != null)
```
NickAragua, line 1502: Is this supposed to apply to battle armor too?

NickAragua (commented):
Looks fine, one question.

megamek/src/megamek/common/Infantry.java
pheonixstorm: no, it only applies to wheeled, tracked, and motorized infantry. Rule first sentence ``If mechanized and motorized infantry...``

Someone on discord did ask if this and the main TW rule should apply to hover as well. I wouldn't think so, but it _IS_ a ground vehicle. Currently MM doesn't allow for a hover CV/SV to get the pavement bonus so that would require a rules question to sort out.

HammerGS: Hover should get it. TW pg 22 defines ground vehicles and hover is included.

pheonixstorm: Ok, that will be also fix a longstanding unknown bug in Tank.java then. It does not have hover listed.

NickAragua (approved):

PR #3983, opened 2022-11-13
RFE 3982 Scale the Client GUI using the GUI scale

kuronekochomusuke:
RFE #3982 Scale the Client GUI using the GUI scale

scale the buttons, status text, panel heading, etc. on the main Client GUI using the GUI scale

image

PR #4011, opened 2022-11-29
fix missing toHitText

kuronekochomusuke:
correct issue caused by  #3983 RFE 3982 Scale the Client GUI using the GUI scale

the text area at bottom was missed, causing it to not show.

image


kuronekochomusuke:
correct issue with defaulting the user search combo box to the current player.

pheonixstorm:
Easier to follow when each commit has a more unique message.

corrected issue with defaulting the user search combo box to the current player.  <~~ would make a better commit message for the second.

NickAragua (approved):

Issue #4013, opened 2022-12-01
NPE turn timer continues to run when you close the main window


kuronekochomusuke:
0.49.10

NPE turn timer continues to run when you close the main window

error happens if you are on the movement or firing phase with a active turn timer, then close the main window.  And then you start a new game the turn timer is still running and will get a NPE error once  the timer runs out.   

need to stop the timer when closing the main window.

14:49:53,227 ERROR [megamek.MegaMek] {AWT-EventQueue-0}
megamek.MegaMek.lambda$main$0(MegaMek.java:61) - Uncaught Exception Detected
java.lang.NullPointerException
	at megamek.client.ui.swing.util.TurnTimer.stopTimer(TurnTimer.java:96)
	at megamek.client.ui.swing.MovementDisplay.stopTimer(MovementDisplay.java:950)
	at megamek.client.ui.swing.ClientGUI.die(ClientGUI.java:1076)
	at megamek.client.ui.swing.ClientGUI$1.windowClosing(ClientGUI.java:551)
	at java.desktop/java.awt.AWTEventMulticaster.windowClosing(AWTEventMulticaster.java:357)
	at java.desktop/java.awt.AWTEventMulticaster.windowClosing(AWTEventMulticaster.java:357)
	at java.desktop/java.awt.Window.processWindowEvent(Window.java:2078)
	at java.desktop/javax.swing.JFrame.processWindowEvent(JFrame.java:298)
	at java.desktop/java.awt.Window.processEvent(Window.java:2037)
	at java.desktop/java.awt.Component.dispatchEventImpl(Component.java:5011)
	at java.desktop/java.awt.Container.dispatchEventImpl(Container.java:2321)
	at java.desktop/java.awt.Window.dispatchEventImpl(Window.java:2772)
	at java.desktop/java.awt.Component.dispatchEvent(Component.java:4843)
	at com.formdev.flatlaf.ui.FlatTitlePane.close(FlatTitlePane.java:768)
	at com.formdev.flatlaf.ui.FlatTitlePane.lambda$createButtons$3(FlatTitlePane.java:277)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1967)
	at java.desktop/javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2308)
	at java.desktop/javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:405)
[...]

PR #4014, opened 2022-12-01
NPE turn timer continues to run when you close the main window

kuronekochomusuke:
#4013 

stop the turn timer when closing the  main window if it is currently active


megamek/src/megamek/client/ui/swing/MovementDisplay.java
```diff
@@ -944,17 +944,21 @@ private void beginMyTurn() {
         tt = TurnTimer.init(this, clientgui.getClient());
     }
 
+    public void stopTimer() {
```
NickAragua, line 947: Is the turn timer in a base class somewhere? If so, we may consider moving this stopTimer function further up the inheritance hierarchy so that we only have to implement it once.

NickAragua (approved):

PR #4015, opened 2022-12-01
AS Conversion Errata and corrections

SJuliez:
More from https://bg.battletech.com/forums/errata/alpha-strike-companion/msg1881442/#msg1881442
Specifically:
- elements with armor 0 will no longer receive BAR
- added exact damage values for IS/CL BA Light Heavy/Mortar including 0.3 M dmg for heavy; they also now count for IF

### DIFF
--- a/megamek/src/megamek/common/weapons/battlearmor/ISBAMortarLight.java
+++ b/megamek/src/megamek/common/weapons/battlearmor/ISBAMortarLight.java
@@ -14,8 +14,10 @@
 package megamek.common.weapons.battlearmor;
 
 import megamek.common.AmmoType;
+import megamek.common.Mounted;
 import megamek.common.TechAdvancement;
 import megamek.common.WeaponType;
+import megamek.common.alphaStrike.AlphaStrikeElement;
 import megamek.common.weapons.Weapon;
 
 /**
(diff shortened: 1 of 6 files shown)


megamek/src/megamek/common/weapons/battlearmor/ISBAMortarLight.java
```diff
@@ -53,4 +55,14 @@ public ISBAMortarLight() {
         techAdvancement.setTechRating(RATING_B);
         techAdvancement.setAvailability(RATING_X, RATING_X, RATING_C, RATING_C);
     }
-}
+
+    @Override
+    public double getBattleForceDamage(int range, Mounted linked) {
+        return (range <= AlphaStrikeElement.SHORT_RANGE) ? 0.276 : 0;
```
NickAragua, line 61: Do we already have a pattern of using these magic numbers for other AS-related calculations?

I suppose it doesn't matter too much right now, since they're only used in the conversion process, it's just something I noticed with this particular pull request - if you wind up needing to update this formula again at some point, it'll be tough to remember the other places where this magic number occurs.

NickAragua: I'll leave it to your discretion as to whether that's worth it or not.

NickAragua (approved):

megamek/src/megamek/common/weapons/battlearmor/ISBAMortarLight.java
SJuliez: Assuming you speak of the 0.276 etc., these are the damage values from Alpha Strike Companion. I once asked (via Hammer) specifically and the numbers from that weapons table supersede anything derived by the older formula approach. So this is much like 
image

MM still implements the old formula approach and I havent removed it entirely only because I wanted to save myself from having to add hundreds of these explicit damage values. In case of the BA heavy mortars the formula gave 0.25 while the table says 0.249, therefore an explicit Override with values.

I hope I didnt misunderstand your point but if so, do say.

PR #4019, opened 2022-12-02
refactor AbstractDialog gui scaling

kuronekochomusuke:
- consolidate common code into AbstractDialog, so that the gui scale is consistently applied
- override when exceptions are needed

needs GUI Scale in meklab and mekhq


kuronekochomusuke:
was cleaning up branches and got the wrong one

Windchild292 (commented):
Has this been tested with MHQ? It will affect there, and MHQ has no access to GUI Preferences (as those are the older MM setup)

kuronekochomusuke:
no, it has not been tested there.

Issue #4035, opened 2022-12-11
RFE - Suggestions for search functionality in the mini report dialog


HammerGS:
I really like the search ability in the dialog. 

![image](https://user-images.githubusercontent.com/13752452/206923321-a8536fea-6f98-4693-a941-8e499a3ee165.png)

Since all our reports are pretty static and inject the unit, weapon, or whatever. Would it be possible to add a keyword search. Things like
- needs
- rolls
- takes
- hit
- falls
- skill roll
- piloting skill

So you could great search strings like. 
**player name  unit name  needs** 

Perhaps adding to the same dropdown as Damage/Destroyed/Phase. 

Or even a way for players to add custom terms to an external file that would populate that drop down.

https://github.com/MegaMek/megamek/pull/4004

SJuliez:
I believe this is implemented.

PR #4039, opened 2022-12-13
report keywords managed in client settings

kuronekochomusuke:
#4035

manage report keywords in client settings for the Mni Report Dialog

can add and remove keywords

image

image


HammerGS:
This is great! Thanks for looking into it.

SJuliez (approved):

Issue #3977, opened 2022-11-12
[49.10] Incorrect introduction date for Hoplite-4B


gsparks3:
Megamek currently has the introduction date for the Hoplite-4B listed as 2682. Per the [MUL entry](http://www.masterunitlist.info/Unit/Details/4330), this should be 2581.

pheonixstorm:
I have this fixed in my next PR which should be ready soon.

PR #4041, opened 2022-12-14
Data fix and code simplification

pheonixstorm:
Data fix is for #3977 

Code change for PR #4000 Consolidated some code removing several if blocks moving them into a single if statement at the beginning of the function which relocated two class casts and removed multiple if statements for boolean variables. Not sure if it will have a positive impact on performance but it does clean up the code some.

This was to include several other data fixes, but they were previously fixed and the tickets were not closed until after.


Windchild292 (changes requested):
Can you please split this into two PRs, as they are two unrelated changes affecting data and code.

Issue #4043, opened 2022-12-14
NPE - MechFileParser


HammerGS:
Running from the Repo - todays date. Getting a ton of the errors below in the log. MegaMek also fails to load a ton of units mostly Dropships.

![image](https://user-images.githubusercontent.com/13752452/207731810-e14e020f-9254-428b-ac47-64d1242cbfae.png)

### Environment
What version of MegaMek does your issue apply to?  Note whether it is the stand alone MegaMek or HQ's MegaMek.
What operating system are you using?  Windows, Mac, or Linux?
What java version are you using?  If unsure, it will state your java version at the top of the megamek.log file.
(Delete the above lines)

### Description
```
15:43:01,217 ERROR [megamek.common.MechFileParser] {Mech Cache Loader}
megamek.common.MechFileParser.<init>(MechFileParser.java:46) - 
java.lang.NullPointerException
	at megamek.common.weapons.autocannons.ACWeapon.getDamage(ACWeapon.java:111)
	at megamek.common.Entity.hasViableWeapons(Entity.java:872)
	at megamek.common.Entity.initMilitary(Entity.java:860)
	at megamek.common.MechFileParser.postLoadInit(MechFileParser.java:777)
	at megamek.common.MechFileParser.parse(MechFileParser.java:155)
	at megamek.common.MechFileParser.<init>(MechFileParser.java:44)
	at megamek.common.MechFileParser.<init>(MechFileParser.java:37)
	at megamek.common.MechSummaryCache.loadMechsFromDirectory(MechSummaryCache.java:621)
	at megamek.common.MechSummaryCache.loadMechsFromDirectory(MechSummaryCache.java:580)
	at megamek.common.MechSummaryCache.loadMechsFromDirectory(MechSummaryCache.java:580)
	at megamek.common.MechSummaryCache.checkForChanges(MechSummaryCache.java:253)
	at megamek.common.MechSummaryCache.loadMechData(MechSummaryCache.java:242)
	at megamek.common.MechSummaryCache.lambda$0(MechSummaryCache.java:82)
	at java.base/java.lang.Thread.run(Thread.java:829)
```
and 

```
15:43:01,478 ERROR [megamek.common.MechFileParser] {Mech Cache Loader}
megamek.common.MechFileParser.<init>(MechFileParser.java:46) - 
java.lang.NullPointerException
	at megamek.common.weapons.lasers.ISERLaserLarge.getLongRange(ISERLaserLarge.java:66)
[...]

PR #4044, opened 2022-12-15
issue 4043 NPE - MechFileParser

kuronekochomusuke:
issue 4043 NPE - MechFileParser
#4043 

- in loadMechsFromDirectory log all errors to file , so it does not hang github Task :megamek:officialUnitList task with popup.
- check if server and options are null before getting the option

will probably need to merge before it fixes the hang on the github Task :megamek:officialUnitList task.  looks like the task is using the merged version, not the version from the pull request

### DIFF
--- a/megamek/src/megamek/common/MechSummaryCache.java
+++ b/megamek/src/megamek/common/MechSummaryCache.java
@@ -639,7 +639,7 @@ public class MechSummaryCache {
                                     .append(failedEquipment.next()).append("\n");
                         }
                     }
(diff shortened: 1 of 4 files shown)


SJuliez (approved):
Looks good to me.

Windchild292 (commented):
This PR removes a necessary null check on the option and adds an unnecessary null check on the GameOptions (which are guaranteed to exist if the server instance isn't null).

I'll be opening a replacement PR shortly, as I was also working on a bugfix for this.

PR #4054, opened 2022-12-17
Code cleanup for #4000

pheonixstorm:
Code change for PR #4000 Consolidated some code removing several if blocks moving them into a single if statement at the beginning of the function which relocated two class casts and removed multiple if statements for boolean variables. Not sure if it will have a positive impact on performance but it does clean up the code some.

### DIFF
--- a/megamek/src/megamek/server/ServerHelper.java
+++ b/megamek/src/megamek/server/ServerHelper.java
@@ -405,7 +405,7 @@ public class ServerHelper {
 	public static void adjustHeatExtremeTemp(Game game, Entity entity, Vector<Report> vPhaseReport) {
         Report r;
         int tempDiff = game.getPlanetaryConditions().getTemperatureDifference(50, -30);
-        boolean heatArmor =false;
+        boolean heatArmor = false;
         boolean laserHS = false;
 
         if (entity instanceof Mech) {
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/server/ServerHelper.java
```diff
@@ -404,36 +404,31 @@ else if ((entity.heat >= 14) && !entity.isShutDown()) {
 
 	public static void adjustHeatExtremeTemp(Game game, Entity entity, Vector<Report> vPhaseReport) {
         Report r;
-		int tempDiff = game.getPlanetaryConditions().getTemperatureDifference(50, -30);
+        int tempDiff = game.getPlanetaryConditions().getTemperatureDifference(50, -30);
+        boolean heatArmor =false;
```
Windchild292, line 408:
```suggestion
        boolean heatArmor = false;
```

Windchild292 (changes requested):
One spacing issue, otherwise this looks good

megamek/src/megamek/server/ServerHelper.java
pheonixstorm: Fixed

Windchild292 (approved):

PR #4055, opened 2022-12-17
fix some gui scale issues

kuronekochomusuke:
- increase the width of the fovHighlightRingsRadii in client setting, so that it can handle large gui scales
- update keyboard shortcut overlay so that it processes any preference changes 
- clean up code in planetary conditions overlay code, so it is more similar to the code in keyboard shortcut overlay


Windchild292 (approved):

HammerGS:
Should we include this in 49.11 since it already has a lot of GUI stuff and this seems to fix/improve those.

kuronekochomusuke:
they are kind of minor, and you can work around them.  but it would be a nicer experience to have them.

PR #4057, opened 2022-12-17
allow report display split pane to resize

kuronekochomusuke:
when you have 2 rows of report phases in the report display it cause the resize of the split pane to freeze.  so the pilot display and chat log are hard to see.  only a few lines visible.

this allows it to resize in this case.

image

### DIFF
--- a/megamek/src/megamek/client/ui/swing/ReportDisplay.java
+++ b/megamek/src/megamek/client/ui/swing/ReportDisplay.java
@@ -106,6 +106,7 @@ public class ReportDisplay extends AbstractPhaseDisplay implements
             panButtons.add(new JLabel(""));
         }
         panelTop.add(panButtons, GBC.eol().fill(GridBagConstraints.HORIZONTAL));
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/client/ui/swing/ReportDisplay.java
```diff
@@ -106,6 +106,7 @@ public void actionPerformed(ActionEvent e) {
             panButtons.add(new JLabel(""));
         }
         panelTop.add(panButtons, GBC.eol().fill(GridBagConstraints.HORIZONTAL));
+        panelTop.setMinimumSize(new Dimension(0,0));
```
Windchild292, line 109:
```suggestion
        panelTop.setMinimumSize(new Dimension(0, 0));
```

Windchild292 (changes requested):
One spacing issue, otherwise looks good

Windchild292 (approved):

PR #4062, opened 2022-12-20
Add OS Thunderbolts

### LINKED ISSUE #2177: Thunderbolt OS Weapons need to be added.
Per this pending errata One-Shot Thunderbolts will need to be added.

https://bg.battletech.com/forums/index.php?topic=70703.msg1644919#msg1644919

SJuliez:
Fixes #2177

Adds OS/IOS Thunderbolt versions for construction and use, tested in MML and MM.

To reduce code duplication this expands the hierarchy of Thunderbolts. ThunderboltWeapon -> Thunderbolt##Weapon -> Thunderbolt##, Thunderbolt##OS, Thunderbolt##IOS where ## is the rack size.

@HammerGS The TechAdvancement for the normal Tbs is probably ok, but would you be able to help me with the TA for the OS and IOS. I copied in some TA from another OS weapon but that's probably off in some way.


SJuliez:
I looked at IO and the advancement tables some more and I improved the TechAdvancement some more. So now ready to merge.

Issue #3961, opened 2022-11-06
Alpha Strike Conversion Feedback


HammerGS:
Opening a thread to track unit errors CGL is finding while reviewing the AS conversion. 

- [ ] AP weapon mounts on Battle get counted as rifles (like conventional infantry) and added to damage calculations.  example, Achileus Light Battle Armor [David](Sqd4) should include the damage for conventional infantry rifles at short range and bring the damage up to 1 instead of 0*. 

- [ ] vehicles need to be checked for JMPW.  Kanga should have JMPW1 for example.
 
- [ ] SHLD missing from Black Knights (BLK-NT-2Y for example)

- [ ] Possible error with TOR calculations are not including Artemis when equipped with it. example AV1-OU

- [ ] BNZ-X missing BT, and definitely explosive (shouldn't have ENE)

- [ ] REAR weapons only use their own heat (not all weapons).  BLR-1G (Red Corsair) should have REAR2/2/- for example.

- [ ] BLR-1S should not have LRM1/1/1.  It shows the calculation for LRM at M correctly, with 0.929 that rounds to nearest tenth as .93.  it doesn't have 1 at medium, so no LRM.

- [ ] AMS doesn't have a direction/firing arc, so it shouldn't be in turret. (Beagle C3I as example)

- [ ] Aerospace fighters shouldn't have OMNI (can't carry mechanized battle armor) they aerospace can have IT, they can't have OMNI.

- [ ] MM's AS conversions are not including HT for conventional infantry as in the conversion report shows it being calculated, but it then doesn't show on card it's not included in the PV calculation either.

SJuliez:
Concerning MEL on Meks: I see that ASCE (=Commanders Edition) says that only Meks with MEL may make physical melee attacks. If that is the final word, ASC p127 should receive errata (says that PM and CV may have MEL). Also, ASCE p45 (states that PM may make Standard or Melee physical attacks).

NCKestrel:
HBK-4G (Shakir), the AC/10 shows up in the AC special ability calculation as 1/1/0, but then says no AC.

NCKestrel:
Kage Light Battle Armor, with squad weapons (where only one suit has the weapon), that weapon should not be multiplied by troop factor.  it just counts once.

NCKestrel:
Lexan Surveillance Helo lists 15/23 as movement,

> p

looks like we're going to change that.  CVs and SVs can get/use MEL.   (yes, PMs can use MEL).

[...]

PR #4067, opened 2022-12-24
Field Gun CI Alpha Strike changes

SJuliez:
AS conversion from #3961 (future errata): Field Gun CI use Field Gun Damage only instead of standard CI damage and may get AC and FLK

### DIFF
--- a/megamek/src/megamek/common/alphaStrike/conversion/ASConvInfantryDamageConverter.java
+++ b/megamek/src/megamek/common/alphaStrike/conversion/ASConvInfantryDamageConverter.java
@@ -42,27 +42,34 @@ public class ASConvInfantryDamageConverter extends ASDamageConverter {
 
     @Override
     protected void processDamage() {
-        report.addEmptyLine();
-        report.addLine("--- Damage:", "");
-        int baseRange = 0;
-        if ((infantry.getSecondaryWeapon() != null) && (infantry.getSecondaryWeaponsPerSquad() >= 2)) {
-            baseRange = infantry.getSecondaryWeapon().getInfantryRange();
-        } else if (infantry.getPrimaryWeapon() != null) {
-            baseRange = infantry.getPrimaryWeapon().getInfantryRange();
+
+        if (infantry.hasFieldWeapon()) {
+            processSDamage();
+            processMDamage();
+            processLDamage();
+            processFrontSpecialDamage(AC);
+            processFrontSpecialDamage(FLK);
+        } else {
+            int baseRange = 0;
+            if ((infantry.getSecondaryWeapon() != null) && (infantry.getSecondaryWeaponsPerSquad() >= 2)) {
+                baseRange = infantry.getSecondaryWeapon().getInfantryRange();
+            } else if (infantry.getPrimaryWeapon() != null) {
+                baseRange = infantry.getPrimaryWeapon().getInfantryRange();
+            }
+            int range = baseRange * 3;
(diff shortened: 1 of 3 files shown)


megamek/src/megamek/common/alphaStrike/conversion/ASConvInfantryDamageConverter.java
```diff
@@ -42,27 +42,34 @@ protected ASConvInfantryDamageConverter(Entity entity, AlphaStrikeElement elemen
...
+            finalSDamage = ASDamage.createDualRoundedUp(getConvInfantryStandardDamage());
+            String maxRangeText = "Range: S";
+            if (range > 3) {
+                finalMDamage = finalSDamage;
+                maxRangeText = "Ranges: S, M";
+            }
+            if (range > 15) {
+                finalLDamage = finalSDamage;
+                maxRangeText = "Ranges: S, M, L";
+            }
```
Windchild292, line 69:
```suggestion
            final String maxRangeText;
            if (range > 15)
                finalLDamage = finalSDamage;
                finalMDamage = finalSDamage;
                maxRangeText = "Ranges: S, M, L";
            } else if (range > 3) {
                finalMDamage = finalSDamage;
                maxRangeText = "Ranges: S, M";
            } else {
                maxRangeText  = "Range: S";
            }
```

Windchild292: This version (the first two parts specifically) prevents misreads, and should range > 15 be assigning the finalMDamage?

megamek/src/megamek/common/alphaStrike/conversion/ASLocationMapper.java
```diff
@@ -78,7 +78,7 @@ public static double damageLocationMultiplier(Entity en, int loc, Mounted mount)
             // Don't count squad support weapons, these are handled separately
             return ((mount.getLocation() <= 1) && !mount.isSquadSupportWeapon()) ? 1 : 0;
         } else if (en instanceof Infantry) {
-            return (loc == mount.getLocation()) ? 1 : 0;
+            return (loc == mount.getLocation()) ^ ((Infantry)en).hasFieldWeapon() ? 1 : 0;
```
Windchild292, line 81:
```suggestion
            return (loc == mount.getLocation()) ^ ((Infantry) en).hasFieldWeapon() ? 1 : 0;
```

Windchild292: Also, is this supposed to be an XOR?

Windchild292 (changes requested):
Some things first

megamek/src/megamek/common/alphaStrike/conversion/ASConvInfantryDamageConverter.java
SJuliez: Yes I'll do that. Infantry damage (disregarding field guns) is the same for S/M/L but extends to either S only, S/M or S/M/L based on the range of the employed infantry weapon. So when the weapon has 20 range, the S dmg value is used for M and L too.

megamek/src/megamek/common/alphaStrike/conversion/ASLocationMapper.java
SJuliez: Yes, that was intentional. But I've rewritten it anyway to be more digestible. It should only count field weapons when it has field weapons and when it hasn't, it can count all its weapons.

Windchild292 (approved):

Issue #4028, opened 2022-12-09
Aerospace Altitude Bomb Attack Scatter rules have been errata'd


Logajam:
### Environment
MM49.10

### Description
When a bombing attack's roll fails, scatter is based on the MOF now.  (TW Errata V9.  Page 47 of that document.)

Megamek is still using the Scatter by 1D6 hexes rules for Altitude bombing attacks.

Noticed this when an altitude bombing attack with a MOF of 2 scattered 6 hexes.  

Dive Bombing attacks are working as intended.

PR #4076, opened 2022-12-28
bombing updates

NickAragua:
-altitude bombs scatter by MoF rather than 1d6 per issue #4028 errata
-micro bombs use gunnery skill for TN instead of anti-mech skill, per TW page 228
-micro bombs only scatter one hex per TW page 228
-allow deserialization of some new classes


SJuliez (approved):
Looks good.

PR #4078, opened 2022-12-29
unit tool tip updates

kuronekochomusuke:
update the unit tool tip with addition information

- add system info to the mini armor display
- add indicator if movement  was modified:
🔥= heat
☂ = planetary conditions
🔧= damage
💣= bomb load
⛨ = shields
⟬▮ = modular armor
- if jump jet or jump boosters or partial wing are damage show counts for them
- show a message for the overall status of the entity
- show a message when unjamming a RAC
- prompt when done moving if you have jammed weapons and did not unjam them.
- add facing and twist

image

image

image

image

image

image


HammerGS:
These look great.

HammerGS:
Posted a teaser on the forums. There was a request

any chance of Bomb count and type on aircraft (and potentially VTOLs) would be great as well, seeing as there's no way of telling that in-game.

kuronekochomusuke:
which VTOL has bombs?  tried searching for one with a bomb bay, but did not find any.

image

HammerGS:
Seem to recall VTOL bomb is an optional rule. So if you enabled it and loaded up a VTOL that it would show.

kuronekochomusuke:
turning on TacOps VTOL Special Attacks allowed bombs to be added

image

HammerGS:
That’s a huge improvement. Do we show what’s being carried? I.e. Dropship showing units and doors, warships shows dropships and fighters

kuronekochomusuke:
it already show that

image

HammerGS:
Cool. It wondered if it did but not near my computer to check.

HammerGS:
Had a couple of comments on the forums. 

> Almost wish there was a option to put a warning message on for moving a unit with a jammed weapon and not unjamming it.  I'm probably the only person forgetful enough to find such a message useful. 
> 
> Does the message change if the weapon being unjammed is not a RAC?

Should we support the message with unofficial unjam UAC?

> does that unjamming message change for a jammed UAC or standard AC when using rapid fire? Princess loves using rapid fire when I have both options turned on (Rapid Fire and unjamming other AC weapons)

Not sure about support around optional rules, but I think it would make sense to cover all the edge cases.

kuronekochomusuke:
the button is Unjam RAC also
image

are there any weapons that can jam that don't  get unjammed when the unofficial setting is used:?

could change the to just descriptions unjam/unjamming, if it unjams all weapon types when the unofficial setting is used.

HammerGS:
> the button is Unjam RAC also image
> 
> are there any weapons that can jam that don't get unjammed when the unofficial setting is used:?
> 
> could change the to just descriptions unjam/unjamming, if it unjams all weapon types when the unofficial setting is used.

I'm a fan anything that simplifies the process and makes things clearer.

kuronekochomusuke:
image

image

kuronekochomusuke:
image

kuronekochomusuke:
image

SJuliez:
This looks great. However, I'm confused by what the part behind movement is showing. Maybe I'm just dense :)

SJuliez:
Concerning the bombers it should be possible to do 

if (entity instanceof IBomber)

and then cast to (IBomber) and use .getBombChoices() on that (all bomb-capable entity classes implement IBomber).

kuronekochomusuke:
the symbols after the movement are an indicator as to why the movement is lower.  

should probably make a key, since some may not be easy to interpret

kuronekochomusuke:
should LandAirMech movement be reduced by bomb load like the other IBomber units?

should QuadMech movement be reduced by shields like BipedMech and TripodMech units?

gsparks3:
Request: is it possible to add unit facing to the tooltip? It's essentially impossible to determine unit facing if there is more than one unit in a single hex, but a tooltip readout would make it unambiguous.

kuronekochomusuke:
image

kuronekochomusuke:
this is ready for review

megamek/src/megamek/client/ui/swing/MovementDisplay.java
```diff
@@ -311,6 +313,11 @@ public MovementDisplay(final ClientGUI clientgui) {
         buttons = new HashMap<>((int) (MoveCommand.values().length * 1.25 + 0.5));
         for (MoveCommand cmd : MoveCommand.values()) {
             String title = Messages.getString("MovementDisplay." + cmd.getCmd());
+            if ((clientgui != null) &&
+                    (title.equals(Messages.getString("MovementDisplay." + MoveCommand.MOVE_UNJAM.getCmd()))) &&
```
SJuliez, line 317: This string comparison looks like it could be replaced by 
cmd == MoveCommand.MOVE_UNJAM

megamek/src/megamek/common/IBomber.java
```diff
@@ -52,6 +52,11 @@ default int getMaxBombSize() {
      * Sets the count of each bomb to zero
      */
     void clearBombChoices();
+
+    /**
+     * @return The load for bombs currently on unit
+     */
+    int getBombLoad(int t);
```
SJuliez, line 59: I assume the name comes from the "// get bomb load" comments but since this method has found its way to the IBomber interface I'd prefer a name that actually says what it does, like reduceMPForBombLoad or something. Also the comment here might preferably say what the parameter is and what the method is supposed to return.

megamek/src/megamek/common/VTOL.java
```diff
@@ -273,6 +273,20 @@ public void clearBombChoices() {
...
+        for (Mounted m : getBombs()) {
+            if (m.getUsableShotsLeft() > 0) {
+                t--;
+            }
+        }
+        if (t < 0) {
+            t = 0;
+        }
+        return t;
+    }
```
SJuliez, line 288: Well I havent tested it but this could probably be a one-liner  :D
return Math.max(0, t - getBombs().stream().filter(m-> m-getUsableShotsLeft() > 0).count());
(really unimportant, it just crossed my mind)

megamek/src/megamek/client/ui/swing/tooltip/UnitToolTip.java
```diff
@@ -98,12 +101,12 @@ private static StringBuilder getEntityTipTable(Entity entity, Player localPlayer
...
             result.append(inGameValues(entity, localPlayer));
             if (pilotInfo) {
-                result.append(PilotToolTip.getPilotTipShort(entity,
-                        GUIPreferences.getInstance().getBoolean(GUIPreferences.SHOW_PILOT_PORTRAIT_TT)));
+                result.append(PilotToolTip.getPilotTipShort(entity, GUIP.getBoolean(GUIPreferences.SHOW_PILOT_PORTRAIT_TT)));
             } else {
                 result.append("<BR>");
             }
         }
+        result.append("\n");
```
SJuliez, line 109: Why are all these \n added?

megamek/src/megamek/client/ui/swing/tooltip/UnitToolTip.java
```diff
@@ -179,62 +185,131 @@ private static boolean hideArmorLocation(Entity entity, int location) {
...
     private static String locationHeader(Entity entity, int location) {
-        return entity.isConventionalInfantry() ?
-                ((Infantry) entity).getShootingStrength() + " Active Troopers" : entity.getLocationAbbr(location);
+        String msg_activetroopers =Messages.getString("BoardView1.Tooltip.ActiveTroopers");
+        return entity.isConventionalInfantry() ? ((Infantry) entity).getShootingStrength() + " " + msg_activetroopers : entity.getLocationAbbr(location);
+    }
+
+    private static StringBuilder sysCrits(Entity entity, int type, int index, int loc, String l) {
+        StringBuilder result = new StringBuilder();;
+        int good = entity.getGoodCriticals(type, index, loc);
```
SJuliez, line 194: This adds e.g. 6 crits for the engine. Would it make sense to only show the crits necessary to destroy the unit instead (as on the record sheet)? This would probably require some rewriting so I'm okay with it as it is now but it crossed my mind when seeing the engine crits in the screenshots.

megamek/src/megamek/client/ui/swing/tooltip/UnitToolTip.java
```diff
@@ -495,37 +616,77 @@ private static StringBuilder weaponList(Entity entity) {
...
         }
-        result.append("</TABLE>");
+        result.append("</TABLE>\n");
+        return result;
+    }
+
+    private static StringBuilder bombList(Entity entity) {
+        StringBuilder result = new StringBuilder();
+
+        if (entity.isBomber()) {
```
SJuliez, line 661: Looks to me like l.661 could already be the instanceof check and the in l.665 could be omitted.

SJuliez (changes requested):
A few comments/questions below, apart from those all ok.

megamek/src/megamek/client/ui/swing/tooltip/UnitToolTip.java
kuronekochomusuke: just makes it easier to see if i did things correctly, can remove if needed

megamek/src/megamek/client/ui/swing/tooltip/UnitToolTip.java
kuronekochomusuke: think xl engines with slots in the left and right torso would also make this complicated

SJuliez (approved):

PR #4084, opened 2023-01-03
Java 17: Manifest File Add-Opens

Windchild292:
The `add-opens` was only applying for the exe files, which means it was missing for JAR uses. I believe this fixes that issue.


Windchild292:
+ PR #4084: Java 17: Manifest File Add-Opens

PR #4092, opened 2023-01-07
Lobby updates

kuronekochomusuke:
Lobby updates

- add deployment zone offset and width to player display, if not default
- add start to unit display
- add deployment zone offset and width to unit display, if not default
- if unit start is not using players start, show in different color
- add direction and distance to display for off board units
- if mech or aero has an ejection seat and auto ejection is turned off, display a warning
- if bot is set to single bind, show any eye indicator in player display
- add game option:  (Unofficial) Default bots to single blind, when using TacOps Double blind
- when applying critical damage to  system components in the edit damage dialog, only apply damage once when the component occupies multiple locations.
- when partial wing has critical damage applied in the  edit damage dialog and you edit damage again load in the damage correctly.
- add Alpha Strike Unit Role to the unit display
- add sorter for Player, Unit Role 

image

image


HammerGS:
Been playing around with these and really like them. Would it be possible to add the unit role somewhere?

image

I often assign Princess roles based on the units and I find I have to go and back forth and would be nice to see them without the extra clicks.

kuronekochomusuke:
image

HammerGS:
Perfect! Thanks.

kuronekochomusuke:
this is ready for review

rjhancock (approved):

HammerGS:
Thanks for these!

PR #4097, opened 2023-01-11
attack tool tip update

kuronekochomusuke:
attack tool tip update

- show attacks that cross the current hex, instead of attacks at current point.  this makes it easier for these to be displayed.
- display attacks after the unit info, so that it does not hide the unit info when there are many attacks.
- collapse multiple shots with the same weapon type into a single row. 

image

image

image


kuronekochomusuke:
this is ready for review

rjhancock (approved):

HammerGS:
Thanks for these!

PR #4102, opened 2023-01-13
SPA: Terrain Master [Nightwalker]

pheonixstorm:
I have been sitting on this for several years in a vain attempt to find a way to allow for the full rules used for Nightwalker. I never could find a solution, and it seems I wasn't the only one as Nick tried as well and gave up on it.


megamek/i18n/megamek/common/options/messages.properties
```diff
@@ -555,6 +555,8 @@ PilotOptionsInfo.option.tm_frogman.displayableName=Terrain Master [Frogman] (Cam
 PilotOptionsInfo.option.tm_frogman.description=Movement in depth 2 or greater water costs 1 MP less.\nAlso grants -1 to all PSR checks (including physicals) as well as -2 to Crush Depth checks.
 PilotOptionsInfo.option.tm_mountaineer.displayableName=Terrain Master [Mountaineer] (CamOps)
 PilotOptionsInfo.option.tm_mountaineer.description=Movement in rubble or rough cost 1 less MP and elevation changes also cost 1 MP less.\nThe mountaineer also recieves a -1 to PSR checks for the given terrain.
+PilotOptionsInfo.option.tm_nightwalker.displayableName=Terrain Master [Nightwalker] (CamOps)
+PilotOptionsInfo.option.tm_nightwalker.description=Nightwalker ignores all darkness move penalties when using Walk/Cruise movement but does not affect gunnery modifiers for lighting conditions.
```
Windchild292, line 559:
```suggestion
PilotOptionsInfo.option.tm_nightwalker.description=Nightwalker ignores all darkness move penalties when using Walk/ Cruise movement but does not affect gunnery modifiers for lighting conditions.
```

megamek/src/megamek/common/Entity.java
```diff
@@ -12810,6 +12810,16 @@ public boolean isNaval() {
                 || (getMovementMode() == EntityMovementMode.SUBMARINE);
     }
 
+    /**
+     * Determines if the pilot has the Nightwalker SPA
+     * @return true when pilots have the SPA and not
```
Windchild292, line 12815:
```suggestion
     * @return true when pilots have the SPA and are not
```

megamek/src/megamek/common/options/OptionsConstants.java
```diff
@@ -167,7 +167,8 @@ public class OptionsConstants {
     // public static final String PILOT_TM_= "tm_"; 
     public static final String PILOT_TM_FOREST_RANGER= "tm_forest_ranger"; 
     public static final String PILOT_TM_FROGMAN= "tm_frogman"; 
-    public static final String PILOT_TM_MOUNTAINEER= "tm_mountaineer"; 
+    public static final String PILOT_TM_MOUNTAINEER= "tm_mountaineer";
+    public static final String PILOT_TM_NIGHTWALKER= "tm_nightwalker";
```
Windchild292, line 171:
```suggestion
    public static final String PILOT_TM_MOUNTAINEER = "tm_mountaineer";
    public static final String PILOT_TM_NIGHTWALKER = "tm_nightwalker";
```

Windchild292: This is an accessibility issue

megamek/src/megamek/common/MoveStep.java
```diff
@@ -2953,16 +2953,20 @@ && getEntity().getPosition().equals(prev)
...
+                    case PlanetaryConditions.L_MOONLESS:
+                        mp += 2;
+                        break;
+                    case PlanetaryConditions.L_PITCH_BLACK:
+                        mp += 3;
+                        break;
+                }
+            } else {
+                setRunProhibited(true);
             }
```
Windchild292, line 2970: Does this ability let you choose if it's used or not? If so, this is likely a problematic way to handle it.

Windchild292 (changes requested):
A few minor things and a major question

megamek/src/megamek/common/MoveStep.java
pheonixstorm: The ability is supposed to allow for full walk movement at night OR run at a -1 to the night move penalty. Both Nick and I have tried to work in both but it can't be done.

As for choice, no. Say you have a 5/8 mech, w/o the SPA you are running 4 hexes at 2mp per hex minimum under a full moon. With the SPA you can walk 5 hexes.

The choice making this was either walk full speed or run at -1 to the penalty. Disabling running seemed the better option of the two for the player

megamek/i18n/megamek/common/options/messages.properties
pheonixstorm: Written with the same spacing usage in the rules. I can change it if you want, just following how it is written elsewhere.

Windchild292: It should never be written without at least one space, and both sides is preferrable. It's an accessibility issue, as without the spacing the words merge together.

pheonixstorm: Ok then, understandable

megamek/src/megamek/common/MoveStep.java
Windchild292: I've been thinking about this today, and it's a huge negative for daytime use. Also, do we not split movement between the various types?

pheonixstorm: Derp, first is a bug. I rewrote this from memory and thought I had it right.

The second I'm not sure what you mean.

Windchild292: How is movement calculated, especially the various levels of movement. I've never looked at it, so I'm wondering how it works as I have ideas for potentially handling it better.

pheonixstorm: Ah, ok. I know weather MP modifiers are handled in each units class while terrain, fog, and night are handled in MoveStep.

MoveStep processes each hex a unit travels over and builds a list of mp spent moving into each hex. The terrain cost is in either Terrain or Terrains, I don't remember. Regardless the function processMovementCost handles all of it. Not sure what handles the color changing when going from walk to run to sprint. I'm not as familiar with it as I used to be so it is hard to properly explain, though with Cross-Country being half done I will have to figure out how it works better if I am to makes tanks swim.

Also, if you check in Tank, you will find functions for getWalkMP, getRunMP, getSprintMP etc that defines each from the rules. If this has been any help I would be surprised. I can read and follow the flow of the code I'm just not very good at explaining it. Nick or Arlith would be of more help. I don't know if neoancient ever touched this section of the code or not, so maybe him as well?

megamek/src/megamek/common/MoveStep.java
```diff
@@ -2953,16 +2953,20 @@ && getEntity().getPosition().equals(prev)
...
+                        mp += 1;
+                        break;
+                    case PlanetaryConditions.L_MOONLESS:
+                        mp += 2;
+                        break;
+                    case PlanetaryConditions.L_PITCH_BLACK:
+                        mp += 3;
+                        break;
+                }
+            } else if ((game.getPlanetaryConditions().getLight() > PlanetaryConditions.L_DUSK)){
```
Windchild292, line 2968:
```suggestion
            } else if (game.getPlanetaryConditions().getLight() > PlanetaryConditions.L_DUSK) {
```

pheonixstorm:
Made the other suggested change and fixed conflict status.

Windchild292 (approved):

Issue #4082, opened 2022-12-31
ClientGUI: Remove Static Resource Calls


Windchild292:
Resource calls should never be static calls, but instead call the proper bundle at the time.

Windchild292:
ClientGUI is the reason for this issue, as it contains a large number of unlocalized values specified as `private static final`, but I know this issue exists in other places in MM. If anyone finds other cases of it, please add it to this issue if it is still open or otherwise open a new issue.

kuronekochomusuke:
most of these can be found by searching for 

private static final String MSG_

there could be a few named differently but most should start with MSG_

i can correct these, since i caused them.   how should these be?

kuronekochomusuke:
would these also be an issue? 

CommonSettingsDialog
private final JCheckBox autoEndFiring = new JCheckBox(Messages.getString("CommonSettingsDialog.autoEndFiring"));

kuronekochomusuke:
these files private static final String Messages

megamek\megamek\src\megamek\client\ui\dialogs\BotConfigDialog.java
megamek\megamek\src\megamek\client\ui\swing\boardview\PlanetaryConditionsOverlay.java
megamek\megamek\src\megamek\client\ui\swing\ChatterBox.java
megamek\megamek\src\megamek\client\ui\swing\ClientGUI.java
megamek\megamek\src\megamek\client\ui\swing\CommonSettingsDialog.java
megamek\megamek\src\megamek\client\ui\swing\lobby\ChatLounge.java
megamek\megamek\src\megamek\client\ui\swing\MiniReportDisplay.java
megamek\megamek\src\megamek\client\ui\swing\RandomMapDialog.java
megamek\megamek\src\megamek\client\ui\swing\RandomMapPanelBasic.java
megamek\megamek\src\megamek\client\ui\swing\ReplacePlayersDialog.java
megamek\megamek\src\megamek\client\ui\swing\ReportDisplay.java
megamek\megamek\src\megamek\client\ui\swing\ResizeMapDialog.java
megamek\megamek\src\megamek\common\commandline\AbstractCommandLineParser.java
megamek\megamek\src\megamek\common\PlanetaryConditions.java

Windchild292:
They should all be inline Message calls, never constants as those require a restart to recreate. Values recreated on initialization are okay, provided the dialogs aren't kept initialized.

kuronekochomusuke:
[...]

PR #4108, opened 2023-01-14
4082: Removing Static Message Constant Uses

Windchild292:
This handles the easier parts of #4082, and subsumes #4089. The last two commits fix resource leaks and invalid logging (as out and err both redirect to logging).


megamek/src/megamek/server/DedicatedServer.java
```diff
@@ -67,21 +65,13 @@ public static void start(String[] args) {
...
-            MegaMek.printToOut(ex.getLocalizedMessage());
             return;
         }
 
         if (null != resolver.saveGameFileName) {
             server.loadGame(new File(resolver.saveGameFileName));
         }
     }
-
-    public static void main(String[] args) {
```
NickAragua, line 83: Are we uh, still able to start a dedicated server?

NickAragua (commented):
Just one question, looks good otherwise.

megamek/src/megamek/server/DedicatedServer.java
Windchild292: Yes, this was just a manual dev use starter... that hasn't worked in years (since Log4j2 was installed at the latest).

PR #4109, opened 2023-01-14
4098: Adding Sprite Camouflage and Damage Export Options

### LINKED ISSUE #4098: RFE - Export Sprite with Camo
I’ve seen this request a few times in the last few years.  

Have the ability to export the image of a sprite with camo or colouring.

Windchild292:
This fixes #4098. Review with whitespace differences off.

### DIFF
--- a/megamek/i18n/megamek/client/messages.properties
+++ b/megamek/i18n/megamek/client/messages.properties
@@ -68,17 +68,17 @@ comboClients.toolTipText=The player whose skill generation setup is being custom
 #### Lobby
 ### ExportUnitSpriteMenu Class
 exportUnitSpriteMenu.title=Export Unit Sprite
-exportUnitSpriteMenu.toolTipText=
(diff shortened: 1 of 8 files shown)


SJuliez (approved):
line 469 in EntityImage looks like it could be removed. Other than that, looks good and works great!

Windchild292 (commented):
I forgot to remove testing code, was having some odd issues with damage export that were because of how I was testing.

HammerGS:
This is really cool! Do you think there would be a way to have it multi-select in the lobby and then save/export a batch at once?

Windchild292:
I considered it while coding, but every way I've thought of has terrible UX. This is because there's no good way to have the same functionality while naming individual files and catching duplicated file names.

HammerGS:
> I considered it while coding, but every way I've thought of has terrible UX. This is because there's no good way to have the same functionality while naming individual files and catching duplicated file names.

Wouldn't something like limit that option to the lobby then leave the naming the same as the unit. Then players could use tools outside of the MegaMek to fix?

PR #4112, opened 2023-01-15
BaseComponents: Adding Nag Dialog Setup

Windchild292:
This migrates the Nag Dialog setup to MM. There's two FIXME's for places where we should have access to frames but we do not right now.

### DIFF
--- a/megamek/i18n/megamek/client/messages.properties
+++ b/megamek/i18n/megamek/client/messages.properties
@@ -54,6 +54,7 @@ AbstractHelpDialog.helpFile=Help File:\u0020
 AbstractHelpDialog.noHelp.title=No Help Available
 
 ### Nag Dialogs
+## BotReadmeNagDialog Class
 BotReadmeNagDialog.title=Please Read The Bot Documentation First
 BotReadmeNagDialog.text=Please read the ai-readme.txt file in the docs/Bot Stuff folder before using the bot. \nNote: The bot does not support all game options. \nWould you like to read the AI documentation now?
 
--- a/megamek/i18n/megamek/client/messages.properties
+++ b/megamek/i18n/megamek/client/messages.properties
@@ -39,11 +39,23 @@ SkillGeneratorType.MANEI_DOMINI.toolTipText=Generate skills for Manei Domini cre
 
 
 ##### UI Resources
+#### Base Components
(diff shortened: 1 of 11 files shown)


megamek/i18n/megamek/client/messages.properties
```diff
@@ -39,12 +39,25 @@ SkillGeneratorType.MANEI_DOMINI.toolTipText=Generate skills for Manei Domini cre
 
 
 ##### UI Resources
+#### Base Components
+#### AbstractMHQNagDialog
+chkIgnore.text=Do not bother me again
+chkIgnore.toolTipText=If selected this nag will no longer show until re-enabled under MekHQ Options.
```
NickAragua, line 45: Is this MekHQ specific?

NickAragua (commented):
One minor question, looks fine otherwise.

megamek/i18n/megamek/client/messages.properties
Windchild292: Dumb copy-paste error, the dialog was created based on MHQ's.

megamek/src/megamek/client/ui/dialogs/nagDialogs/BotReadmeNagDialog.java
```diff
@@ -0,0 +1,39 @@
...
+public class BotReadmeNagDialog extends AbstractNagDialog {
+    //region Constructors
+    public BotReadmeNagDialog(final JFrame frame) {
+        super(frame, "BotReadmeNagDialog", "BotReadmeNagDialog.title",
+                "BotReadmeNagDialog.text", MMConstants.NAG_BOT_README);
+    }
+    //endregion Constructors
+
+    @Override
+    protected boolean checkNag() {
```
SJuliez, line 36: Does this need an override? It looks like it would already work in the super class.

SJuliez (approved):
All good. One question/comment.

megamek/src/megamek/client/ui/dialogs/nagDialogs/BotReadmeNagDialog.java
Windchild292: It does if we need to check other things, which is required in the MHQ implementations and likely for some MM implementations.

PR #4115, opened 2023-01-15
initiative reports update

### LINKED ISSUE #4060: RFE Deployment report
RFE Deployment report

- add a report to the round reports that shows the units to deploy broken down by each deployment round.  
- as units are deployed, don't display them in the next round's report.
- if double bind setting is active, show ??? instead of the names of units you are not allowed to know about.
- show deployment zone for each unit

this would help in verifying that every thing was setup correctly at the start of the game.   and help remember what is left to deploy as the game progresses.

mockup:
image

kuronekochomusuke:
initiative reports update fixes #4060 

- when using (Unofficial) Individual Initiative, do not show rolls for units that will be deployed in later rounds.
- add a section to show the remain units to deploy, break down by round

image

image


NickAragua (approved):
Looks good.

PR #4118, opened 2023-01-17
SPA Cross-Country

pheonixstorm:
With the exception of water movement this SPA is completed. Waiting on ruling for how tanks in depth 1 water should be treated. I'm thinking they should have the same LOS and targeting rules as amphibious tanks, but you never know sometimes.


megamek/src/megamek/common/Terrain.java
```diff
@@ -386,29 +388,52 @@ public int movementCost(Entity e) {
...
                     mp -= 1;
                 }
                 return Math.max(0, mp);
             case Terrains.WOODS:
                 mp = level;
+                if (isCrossCountry) {
+                    if ((level == 1) && ((moveMode == EntityMovementMode.HOVER) || (moveMode == EntityMovementMode.WHEELED))) {
+                        mp *= 2;
+                    } else {
+                        mp *= 2;
```
Windchild292, line 420: This just assigns the same value for both.

megamek/src/megamek/common/Terrain.java
```diff
@@ -386,29 +388,52 @@ public int movementCost(Entity e) {
                 } else {
                     mp = 1;
                 }
+
+                if (isCrossCountry) {
+                    if (((moveMode == EntityMovementMode.HOVER) || (moveMode == EntityMovementMode.TRACKED))
+                            && (level == 6)) {
+                        mp *= 2;
+                    } else if ((moveMode == EntityMovementMode.WHEELED)) {
+                        mp *= 2;
```
Windchild292, line 397: Is this meant to be the same? If so, maybe look at using an or statement?

megamek/src/megamek/common/Terrain.java
```diff
@@ -494,12 +526,24 @@ public int movementCost(Entity e) {
...
                     mp = 1;
                 }
+
+                if (isCrossCountry) {
+                    if (((moveMode == EntityMovementMode.HOVER) || (moveMode == EntityMovementMode.TRACKED))
+                            && (level == 2)) {
+                        mp *= 2;
+                    } else if ((moveMode == EntityMovementMode.WHEELED)) {
+                        mp *= 2;
+                    }
```
Windchild292, line 536: Is this meant to be the same? If so, maybe look at using an or statement?

Windchild292 (changes requested):
Few things first

megamek/src/megamek/common/Terrain.java
pheonixstorm: Wheeled vehicles aren't allowed in rubble while all ground vehicles are not allowed in ultra rubble.

megamek/src/megamek/common/Terrain.java
pheonixstorm: Tracked vehicles are allowed in light woods while the others aren't

megamek/src/megamek/common/Terrain.java
pheonixstorm: Wheeled vehicles aren't allowed in rough while all ground vehicles are not allowed in ultra rough.

pheonixstorm:
I've started a new branch to add in the isXXX calls so this will get cleaned up after it is merged along with the original terrain code and possibly some other areas.

megamek/src/megamek/common/Terrain.java
Windchild292: That isn't what this code does. It just assigns mp to mp * 2 if isCrossCountry is true.

pheonixstorm: Fixed

pheonixstorm:
Is there anything else I need to fix so this can get approved?

pheonixstorm:
Conflict fixed

megamek/src/megamek/common/Terrain.java
```diff
@@ -386,29 +388,52 @@ public int movementCost(Entity e) {
                 } else {
                     mp = 1;
                 }
+
+                if (isCrossCountry) {
```
SJuliez, line 392: Sorry to chime in. 
This only checks the movement mode. WHEELED is also used by QuadVees in wheeled mode. A vehicle pilot probably shouldn't command a qv but MM doesn't prevent this. I wonder if e.isVehicle() should be added to the crossCountry boolean definiton.

Also, I agree with Windchild that repeating mp *= 2 really calls for OR-ing the two conditions or it looks like an error.

megamek/src/megamek/common/Terrain.java
```diff
@@ -386,29 +388,52 @@ public int movementCost(Entity e) {
...
+
                 if ((e.hasAbility(OptionsConstants.INFANTRY_FOOT_CAV)
                         && (moveMode == EntityMovementMode.INF_LEG))) {
                     mp -= 1;
                 }
                 return Math.max(0, mp);
             case Terrains.WOODS:
                 mp = level;
+                if (isCrossCountry) {
+                    if ((level == 1) && ((moveMode == EntityMovementMode.HOVER) || (moveMode == EntityMovementMode.WHEELED))) {
```
SJuliez, line 417: Same as the other, OR-ing this looks even easier as level can't be 0

megamek/src/megamek/common/Terrain.java
```diff
@@ -494,12 +526,24 @@ public int movementCost(Entity e) {
                 } else {
                     mp = 1;
                 }
+
+                if (isCrossCountry) {
+                    if (((moveMode == EntityMovementMode.HOVER) || (moveMode == EntityMovementMode.TRACKED))
```
SJuliez, line 531: as the others

SJuliez (changes requested):
I'd like to merge this but I have some comments (agreeing with Windchild on this)

megamek/src/megamek/common/Terrain.java
pheonixstorm: The first one for woods was a bug that was fixed. I can probably try to do the if/else or if/else if better, but to cram it all in a single if would make it look rather confusing imo.

As for quadvee. A check for isVee will get added just to make sure it ignores any weirdness like that.

megamek/src/megamek/common/Terrain.java
pheonixstorm: But it can be greater than one, and certain vehicles can't travel is certain woods levels. Woods have 3 levels, 1 is light woods, 2 is heavy woods, and 3 is ultra heavy woods. Hover and Wheeled vehicles cannot enter woods at all while tracked vehicles can only enter light woods.

SJuliez: Hmm I feel we're talking past each other here.
With crossCountry, what this checks is:
if Level 1 and h/w --> mp*2
-also- if Level > 1 --> mp*2
That is exactly the same as if (Level 1 and h/w) -or- Level > 1 --> mp*2
Which is the same as: if level > 1 -or- h/w --> mp*2 (assuming level must be > 0)
(I have not thought about the rules, only the code as it is)

SJuliez: italic text not intentional

megamek/src/megamek/common/Terrain.java
SJuliez: I agree these are usually ugly to read with all the ((()()((()((()))). But having the same result multiple times screams error or legacy issue. Maybe a solution is to add a 
final boolean allowedHoverWheeled = ((moveMode == EntityMovementMode.HOVER) || (moveMode == EntityMovementMode.TRACKED)) && (level == 6);
and then if (allowedHoverWheeled || (moveMode == EntityMovementMode.WHEELED)) {
This is still shorter than the if else and readable enough I guess. What do you say?

SJuliez: Eh hoverTracked of course

pheonixstorm: It could work, I will split the changes into multiple commits for review to make it easier to revert if it doesn't look right and you decide against it. Quadvee will be first then the others.

pheonixstorm:
Changes done.

SJuliez (approved):

HammerGS:
https://www.battletech.com/forums/index.php/topic,80157.0.html

PR #4121, opened 2023-01-18
LocalParser: Tearing Out Unused Code

Windchild292:
We never use this, as it's all inline.


SJuliez (approved):
Wow the LocaleParser was antique.

PR #4125, opened 2023-01-20
button tool tip for hot keys

kuronekochomusuke:
- on the phase display buttons, update the tool tip to add the hot key bind description, when the button has a hot key.
- correct issue with the done hot key not working in the report phases
- reset defaultKeyBinds.xml to the defaults defined in KeyCommandBind
- set quick save to Ctrl+Shift+S and quick load to Ctrl+Shift+L so it does not conflict with the LOS Setting hot key Ctrl+Alt+L

image


megamek/src/megamek/client/ui/swing/DeployMinefieldDisplay.java
```diff
@@ -56,6 +60,17 @@ public String getCmd() {
         public String toString() {
             return cmd;
         }
+
+        public String getHotKeyDesc() {
```
SJuliez, line 64: This code block looks like it could be improved.

megamek/src/megamek/client/ui/swing/FiringDisplay.java
```diff
@@ -185,6 +238,42 @@ public FiringDisplay(final ClientGUI clientgui) {
...
+            newButton.addActionListener(this);
+            newButton.setActionCommand(cmd.getCmd());
+            newButton.setEnabled(false);
+            buttons.put(cmd, newButton);
+        }
+        numButtonGroups = (int) Math.ceil((buttons.size() + 0.0) / buttonsPerGroup);
+    }
+
+    @Override
+    protected void setButtonsTooltips() {
```
SJuliez, line 256: This is repeated a lot with minor differences. It looks like it could be unified as a static method to StatusBarPhaseDisplay.

megamek/src/megamek/client/ui/swing/FiringDisplay.java
```diff
@@ -185,6 +238,42 @@ public FiringDisplay(final ClientGUI clientgui) {
...
+            if (!tt.isEmpty()) {
+                String title = Messages.getString("FiringDisplay." + cmd.getCmd());
+                tt = guiScaledFontHTML(uiLightViolet()) + title + ": " + tt + "</FONT>";
+                tt += "<BR>";
+            }
+            if (Messages.keyExists(ttKey)) {
+                String msg_key = Messages.getString(ttKey);
+                tt += guiScaledFontHTML() + msg_key + "</FONT>";
+            }
+            String b = "<BODY>" + tt + "</BODY>";
```
SJuliez, line 269: Not sure if this is intentional and for readability but l. 269 and 270 do look a bit verbose. At the very least they could be moved into the if-block.

SJuliez (changes requested):
Good stuff. I have a few requests and comments for this. A personal request would be to move the LOS settings to Ctrl-Alt-L and leave the quick-save and -load at Ctrl-Shift-S/L. I don't know, maybe people who play more and develop less see this differently.

megamek/src/megamek/client/ui/swing/FiringDisplay.java
SJuliez: The few dependencies could be injected as parameters or obtained with overridden methods like getCommands()

megamek/src/megamek/client/ui/swing/FiringDisplay.java
kuronekochomusuke: just makes it easier to see that the tags are added in matching pairs.  adding the start tag, then some time later adding the end tag makes it easy to miss one.  also makes it easier to swap them out if needed, to have the set only on one line.

megamek/src/megamek/client/ui/swing/DeployMinefieldDisplay.java
```diff
@@ -35,31 +38,51 @@ public class DeployMinefieldDisplay extends StatusBarPhaseDisplay {
...
+            priority = p;
+        }
         
         @Override
         public String toString() {
-            return cmd;
+            return Messages.getString("DeployMinefieldDisplay." + getCmd());
+        }
+
+        public String getHotKeyDesc() {
```
SJuliez, line 77: why not return ""; ?

megamek/src/megamek/client/ui/swing/PhysicalDisplay.java
```diff
@@ -91,6 +94,22 @@ public void setPriority(int p) {
         public String toString() {
             return Messages.getString("PhysicalDisplay." + getCmd());
         }
+
+        public String getHotKeyDesc() {
+            String result = "";
+
+            switch (this) {
```
SJuliez, line 101: Why not use
if (this == PHYSICAL_NEXT) {
?

megamek/src/megamek/client/ui/swing/ReportDisplay.java
```diff
@@ -63,6 +66,21 @@ public void setPriority(int p) {
         public String toString() {
             return Messages.getString("ReportDisplay." + getCmd());
         }
+
+        public String getHotKeyDesc() {
+            String result = "";
+
+            switch (this) {
+                case REPORT_REPORT:
```
SJuliez, line 74: ditto

megamek/src/megamek/client/ui/swing/SelectArtyAutoHitHexDisplay.java
```diff
@@ -77,6 +80,17 @@ public void setPriority(int priority) {
         public String toString() {
             return Messages.getString("SelectArtyAutoHitHexDisplay." + getCmd());
         }
+
+        public String getHotKeyDesc() {
```
SJuliez, line 84: This can be reduced to return "";

megamek/src/megamek/client/ui/swing/PrephaseDisplay.java
```diff
@@ -98,6 +100,25 @@ public void setPriority(int p) {
...
         }
+
+        public String getHotKeyDesc() {
+            String result = "";
+
+            String msg_next= Messages.getString("Next");
+            String msg_previous = Messages.getString("Previous");
+
+            switch (this) {
+                case PREPHASE_NEXT:
```
SJuliez, line 111: This can be shortened by using if (this == PREPHASE_NEXT) {

SJuliez (changes requested):

SJuliez (approved):

PR #4128, opened 2023-01-22
status bar update for players remaining

kuronekochomusuke:
- update status bar on reporting phases to show players who are not done.
- update status bar on deployment, targeting,  movement, firing, and physical phases to show players with next turns.
- add a advanced setting to limit the number of players for these.  set to 0 to not show any.

image

image


SJuliez:
I'm not quite sure what to look for here. The [qwret, qasf] text? If so, may I suggest adding some helpful text because it doesn't look self-explanatory to me. Like: "Next turns: " or "Unfinished players: " or something. Also, if there are more turns than shown it might make sense to add a ", ..." at the end.

rjhancock (approved):

PR #4130, opened 2023-01-22
show movement for all entity types on unit tool tip

kuronekochomusuke:
- show movement mode for all entity types on unit tool tip.  to make it easier to see what mode unit is in.  most useful for units that can convert modes.
- correct issue of movement reduced bomb indicator incorrectly showing on LAMs.

image

image

### DIFF
--- a/megamek/src/megamek/common/LandAirMech.java
+++ b/megamek/src/megamek/common/LandAirMech.java
@@ -1130,7 +1130,7 @@ public class LandAirMech extends BipedMech implements IAero, IBomber {
     @Override
     public int reduceMPByBombLoad(int t) {
         // bombs don't impact movement
-        return 0;
+        return t;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/common/LandAirMech.java
```diff
@@ -1130,7 +1130,7 @@ public void clearBombChoices() {
     @Override
     public int reduceMPByBombLoad(int t) {
         // bombs don't impact movement
-        return 0;
```
Windchild292, line 1133: That goes against the comment, and doesn't appear related?

Windchild292 (changes requested):
It looks like there's an unintended change included in this.

Windchild292:
I'm just blind apparently, but does that cause issues with the actual movement given the very obvious comment above the changed LAM line.

megamek/src/megamek/common/LandAirMech.java
kuronekochomusuke: the movement it not impacted so, so it should return the same value.

Windchild292 (approved):

PR #4140, opened 2023-01-26
Issue 2220: units dismount to adjacent hexes from SV

### LINKED ISSUE #2220: 47.9 - Unloading units from Bays in Support Vehicles is not rules-correct
On p. 207 of _Total Warfare_ (2020 edition, 6th printing), it states ...

> "Unless specifically noted otherwise, units use the same rules to mount and dismount from Support Vehicles as for mounting and dismounting from grounded aerospace units (see _Carrying Units_, p. 89, in the _Aerospace Movement_ section)."

And on p. 91 of _Total Warfare_ (2020 edition, 6th printing), it states ...

> "Units carried by Small Craft and DropShips dismount into adjacent hexes (any adjacent hex, though the hex must be within two levels of the level of the underlying hex(s) of the aerospace unit), with a facing of their choice, and the dismounting unit must be able to enter the terrain (though a unit may use the Minimum Movement rule, p. 49, to enter a hex, but it still cannot be prohibited from entering that hex). The dismounting unit may not violate the stacking limits."

I can't find anything stating that this 'dismount into adjacent hex' rule is also applied for units dismounting Support Vehicles. (While it is specifically stated that to mount a (non-Large) SV, the mounting unit must be in the same hex, there appears to be no statement that units dismount into the same hex.)

However, in Standalone MM 0.47.9, a unit (vehicle or mech) unloaded from a Vehicle Bay or Mech Bay in a Support Vehicle is unloaded into the same hex as the Support Vehicle - even where this would violate stacking limits (e.g. a vehicle being unloaded from a Large Support Vehicle). Is this an error, or have I missed an exemption somewhere?

SJuliez:
Fixes #2220


megamek/src/megamek/client/ui/swing/MovementDisplay.java
```diff
@@ -2646,7 +2646,7 @@ private synchronized void updateLoadButtons() {
         // we will say that it is not possible to unload units in such a situation.
         if (finalCoordinatesOnBoard) {
             for (Entity en : loadedUnits) {
-                if (en.isElevationValid(unloadEl, hex) || (en.getJumpMP() > 0)) {
+                if (ce.isSupportVehicle() || en.isElevationValid(unloadEl, hex) || (en.getJumpMP() > 0)) {
```
NickAragua, line 2649: Not sure about this logic - why would we ignore the valid elevation/jump MP check for support vehicles?

SJuliez: Hmmm you caught me there. It kinda works because this only activates the button while the legal unload hexes are determined elsewhere so as the worst result the list of unload positions shown to the player would be empty. Not exactly an improvement though. The elevation check is invalid for SV unload as it only checks the hex of the transport. I'll have to take a closer look. This method looks like a patchwork of stuff to me.

megamek/src/megamek/client/ui/SharedUtility.java
```diff
@@ -905,19 +905,12 @@ public static String[] getDisplayArray(List<? extends Targetable> entities) {
...
 
-    public static @Nullable Targetable getTargetPicked(List<? extends Targetable> targets,
-                                                       String input) {
-        if (input == null) {
+    public static @Nullable Targetable getTargetPicked(@Nullable List<? extends Targetable> targets,
+                                                       @Nullable String chosenDisplayName) {
+        if ((chosenDisplayName == null) || (targets == null)) {
             return null;
+        } else {
+            return targets.stream().filter(t -> chosenDisplayName.equals(t.getDisplayName())).findAny().orElseGet(null);
```
Windchild292, line 913:
```suggestion
            return targets.stream().filter(t -> chosenDisplayName.equals(t.getDisplayName())).findAny().orElse(null);
```
This will NPE if called with null. Optional::orElse is the proper method to call.

megamek/src/megamek/client/ui/swing/MovementDisplay.java
```diff
@@ -2651,120 +2651,109 @@ private void updateBombButton() {
...
-        boolean legalGear = ((gear == MovementDisplay.GEAR_LAND)
-                             || (gear == MovementDisplay.GEAR_TURN)
-                             || (gear == MovementDisplay.GEAR_BACKUP)
-                             || (gear == MovementDisplay.GEAR_JUMP));
-        int unloadEl = cmd.getFinalElevation();
-        Hex hex = ce.getGame().getBoard().getHex(cmd.getFinalCoords());
-
-        boolean finalCoordinatesOnBoard = ce.getGame().getBoard().contains(cmd.getFinalCoords());
-        boolean canUnloadHere = false;
+        final boolean canMount = isFinalPositionOnBoard() && !ce.isAirborne() && (mpUsed <= Math.ceil((ce.getWalkMP() / 2.0)))
```
Windchild292, line 2722:
```suggestion
        final boolean canMount = isFinalPositionOnBoard() && !ce.isAirborne() && (mpUsed <= Math.ceil(ce.getWalkMP() / 2.0))
```

Windchild292 (changes requested):
The logic seems solid, but there's an NPE in it.

SJuliez:
Updated.

Windchild292 (approved):

PR #4150, opened 2023-02-02
Fixes #4148 Infantry Fortify option takes one turn instead of three

### LINKED ISSUE #4148: Infantry Fortify option takes one turn instead of three (At least 0.42.2)
Discovered this last night, the code is written in a way that both dig in and fortify are taking the same amount of time to complete instead of 1 and 3 turns respectively.

I'm looking into a fix and may have a PR up soonish.

pheonixstorm:
Minor bugfix, fixes #4148


pheonixstorm:
Waiting for a little guidance from Hammer on if the turn the dig in or fortify buttons should count as the first turn or if those buttons should be moved to the physical phase to represent an end phase declaration. Will depend on how clear the TT rules are on this point.

Windchild292 (approved):

PR #4161, opened 2023-02-06
issue 4158

### LINKED ISSUE #4158: [0.49.11] Error message in phase after firing from trailer ammunition
Bundled Megamek with MekHQ 0.49.11

image

[before.sav.gz](https://github.com/MegaMek/megamek/files/10613924/before.sav.gz)
Name: Test

The error only seems to occur after switching ammo feed to trailer.

kuronekochomusuke:
Fixes #4158

- save the ammo munition type with the weapon attack action, so that it can later be used to test if it is homing ammo.  the existing check was using the ammo id from the trailer and trying to look it up on the main unit causing the error.  the new code does the look up when the ammo is chosen and later does the check with the stored value.
- correct a reporting issue with firing ammo from a trailer, which causes the entity owner to be null.   set a default color and unknown owner.


SJuliez (changes requested):
OK. Have a few comments.

Issue #3026, opened 2021-07-22
RFE - Fieldwork Vehicles


Sliprunner:
The 'Fortify' action available to create a fortified position after a few turns is presently locked to infantry only.

Is it possible to expand this feature as per TacOps errata to include vehicles mounting appropriate Fieldworks equipment? (Ex: Bulldozer or Backhoe). This would apply to the following Cannon vehicles: Brunel, Buffel, 'Engineering Vehicle', Kallon, Lesseps, O-65/66 'Oppie', Hesiod , and Rhino FST EC; Ideally it would apply to any vehicles with Bulldozer (F_bulldozer) or Backhoe (F_backhoe)

If it assists, relevant code appears to be
MovementDisplay.Java, Lines 145 and 896-907 [Controls if the fortify button is visible and what conditions it's enabled]
MoveStep.Java, Lines 2121-2152 [Defines to Fortify MoveStep]
Infantry.Java, Lines 150-157 [Enum states], and Lines 1792-1809 [Ticking 'Dug In' state]
I am unsure where it is defined how a hex is updated to 'fortified' status from the fortify command.

I understand if this admittedly niche option is not desired, Thank you for your time and effort

gsparks3:
Additional note: the ruling allowing vehicles to construct fieldworks is found in the TO (old version) Errata Document, and presumably the new TO as well, although I do not have access to that version. 

It is quite straightforward:
`Vehicles with a bulldozer, backhoe, or any piece of equipment ruled as equivalent may be considered to have the fieldworks ability, allowing them to construct field fortifications in a manner identical to Trench/Fieldworks Engineers (see p. 341).`

pheonixstorm:
This would make for a good addition.

pheonixstorm:
Adding, there is code that will update the map to change or add new terrain. It is mostly used for snow, ice, magma, and mud iirc. May be a few I haven't thought of.

Sliprunner:
Given the recent snapshot, ended up looking at this again and can see how it could be implemented, I just don't have the build environment to do so. Everything is currently contained in GameManager.Java, MovementDisplay.java, and Tank.java; I may work on this if I find time to get things setup. What I'm seeing thus far is to do these potential changes:

Tank.Java: Add dugIn int's; copy 'dugIn' progress from Infantry.newRound to Tank.newRound; copy .isEligbleFor, .setDugIn, and .getDugIn from Infantry.Java to Tank.Java
[...]

PR #4164, opened 2023-02-08
Adds ability to use tanks for fieldworks

pheonixstorm:
There is an odd glitch with the more button that will sometimes cause the more button to remain greyed out. Hitting backspace will reset this. Otherwise everything works as it should. If you move fortify into the main button space it can be used without the need for using backspace. Or the more button.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/MovementDisplay.java
+++ b/megamek/src/megamek/client/ui/swing/MovementDisplay.java
@@ -872,15 +872,14 @@ public class MovementDisplay extends StatusBarPhaseDisplay {
 
         // Infantry and Tank - Fortify
         if (isInfantry
-            && ce.hasWorkingMisc(MiscType.F_TOOLS, MiscType.S_VIBROSHOVEL)) {
+            && ce.hasWorkingMisc(MiscType.F_TRENCH_CAPABLE)) {
             // Crews adrift in space or atmosphere can't do this
             if (ce instanceof EjectedCrew && (ce.isSpaceborne() || ce.isAirborne())) {
                 getBtn(MoveCommand.MOVE_DIG_IN).setEnabled(false);
             } else {
                 getBtn(MoveCommand.MOVE_FORTIFY).setEnabled(true);
             }
-        } else if (isTank && (ce.hasWorkingMisc(MiscType.F_BULLDOZER)
-                || ce.hasWorkingMisc("Backhoe"))) {
+        } else if (isTank && ce.hasWorkingMisc(MiscType.F_TRENCH_CAPABLE)) {
             getBtn(MoveCommand.MOVE_FORTIFY).setEnabled(true);
         } else {
             getBtn(MoveCommand.MOVE_FORTIFY).setEnabled(false);
--- a/megamek/src/megamek/client/ui/swing/MovementDisplay.java
+++ b/megamek/src/megamek/client/ui/swing/MovementDisplay.java
@@ -60,7 +60,7 @@ import static megamek.client.ui.swing.util.UIUtil.uiLightViolet;
 
 public class MovementDisplay extends StatusBarPhaseDisplay {
     private static final long serialVersionUID = -7246715124042905688L;
-    
+
     // Defines for the different flags
     public static final int CMD_NONE = 0;
     public static final int CMD_MECH = 1;
@@ -134,7 +134,7 @@ public class MovementDisplay extends StatusBarPhaseDisplay {
         MOVE_SELF_DESTRUCT("moveSelfDestruct", CMD_NON_INF),
         // Infantry only
         MOVE_DIG_IN("moveDigIn", CMD_INF),
-        MOVE_FORTIFY("moveFortify", CMD_INF),
+        MOVE_FORTIFY("moveFortify", CMD_INF | CMD_TANK),
         MOVE_TAKE_COVER("moveTakeCover", CMD_INF),
         MOVE_CALL_SUPPORT("moveCallSuport", CMD_INF),
         // VTOL attacks, declared in the movement phase
@@ -319,7 +319,7 @@ public class MovementDisplay extends StatusBarPhaseDisplay {
      * A local copy of the current entity's loaded units.
      */
     private List<Entity> loadedUnits = null;
-    
+
     /**
      * A local copy of the current entity's towed trailers.
      */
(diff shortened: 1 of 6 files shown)


pheonixstorm:
This should satisfy #3026

megamek/src/megamek/client/ui/swing/MovementDisplay.java
```diff
@@ -878,9 +879,13 @@ private void updateButtons() {
             } else {
                 getBtn(MoveCommand.MOVE_FORTIFY).setEnabled(true);
             }
+        } else if (isTank && (ce.hasWorkingMisc(MiscType.F_BULLDOZER)
```
SJuliez, line 882: Suggest using EquipmentTypeLookup.BULLDOZER and EquipmentTypeLookup.BACKHOE here (I have no idea which equipment would be "ruled equivalent" as per TO, so I'd rather be specific)

(A real good use of those flags would be adding a TRENCH_CAPABLE flag to Tools, Vibroshovel, Bulldozer and Backhoe and checking that flag, but that's not something for a lil change request)

megamek/src/megamek/client/ui/swing/MovementDisplay.java
```diff
@@ -804,6 +804,7 @@ private void updateButtons() {
         }
         boolean isMech = (ce instanceof Mech);
         boolean isInfantry = (ce instanceof Infantry);
+        boolean isTank = (ce instanceof Tank);
```
SJuliez, line 807: It is possible now to do ce.isVehicle() or ce.isSupportVehicle() or ce.isCombatVehicle, so no need to cache this I guess. I suppose isVehicle is the correct one here.

megamek/src/megamek/client/ui/swing/MovementDisplay.java
SJuliez: In that case this whole if {} block could be reduced to one line

megamek/src/megamek/common/Tank.java
```diff
@@ -83,6 +84,13 @@ public class Tank extends Entity {
     public static final int CRIT_TURRET_LOCK = 13;
     public static final int CRIT_TURRET_DESTROYED = 14;
 
+    //Fortify terrain just like infantry
```
SJuliez, line 87: Since this copies the infantry code (mostly) I suggest removing it from both and placing it into Entity unless there is some good reason against that. Entity can respond with 0 or false to all such calls, overridden in Inf&Tank. This removes the need for class casts when calling dugIn().

SJuliez: This includes the methods added to Tank of course

megamek/src/megamek/server/GameManager.java
```diff
@@ -6960,6 +6960,18 @@ private void processMovement(Entity entity, MovePath md, Map<EntityTargetPair,
...
+            // check for tank fortify
+            if (entity instanceof Tank) {
+                Tank tnk = (Tank) entity;
+                if (step.getType() == MovePath.MoveStepType.FORTIFY) {
+                    if (!tnk.hasWorkingMisc(MiscType.F_BULLDOZER)
+                            && !tnk.hasWorkingMisc("Backhoe")) {
+                        sendServerChat(entity.getDisplayName()
+                                + " failed to fortify because it is missing suitable equipment");
+                    }
+                    tnk.setDugIn(Tank.DUG_IN_FORTIFYING1);
```
SJuliez, line 6972: Is this missing a return statement after the error chat message?

SJuliez (changes requested):
This will be a cool addition. I have a few comments.

megamek/src/megamek/client/ui/swing/MovementDisplay.java
pheonixstorm: Want me to change it now or do it in a separate PR where I can change all of the older code at once? Windchild292 mentioned something similar in another PR of mine about moveMode,

I will be doing a PR to change out all instances of code like `moveMode == EntityMovementMode.INF_LEG`

megamek/src/megamek/server/GameManager.java
pheonixstorm: The original code did not have it so I don't know. Also, there is another spot like this from the original code on line 6955 and several others higher up.

megamek/src/megamek/client/ui/swing/MovementDisplay.java
SJuliez: If you want you can add it to that other todo.

megamek/src/megamek/server/GameManager.java
SJuliez: Well looking at it, certainly not a return. It's strange that when the unit somehow hasn't the gear anymore it would send a message but then set the unit to FORTIFYING1 all the same. I think the problem is that this never happens with infantry, how should they lose their shovels so it's never tested. Problem for another day.

pheonixstorm: The original dig in and fortify code is from 2006, who knows why it was written that way...

megamek/src/megamek/client/ui/swing/MovementDisplay.java
pheonixstorm: Yeah, that would make it easier. F_TOOLS is a flag itself, so instead we would just need to add F_TRENCH_CAPABLE to those three equipment for now. It would also eliminate the need for testing the sub types.

megamek/src/megamek/common/Tank.java
pheonixstorm: So copy out the getter and setter along with the variables? Then add Override to the Infantry and Tank functions?

pheonixstorm: So I tried that and I run into other problems with class casts being intertwined within those blocks from other sections of code. This is something that may just require a complete cleanup as another separate PR, probably by someone with more Java knowledge than me considering how old the code is. The only section that I can quickly get in place will be for Tank, else I will never find an end to this rabbit hole.

Plus it becomes, how much do we want to bloat entity to avoid class casts. For now my vote is to leave it alone.

SJuliez: OK, then let's leave it. Entity is too big that is true, but the solution shouldn't be not using proper overrides, also it's nice not to have to use class casts. But that's for another cleanup then.

pheonixstorm:
Unless there is anything else to fix, this should be ready to merge

SJuliez (approved):

PR #4166, opened 2023-02-08
Adding Strategic Battle Force Conversion

SJuliez:
Allows SBF conversion of a company-size force. Accessed from the lobby context menu in the force view.


HammerGS (approved):
Tested and has been working fine.

PR #4167, opened 2023-02-09
Separating Client

SJuliez:
This separates Client into a game-independent Client with associated interface and a game-specific TwGameClient. The game-independent Client (which I didn't rename to AbstractClient here to avoid changes in yet more files) should work as a base for a Client for AS, BF, SBF games and handle the connection stuff. There's probably more that could be moved into the game-independent Client.


SJuliez:
This is (unfortunately) for now irredeemable from merge conflicts. Therefore closing

PR #4183, opened 2023-02-17
swap advance setting to panel height from visible row count, handles …

kuronekochomusuke:
- swap advance setting for for the weapon panel, to use panel height instead of using visible row count.  
- when using the visible row count, the panel would jump between different sizes when resizing the window.
- add additional panels, so that it handles small widths better.

image
image

### DIFF
--- a/megamek/src/megamek/client/ui/swing/unitDisplay/WeaponPanel.java
+++ b/megamek/src/megamek/client/ui/swing/unitDisplay/WeaponPanel.java
@@ -400,7 +400,6 @@ public class WeaponPanel extends PicMap implements ListSelectionListener, Action
         WeaponListMouseAdapter mouseAdapter = new WeaponListMouseAdapter();
         weaponList.addMouseListener(mouseAdapter);
         weaponList.addMouseMotionListener(mouseAdapter);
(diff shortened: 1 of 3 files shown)


megamek/src/megamek/client/ui/swing/unitDisplay/WeaponPanel.java
```diff
@@ -375,38 +374,50 @@ public void sort(Comparator<Mounted> comparator) {
...
+
         gridy++;
 
         // weapon list
         weaponList = new JList<>(new DefaultListModel<>());
         WeaponListMouseAdapter mouseAdapter = new WeaponListMouseAdapter();
         weaponList.addMouseListener(mouseAdapter);
         weaponList.addMouseMotionListener(mouseAdapter);
-        weaponList.setVisibleRowCount(GUIP.getUnitDisplayWeaponListCount());
+        //weaponList.setVisibleRowCount(GUIP.getUnitDisplayWeaponListCount());
```
Windchild292, line 403: this can be removed

Windchild292 (commented):
Some leftover code, otherwise this looks good

Windchild292 (approved):

PR #4193, opened 2023-02-19
show amount of heat over capacity on the weapon tab

kuronekochomusuke:
- show amount of heat over capacity on the weapon tab.  to help determine heat penalties and TSM heat quicker.

image


HammerGS:
Tested and working with no issues

PR #4194, opened 2023-02-19
show movement mode before move indicators

kuronekochomusuke:
- show movement mode before movement indicators
- instead of indicators ⬝ movement mode ⬝ indicators 

image


HammerGS:
Tested and working with no issues

HammerGS (approved):

PR #4199, opened 2023-02-22
Feature: Working Bridgelayer

pheonixstorm:
This PR still needs a lot of work, but it also has a few issues I need help with finding solutions to improve the code.

1. Need to set exits: This can easily be done manually I believe, but I would rather find a way to do it within the software that already exsists.
2. There is an NPE that pops up when you mouseover the new bridge hex (log below). I haven't looked into it yet so I may find a fix before it gets a once over.
3. Abstract the getter and setter and override in the appropriate entity class or just let entity handle everything as coded now?

Items still to do: TW p131
1. Create a second bridge terrain BRIDGE2 to use for bridges over water. Per rules bridges created using a bridgelayer have double the weight capacity of the CF listed only if placed on the water. If the land hex is higher it remains unchanged.
2. SV vehicles with bridgelayers can have the bridgelayer destroyed if the turret is hit.
3. Setup infantry bridging engineers. So far I have touched this yet.
4. Finish coding water bridging. So far only Heavy Bridgelayer is working

NOTE: when completed this will be closed, cleaned up, rewritten, and reposted.

### DIFF
--- a/megamek/i18n/megamek/client/messages.properties
+++ b/megamek/i18n/megamek/client/messages.properties
@@ -2208,6 +2208,7 @@ MovementDisplay.moveHullDown=Hull Down
 MovementDisplay.moveJump=Jump
 MovementDisplay.moveSwim=UMU
 MovementDisplay.MoveLaunch=Launch
+MovementDisplay.moveLayBridge=Lay Bridge
 MovementDisplay.moveLayMine=Lay Mine
 MovementDisplay.moveLoad=Load
 MovementDisplay.MoveLand=Land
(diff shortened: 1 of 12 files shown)


pheonixstorm:
Forgot a couple things.
[megamek.log](https://github.com/MegaMek/megamek/files/10799053/megamek.log)
bridge

pheonixstorm:
So I tried again and went to a new movement phase and came across some new NPEs from Princess
[megamek.log](https://github.com/MegaMek/megamek/files/10799294/megamek.log)

rjhancock:
This hasn't been touched in almost 2 years. Closing for now. When updated, please re-open.

PR #4207, opened 2023-02-25
add filters tech level and tech base

kuronekochomusuke:
- add filter tech level 
- add filter tech base

image

### DIFF
--- a/megamek/src/megamek/client/ui/swing/unitSelector/TWAdvancedSearchPanel.java
+++ b/megamek/src/megamek/client/ui/swing/unitSelector/TWAdvancedSearchPanel.java
@@ -399,11 +399,11 @@ public class TWAdvancedSearchPanel extends JPanel implements ActionListener, Ite
         }
     }
 
-    private void loadTriStateItem(String[] s, JList l, int count) {
+    private void loadTriStateItem(List<String> s, JList l, int count) {
         DefaultListModel dlma = new DefaultListModel();
 
-        for (int i = 0; i < s.length; i++) {
-            dlma.addElement(new TriStateItem("\u2610", s[i]));
+        for (int i = 0; i < s.size(); i++) {
+            dlma.addElement(new TriStateItem("\u2610", s.get(i)));
         }
 
         l.setModel(dlma);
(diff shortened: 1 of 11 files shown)


HammerGS (approved):
Testing this and works fine.

SJuliez:
Just curious: Do these two fields lead to different results from the fields we have in the unit selector main window?

kuronekochomusuke:
yes, it is different.   The new Tech Level and Tech Base come from the values set in MegaMekLab and are shown on the summary tab of the mech preview in the add a combat unit search.  

Not sure where the Type: in the add a combat unit search comes from.

kuronekochomusuke:
would think the tech level should be the same, but they are not.  

could be these are old files that need to be updated, to have the tech level set correctly.

Type: unofficial
image

Tech Level: unofficial
image

megamek/src/megamek/client/ui/swing/unitSelector/TWAdvancedSearchPanel.java
```diff
@@ -445,6 +451,8 @@ private JPanel createBasePanel() {
         loadTriStateItem(Mech.COCKPIT_STRING, listCockpitType, 7);
         loadTriStateItem(EquipmentType.structureNames, listInternalsType, 7);
         loadTriStateItem(Engine.getEngineTypes(), listEngineType, 5);
+        loadTriStateItem(SimpleTechLevel.getDescriptions(), listTechLevel, 5);
```
SJuliez, line 454: I have noticed in a previous PR but didn't comment on it. Would you mind adding a version of this that ccepts a List<String>. Lists are generally preferable to arrays for safety and general ease of handling and well, modernity. Generating that list im SimpleTechLevel is at least one method call shorter then. The getEngineTypes method would also be simpler.

SJuliez (changes requested):

megamek/src/megamek/client/ui/swing/unitSelector/TWAdvancedSearchPanel.java
SJuliez: a `List<String>`
the formatting swallowed that :)

SJuliez (approved):

PR #4227, opened 2023-03-06
Issue #3696: Prevent altitude values from influencing space combat

### LINKED ISSUE #3696: [0.48.0] Mysterious added range in space-map combat
### Description
During some of our 0.48 campaign game space battles, we have noticed that occasionally units would register as being at a higher range than they should be. The effect has been erratic, but I managed to capture it in a recent game. In this case, despite the Avenger clearly being 6 hexes from the Union PWS, the range in the weapon attack dialogue is noted as 11 hexes. The effect seems specific to certain units - all attacks against the Union PWS are registered as being 5 hexes further than they actually are, in some cases causing Out of Range notifications when visibly in range per the field of fire indicators. I am not aware of any rule that would cause this.

image

It is possible that Altitude is somehow erroneously involved, when it should have no effect on the space map - attacks at the Union PWS were noted as using the "below" table, as shown here. This makes some sense as a potential cause of the bug, as on the ground or atmospheric map Altitude affects the range to target beyond the visible hex-to-hex distance, but it should have no effect on the space map. I attempted to reproduce the bug by adjusting starting Altitude in the lobby for space-map games, but was unable to reproduce the bug deliberately, so no guarantees. 

image

### Files
Unfortunately I do not have logs, but the issue reproduces on loading the saved game, and persisted throughout the fight until the Union PWS was destroyed. It is possible that other units in the same battle showed the effect as well, but I have only confirmed it on the unit shown.
[autosave_2022-06-04_14-37-04.sav.gz](https://github.com/MegaMek/megamek/files/8840966/autosave_2022-06-04_14-37-04.sav.gz)

SJuliez:
I wasn't able to reproduce the issue. I think all Aeros keep their starting altitude of 5 in space but it seems possible to lose altitude and then air to air attacks get increased effective ranges and incorrect hit tables due to altitude difference. This PR
- adds a spaceborne check to prevent effective range from being influences by altitude in space 
- makes Entity return altitude 0 when in space (which should prevent all those issues) and give a log warning when altitude is retrieved for a spaceborne unit (since that shouldn't be done).

Fixes #3696

### DIFF
--- a/megamek/src/megamek/common/Compute.java
+++ b/megamek/src/megamek/common/Compute.java
@@ -1738,7 +1738,7 @@ public class Compute {
         }
 
         // air-to-air attacks add one for altitude differences
-        if (Compute.isAirToAir(attacker, target)) {
+        if (Compute.isAirToAir(attacker, target) && !attacker.isSpaceborne()) {
             int aAlt = attacker.getAltitude();
             int tAlt = target.getAltitude();
             if (target.isAirborneVTOLorWIGE()) {
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -13061,7 +13061,12 @@ public abstract class Entity extends TurnOrdered implements Transporter, Targeta
 
     @Override
     public int getAltitude() {
-        return altitude;
+        if (isSpaceborne()) {
+            LogManager.getLogger().warn("Altitude retrieved for a spaceborne unit!");
+            return 0;
+        } else {
+            return altitude;
+        }
     }
 
     public void setAltitude(int a) {


SJuliez:
Eh there is also the issue of wrong hit tables chosen from altitude differences.

Duuvian:
I think I found a bug related to this.
Note taken at time of bug:
Bots in a space battle "breakthrough" where an enemy dropship has to make it to the other side did not do anything and it did not advance to other units' turn to move. Kicking the bot resulted in a temporary freeze, and the next unit also did not move.

When I went to gather the logs I noticed the megamek log was just under 9 gigabytes large. I did not attach this log for this reason. On a third attempt to see if the logsize could be lowered, I closed megamek after about 20 seconds and found the attached logs. The log suggests a bug with receiving elevation for aerospace in a space map that causes a loop or something. I'll attach this smaller log file if it fits onto github though it is still massive for a text file. (Note: it does not, it's limited to 25mb and 756 MB was attempted to upload). Here is a screenshot of the 9gb logfile:

9gb log file

I attached the save file, it has no customs. To replicate the circumstances, assign all the player units in the scenario to the integrated allies bot before starting the match. The map is huge so you may want to lower that to 50x50 or something; I did this on my second attempt as I thought it could have been caused by too large of a map but the bug still occurred.

[Pleasurebots30230425.cpnx.gz](https://github.com/MegaMek/megamek/files/11352775/Pleasurebots30230425.cpnx.gz)

PR #4228, opened 2023-03-07
search updates

### LINKED ISSUE #4203: RFE - QoL stuff for Create Random Army tabs.
Looking for some QoL stuff for the Create Random Army. 
image

Would it be possible to get the red sections to be sortable A-Z/High to Low/ that type of thing.

FORMATION BUILDER
image
As above with red sections sortable. 

FORCE GENERATOR
image
As above with the red and yellow. 

In all the screens where a unit is present it would be useful to show the Name, Tech Base, BV, AS Role, and a way to pull up a selected units Unit Viewer Information.

### LINKED ISSUE #4230: Advanced Search won't find Patchwork armor
In  49.12 if searching for Patchwork no unit shows up. But the Aquagladius AQS-5 MAM and  Spindrift SDT-1A Aquatic SecurityMech have it. 

Might be related to it not being labelled in the unit summary.
image

### LINKED ISSUE #4231: [BUG] 0.49.12 Advanced Search - Some Unit Roles Missing
0.49.12

There are some unit roles missing from the Alpha Strike search options:

- Sniper
- Scout
- Missile Boat
- Skirmisher

kuronekochomusuke:
Create Random Army
- add tech base and unit role columns to unit table
- add view, view bv, view cost to unit table
- add column sort to unit table
- add tech base and unit role columns to army table
- add view, view bv, view cost to army table
- add column sort to army table
- add unit role columns to rat gen table
- add view, view bv, view cost to rat gen table
- add column sort to rat gen table
- add unit role columns to force gen table
- add view, view bv, view cost to force gen table
- add column sort to force gen table

image

AS search

-  correct issue with missing unit roles
image

Advance Search

- add patchwork filter
- add filters industrial, naval, support vehicle
- add filter aerospace fighter
- add filter Doomed On Ground
- add filter Doomed In Atmosphere
- add filter Doomed In Space
- add filter Doomed In Extreme Temp
- add filter Doomed In Vacuum

image
image

fixes #4203
fixes #4231
fixes #4230

### DIFF
--- a/megamek/src/megamek/common/MechSummary.java
+++ b/megamek/src/megamek/common/MechSummary.java
@@ -67,6 +67,9 @@ public class MechSummary implements Serializable, ASCardDisplayable {
     private String extinctRange;
     private boolean canon;
     private boolean patchwork;
+    private boolean doomedOnGround;
+    private boolean doomedInAtmosphere;
+    private boolean doomedInSpace;
     private boolean clan;
     private boolean support;
     private int walkMp;
@@ -198,6 +201,18 @@ public class MechSummary implements Serializable, ASCardDisplayable {
 
     public boolean isPatchwork() {
         return patchwork;
+    }
+
+    public boolean isDoomedOnGround() {
+        return doomedOnGround;
+    }
+
+    public boolean isDoomedInAtmosphere() {
+        return doomedInAtmosphere;
+    }
+
+    public boolean isDoomedInSpace() {
+        return doomedInSpace;
     }
 
     public boolean isClan() {
@@ -892,6 +907,18 @@ public class MechSummary implements Serializable, ASCardDisplayable {
 
     public void setPatchwork(boolean patchwork) {
         this.patchwork = patchwork;
+    }
+
+    public void setDoomedOnGround(boolean doomedOnGround) {
+        this.doomedOnGround = doomedOnGround;
+    }
+
+    public void setDoomedInAtmosphere(boolean doomedInAtmosphere) {
+        this.doomedInAtmosphere = doomedInAtmosphere;
+    }
+
+    public void setDoomedInSpace(boolean doomedInSpace) {
+        this.doomedInSpace = doomedInSpace;
     }
 
     public void setClan(boolean clan) {
--- a/megamek/src/megamek/common/MechSummary.java
+++ b/megamek/src/megamek/common/MechSummary.java
@@ -70,6 +70,8 @@ public class MechSummary implements Serializable, ASCardDisplayable {
     private boolean doomedOnGround;
     private boolean doomedInAtmosphere;
     private boolean doomedInSpace;
+    private boolean doomedInExtremeTemp;
+    private boolean doomedInVacuum;
     private boolean clan;
     private boolean support;
     private int walkMp;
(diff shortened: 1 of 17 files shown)


megamek/src/megamek/common/MechSummary.java
```diff
@@ -307,6 +316,22 @@ public boolean getMilitary() {
         return military;
     }
 
+    public boolean getIndustrial() {
```
SJuliez, line 319: MechSummary already has isIndustrialMek(), isSupportVehicle() and isAerospaceFighter(). 
isNaval() can be added to BTObject and implemented in ASCardDisplayable as movement code = "n" and in Tank there is a similar method that can be forwarded to.

SJuliez: MechSummary, AlphaStrikeElement and Entity all implement BTObject. I prefer isX() methods over ETYPE checks as methods have a comment that tells me what they return while with the ETYPE checks I never know for which units it'll come out as true without looking exactly where they are used.

SJuliez (changes requested):

megamek/src/megamek/common/MechSummary.java
kuronekochomusuke: i did it as an etype so that, i could get it to OR with the other ones correctly.

SJuliez: I don't understand. I see that you are using 
entityTypes = entityTypes | Entity.ETYPE_AEROSPACEFIGHTER
etc. and that's ok. 
But you want to add
public boolean getSupportVehicle() {
        return supportVehicle;
    }
to MechSummary when it already inherits the method isSupportVehicle which does exactly the same and ditto for AerospaceFighter and IndustrialMek. I fail to see the gain or where the existing methods couldn't be used in the place of the new ones.

SJuliez (changes requested):

megamek/src/megamek/common/MechSummary.java
SJuliez: wow this throws me, I didn't want to add a second review...

kuronekochomusuke: thanks, this make it more clear what you asking.

SJuliez (changes requested):
Just small things!

HammerGS:
I'm super curious how the Doomed filter will work? Is it for showing unit that would be instantly killed in a certain condition?

kuronekochomusuke:
yes, it uses the same value that is displayed in the lobby.   

you can do an exclude search to get only units that will not be doomed.   so if you know you are doing a space map, you can get only units to are good for space.

kuronekochomusuke:
there is also doomedInExtremeTemp and doomedInVacuum, should I add those also?

HammerGS:
Not sure how much use these specific ones with get but best to have all them.

HammerGS:
Did the requested changes get done? I think so?

kuronekochomusuke:
i changed direction for the naval into water only.  this is ready for review.

SJuliez:
@kuronekochomusuke Heya, according to our talk: The BTObject methods hasGround/Air/WaterMovement() would be a better fit for ASCardDisplayable than BTObject as they check AlphaStrike movement only. They will still be usable for MechSummary that way. Also, ideally, please add a comment to them showing the movement types checked like "Checks for h/a/i/v movement types" or something. Thanks!

PR #4250, opened 2023-03-16
Server-side Smoke Processing Fixes

### LINKED ISSUE #2102: Smoke missile smoke dissipates one turn earlier than it should
According to https://bg.battletech.com/forums/index.php?topic=69608.0 a smoke missile cloud shot on Turn 1 should clear on Turn 4. But as of 0.46.1 it goes away on Turn 3.

From what I see in the code, it's because the smoke duration counter is initially set to 3 in a fire phase, but then affected by the end phase of the same turn. Therefore, changing it to 4 should help.

### LINKED ISSUE #3896: [Bug] Weapon-deployed smoke should not drift/dissipate on the turn it is deployed.
This is possibly related to #2102, but is a different issue.

Weapon-deployed smoke should not drift or dissipate in the End Phase of the Turn in which it lands, as per [this rules query](https://bg.battletech.com/forums/tactical-operations/timing-of-initial-smoke-dissipation-check-after-placement/msg1715086/)

However, at least in 0.49.7, if the **Smoke Drift** option is enabled, the weapon-deployed smoke drifts/dissipates in the End Phase of the Turn in which it lands.

(If **Smoke Drift** is not enabled, it (correctly) does not drift/dissipate in the turn it is deployed - but as per #2102, it does dissipate early.)

(Attached save is an example - all smoke drifted, and two hexes dissipated.)

[SmokeTest.sav.gz](https://github.com/MegaMek/megamek/files/9543661/SmokeTest.sav.gz)

SJuliez:
Fixes #2102 
Fixes #3896


megamek/src/megamek/common/Board.java
```diff
@@ -607,19 +610,43 @@ public boolean contains(int x, int y) {
...
+     * Returns the Hex at the given Coords, both of which may be null.
+     *
+     * @param coords the Coords to look for the Hex
      * @return the Hex at the specified Coords, or null if there is not a hex there
      */
-    public @Nullable Hex getHex(final @Nullable Coords c) {
-        return (c == null) ? null : getHex(c.getX(), c.getY());
+    public @Nullable Hex getHex(final @Nullable Coords coords) {
+        if (coords == null) {
+            LogManager.getLogger().warn("Method called with null Coords!");
```
NickAragua, line 628: Just gotta warn you, this will slow down processing in "performance sensitive areas" everything even if the debug level is set so that the logging won't occur - so I don't know if we actually want to be doing this logging.

NickAragua (changes requested):
Normally, I'm all about logging, but for a method that's called as often as getHex(), I think we want to err on the side of caution, even if it's just logging for situations with null coordinates.

megamek/src/megamek/common/Board.java
SJuliez: Yeah, removed it. Looked good on paper.

NickAragua (approved):

PR #4256, opened 2023-03-19
issue3968 multiple units

kuronekochomusuke:
- when there are multiple units in a hex, display Multiple Units... instead of one of the names of the units in the space.

image

image

### DIFF
--- a/megamek/i18n/megamek/client/messages.properties
+++ b/megamek/i18n/megamek/client/messages.properties
@@ -489,6 +489,7 @@ BoardView1.EVADE=EVADE
 BoardView1.ROLLED=ROLLED
 BoardView1.sensorReturn=Sensor Return
 BoardView1.LowFoliage=FOLIAGE
+BoardView1.multipleUnits=Multiple Units...
 
 BoardView1.Tooltip.AbbreviationRear=R
 BoardView1.Tooltip.AbbreviationSensors=S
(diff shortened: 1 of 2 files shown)


Thom293:
Very cool!

NickAragua:
In most cases, there are only two units that can be stacked in a hex - in which case, I'd rather have an extra line of text telling me what's in there instead of having to hover over the hex. There are some exceptions such as air/space battles or infantry/turrets in which case those restrictions are ignored. Can we set the threshold at "more than two units in a hex" before we display this message?

kuronekochomusuke:
would need to change the code to display the name for the 2 units.  currently it only displays the name of one of them.

gsparks3:
Is this options-based? I'm pretty sure I've seen the game display nameplates for all units in a hex with multiple units in said hex. Or is it specific to turrets only?

kuronekochomusuke:
it looks like they are not get the #number added to the name when there are multiple turrets with the same name.  this may be causing the issue.

kuronekochomusuke:
- allow up to 4 units before Multiple Unit.. and allow turrets to display multiple

image

image

kuronekochomusuke:
i'm not seeing a option for the multiple.  there was an issue with getting the entities in the current hex, with turrets.

gsparks3:
I checked, in 49.12 with all default settings it displays multiple nameplates. No idea why you don't see them, Thom, outside of not having multiple units in the same hex.

Thom293:
Thanks. Will double check tomorrow.

PR #4268, opened 2023-03-23
Berserk Princess

Thom293:
Increases Princess Aggression on highest 3 notches of the Aggression Slider.  A full Berserk Setting is now truly Berserk.  On the highest setting, she will attempt to move mechs to zero range and kick your face in.  The next two settings do increase Aggression and Close range mechs actually move to close range and fire.  The long range mechs will still not approach though.  This may change against different enemies, but both settings are an improvement over default.  Only the last 3 settings on the slider were changed.  The remainder were unchanged, so this only applies for people that crank up aggression.  I just changed the Hyper_Aggression_Values setting.  Thats it. 

NOTE:  All tests were against a Union C grounded Dropship.  Her behavior may change against higher or lower firepower.

WHY:  As reported in the community and confirmed by tests by @PhoenixHeart512 and I, Princess was not aggressive against high firepower, even on the highest aggression settings.  Formerly, against massive firepower, Princess would not properly approach and often get "stuck" on enemy range brackets and not approach past certain damage/range thresholds, even with short range mechs.  This was a problem.  This hopefully fixes or at least reduces that problem.

Changes:
5 --> 500.  The highest setting.  She will move to zero with all mechs - even Sniper-types, and attempt to kick your face in.
4.5 --> 50.  Second Highest.  She will move short range mechs into firing range and fire.  Long range mechs still got stuck at extreme range, at least against a Union C.
4.0-->10.  Third Highest.  She will move short range mechs into firing range and fire.  Long range mechs still got stuck at extreme range, at least against a Union C.  This behavior was identical to the one above, but I presume will change slightly depending on the enemy.  Its still an improvement over default though.

Old Behavior:  

image

New Behavior on Max Setting - red and dark blue team have new setting.

image

### DIFF
--- a/megamek/src/megamek/client/bot/princess/BehaviorSettings.java
+++ b/megamek/src/megamek/client/bot/princess/BehaviorSettings.java
@@ -80,9 +80,9 @@ public class BehaviorSettings implements Serializable {
             2.5,
             3,
             3.5,
-            4,
-            4.5,
-            5};
+            10,
+            50,
+            500};
     static final double[] HERD_MENTALITY_VALUES = {
             0.1,
             0.2,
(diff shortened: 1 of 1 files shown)


Thom293:
Hhhmmm.  I only changed the 3 numbers listed in my post.  I dont know why it shows 11 changes.  I did not make those others.  I dont know why it is showing them.  I had updated my repo before this, but perhaps it is out of sync?  If there is something I need to change, please let me know.

PR #4269, opened 2023-03-23
Remove anti aliasing as a gui option; use high quality rendering

SJuliez:
As the title says. Having AA as an option seems very outdated and the option was not even checked consistently and not at all in MML/MHQ. So this PR removes it in favor of general high quality rendering.


Windchild292 (approved):
Approved, but can you also remove the lines from the resource files?

SJuliez:
Yup.

PR #4270, opened 2023-03-23
issue4267 copy initiative to ejected crew

### LINKED ISSUE #4267: IllegalArgumentException thrown mid-combat
### Environment
        Starting MegaMek v0.49.12
        Build Date: 2023-03-04T15:27:34.905712500
        Origin Project: MekHQ
        Java Vendor: Homebrew
        Java Version: 11.0.18
        Platform: Mac OS X 13.2.1 (aarch64)
        System Locale: en_US

### Description
While in combat in MegaMek, MekHQ throws a java.lang.IllegalArgumentException, causing MegaMek to stop. Reloading the previous round's save does not allow me to proceed.

### Files
MegaMek autosave file, Campaign Savegame, custom units, and logs are in issue_report_files.tgz

[issue_report_files.tgz](https://github.com/MegaMek/mekhq/files/11045090/issue_report_files.tgz)

kuronekochomusuke:
- copy initiative to ejected crew
- log if entity does not have initiative rolled before determining turn order.

fixes #4267

### DIFF
--- a/megamek/src/megamek/server/GameManager.java
+++ b/megamek/src/megamek/server/GameManager.java
@@ -2763,7 +2763,10 @@ public class GameManager implements IGameManager {
             entities = game.getEntitiesVector();
         }
 
-        String noInitiative = entities.stream().filter(e -> e.getInitiative().size() == 0).map(Object::toString).collect(Collectors.joining(";"));
+        String noInitiative = entities.stream()
+                .filter(e -> e.getInitiative().size() == 0)
+                .map(Object::toString)
+                .collect(Collectors.joining(";"));
 
         if (!noInitiative.isEmpty()) {
             LogManager.getLogger().error("No Initiative rolled for: " + noInitiative);


megamek/src/megamek/server/GameManager.java
```diff
@@ -2762,6 +2762,13 @@ private void determineTurnOrderIUI(GamePhase phase) {
         } else {
             entities = game.getEntitiesVector();
         }
+
+        String noInitiative = entities.stream().filter(e -> e.getInitiative().size() == 0).map(Object::toString).collect(Collectors.joining(";"));
```
Windchild292, line 2766: Can we split this into four lines for readability? filter, map, and collect would be better done on separate lines.

megamek/src/megamek/server/GameManager.java
```diff
@@ -2762,6 +2762,13 @@ private void determineTurnOrderIUI(GamePhase phase) {
...
+
+        if (!noInitiative.isEmpty()) {
+            LogManager.getLogger().error("No Initiative rolled for: " + noInitiative);
+        }
```
Windchild292, line 2770: Should we escape or exit out here?

Windchild292 (commented):
Two comments first

megamek/src/megamek/server/GameManager.java
kuronekochomusuke: it mostly seems to work if there there is no initiative is assigned.   just seems to error when it  interacts with the other initiatives just right.   just putting this in to hopefully help if somewhere else is not assigning it.

also just escaping or erroring here would probably break the turn, since nothing would be assigned to the turn object.

kuronekochomusuke: i tried filtering out the ones with no initiative, but it was then causing errors later when checking for the turn in the entities.

SJuliez (approved):

PR #4278, opened 2023-03-30
AS Conversion and Cards

SJuliez:
see #3961 
- Corrects the slightly different background colors used on the cards
- IS ER Proto LL now uses 15 heat for AS
- BA with Artillery tubes now receive ARTBA
- Allows armor pip wrapping for high armor values for aeros and improves the armor pip arrangement when wrapping

Changes are per commit


HammerGS:
Pulled the branch and tested and everything looks good to me.

PR #4282, opened 2023-03-31
add button to mini map that shows and hides the symbols on the map

kuronekochomusuke:
- add button to mini map that shows and hides the symbols on the map
- add popup menu for the zoom, height, and symbol actions

S = show symbols
image
N = no symbols
image

image

### DIFF
--- a/megamek/i18n/megamek/client/messages.properties
+++ b/megamek/i18n/megamek/client/messages.properties
@@ -2193,6 +2193,8 @@ Minimap.GroundHeightLabel=G
 Minimap.NoHeightLabel=N
 Minimap.TotalHeightLabel=T
 Minimap.X=X
+Minimap.SymbolsLabel=S
+Minimap.NoSymbolsLabel=N
 
 #Mini round report display
 MiniReportDisplay.Damage=Damage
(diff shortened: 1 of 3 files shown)


SJuliez:
I'm not so sure about this one. I imagine we're going to get an issue at some point saying "minimap broken: shows no units" because that button is easy to miss.

HammerGS:
I think this is an interesting idea, but agree with Juliez. I've wondered if there was an overlay or tooltip that could be created to show what options are on with the minimap.

kuronekochomusuke:
how about a menu for the buttons on the bottom? 

think most people don't know about the existing button for showing the hex height.  and sometimes they can be cut off when you zoom out with a big map.

gsparks3:
I definitely was not aware that you could display elevation on the minimap like that. Making all the options more obvious could be useful.

HammerGS:
> how about a menu for the buttons on the bottom?
> 
> think most people don't know about the existing button for showing the hex height. and sometimes they can be cut off when you zoom out with a big map.

I think that would be a great idea.

HammerGS:
Just tested this and the menu makes a big difference. My only issue was the map opened at the smallest zoom level and the menu option to increase zoom didn't work. But once I manually increased it, it started working.

kuronekochomusuke:
where you trying to zoom in out zoom out?  at the smallest size you need to zoom in to make it bigger.  zoom out is disabled, since that is the smallest size.

HammerGS:
> were you trying to zoom in or zoom out? at the smallest size you need to zoom in to make it bigger. zoom out is disabled, since that is the smallest size.

That might have been it. I rarely play with the Mini-map

SJuliez (approved):

PR #4288, opened 2023-04-05
Optional "Ignore Nag" button to allow ready() call even if getNagForNoAction is enabled

pakfront:
The "Confirm done if no action" optional nag dialog is very useful, but gets tedious fast. This PR adds a new button to the phase display button bar that allows you to skip the dialog. 
The intent is to allow a user have similar functionality to the nag with clearer feedback and less mouse movement. It only appears if the no-action-nag option is checked on.
This would apply to the phases currently using the no-action-nag option: Move, Targeting, Firing and Physical.

If you have not entered any actions the bar will look like the image below. The usual Done button is disabled, but the new Skip is enabled.  Clearing the action stack (with ESC or Backspace) will also result in this. 
noaction

Once you have entered actions, it will look like the image below. The Done button is enabled, the Skip button is disabled.
action

In addition, even with the option off, I have made an attempt to have the Done button give more feedback indicating if pressing it will result in a no action turn.
nomove
move

Alternatively I could make it so the Done button is always enabled, and, same as now, pressing it with no actions entered will result in the nag. The new Skip button only be enabled if there are no actions, but force skip the nag.

The new Skip button has the keyboard command CTRL-SHIFT-Enter

I also added some superscript symbols to the Move button to indicate if a PSR (*) or potential damage (!) will happen. I'd like to add one for collapsing buildings/basements as well but that would take a bit of work in MoveStep in a future PR.

Code notes - two new classes to handle the new feature 
ActionPhaseDisplay handles the updating of the Done and Ignore Nag buttons for the Move, Target, Fire and Physical phases.
AttackPhaseDisplay moves some code that was shared across the Target, Firing and Physical displays, notably handing the attacks queue. Probably more could be moved into it from those subclasses.
Other phase display classes have a few common things, like tooltips, moved into StatusBarPhaseDisplay super class. This super class also has some large functions split into smaller ones so they can be overriden.


pakfront:
This is ready for review.

SJuliez:
I notice that the text Skip Movement is somehow written behind that button:
image

pakfront:
> I notice that the text Skip Movement is somehow written behind that button: 
good catch, I'll look into it

kuronekochomusuke:
looks like the tooltips are not set correctly on these buttons.    Move looks to have the DONE_NO_ACTION one and Skip Move done not have a tooltip.

image

pakfront:
> I notice that the text Skip Movement is somehow written behind that button: 
I'm not able to repro that. Did you have any client settings that are not default such as number of buttons? Are you on Windows?

SJuliez:
I am on Windows and have I believe 12 buttons per row or 15.

megamek/src/megamek/client/ui/swing/ActionPhaseDisplay.java
```diff
@@ -0,0 +1,172 @@
...
+            } else {
+                butDone.setText("<html><b>" + noActionLabel + "</b></html>");
+            }
+        }
+        butIgnoreNag.setText("<html><b>" + noActionLabel + "</b></html>");
+        updateDonePanelEnabled();
+    }
+
+    protected void updateDonePanelEnabled() {
+        // Note for reviewers: currently this toggles between the two buttons being active,
```
SJuliez, line 159: Better remove this comment

megamek/src/megamek/client/ui/swing/MovementDisplay.java
```diff
@@ -976,12 +972,33 @@ private void updateAeroButtons() {
...
+        } else {
+            boolean psrCheck = (!SharedUtility.doPSRCheck(cmd).isBlank())
+                    || (!SharedUtility.doThrustCheck(cmd, clientgui.getClient()).isBlank());
+            boolean damageCheck = cmd.shouldMechanicalJumpCauseFallDamage()
+                    || cmd.hasActiveMASC()
+                    || (!(ce() instanceof VTOL) && cmd.hasActiveSupercharger())
+                    || cmd.willCrushBuildings();
+
+            String moveMsg =  Messages.getString("MovementDisplay.Move")
+                    + "<sup>"
```
SJuliez, line 989: The PSR * looked a bit misplaced on the button to me. Maybe remove the sup?

megamek/src/megamek/client/ui/swing/TargetingPhaseDisplay.java
```diff
@@ -195,16 +192,22 @@ public TargetingPhaseDisplay(final ClientGUI clientgui, boolean offboard) {
...
-        butDone.setEnabled(false);
-
         setupButtonPanel();
 
         registerKeyCommands();
     }
 
+    @Override
+    protected void setDoneButtonValid() {
+        // could add cmd
```
SJuliez, line 202: Not sure what this comment means

megamek/src/megamek/client/ui/swing/util/KeyCommandBind.java
```diff
@@ -77,6 +77,7 @@ public enum KeyCommandBind {
     /** Used to cancel moves/fires/chatterbox */
     CANCEL("cancel", false, VK_ESCAPE, 0, true),
     DONE("done", false, VK_ENTER, CTRL_DOWN_MASK, true),
+    DONE_NO_ACTION("doneNoAction", false, VK_ENTER, CTRL_DOWN_MASK | SHIFT_DOWN_MASK, true),
```
SJuliez, line 80: Please add an entry for this to mmconf/defaultkeybinds.xml. We've had fresh MM downloads fail to run because the xml entry for a new keybind was missing even though it worked perfectly in the IDE.

megamek/src/megamek/client/ui/swing/AttackPhaseDisplay.java
```diff
@@ -0,0 +1,74 @@
...
+
+public abstract class AttackPhaseDisplay extends ActionPhaseDisplay {
+    // client list of attacks user has input
+    protected Vector<EntityAction> attacks;
+
+    protected AttackPhaseDisplay(ClientGUI cg) {
+        super(cg);
+    }
+
+    abstract protected void setDoneButtonValid();
```
SJuliez, line 35: It would be good form to add a comment what an abstract method is supposed to be doing when overridden and/or when it is called.

SJuliez (changes requested):
Sorry for not looking at this for so long so it picked up a conflict. I have a few minor requests.

megamek/src/megamek/client/ui/swing/TargetingPhaseDisplay.java
pakfront: old note to self, removed

megamek/src/megamek/client/ui/swing/AttackPhaseDisplay.java
pakfront: Nothing like writing docs to see how it could be simplified. I moved most of the repeated button managing into a single method on the ActionPhaseDisplay class. Now the child of classes handle the labeling and enabling of the Done and Skip buttons while the super class does the bookkeeping. This also makes it more reliable to reset the button labels and states after a change in client settings.

megamek/src/megamek/client/ui/swing/util/KeyCommandBind.java
pakfront: @SJuliez How do I add a binding that has multiple modifier keys, in this case "ctrl+shift enter?"

SJuliez (approved):

PR #4295, opened 2023-04-08
About-Dialog: Refactor and improvements

SJuliez:
Not that this is terribly important but since the About dialog had a few issues I gave it a rework.

image


HammerGS:
Looks good to me, one tweak the term Classic BattleTech isn't used anymore. It's just BattleTech or use Table top BattleTech.

PR #4296, opened 2023-04-08
client settings updates, convert advanced settings

kuronekochomusuke:
- convert some of advanced settings
- add Game Board tab
- add Mini Map tab
- add Report tab
- add Overlays tab
- clean up code to use getX() functions
-

image
image
image
image
image
image
image


HammerGS:
Downloaded this and I think it's a great idea. I'm wondering if we could do something about this.
image

We show the list vertical but in game it's horizontal. Would there be away to display what it would look like in game?

image

kuronekochomusuke:
there is a setting for the JList setLayoutOrientation(JList.HORIZONTAL_WRAP) that will display them horizontally, but i am not seeing a good may to have the number of columns be equal the number of buttons per row setting.  

it may be more confusing without the columns set correctly.  probably need to do it in a separate PR, since it may need to be switched over to a JTable to do in right. 

I have 12 buttons per row, which is 2 rows of buttons, but here it is doing (8, 8) instead of (12,4)
image

HammerGS:
Worth a try. I know its a setting more players should use and look at to help with some of the UI issues.

kuronekochomusuke:
thinking about it more, not all of them are shown at the same time any way.   some are only shown for certain unit types, so it may not matter that much.

HammerGS:
> thinking about it more, not all of them are shown at the same time any way. some are only shown for certain unit types, so it may not matter that much.

Yeah, another thing I've long thought of on this screen a way to define configuration by unit type, but I think that would be a nightmare to implement and possibly use.

HammerGS:
Another thought on this, would it be possible to do a search like we do for game options?

kuronekochomusuke:
think most of it could work similar to the game options.  

some parts may not be easily done, like this part.

image

HammerGS:
Would it be finding the Check box sections, or the options that depend on the check boxes being selected? I would think just finding the check boxes would be good.

kuronekochomusuke:
the game options are built dynamically, that allows the search to be done easier.  would need to convert the client settings to built dynamically.    

would also need to figure out if the above could be built dynamically without too much trouble.  the different java control types used and the dependencies.

SJuliez (approved):

Issue #3813, opened 2022-07-25
Bug: 49.8 Repeating TSEMP does not repeat


MarcJC:
[CWS Bugged.zip](https://github.com/MegaMek/megamek/files/9184809/CWS.Bugged.zip)
(EDIT - fixed zip to include legacy log with actual data)

I played a game of MM where I had a custom mech with a repeating TSEMP 
The weapon should be usable each round (as it repeats - and also blows up on a 2) but the weapon only fired every other turn as it had to recharge like the standard TSEMP cannon

Saklad5:
Any progress on fixing this? It shouldn't be that hard: perhaps I could take a stab at it?

Saklad5:
@HammerGS I think I've fixed this specific issue, though I'd like to add the downside too before merging.

I should be able to model this off the code for PPC capacitors, right?

Saklad5:
This should be closed, right?

PR #4298, opened 2023-04-09
Implement repeating TSEMPs

Saklad5:
Repeating TSEMPs, as the name implies, can fire every turn.

Like any self-respecting RISC project, they also explode: in this case, on a roll of 2.

### DIFF
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -6399,13 +6399,13 @@ public abstract class Entity extends TurnOrdered implements Transporter, Targeta
             setTsempEffect(MMConstants.TSEMP_EFFECT_NONE);
         }
 
-        // TSEMPs can fire every other round, so if we didn't fire last
-        //  round and the TSEMP isn't one-shot, reset it's fired state
+        // Standard TSEMPs can fire every other round, so if we didn't fire last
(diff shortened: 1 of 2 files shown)


Saklad5:
Closes https://github.com/MegaMek/megamek/issues/3813

SJuliez:
Heya, @Saklad5 the code looks alright. Have you given this a test run by setting the roll to 2?

Saklad5:
> Heya, @Saklad5 the code looks alright. Have you given this a test run by setting the roll to 2?

Not yet. Could you explain how to do so?

SJuliez:
I mean adding a line to the code that makes sure the roll happens to be 2 (e.g. just before the check, or commenting out the check for testing, I mean you can do without but might have to shoot a lot to get it to explode); then starting a game with a unit wielding said TSEMP and seeing if it explodes in a satisfying way and gives a fitting report message and the unit behaves the round after like it should (not being able to fire the TSEMP again and so on)

NickAragua:
@Saklad5 did you wind up running that test to get it to explode?

Saklad5:
@NickAragua 
image

It worked fine, though it turned out there was an issue with the downtime skip. I fixed that, so I'll vouch for this being good to go.

SJuliez (approved):

PR #4304, opened 2023-04-09
Reports

SJuliez:
For generating Reports, this PR adds static calls and the ability to chain call settings, so that e.g.

```
r = new Report(7030);
r.type = Report.PUBLIC;
r.newlines = 0;
reports.addElement(r);
```

becomes
`reports.addElement(Report.publicReport(7030).noNL());`

See the code changes for some examples how this makes some code much more readable and to the point (imo)


NickAragua (approved):
I suspect the ease/difficulty of reading/writing report code is in the eye of the beholder, but if having the option to do it this way makes it easier for you to read/write, it's good enough for me.

PR #4310, opened 2023-04-11
show a chat message when changing settings

kuronekochomusuke:
show a chat message when changing settings to show who made the change
- when changing selected map
- when changing map type (ground, low altitude, space)
- when changing bot settings
- when changing skill generation settings
- when adding units from mul
- when adding units from add a combat unit
- when adding units from create random army
- when adding units from paste from clipboard

image


SJuliez (changes requested):
OK, one request: Please change the msg code so that messages say [12 units] instead of [units 12]

SJuliez (approved):

PR #4311, opened 2023-04-11
Add Sneede's Hybrid Rifleman variants from the Snord's Irregulars focepack

pavelbraginskiy:
I followed the record sheets for the forcepacks, here: https://bg.battletech.com/download/ForcePack%20Record%20Sheets%20Snords%20Irregulars.pdf

The existing unofficial Rifleman Sneede-FLU is almost identical to the newly official Hybrid Rifleman RFL-3N (Sneede), but the Sneede-FLU was created with FrankenMech construction rules while the new one is an approximation under standard construction rules.


HammerGS:
Per Greekfire sheets these are built off are wrong updated sheets at some point.

PR #4313, opened 2023-04-11
add more client settings

kuronekochomusuke:
add more client settings

- main
 caution and precaution color (align colors between unit tool tip, unit over view and entity sprite)
- game board 
 add field of fire colors
 add combo box for move font type and style
 unit text, unit selected, unit valid colors
- overlay 
  key binds settings to use planetary condition settings
 more unit overview settings

image
image
image


SJuliez:
I appreciate moving the advanced settings to accessible settings. IMHO some of these may be a bit of overkill, seeing that we already have a ton of settings. Personally, I'd make the overlay settings the same for keybinds and weather as far as possible (text colors, show header). Also the colors on the unit overview should be the same as those used on labels (if they aren't already).
@HammerGS What do you think?

For font selection, there's a FontComboBox that you might want to use. Alternatively you can copy the code from ConfigurableASCardPanel for a less involved font chooser.

kuronekochomusuke:
which labels?

HammerGS:
> I appreciate moving the advanced settings to accessible settings. IMHO some of these may be a bit of overkill, seeing that we already have a ton of settings. Personally, I'd make the overlay settings the same for keybinds and weather as far as possible (text colors, show header). Also the colors on the unit overview should be the same as those used on labels (if they aren't already). @HammerGS What do you think?
> 
I'm a big fan of consistent coloring whenever possible. I do like how this is shaping up but will be a big change for players. I almost wonder if helper text where possible wouldn't be a bad idea

SJuliez:
> which labels?

The unit labels on the map. They also get colored by valid, moved etc. They probably use the same colors as the unit overview already. But might be good naming this accordingly in the settings dialog.

SJuliez:
I have no objections to the code. So comment when you'll finished with any remaining changes.

kuronekochomusuke:
im done, unless yall have some specific ones that need tooltips.

megamek/src/megamek/client/ui/swing/boardview/FieldofFireSprite.java
```diff
@@ -65,18 +63,35 @@ public class FieldofFireSprite extends MovementEnvelopeSprite {
...
+        Color c = getFieldOfFireColor(rangeBracket);
+        fillColor = new Color(c.getRed(), c.getGreen(), c.getBlue(), borderOpac);
         this.rangeBracket = rangeBracket;
     }
-    
+
+    public static Color getFieldOfFireColor(int rangeBracket) {
+        // colors for Min,S,M,L,E ranges
+        switch (rangeBracket) {
+            case 0:
```
NickAragua, line 80: If practical, can we re-use the existing range bracket constants rather than magic numbers?

NickAragua (commented):
Looks good, just one potential request to avoid magic numbers.

NickAragua (approved):

PR #4318, opened 2023-04-13
Replace Target Selection Combobox with more detailed dialog.

pakfront:
The current dialog that pops up when multiple units are in a hex is not detailed. That makes it hard to choose which target you want to target, especially if they have similar names. This PR replaces most of the uses of the combobox dialog with a more useful dialog that has two modes: a summary mode and a detailed mode, which can be toggled with a button in the top left.

Summary Mode:
summary1
summary2

Details Mode:
details1
details2

In addition, I have generalized the dialog so it can do multi-selection and handle various objects types. For example, I have changed the load/unload tools to use this new dialog as well. I intend to use this dialog or a derivative for use with some GM tools. 

I also added target info to the "Aim At" Dialog.
aimat


pakfront:
This is ready for review.

kuronekochomusuke:
would it be better for inGameValues to call getOneLineSummary to get that part?  so that the code is not duplicated it 2 spots.

pakfront:
> would it be better for inGameValues to call getOneLineSummary to get that part? so that the code is not duplicated it 2 spots.

The HTML formatting between the one line and the ingamevalues usage is different enough that I chose to separate it.

SJuliez:
I like this dialog. Much better than the old combobox. A few observations:
- ClientDialog is pretty dated. The choice dialog could be based off of AbstractDialog or AbstractButtonDialog which would mean the dialog would remember its position and size. Not sure if this is useful for such a dialog though. Might be better to just show it centered.
- It would be good to spend some effort on comments for methods that need an override.
- Could the showDialog() be rewritten as an override of setVisible()? Less chance for error in using the dialog. 
- It seems like some ActionListener definitons could be replaced by xyzButton.addActionListener(() -> whatEverMethod());
- I'm not sure if the EntityChoiceDialog is necessary as the Targetable one would already work for Entity. A downside is that the return value is List<Targetable> instead of List<Entity>, which is an inconvenience.
- Does the dialog handle a bigger number of targets well when showing details (i.e. the dialog gets really long)?

pakfront:
Thanks for the review, very helpful.
> * ClientDialog is pretty dated. The choice dialog could be based off of AbstractDialog or AbstractButtonDialog
I switched to using AbstractButtonDialog, and that works wll enough, though unlike my previous implementation it leaves an OK button there even if single select is ON. This is harmless.

> * It would be good to spend some effort on comments for methods that need an override.
I've added some, let me know if you thin it needs more.

> * Could the showDialog() be rewritten as an override of setVisible()?
Switching super class to AbstractButtonDialog now uses the AbstractButtonDialog class behavior.

> * It seems like some ActionListener definitons could be replaced by xyzButton.addActionListener(() -> whatEverMethod());

> * Possibly, the entity choice dialog could override the Targetable one with less effort?
I think having individual classes for each is preferable. There is not too much duplicated code and it simplifies usage of each and helps prevent code errors with strict typing and less casting.

> * Does the dialog handle a bigger number of targets well when showing details
Yes, the button panel is inside a scrollbar and uses the new AbstractDialog fit() I added to ensure it stays in the screen. Fit() resizes the dialog, but only re-centers it if it would go out of bounds. This gives better behavior for this window.

megamek/src/megamek/client/ui/swing/boardview/BoardView.java
```diff
@@ -5750,25 +5743,25 @@ public void appendTerrainTooltip(StringBuffer txt, @Nullable Hex mhex) {
...
-                String msg_tf =  Messages.getString("BoardView1.Tooltip.TF");
-                name += (tf > 0) ? " (" + msg_tf + ": " + tf + ")" : "";
-                sTerrian += name + "<BR>";
-            }
-        }
-
-        result += guiScaledFontHTML(UIUtil.uiBlack()) + sTerrian + "</FONT>";
-        txt.append(result);
+        txt.append(HexTooltip.getTerrainTip(mhex));
+//        Coords mcoords = mhex.getCoords();
```
NickAragua, line 5747: Let's get this cleaned up

megamek/src/megamek/client/ui/swing/tooltip/HexTooltip.java
```diff
@@ -0,0 +1,172 @@
+package megamek.client.ui.swing.tooltip;
```
NickAragua, line 1: Think this needs the standard megamek copyright header

megamek/src/megamek/client/ui/swing/TargetChoiceDialog.java
```diff
@@ -0,0 +1,159 @@
...
+        this.firingEntity = firingEntity;
+        // initialize must be called after all member variables set
+        initialize();
+    }
+
+    @Override
+    protected void detailLabel(JToggleButton button, Targetable target) {
+        if (target instanceof Entity) {
+            button.setText("<html>" + infoText(target) + UnitToolTip.getEntityTipVitals((Entity) target, null) + "</html>");
+        } else if (target instanceof BuildingTarget) {
```
NickAragua, line 78: I seem to recall that there's a TargetType property that covers this sort of thing; I could be wrong though.

NickAragua (changes requested):
Looks good, just needs a little cleanup and one potential set of changes that can be left to your discretion as the performance impact is probably negligible (given that it's a user interaction).

megamek/src/megamek/client/ui/swing/TargetChoiceDialog.java
pakfront: There is a Targetable.getTargetType() but it is more specific and maybe not useful for this code, which is just passing onto the Tooltip helpers to handle the details. 
For instance,  BuildingTarget can have several TargetTypes, including TYPE_BLDG_IGNITE, TYPE_BLDG_IGNITE, TYPE_BLDG_TAG. Does this code care a new one is added or removed or the flags get redone? I don't think it does, so keeping the instanceof is more maintainable. But, of course, will do as you guys think correct.

pakfront: Actually - I think maybe I should move this if statement into a static method in tooltips, as I have encountered another case where I want to handle this. Please hold off on review (sorry, can;t seem to turn this back into a draft)

pakfront: @NickAragua What do you think of the keeping the instanceof for maintainability? Like I said, either way is fine with me, just let me know.

megamek/src/megamek/client/ui/swing/TargetChoiceDialog.java
```diff
@@ -0,0 +1,163 @@
...
+        this.firingEntity = firingEntity;
+        // initialize must be called after all member variables set
+        initialize();
+    }
+
+    @Override
+    protected void detailLabel(JToggleButton button, Targetable target) {
+        button.setText("<html>" + infoText(target) + UnitToolTip.getTargetTipDetail(target,  clientGUI.getClient().getBoard(), clientGUI) + "</html>");
+
+//        if (target instanceof Entity) {
```
NickAragua, line 78: Not really a fan of this commented out code.

megamek/src/megamek/client/ui/swing/TargetChoiceDialog.java
```diff
@@ -0,0 +1,163 @@
...
+//        } else {
+//            summaryLabel(button, target);
+//        }
+    }
+
+    @Override
+    protected void summaryLabel(JToggleButton button, Targetable target) {
+        button.setText("<html>" + infoText(target) + UnitToolTip.getTargetTipSummary(target,  clientGUI.getClient().getBoard()) + "</html>");
+
+//        String result = infoText(target);
```
NickAragua, line 93: Clean up in aisle 93-99

NickAragua (changes requested):
Just a little cleanup.

NickAragua (approved):

PR #4319, opened 2023-04-13
add original armor and internal structure values

kuronekochomusuke:
- add original armor and internal structure total values

image


HammerGS:
Tested and seems to work.

SJuliez (approved):

PR #4338, opened 2023-04-18
add weapon quirks to unit summary panel

### LINKED ISSUE #2333: Show Weapon Quirks in various tool tips.
Currently in the various views we only show chassis quirks can we get the weapon quirks to show. Would be 

### Environment
What version of MegaMek does your issue apply to?  47.11
What operating system are you using?  Windows, 

image  
image

kuronekochomusuke:
- add weapon quirks to unit summary panel

image
image

fixes #2333

### DIFF
--- a/megamek/src/megamek/common/MechView.java
+++ b/megamek/src/megamek/common/MechView.java
@@ -453,30 +453,34 @@ public class MechView {
             }
         }
         sBasic.add(new SingleLine());
-        
-        StringJoiner quirksList = new StringJoiner("<br/>\n");
-        Quirks quirks = entity.getQuirks();
-        for (Enumeration<IOptionGroup> optionGroups = quirks.getGroups(); optionGroups.hasMoreElements();) {
-            IOptionGroup group = optionGroups.nextElement();
-            if (quirks.count(group.getKey()) > 0) {
-                for (Enumeration<IOption> options = group.getOptions(); options.hasMoreElements();) {
-                    IOption option = options.nextElement();
-                    if (option != null && option.booleanValue()) {
-                        quirksList.add(option.getDisplayableNameWithValue());
-                    }
-                }
-            }
-        }
-        if (quirksList.length() > 0) {
-            ItemList list = new ItemList(Messages.getString("MechView.Quirks"));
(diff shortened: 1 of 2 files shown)


HammerGS:
This is great and long requested. Have a suggestion, possible issue and a something needing some opinions. 

Could there be a line space between the unit quirks and weapon quirks? I find the no space makes it easy to miss the weapon quirk.

image

This is from the War Dog WR-DG-02FC
image

I think there is a wrapping issue with the Streak staying on the same line.

This is more of an opinion one and would love other thoughts. When a unit has fluff the quirks are before the fluff. I feel like quirks should be after the all the fluff text.

image

kuronekochomusuke:
seems like they would be not seen if at the bottom.

JonhyBurner:
As a user, my opinion is that quirks are affected by game mechanics, but fluff affects no rules, so it makes sense for all the quirks to come right after the unit's summary. Some fluff text is also detailed (and long), and looking for the quirks would be a chore.

megamek/src/megamek/common/MechView.java
```diff
@@ -475,6 +472,41 @@ public MechView(final Entity entity, final boolean showDetail, final boolean use
             list.addItem(quirksList.toString());
             sFluff.add(list);
         }
+
+        Game game = entity.getGame();
+
+        if ((game == null) || game.getOptions().booleanOption(OptionsConstants.ADVANCED_STRATOPS_QUIRKS)) {
```
NickAragua, line 478: What's the "real life" situation where the game is null? Do we actually want to default to displaying quirks in that situation?

kuronekochomusuke: this is in mekhq and megameklab.   i would think you would want them displayed then.

SJuliez:
I think I agree with @JonhyBurner in that the quirks affect gameplay while the fluff doesn't meaning the quirks should go above fluff.

megamek/src/megamek/common/MechView.java
NickAragua: ok

NickAragua (approved):

PR #4343, opened 2023-04-19
add player colour setting

### LINKED ISSUE #2569: Add PlayerColour Selection GUI
This calls for the creation of a GUI within which one can select a PlayerColour.

kuronekochomusuke:
- add player colour settings

image

fixes #2569


pakfront:
All the user facing strings use the American "Color" rather than the British "Colour"

NickAragua:
Wait a sec, don't players already have the ability to pick their own colors/camo patterns?

kuronekochomusuke:
changing  camo only changes the look of the unit sprites.  player colours are used for "player" text also.   this allows you to update the one used for the "player" text. 

some of the current ones are hard to read with using some of the themes.

NickAragua:
I'm feeling like there might wind up being some confusion here on what this stuff does and why it shows up in some places but not others.

kuronekochomusuke:
which places does it not show up?

it is used in the default camo, unit tool tip, reports, attack sprite,  c3 sprite, ecm color, mini map symbols.

currently when you change camo only the camo changes, not  all the other places.

NickAragua:
I guess I'm not really explaining my problem with this adequately. So the first question I have is - what problem exactly are we trying to solve here? The parent issue provides nothing in the way of explanation and is not immediately obvious. Right now, it looks like you're adding the ability to have all blue players to display as red etc. I don't know that this makes any sense.

In general can see the value of setting text/icons/etc for particular players to a particular color. However, this should not be a client setting, but something configurable per player at game time - as a client setting, it's persistent across games, so if you define red player as blue, all your red players will show as blue and it will be... well, I don't know how that really improves the user experience any, especially when you can define multiple players as blue. I think we need to spend a little more time trying to figure out what problem we're trying to address here, which may be better done in discord with a wider audience.

kuronekochomusuke:
the main problem is trying to solve it to make text that is using the player colours easier to read.   this mostly affects the reports and unit tool tip.

kuronekochomusuke:
it improves things by only having to fix the bad colors once.    currently you are stuck with the current set of player colors with no way to change or exclude the ones that are hard to read.

pakfront:
Maybe changing the name "PlayerColourYellow" to something like PlayerA or Player01 would make it more understandable.
Or - and I like this better, but it is more work, we come up with a few fixed palettes that work with the various skins and LAFs. The user could then chose from some light-mode palettes and dar-mode-palettes

NickAragua:
Ok, I think I understand what you're trying to do here. I'm not a huge fan of the way the configuration is presented in the UI - it doesn't really tell me what the purpose of those settings is; and also feels pretty cluttered with the 20-30 color picker options. Maybe something like this instead: 

Player Color Mapping:
[Dropdown with list of colors that can be mapped]: [color selector button with the mapped color for the currently selected dropdown]

When you change the dropdown, it changes the value of the color selector; changing the value of the color selector updates the color mapping for the original color. That way you just have a single line that you can use to map an arbitrary number of colors.

We'll also want to add a tooltip or something explaining what these color mappings do.

kuronekochomusuke:
for me being able to see all the colors together helps in choosing colors that work well together.  one of the problems i have with some the current colors is some of them look too alike.

NickAragua:
All right, I can buy that argument. 

Can we change the code to dynamically append the color picker controls (based on the contents of an enum or something similar - if you look at some of the enums being used in megamek/mekhq, you can embed a lot of data in them including perhaps default colors) rather than having a hard-coded set? That way if the set of colors changes for any reason, it'll be a little easier deal with, code-wise. And also it'll reduce the line count considerably. 

Ideally we'd also do something similar with that 20-clause switch statement; it's generally not considered ideal software engineering practice.

And we'll definitely want a tooltip or something above that color mapping array to explain exactly what it affects. I can see people complaining in the future about how it doesn't swap camo colors in addition to the text.

NickAragua (approved):
Works for me codewise.

PR #4348, opened 2023-04-22
Show weapon location in tooltip for multi-location entities

pakfront:
This adds the weapon location to the tooltips. I find this very useful info to know. Without this change,  I had to select and view a unit in the UnitDisplay.
weaponlocations

This is optional, and can be toggle in the client settings UI:
Capture


NickAragua (approved):

SJuliez:
Can we maybe give this a client setting? I understand that this information can be interesting but it's a tooltip and there's probably a point where we have to stop adding info to it, especially if it makes the tooltip longer than it already is.

HammerGS:
I think it's useful information. But we've add so much to the tooltips I'm concerned they will become too large or complex to easily understand. In addition I'm worried about the impacts to with smaller screens. 

I agree with Juliez that this might be one for a client setting on. Or perhaps there could be improvements to the Field of Fire display to better show and define the arcs. That option could help with players who think they are having LOS bugs.

pakfront:
Sounds good, I'll add a client option.

pakfront:
I added the Client option (see description) and confirmed it works and saves with other options.

NickAragua (approved):

PR #4356, opened 2023-04-24
MTF Files: fix system model fluff entries

RexPearce:
Several MTF files specify system model fluff using 'systemmodel:' entries.

However, the MtfFile loader expects 'systemmode:' for these entries instead: 
https://github.com/MegaMek/megamek/blob/c7dad145036c6a54cf39fe785af2924574b07bb4/megamek/src/megamek/common/loaders/MtfFile.java#L116

so these entries are ignored.

The PR replaces 'systemmodel:' with 'systemmode:'. I also added the system model info to the Puma Prime config.

Note: Probably a copy & paste error, since 13 of the 14 affected files are Puma alt configs.


HammerGS:
The L and the : are side by side on the keyboard. I work very fast when I do these. Have you confirmed no other units are impacted with this change.

RexPearce:
I wonder if 'systemmode:' in the code was originally the typo - it'd make more sense for the tag to have been 'systemmodel:'. But, easier to change the 14 affected files to 'systemmode:' than change the code and the 100 or so files that use 'systemmodel:'.

Yeah, I scanned all the mtf files and the 14 files updated here were the only ones impacted.

PR #4362, opened 2023-04-26
detailed unit counts

kuronekochomusuke:
- "units remaining" only count units, exclude ejected crew in the count and percentage
- display counts for units : light damage, moderate damage, heavy damage, crippled, destroyed, fled, crew ejected, crew trapped, and crew killed
- display counts for ejected crew: active, picked up by team, picked up by enemy team, killed, and fled

image

### DIFF
--- a/megamek/i18n/megamek/common/report-messages.properties
+++ b/megamek/i18n/megamek/common/report-messages.properties
@@ -958,8 +958,8 @@
 7016=<data>:
 7017=<data> / <data> BV remaining (<data>%), <data> BV fled.
 7018=<data> / <data> units remaining (<data>%).
-7019=units: <data> light damage, <data> moderate damage, <data> heavy damage, <data> crippled, <data> destroyed, <data> fled, <data> crew ejected, <data> crew trapped, <data> crew killed
-7020=ejected crew: <data> active, <data> picked up by team, <data> picked up by enemy team, <data> killed, <data> fled
+7019=Units: <data> light damage, <data> moderate damage, <data> heavy damage, <data> crippled, <data> destroyed, <data> fled, <data> crew ejected, <data> crew trapped, <data> crew killed
+7020=Ejected crew: <data> active, <data> picked up by team, <data> picked up by enemy team, <data> killed, <data> fled
 7023=<newline>Survivors are:
 7025=<data> (<data>)
 7030=Pilot :
--- a/megamek/i18n/megamek/common/report-messages.properties
+++ b/megamek/i18n/megamek/common/report-messages.properties
@@ -955,7 +955,7 @@
 7005=A strange game. The only winning move is not to play.<newline>
 7010=Winner is: <data><newline>
 7015=Winner is: TEAM #<data><newline>
-7016=<data>: <data> / <data> BV remaining (<data>%), <data> BV fled. <data> / <data> units remaining (<data>%).
+7016=  <data>: <data> / <data> BV remaining (<data>%), <data> BV fled.  <data> / <data> units remaining (<data>%).  <data> units fled, <data> ejected crew, <data> ejected crew fled
 7020=<newline>Survivors are:
 7025=<data> (<data>)
 7030=Pilot :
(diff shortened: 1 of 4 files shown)


kuronekochomusuke:
would probably need to use 2 lines to keep it for getting too long and hard to read.

HammerGS:
That could work if it was stacked and indented a bit. 

1 Light Damage
2 Medium Damage
1 Heavy Damage
1 Crippled
2 Destroyed
2 Fled

For ejected would something like
2 Destroyed (1 ejected)

Tossing ideas here is all :)

pakfront:
Possibly, to avoid clutter, if the count for a category is 0, don't list it.

kuronekochomusuke:
that is kind of hard with the way reports are built.  you would need a template for every combination.

here are the templates for this section:
```
7017=<data> / <data> BV remaining (<data>%), <data> BV fled.
7018=<data> / <data> units remaining (<data>%).
7019=units: <data> light damage, <data> moderate damage, <data> heavy damage, <data> crippled, <data> destroyed, <data> fled, <data> crew ejected, <data> crew killed
7020=ejected crew: <data> active, <data> picked up by team, <data> picked up by enemy team, <data> killed, <data> fled
```

pakfront:
Ah, yes. Well, how about if ALL values for a report line are 0, skip it.

pakfront:
you can use html tags on the "data" so you can make 0 black, and highlight any non-0 value somehow.

SJuliez:
You can string together reports by having them use no linefeed. The FireProcessor does this to create messages like "Smokecloud at 2,5 moves to 3,5 and dissipates completely." With this you could omit 0 entries in the lists. Every entry could end in a ';'. Are you interested in trying this?

kuronekochomusuke:
i personally prefer the 0 to keep the display consistent.    would also need to move more reports around id for other reports to make room for all id needed to do this.  

but either way is fine with me, if people prefer it only showing ones that are non zeros.

HammerGS:
I kinda like the display as shown right now.

megamek/i18n/megamek/common/report-messages.properties
```diff
@@ -955,8 +955,12 @@
 7005=A strange game. The only winning move is not to play.<newline>
 7010=Winner is: <data><newline>
 7015=Winner is: TEAM #<data><newline>
-7016=<data>: <data> / <data> BV remaining (<data>%), <data> BV fled. <data> / <data> units remaining (<data>%).
-7020=<newline>Survivors are:
+7016=<data>:
+7017=<data> / <data> BV remaining (<data>%), <data> BV fled.
+7018=<data> / <data> units remaining (<data>%).
+7019=units: <data> light damage, <data> moderate damage, <data> heavy damage, <data> crippled, <data> destroyed, <data> fled, <data> crew ejected, <data> crew trapped, <data> crew killed
```
SJuliez, line 961: Please start these two lines with a capital letter.

megamek/src/megamek/common/Player.java
```diff
@@ -467,6 +467,75 @@ public int getEntityCount() {
                 .filter(entity -> !entity.isDestroyed() && !entity.isTrapped()).count());
     }
 
+    public int getUnitCount() {
+        return Math.toIntExact(game.getPlayerEntities(this, false).stream()
+                .filter(entity -> !entity.isDestroyed() && !entity.isTrapped() && !(entity instanceof EjectedCrew)).count());
+    }
+
+    public int getUnitLightDamageCount() {
```
SJuliez, line 475: It seems like using the damage type as a parameter here would save some repetition.
int getDamagedUnits(int damage) ...... entity.getDamageLevel() == damage

megamek/src/megamek/common/Player.java
```diff
@@ -507,6 +576,32 @@ public int getFledBV() {
         return bv;
     }
 
+    public int getFledUnitsCount() {
+        //TODO: I'm not sure how squadrons are treated here - see getBV()
+        Enumeration<Entity> fledUnits = game.getRetreatedEntities();
```
SJuliez, line 581: If you want, this would absolutely warrant adding a method to Game that returns a List instead of that silly enumeration and using that to rewrite this method.

SJuliez (changes requested):
I have two requests and one suggestion for this, but nothing major.

SJuliez (approved):

PR #4368, opened 2023-04-28
Snow mapgen option

NickAragua:
Adds the option to put random snow fields on the map;

Also takes advantage of that option by adding a bunch of arctic map templates that can be used independently or in AtB/Stratcon


megamek/src/megamek/common/MapSettings.java
```diff
@@ -170,6 +170,19 @@ public class MapSettings implements Serializable {
     @XmlElement(name = "SANDMAXHEXES")
     private int maxSandSize = 2;
 
+    /** how much sand spots at least */
```
SJuliez, line 173: the comments should say "snow"

SJuliez (approved):
Looks fine code-wise.

Thom293:
Is this an option we have to manually enable, or with stratcon maps just generate with snow and arctic maps now?  Looking forward to this one.

PR #4379, opened 2023-04-29
calculate heat over using modified heat sink value

kuronekochomusuke:
- calculate heat over using modified heat sink value, when heat sinking has been reduced.

image


SJuliez (changes requested):
There's nothing wrong with the code, but do you think it is possible to extract HeatDisplayHelper from Entity and place getHeatCapacityForDisplay() into it? Seeing how long Entity already is, anything not added to it is a win. Also, getHeatCapacityForDisplay() does instanceof checking anyway and so doesn't benefit much from being in Entity.

SJuliez (approved):

PR #4383, opened 2023-05-01
allow entity images to show on reports when using double blind

### LINKED ISSUE #2981: RFE - Unit sprites in Round Report in Double Blind
At present (0.48.0), the unit sprites are not displayed in the Round Report when the Double Blind option is on.

This is the case both for the player's own units, and for other units.

This suggestion is that unit sprites be displayed for the Client player's (or player's team's) units, as those would not be a mystery.
Units on other teams that are seen by the player's (or player's team if 'shared view' is on) should also be displayed.
Units that are not seen by the player's team could be represented in the Round Report by the 'Sensor Trace' sprite, next to the "????????????????" text currently used.

This would ease comprehension of the Round Report for those players using Double Blind.

kuronekochomusuke:
- allow entity images to show on reports when using double blind
- show ? image when entity is hidden

image

fixes #2981

### DIFF
--- a/megamek/src/megamek/client/Client.java
+++ b/megamek/src/megamek/client/Client.java
@@ -27,7 +27,6 @@ import megamek.client.generator.skillGenerators.ModifiedTotalWarfareSkillGenerat
 import megamek.client.ui.IClientCommandHandler;
 import megamek.client.ui.swing.GUIPreferences;
 import megamek.client.ui.swing.boardview.BoardView;
-import megamek.client.ui.swing.util.UIUtil;
 import megamek.common.*;
 import megamek.common.Building.DemolitionCharge;
 import megamek.common.actions.*;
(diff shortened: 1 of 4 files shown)


HammerGS:
This is going to be very popular!

megamek/src/megamek/client/Client.java
```diff
@@ -965,6 +966,22 @@ protected void receiveBoard(Packet c) {
         game.setBoard(newBoard);
     }
 
+    private void createDoubleBlindHiddenImage() {
```
SJuliez, line 969: Please move this to ImageUtil.

SJuliez (changes requested):
Just a small request. I think Client is a very specific class and should ideally remain free of miscellaneous stuff.
Have you tested this with double blind off or single blind, just as a sanity check? Code always breaks in unlikely ways.

Issue #4384, opened 2023-05-01
RFE: Implement chaff pods


Saklad5:
Chaff pods are currently in MegaMek, but are not functional.

The rules may be found on page 111, and consist of a few distinct effects that occur at the end of the movement phase:

* Fill the unit's hex with a cloud of chaff that acts like light smoke
* Apply a single hex of ECM, which cannot be countered with ECCM
* Act as AMS against all incoming missile attacks except Rocket Launchers and MRMs

These effects are all removed during the end phase of the same round, and do not work without an atmosphere (not in space, not in vacuum).

Even a partial implementation would still be an improvement.

Saklad5:
@Lu9us This appears broken in the 0.49.13 snapshot, seemingly due to the ECM's lack of affiliation.

```
07:04:08,587 ERROR [megamek.client.Client] {AWT-EventQueue-0}
megamek.client.Client.handlePacket(Client.java:1666) - Failed to parse Packet command ENTITY_UPDATE
java.lang.NullPointerException: Cannot invoke "megamek.common.Player.isEnemyOf(megamek.common.Player)" because "this.owner" is null
	at megamek.common.ECMInfo.addAlliedECMEffects(ECMInfo.java:343)
	at megamek.client.ui.swing.boardview.ECMEffects.getHexColor(ECMEffects.java:96)
	at megamek.client.ui.swing.boardview.BoardView.processAffectedCoords(BoardView.java:5344)
	at megamek.client.ui.swing.boardview.BoardView.updateEcmList(BoardView.java:5306)
	at megamek.client.ui.swing.boardview.BoardView$10.gameEntityChange(BoardView.java:4965)
	at megamek.common.event.GameEntityChangeEvent.fireEvent(GameEntityChangeEvent.java:106)
	at megamek.common.Game.processGameEvent(Game.java:2985)
	at megamek.common.Game.setEntity(Game.java:1274)
	at megamek.client.Client.receiveEntityUpdate(Client.java:1049)
	at megamek.client.Client.handlePacket(Client.java:1422)
	at megamek.client.Client$1.lambda$packetReceived$0(Client.java:174)
	at java.desktop/java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:318)
	at java.desktop/java.awt.EventQueue.dispatchEventImpl(EventQueue.java:773)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:720)
[...]

PR #4388, opened 2023-05-02
Add Chaff Pod functionality to the game 

Lu9us:
This pull request at least partially implements #4384 and allows units to deploy chaff on the their final movement position (as per the rules) by adding a 0 MP movement node to the end of the path and is inspired by how unjamming a RAC behaves.


Saklad5:
Could you elaborate on which functionality you are implementing?

Saklad5:
I imagine the AMS not affecting MRMs and Rocket Launchers will be a minor nightmare to implement.

Speaking of which, are you reusing the code from the vehicular grenade launchers? They have a version of chaff which is mostly the same, minus the AMS and deployment method. Oh,and the ECM is a bit unusual, in that it is immune to ECCM.

Lu9us:
The method of deployment is new but it uses the same kind of smoke as the chaff grenades which brings with it the ECM.

megamek/src/megamek/server/GameManager.java
```diff
@@ -7359,6 +7359,24 @@ private void processMovement(Entity entity, MovePath md, Map<EntityTargetPair,
...
+                List<Mounted> chaffDispensers = entity.getMiscEquipment(MiscType.F_CHAFF_POD)
+                        .stream().filter(dispenser -> dispenser.isReady())
+                        .collect(Collectors.toList());
+                if (chaffDispensers.size() > 0) {
+                    chaffDispensers.get(0).setFired(true);
+                    createSmoke(curPos, 1, 1);
+                    Hex hex = game.getBoard().getHex(curPos);
+                    hex.addTerrain(new Terrain(Terrains.SMOKE, SmokeCloud.SMOKE_CHAFF_LIGHT));
+                    sendChangedHex(curPos);
+                    r = new Report(2512)
```
SJuliez, line 7372: There is report 5187; is this applicable here instead of a new one?

SJuliez (changes requested):
I have a few remarks and questions

megamek/src/megamek/server/GameManager.java
Lu9us: Different context here is that unit is deploying chaff rather than a grenade simply filling the hex with chaff smoke ?

megamek/src/megamek/server/GameManager.java
```diff
@@ -7359,6 +7359,24 @@ private void processMovement(Entity entity, MovePath md, Map<EntityTargetPair,
                 }
             }
 
+            if (step.getType() == MovePath.MoveStepType.CHAFF) {
+                List<Mounted> chaffDispensers = entity.getMiscEquipment(MiscType.F_CHAFF_POD)
+                        .stream().filter(dispenser -> dispenser.isReady())
+                        .collect(Collectors.toList());
+                if (chaffDispensers.size() > 0) {
+                    chaffDispensers.get(0).setFired(true);
```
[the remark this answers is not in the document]
Lu9us: I assume so, from what I understand when MegaMek sends the entity packet back from the server after doing the movement work it doesn't filter any equipment changes. Testing it in game seems to confirm this too.

megamek/src/megamek/server/GameManager.java
```diff
@@ -7359,6 +7359,24 @@ private void processMovement(Entity entity, MovePath md, Map<EntityTargetPair,
...
+                    createSmoke(curPos, 1, 1);
```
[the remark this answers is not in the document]
Lu9us: yep, fixed that good spot 👍

megamek/src/megamek/common/Entity.java
```diff
@@ -4418,6 +4418,19 @@ public boolean hasMisc(BigInteger flag) {
         return false;
     }
 
+    public List<Mounted> getMiscEquipment(BigInteger flag) {
```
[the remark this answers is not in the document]
Lu9us: `hasWorkingMisc` can replace the check in `updateChaffButton`.

However in the `GameManager` where the Chaff is actually deployed `hasWorkingMisc` doesn't provide a returned list of equipment like this, I could do the same work through `getMisc` and filtering but it wouldn't be as neat.

megamek/src/megamek/client/ui/swing/boardview/BoardView.java
```diff
@@ -5239,9 +5239,12 @@ public void updateEcmList() {
             }
 
             // hidden enemy entities don't show their ECM bubble
-            if (ecmInfo.getEntity().getOwner().isEnemyOf(localPlayer) && ecmInfo.getEntity().isHidden()) {
-                continue;
-            }
+                if (ecmInfo.getEntity() != null
```
Lu9us, line 5242: Bug I encountered here when I create an ECM field without an entity tied to it, this hidden entity check will cause a null pointer exception. I confirmed this happens with both the new Chaff pods and the Vehicular Grenade Launcher chaff grenades.


Not sure if additional work is needed to hide ECM fields without an entity associated with them.

SJuliez (approved):

megamek/src/megamek/client/ui/swing/boardview/BoardView.java
SJuliez: Don't know either. Let's see what happens.

SJuliez:
Thanks for the contribution!

PR #4394, opened 2023-05-03
Issue 4345: Nag dialog behavior, Add nag reset menu item

### LINKED ISSUE #4345: Princess Help text opens constantly.
Getting lots of reports and seeing this myself. Nothing in the logs. Started in 49.12

I'm getting the bot help text every time I add one now. Is that intended or did I accidentally turn a setting on?

### Environment
What version of MegaMek does your issue apply to?  49.12

SJuliez:
Fixes #4345

This makes a nag dialog of the AbstractNagDialog style return CANCELLED instead of CONFIRMED when it is in "dont bother" mode. When this system is used for more than the bot nag, it's possible that we need to also save the last user input (CANCEL or OK) to always return that when the dialog isnt shown, but currently that's not possible. When someone checks "dont bother" they likely dont want to see the bot readme again.

This also adds a menu point under Help to reset all nags of the AbstractNag system (currently only the bot nag), so the dialog appears again. This would be nice to automatically find all such nags but I had no idea how.


megamek/i18n/megamek/client/messages.properties
```diff
@@ -976,6 +976,7 @@ CommonMenuBar.GameMenu=Game
 CommonMenuBar.helpSkinning=Skinning How-to
 CommonMenuBar.Inferno=Inferno({0})
 CommonMenuBar.helpAbout=About MegaMek
+CommonMenuBar.helpResetNags=Reset Readme Settings
```
NickAragua, line 979: Not really sure that this text accurately describes what the functionality does. Maybe "Reset Nag Settings" instead?

SJuliez: I had no idea what to call it, even less so for German. OK let's try Nag :)

NickAragua (approved):

PR #1227, opened 2019-01-17
Enable princess to unjam weapons

NickAragua:
Sometimes, a tank or a turret will take damage and a weapon will become jammed. This pull request enables the bot to consider unjamming its weapons rather than sitting there with jammed weapons and not doing anything.

PR #4404, opened 2023-05-06
Fix MegaMekLab issue #1227 regarding "ArrayIndexOutOfBoundsException when loading unit"

Sleet01:
Bug was originally filed on MegaMekLab but is in fact in the MTF file load code.  MegaMekLab appears to do everything correctly with regards to Variable Size components in Superheavies; this fixes the one remaining path that is not correctly accounting for this edge case.


NickAragua (approved):
This is fine - normally, the person merging the pull request handles updating history.txt directly, as it's updated very frequently and you have to constantly keep doing merges to keep it up to date otherwise.

Sleet01:
> This is fine - normally, the person merging the pull request handles updating history.txt directly, as it's updated very frequently and you have to constantly keep doing merges to keep it up to date otherwise.

Apologies; one of the wiki page said that the contributor should do that.  I'll refrain in the future.

PR #4406, opened 2023-05-06
allow force to work when using double blind

### LINKED ISSUE #4253: double bind removes units from the forces
0.49.12

double bind removes units from the forces.

forces assigned in lobby
image

while deploying units are still in the forces assign in the lobby
image

once all units are deployed, the force is removed from the units
image

kuronekochomusuke:
- allow force to work when using double blind

fixes #4253


megamek/src/megamek/client/ui/swing/tooltip/UnitToolTip.java
```diff
@@ -117,10 +118,6 @@ private static StringBuilder getEntityTipTable(Entity entity, Player localPlayer
         if (details && (mapSettings != null)) {
             sPilotInfo += deploymentWarnings(entity, localPlayer, mapSettings) + "<BR>";
         } else {
-            if (pilotInfo) {
```
NickAragua, line 120: I don't particularly like removing safeguards like this.

kuronekochomusuke:
that change allows force to show on the unit display and on the lobby toot
tip.  before it was only displayed on the game board tool tip.  is there a
reason it should not be displayed on those tool tips?

NickAragua:
Never mind, I misread the code. Should be good to go.

PR #4407, opened 2023-05-06
Issue 4314: Correct missing long range issue with BA Micro GL

### LINKED ISSUE #4314: [0.49.12] BA Micro GL has zero range
Bundled Megamek with MekHQ 0.49.12

p305 TW 9th printing shows the Micro GL should have a range of 0/1/2

MM currently only allows firing at units in the same hex with the Micro GL and lists the range as zero, although the UI shows the correct range brackets.

[test.sav.gz](https://github.com/MegaMek/megamek/files/11208334/test.sav.gz)
name: 'test'

image

SJuliez:
Fixes #4314 
The ranges were not showing because the code treated it like a torpedo and showed its water ranges; also it seems that ranges of 1/2/0 don't work while 1/2/2 works correctly. I don't know if a weapon without a long range has an extreme range; I copied the present values from the similar BA Heavy MG.


NickAragua (approved):

Gribbly1:
Many thanks!

PR #4412, opened 2023-05-08
Remove "(Standard)" as model to follow the MUL

SJuliez:
As requested. We should probably only merge this when we also have the required addition to the name change list.


HammerGS:
Merging this so I can work on the name change list.

PR #4419, opened 2023-05-13
Issue 4411: Cruise Missile damage falloff in the Unit Display

### LINKED ISSUE #4411: Cruise Missiles (possible other artillery) not respecting radius
From Discord. 

image

Per Tac Op:AR pg 152 Cruise Missiles should have a max radius of 4. The damage drop off is correct but the blast radius is well past the four. 

Digging around in the https://github.com/MegaMek/megamek/blob/master/megamek/src/megamek/common/weapons/AreaEffectHelper.java

I see references to FAE and area of effect, but not the other weapons. Wondering if data was missed or code was missed. 

### Environment
What version of MegaMek does your issue apply to?  0.49.13-Dev

SJuliez:
Fixes #4411 
Assuming this is only about the text in the Unit Display which used a hardcoded damage falloff of 10


HammerGS (approved):
Seemed to work for me.

PR #4423, opened 2023-05-15
Add more target info to the WeaponsPanel of the Unit Display

pakfront:
This adds a section with current target info to the Weapon Panel. It is separated by a split pane to allow user to adjust details they need to see. This is a major refactoring of the WeaponPanel class to make it simpler to debug individual components, and share layout settings between components. This should help with any further work needed on panel.

There is a known issue with the WeaponPanel hanging on to old target data between rounds. Apparently this has been the case even before this PR. I plan on addressing this and some other issues in my next PR, which will require a refactor I'd rather keep out of this PR.

noaim

aim


SJuliez (approved):

SJuliez:
There are conflicts.

Issue #4424, opened 2023-05-15
Stingray ammo is misspelled


Saklad5:
### Environment
MegaMek Version: MekHQ v0.49.12 Development Snapshot
Operating System: Mac
Java Version: 19.0.2

### Description

The ammunition for Stingray Sub-capital Missiles is currently named "Stringray Ammo", which is clearly a typo.

PR #4427, opened 2023-05-17
Fix: #4424 type of Stingray Ammo

HoneySkull:
Fixed an ammo spelling mistake:  Stringray -> Stingray.


SJuliez:
A few units use this ammo. They'll not load correctly when their files aren't corrected as well.

image

HoneySkull:
Good point.  I only searched in the .java files.  Thanks, I'll finish the sweep.

HoneySkull:
These .blk files exist in megameklab and MekHQ repositories as well, I'll update them.

PR #4429, opened 2023-05-18
Some Map Setup/Board Fixes

SJuliez:
- Fixes some broken multi-board map setups (those that can be loaded from the lobby) due to board renames
- Rewrite all map setups to have only / instead of \ in paths
- Adds a few more map setups from existing boards
- replaces all \ in board paths with / when the board dir is scanned so they don't have to be removed later
- removes a few building:5 (not implemented) from boards (I actually got an NPE from one but this seemed a special case, mostly building:5 shows no graphics and doesnt generate errors)


NickAragua (approved):
Looks fine code-wise, except for the weird compilation errors.

SJuliez:
@NickAragua Sorry for that, the silly author of this PR of mine actually managed not to test-build it after the last change and before making it a PR (I thought only other people did that). Also, I'll separate this. The map setups are not dangerous to merge but the other stuff is so I'd leave this till after the next release.

PR #4446, opened 2023-05-20
show teams before players on bv count report section

kuronekochomusuke:
- show teams before players on bv count report section

image

### DIFF
--- a/megamek/src/megamek/server/GameManager.java
+++ b/megamek/src/megamek/server/GameManager.java
@@ -1631,8 +1631,8 @@ public class GameManager implements IGameManager {
     }
 
     private void bvReports(boolean checkBlind) {
-        Vector<Report> playerReport = new Vector<>();
-        Vector<Report> teamReport = new Vector<>();
+        List<Report> playerReport = new ArrayList<>();
+        List<Report> teamReport = new ArrayList<>();
         HashMap<Integer, BVCountHelper> teamsInfo = new HashMap<>();
 
         for (Team team : game.getTeams()) {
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/server/GameManager.java
```diff
@@ -1631,6 +1631,8 @@ public BVCountHelper() {
     }
 
     private void bvReports(boolean checkBlind) {
+        Vector<Report> playerReport = new Vector<>();
```
SJuliez, line 1634: Ideally, we'd use the modern collection classes, at least when we can do so without having to do a grand refactor. This seems like it could work with List<Report> playerReport = new ArrayList<>()

megamek/src/megamek/server/GameManager.java
```diff
@@ -33555,6 +33564,10 @@ public void addReport(Report report) {
         vPhaseReport.addElement(report);
     }
 
+    public void addAllReport(Vector<Report> reports) {
```
SJuliez, line 33567: I'm all for extracting methods but this doesn't seem to add anything over just writing the addAll in place of the method call, or does it?

SJuliez (commented):
Just 2 comments.

megamek/src/megamek/server/GameManager.java
kuronekochomusuke: not really was just making it similar to and not accessing vPhaseReport directly

    public void addReport(Report report) {
        vPhaseReport.addElement(report);
    }

SJuliez (approved):

PR #4451, opened 2023-05-23
when tacops sensors is not on, show visual range in tool tip

kuronekochomusuke:
- when tacops sensors is not on, show visual range in tool tip
- show second range value when dusk dawn also
- allow search light to be turned on when dusk / dawn

image


HammerGS:
Tested and seems working.

PR #4468, opened 2023-05-29
limit the width of the unit tooltip so that it does not grow really wide

kuronekochomusuke:
- limit the width of the unit tooltip, so that it does not grow really wide.  

image
image


megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
```diff
@@ -985,6 +986,16 @@ private JPanel getUnitDisplayPanel() {
         row.add(unitTooltipSeenbyCbo);
         comps.add(row);
 
+        SpinnerNumberModel munitTooltipMaxWidth = new SpinnerNumberModel(GUIP.getUnitToolTipMaxWidth(), 0, 20000, 1);
```
NickAragua, line 989: 20,000 what? Unless it's like nano-pixels, that still seems entirely too wide.

kuronekochomusuke: html width, is 2000 a better max value?

SJuliez: I wonder if not giving it a client setting (for once) would be better here. It doesn't feel like this would be something I'd want to concern myself with as a player. Just set it to a sensible width (with gui scaling)?

kuronekochomusuke:
that is how it was before, at min scale it wider than it needs to be, but probably not a big deal.

image

NickAragua (approved):

PR #4470, opened 2023-05-29
add client setting for Max Range to Draw Sensor / Visual Rings

### LINKED ISSUE #4467: [0.49.13] Lag when using large craft and TacOps Sensor Rules
0.49.13
Windows 11
Java 11

When using large craft such as WarShips and DropShips with the TacOps Sensors rule enabled the program experiences significant lag and delays responding to input. I originally thought only the space map was affected, but additional testing this morning presented issues on the low altitude and ground maps as well. Toggling this rule off returns to the program to expected behavior. On the ground map, the lag seems to only persist through the initial deployment phase, but afterwards the program continues to run normally. On space and low altitude maps the lag continues through all following phases and turns.

This does not seem to be directly related to #4462, as this problem exists whether High Quality Graphics are on or off.

kuronekochomusuke:
- code improvements to help with lag in the case when sensor range is 1000+

fixes #4467


megamek/src/megamek/client/ui/swing/boardview/BoardView.java
```diff
@@ -6499,9 +6500,11 @@ public void setSensorRange(Entity entity, Coords c) {
 
         for (RangeHelper rangeH : lBrackets) {
             sensorRanges.add(new HashSet<>());
+            int rangeMin = Math.min(rangeH.min, GUIP.getSenorMaxDrawRange());
```
NickAragua, line 6503: "Senor" Max Draw Range?

NickAragua (changes requested):
Spelling fix. Also, discussion point: do we really need the extra setting? Why not just cap the draw range to the bigger of board width/height and call it a day? I don't think I've ever seen a board bigger than 200x200.

kuronekochomusuke:
would just capping still lag people with slow computers on large maps?  the setting allows it be be dialed back to what your computer can handle.

kuronekochomusuke:
it calculates out to max range, then removes them any not on the board, before it tries to draw anything.   it grows larger the further out you are, since it adds a other circle of points each step.  

I have added logic to only calculate out to max range from your position on the board, that should help.

SJuliez:
I'd like to second @NickAragua suggestion to not give this a client setting if it can be avoided. The algorithm for determining the hexes to paint can be improved a lot here I think. If I'm not mistaken, the thing to draw here is a ring of hexes at the respective range, right? So the only hexes that can possibly have a sprite are those at exactly the range and those on the board edge. Those on the board edge could be omitted because that doesn't add any relevant information (obviously, any range ends at the board edge). First, check if X > board diagonal. If so, draw nothing. Otherwise, add those at range X, not all those from range 0 to X. Then remove any not on the board. Then to determine the edges to draw just compare the surrounding hexes if they're more distant than X. Unless range X = 0, only a maximum of 3 edges can ever be there. This should speed up things and I'm sure there is more potential improvement. I hope I made not too many errors in this.

kuronekochomusuke:
with sensors currently, the start is drawn on the near side of the hex, and the end is on the far edge.  this helps showing they are one range.  not sure your algorithm would handle that.    

also drawing the ones along the board edge help show that the two ranges are connected and the sensors are active between the two.  

drawing along the board edge shows that you can see up to that edge.

NickAragua (approved):
It's a major improvement over what was there before anyway. If the person reporting the issue is still experiencing slowness afterwards, then we can work on further improvements in the spirit of those outlined by @SJuliez, in terms of not looping through every hex on the map.

kuronekochomusuke:
also if you turn it off, nothing will calculate now.

PR #4509, opened 2023-06-09
Adding MHQ's portraits to MM

SJuliez:
This only adds MHQ's portraits to MM and changes the gitignore accordingly.


NickAragua:
I'm going to assume this is ok as I have no intention of scrolling through 1400 files, lol.

pheonixstorm:
With the portraits being added to MM, do we need to have them as part of HQ now?

SJuliez:
Yes, at least as things are now, shared data is added to all of the three apps that need it.

PR #4510, opened 2023-06-09
Issue 206: Weapon Specialist

### LINKED ISSUE #206: Dropships using bay weapons in the air incompatible with weapon specialist SPA
Even if you have the unofficial rule checked, when editing a dropship pilot in the megamek lobby, the weapon specialist SPA only allows you to use weapons by bay (e.g. artillery bay, PPC bay, laser bay, etc) instead of the individual weapons that you would be using (arrow IV, ER PPC, Large Laser, etc). You can set it to weapon specialist (artillery bay) but it has no effect when the grounded dropship is attempting to fire the individual weapons when grounded (such as arrow IV).

The dropship in the save was set to weapon specialist (artillery bay) in the lobby.

[#battletech.zip](https://github.com/MegaMek/megamek/files/271891/battletech.zip)

SJuliez:
- corrects allowed weapons for the Weapon Specialist SPA: no bay weapons as per discussion with Hammer, also no Screen Launchers and AMS as they don't use a to-hit roll
- gives the bonus for weapon bays when a bay exclusively contains the exact weapon type of a specialist
- corrects allowed weapons for the Sandblaster SPA
- shows the chosen weapon of a specialist in the pilot tooltip

Fixes #206


megamek/src/megamek/common/PilotSPAHelper.java
```diff
@@ -0,0 +1,95 @@
+package megamek.common;
```
NickAragua, line 1: Do we want to put the copyright thing at the top of the file?

megamek/src/megamek/common/PilotSPAHelper.java
```diff
@@ -0,0 +1,95 @@
...
+
+    /** @return True when the given Mounted equipment is a valid choice for the Sandblaster SPA. */
+    public static boolean isSandblasterValid(Mounted mounted) {
+        return isSandblasterValid(mounted.getType());
+    }
+
+    /** @return True when the given EquipmentType is a valid choice for the Sandblaster SPA. */
+    public static boolean isSandblasterValid(EquipmentType equipmentType) {
+        return (equipmentType instanceof WeaponType)
+                && ((equipmentType instanceof UACWeapon) || (equipmentType instanceof LBXACWeapon)
```
NickAragua, line 62: One comment here, which is that the "rapid fire autocannons" weapon allows standard ACs to behave like UACs (at the risk of exploding), so I think standard autocannons should be included in this.

NickAragua (changes requested):
Looks good, just one quibble about standard autocannons.

megamek/src/megamek/common/PilotSPAHelper.java
SJuliez: As a suggestion, I have added a GameOptions parameter here based on which all ACs may be allowed. Both MM and MHQ can supply GameOptions and so it'll reflect the state of the game/campaign. The GameOptions can be nulled if necessary but I have not added versions without the additional parameter as we'd really want the GameOpts to be given. I think MML won't need these methods.

Aside from that I've removed Streaks from Sandblaster as they don't use the cluster table and C3 from the weapon specialist (hehe).

SJuliez: @NickAragua Any thoughts?

NickAragua: Sorry, lost track of this one.

I think there are some situations (mostly involving high powered ECM as I recall) where Streak launchers revert to cluster rolls, so there may be a case for Sandblaster applying to them.

NickAragua (approved):
I suspect some people may have complaints about the removal of AMS (given the "AMS as machine gun" option) and infantry attacks, but that's easy enough to change if a rule clarification comes through.

SJuliez:
Good point with AMS, I've changed that to allow AMS with the AMS as weapon game option.

Issue #4501, opened 2023-06-08
Attack Sprite ToHit is wrong after deleting an attack


pakfront:
### Environment
49.14

### Description
The AttackSprite tooltip will always show the To Hit number for attacks as Impossible if you delete an existing attack with Backspace. This is because the ToHit .getValue() for an attack changes to Impossible after a new attack is added. Reason behind this is unknown by me, and it may be working as intend but is also causing UI issues elsewhere. Possible fix, if this is intended, is to cache the ToHit value somehow. 
In addition, it would be good to change BoardView.attackSprites from ArrayList to HashMap as there is a lot of iterating looking for matches, and rebuilding which defeats caching.
There are also a few other minor issues with AttackSprite that could be cleaned up here, and a chunk of the text generation code moved to a central, static area so it can be share with other UI.

![attackspritebug](https://github.com/MegaMek/megamek/assets/1423800/8278b259-a0b3-4870-8f25-f1fef571b05a)

Issue #4486, opened 2023-06-04
Better feedback on Movement Phase MovePath contents


pakfront:
### Environment
49.14

### Description
It is easy to build a complicated move path that the user does not fully understand the implications of.
There are some movement commands which, on their own, don't make a lot of sense but meet the criteria of a valid move.
Example - hitting Climb Mode puts a command into the MovePath. So if you just press climb mode, but don't actually move hexes or turn, the current setup thinks that is a valid move to submit. Technically, it is valid, but probably not what the user expects. 

We could add function that does a bunch of tests to sanity test the moves. There is already one, clipToPossible(), that trims down a movepath to only legal moves. It might be modifiable to have it trim these nonsense moves, or we add a new method that is more aggressive.
I also think we should add an client option to show a text breakdown of your move steps. It enhances player understanding of what is happening in terms of the game rules.

PR #4513, opened 2023-06-10
Turn Details Overlay on Boardview

pakfront:
This displays an optional overlay on the board view that gives the specific Moves or Actions a unit will do when the turn is Done.
This is very helpful in understanding what will happen in a turn, and makes debugging turn behavior a bit easier.
attackoverlay

moveoverlay
In the Move overlay, any set of moves that require a PSR get a * appended to their entry. Any illegal moves, such as moves beyond MP limit, are greyed out. There are also unicode icons for some move actions.
This fixes https://github.com/MegaMek/megamek/issues/4486

I optimized and centralized the code to calculate and itemize the user displayed ToHits for this feature and AttackSprites. This fixes https://github.com/MegaMek/megamek/issues/4501.
An important change - in FiringDisplay I switched the order of adding attacks to Game and to the local Log so that the Game add is first. This fixes the issues where the ToHit mods displayed to the user were not calculated correctly when the attack was first added. This should have not break anything, it is more correct and when the ToHits are sent to the server, they go through another recalc pass that verifies the values.

 There are cases where the display differs from the actual results. Without considerable and possibly disruptivework, these are unavoidable due to the way MegaMek works:
- In Move phase, it will display the Moves in the order they are entered. This may not match the actual execution order.
 Notably, it seems in the game all altitude changes are reordered to happen at the beginning of the turn on Done submit. This is not reflected in the Move list, So moves made after a number of alt changes may not have enough MP. This is true of the boardview movement sprite display also, so is a general issue.

- in Firing phase, I put a chunk of work into getting the ToHit modifiers to display and update correctly, but the penalty for Spotting still is not applied to the user displayed ToHit. It is correctly calculated when the Turn is executed though.


megamek/src/megamek/client/ui/swing/tooltip/EntityActionLog.java
```diff
@@ -0,0 +1,202 @@
+package megamek.client.ui.swing.tooltip;
```
SJuliez, line 1: Please add the copyright notice to all new classes. It's in https://github.com/MegaMek/megamek/wiki/MegaMek-Coding-Style-Guide

Also, as general remarks, I'll add it here because I do not want to generate a big number of change requests for rather small issues:
- please try to follow the other coding guidelines in there
- please make sure that new methods and fields, like in this class but also elsewhere, have the access type they need to have. I am not sure all of these are intentionally package-private.
- remove commented out code unless there's a really good reason to leave it (actually there's rarely. It just gets stale fast)

megamek/src/megamek/common/actions/WeaponAttackAction.java
```diff
@@ -245,12 +245,25 @@ public ToHitData toHit(Game game) {
                 game.getTarget(getOriginalTargetType(), getOriginalTargetId()), isStrafing(), isPointblankShot());
     }
 
-    public ToHitData toHit(Game game, List<ECMInfo> allECMInfo) {
+    /**
+     *
+     * @param game
+     * @param evenIfAlreadyFired true if evaluating hit change even if the weapon has already fired or is otherwise unable to fire
```
SJuliez, line 251: hit change?

SJuliez (changes requested):
Few minor things.

I like the AbstractBoardOverlay and the EntityAttackLog classes. Both neat as well as the PR itself of course.

megamek/src/megamek/client/ui/swing/tooltip/EntityActionLog.java
pakfront: will do. I thought I got rid of all commented out code, but must have missed something. which file was it in?

pakfront: Ah, I found one that snuck by in EntityActionLog, line 38. Ill keep looking.

pakfront: I added copyright, think I fixed all the accessors, etc.

SJuliez (approved):

PR #4514, opened 2023-06-10
Map generator upgrades

NickAragua:
Kind of a large pull request, but it adds a lot of stuff to the random map generator.

-Tundra
-Ultra Rough
-Ultra Rubble (and greater regular rubble variety)
-Ultra Woods
-Jungle (standard/heavy/ultra)


megamek/src/megamek/client/ui/swing/RandomMapPanelBasic.java
```diff
@@ -769,6 +844,10 @@ public static void makeCompactGrid(Container parent,
             y = Spring.sum(y, Spring.sum(height, Spring.constant(yPad)));
         }
 
+        // for some reason the bottom border runs over the bottom textboxes in
+        // some of these, so we add an experimentally established padding to the bottom
+        y = Spring.sum(y, Spring.constant(13));
```
SJuliez, line 849: The reason is the call to adjustDialog to adapt it to the gui scale, which doesn't relayout the dialog. I've fought with this too and can't say I won. A slightly less hacky way to deal with it is to add a pack() in RandomMapDialog after l.476, directly after the call to adaptToGUIScale().

SJuliez (approved):
Just one comment

PR #4516, opened 2023-06-10
Icons

SJuliez:
Adds DropShip icons: large and small Buccaneer, small Hannibal
Also prevents the camo overlay (the shading and highlighting) from being used on multi-hex icons as it adds strange shadows for these

image


HammerGS:
Tested and looks great to me.

HammerGS (approved):

PR #4534, opened 2023-06-16
Fixes #4305 Environmental Specialist SPA not reducing penalties from high wind

### LINKED ISSUE #4305: [0.49.12] Environmental Specialist SPA not reducing penalties from high wind
Bundled Megamek with MekHQ 0.49.12

Environmental Specialist, CamOps p74
'Storm' wind conditions TO: AR p34

For example, a hover unit with the Wind version of the SPA turning while flanking is not receiving any benefit:
image

A mech with the same SPA also has the full penalty turning while running on pavement:
image

Name: 'Test'

[test.sav.gz](https://github.com/MegaMek/megamek/files/11186667/test.sav.gz)

Rain and Snow appear to be working fine:
image

image

pheonixstorm:
Repost, this corrects the conflicts from the previous version.

### DIFF
--- a/megamek/i18n/megamek/common/messages.properties
+++ b/megamek/i18n/megamek/common/messages.properties
@@ -619,3 +619,7 @@ MiscType.theaters=theaters
 MiscType.theater=theater
 MiscType.tons=tons
 MiscType.ton=ton
+
+PilotingSPA.EnvSpec.RainSpec=Rain Specialist
+PilotingSPA.EnvSpec.SnowSpec=Snow Specialist
+PilotingSPA.EnvSpec.WindSpec=Wind Specialist
--- a/megamek/src/megamek/common/Aero.java
+++ b/megamek/src/megamek/common/Aero.java
@@ -118,13 +118,13 @@ public class Aero extends Entity implements IAero, IBomber {
     // ignored crew hit for harjel
     private int ignoredCrewHits = 0;
     private int cockpitType = COCKPIT_STANDARD;
-    
+
     //Autoejection
     private boolean autoEject = true;
     private boolean condEjectAmmo = true;
     private boolean condEjectFuel = true;
     private boolean condEjectSIDest = true;
-    
+
     private boolean ejecting = false;
 
     // track straight movement from last turn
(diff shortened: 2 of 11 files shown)


SJuliez:
Sorry, I looked at it a little closer this time and I'm not sure I understand this correctly. I assume this is about the CamOps Environmental Specialist, p.74, right? It's unclear to me why the weather conditions have to be No Weather for some of the Wind specialist effects. E.g. Rain doesnt seem to affect a wind specialist nor does it seem to affect the weather modifier from game.getPlanetaryConditions().getMovementMods(this);

Also, I understand that this is legacy code. But from the rules it seems that what we need to do is to halve the weather modifier, specifically for wind (for a wind spec). game.getPlanetaryConditions().getMovementMods(this) already passes in the Entity, so it might be much better to add these SPA checks right there in PlanetaryConditions where they can be applied exactly as needed instead of reverse engineering the modifier them in Aero/Tank etc.

Unless I misunderstand this all, this would be my suggestion. I admit it's a bit of work tho and it would make sense only when moving over the other specs like Snow spec.

pheonixstorm:
The WEATHER_NONE is there to keep Wind Specialist from being applied to other type of weather even if they include wind. Those other specialties already include modifiers for such conditions. With wind by itself I could put it in getMovementMods, which would cut down on some code,

All of this still doesn't answer my original question though, the first public int getCurrentThrust() is proper, that was there when I wrote the code originally. The next function public int getCurrentThrust(MPCalculationSetting mpCalculationSetting) is new, so I am not sure if I should have added the check or not in this case.

SJuliez:
Yes also in getCurrentThrust.

SJuliez:
@pheonixstorm Heya, any input from my side missing?

pheonixstorm:
No, the question was settled so I know it goes in both places, just haven't moved the wind spec section over to getMovementMods() if that is what you want me to do.

SJuliez:
Nah it's alright for now if you don't want to work any more on it. Just give a shout when you think it's all finished.

pheonixstorm:
It's done for now then. Everything should be fixed until some new feature comes a long to break things.

SJuliez (approved):

PR #4538, opened 2023-06-17
Board Overlays: numerous problems

### LINKED ISSUE #4537: Uncaught Exception on preview game map - 0.49.14-SNAPSHOT
### Environment
Build Date: 2023-06-16T01:30:11.892597800
Origin Project: MegaMek
Java Version: 17.0.7

### Description
Uncaught Exception when viewing map preview in MM. Verified Issue not present in 49.13

Maybe something to do with pr #4514

To recreate:
>Run fresh MM install
>Start a new game
>Select map
>Preview game map

### Files
[megamek.log](https://github.com/MegaMek/megamek/files/11774731/megamek.log)

SJuliez:
Corrects numerous problems, hopefully without introducing too many new ones
- prevents updating the done button and turn summary in phase displays that aren't active
- streamlines turning overlay visibility on/off 
- prevents adding overlays in boardviews that dont need them (it seems safer to create them but it isn't, I immediately got NPEs from listener calls for overlays that were never used, no less than three separate TurnDetailsOverlays)
- ~~removes event listening from overlays (let the boardview have control)~~
- fixes #4537

### DIFF
--- a/megamek/src/megamek/client/ui/swing/boardview/AbstractBoardViewOverlay.java
+++ b/megamek/src/megamek/client/ui/swing/boardview/AbstractBoardViewOverlay.java
@@ -14,48 +14,31 @@
 package megamek.client.ui.swing.boardview;
 
 import megamek.client.ui.IDisplayable;
-import megamek.client.ui.Messages;
 import megamek.client.ui.swing.ClientGUI;
 import megamek.client.ui.swing.GUIPreferences;
-import megamek.client.ui.swing.util.KeyCommandBind;
+import megamek.client.ui.swing.util.StringDrawer;
 import megamek.client.ui.swing.util.UIUtil;
 import megamek.common.Game;
-import megamek.common.KeyBindParser;
 import megamek.common.enums.GamePhase;
 import megamek.common.event.GameListener;
 import megamek.common.event.GameListenerAdapter;
 import megamek.common.event.GamePhaseChangeEvent;
 import megamek.common.event.GameTurnChangeEvent;
-import megamek.common.preference.IPreferenceChangeListener;
-import megamek.common.preference.PreferenceChangeEvent;
 import megamek.common.util.ImageUtil;
 import org.apache.logging.log4j.LogManager;
 
-import java.awt.font.TextAttribute;
-import java.text.AttributedString;
-import java.awt.AlphaComposite;
-import java.awt.Color;
-import java.awt.Composite;
-import java.awt.Font;
-import java.awt.FontMetrics;
-import java.awt.Graphics;
-import java.awt.Graphics2D;
-import java.awt.Image;
-import java.awt.Rectangle;
+import java.awt.*;
 import java.util.List;
+import java.util.stream.Collectors;
 
 /**
- * An overlay for the Boardview that displays a selection of Planetary Conditions
- * for the current game situation
- *
- *
+ * This is a framework for boardview overlays that can show a list of data, conditions etc.
  */
-public abstract class AbstractBoardViewOverlay implements IDisplayable, IPreferenceChangeListener {
+public abstract class AbstractBoardViewOverlay implements IDisplayable {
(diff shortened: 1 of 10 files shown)


megamek/src/megamek/client/ui/swing/boardview/AbstractBoardViewOverlay.java
```diff
@@ -91,11 +74,9 @@ public AbstractBoardViewOverlay(Game game, ClientGUI cg, Font font, String heade
         currentPhase = game.getPhase();
         game.addGameListener(gameListener);
         clientGui = cg;
-        KeyBindParser.addPreferenceChangeListener(this);
-        GUIP.addPreferenceChangeListener(this);
 
         Color colorTitle = GUIP.getPlanetaryConditionsColorTitle();
-        header = String.format("#%02X%02X%02X %s",
+        header = String.format("#%02X%02X%02X%s",
                 colorTitle.getRed(), colorTitle.getGreen(), colorTitle.getBlue(), headerText);
```
pakfront, line 80: This can use the utility function AbstractBoardOverlay.colorToHex()

megamek/src/megamek/client/ui/swing/boardview/TurnDetailsOverlay.java
```diff
@@ -67,10 +57,6 @@ protected void gameTurnOrPhaseChange() {
         super.gameTurnOrPhaseChange();
     }
 
-    @Override
-    protected void setVisibilityGUIPreference(boolean value) {
```
pakfront, line 71: Where did the setVisibilityGUIPreference logic get moved to?

pakfront (commented):
Thanks for cleaning this up. I had tested it by running games myself, did not encounter the NPEs, but  obviously they are there.  (edit, ah the Preview Map from the Map selection UI. I was not even aware that existed.)

megamek/src/megamek/client/ui/swing/boardview/TurnDetailsOverlay.java
pakfront: and does this still respond to GUIP changes with the IPreferenceChangeListener removed?

SJuliez: not finished yet! I first moved the reaction into BoardView but I've since changed my viewpoint. I'll move it back into the Overlay. The logic for this is that the menu item for any of the overlays only changes the GUIP setting and everything else happens as a preferenceChangeEvent reaction. We don't have many static event generators but this one we must make use of :)

megamek/src/megamek/client/ui/swing/boardview/AbstractBoardViewOverlay.java
SJuliez: Aye

SJuliez:
I think this is now ready...

SJuliez:
The NPE on the board preview was only a tooltip problem and had nothing to do with the turn overlay.

The overlays now have their listeners again :) but in return they aren't created with every boardview. The lobby preview doesn't create them but the board editor does although it only uses the keybinds overlay. I simplified the constructor a bit. Every boardview must have a game with a board though not necessarily a clientgui but all these can be accessed from the boardview itself (in an obviously-bad way by directly accessing the public field. Something to correct on another day).

I think the main thing is what I wrote above: Toggling an overlay in CommonMenuBar only directly toggles the GUIP setting, the rest follows from the event (even the checkmark on the menu item is set from the event in CommonMenuBar in case it originates somewhere else). The setVisibilityGUIPreference didn't seem necessary anymore.

I also made the updateDoneButton method in AbstractActionDisplay (was this the name?) only act when it is not ignoring events. I hope this doesnt create problems but it seemed to work. I noticed the turn events being overwritten by phase displays for not-current-phases.

pakfront (approved):
Looks good.

PR #4549, opened 2023-06-22
Unit Selector

SJuliez:
Changes to the unit selector:
- use a normal font again instead of the awful monospaced
- add some formatting to price and tonnage, controlled by the selected language (at least in MM, probably always US for now in MML/MHQ)
- centers some of the columns
- allows multiple space-separated search tokens in the text search (it is now possible to search for "Achi Sqd5" to find all Sqd5 Achileus BAs)

image

### DIFF
--- a/megamek/src/megamek/client/ui/swing/dialog/AbstractUnitSelectorDialog.java
+++ b/megamek/src/megamek/client/ui/swing/dialog/AbstractUnitSelectorDialog.java
@@ -14,7 +14,7 @@
  */
 package megamek.client.ui.swing.dialog;
 
-import megamek.MMConstants;
+import megamek.MegaMek;
 import megamek.client.ui.Messages;
 import megamek.client.ui.dialogs.BVDisplayDialog;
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/client/ui/swing/dialog/AbstractUnitSelectorDialog.java
```diff
@@ -854,16 +862,10 @@ public Class<?> getColumnClass(int col) {
             return getValueAt(0, col).getClass();
         }
 
-        @Override
-        public boolean isCellEditable(int row, int col) {
```
NickAragua, line 858: I assume this is false in the base class?

NickAragua (approved):
Looks fine, just one question.

megamek/src/megamek/client/ui/swing/dialog/AbstractUnitSelectorDialog.java
SJuliez: Yes. (IDEA told me this is equal to the base class)

PR #4572, opened 2023-06-29
Unit Selector: Speed up row height calculation

SJuliez:
I noticed that drag-resizing the unit selector table is very sluggish now when many units are displayed which is due to the row height calculation listeners. This fixes it (the unit selector table fortunately has uniform row height)


NickAragua (approved):
I hate to ask, but what was it doing before?

SJuliez:
It calculated all row heights individually, so for each row it cycled all columns, got the height from the renderer and used the largest of these. I just copied that code. I hadn't thought it's a problem and it isn't when done once, but when dragging the splitpane slider it's done very often. When the unit selector shows 10000 entries, it's too slow. It's possible that I'm overdoing things here but I found it difficult to convince a JTable to adapt the row heights when I want it to and the listener solution at least works reliably.

PR #4576, opened 2023-06-30
Gamemaster mode can edit unit during game

pakfront:
This allows a player with the Gamemaster (GM) role to use the right click menu to:
- open the UnitEditorDialog to add or remove damage from an Entity. This will update ammo counts and weapon usability but will NOT trigger unit immediate effects such as death, ejection, explosions, etc. (this is an issue with editing damage in the lobby as well.)
- open the CustomMechDialog (aka Configure dialog) to change pilot name and stats, change prone/hidden, everything you can do from the lobby except changing deployment zones. 

To do this, the player will have to become a Gamemaster by 
- using the **/gm** command 
- then having all other human players **/allowGM**.

FYI adding support for editing pilot health, buildings and hexes will take a bit more work but is possible in another PR


NickAragua (commented):
Nice. 

I think this may have some unintended effects if the unit has been damaged; we'll want to test around that, especially things like blown-off limbs, ammo explosions and crew effects.

BLR-IIC:
> Nice.
> 
> I think this may have some unintended effects if the unit has been damaged; we'll want to test around that, especially things like blown-off limbs, ammo explosions and crew effects.

I wonder how this would work with units being carried on other units, like mechanized BA, infantry in APCs, or dropships docked with jumpships...

pakfront:
> I think this may have some unintended effects if the unit has been damaged; we'll want to test around that, especially things like blown-off limbs, ammo explosions and crew effects.

 This is expected with the current PR. I think this is useful enough that we could put it in as in. We should create new PR to handle the affects, as it's fairly complex. For example, currently if you edit the damage:
  - destroying a side torso does not disable the corresponding arm
  - destroying the CT will not disable the unit. 
  - marking ammo as damaged does not trigger ammo explosions
  - damaging the head does not trigger damage rolls on the pilot

pakfront:
I added the Configure/CustomMech dialog as well. Lots of cool stuff you can do with it mid-game as GM such as change pilot names (surprise, it's Natasha!), unhide Princess units, hide units mid-game (not legal, I know, but great for RP)

NickAragua (approved):
Looks fine code-wise

PR #4587, opened 2023-07-01
BattleArmor: Allow variable size Cargo Lifter Manipulators

SJuliez:
As the title says, this allows BA Cargo Lifters to be of variable size, see MegaMek/megameklab#1269
Also adapts the few BA that have them.


NickAragua (approved):

NickAragua:
Looks like that name changes list validation is failing

PR #4617, opened 2023-07-11
Added null check to guard against NPE when offboard artillery target has disengaged

### LINKED ISSUE #4553: 0.49.13 offboard artillery Disengage causes uncaught nullpointer exception
Ver. 0.49.13, MekHQ campaign MegaMek
Windows 10
JAVA 20.0.1

Disengaging any of my off-board artillery causes a null pointer exception.

Load save file for Irregular forces scenario. replace players with AI. disengage any arty and then finish the arty phase. ending of arty phase should trigger bug.

Is Replicable without loading the game through the mekHQ save. 

Player name: Golden Kashicks
[Arty Disengage Bug.zip](https://github.com/MegaMek/megamek/files/11840825/Arty.Disengage.Bug.zip)

neoancient:
The artillery attack action normally targets hexes, but when the target is an offboard artillery unit the target is the unit instead. If the target disengages during the same phase that it is targeted, this will cause a null pointer exception.

Fixes #4553.

### DIFF
--- a/megamek/src/megamek/client/bot/princess/ArtilleryTargetingControl.java
+++ b/megamek/src/megamek/client/bot/princess/ArtilleryTargetingControl.java
@@ -430,8 +430,8 @@ public class ArtilleryTargetingControl {
             // calculate damage: damage - (10 * distance to me), floored at 0
             // we only say that it will actually be damage if the attack coming in is landing right after the movement phase
             double actualDamage = 0.0;
-            
-            if (aaa.getTurnsTilHit() == 0) {
+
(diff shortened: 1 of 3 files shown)


Windchild292 (approved):

neoancient:
I thought there might be some more problems so I did more testing. I've added a guard against an NPE when the artillery strike lands and in the bot's pathing. I also found and fixed a potential NPE unrelated to offboard units disengaging, when an offboard vehcile is hit and it checks whether its non-existing hex contains ice.

Windchild292 (approved):

PR #4622, opened 2023-07-13
FluffImageHelper

SJuliez:
Corrects the path chooser for AS Card fluff images so that all images are now found. Also a code duplication cleanup.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/util/FluffImageHelper.java
+++ b/megamek/src/megamek/client/ui/swing/util/FluffImageHelper.java
@@ -184,9 +184,6 @@ public class FluffImageHelper {
     private FluffImageHelper() { }
 
     private static String getImagePath(final ASCardDisplayable element) {
-        if ((element.getASUnitType() == ASUnitType.SV) && (element.hasMovementMode("a"))) {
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/client/ui/swing/util/FluffImageHelper.java
```diff
@@ -184,6 +184,9 @@ protected static Image loadFluffImage(final Entity unit) {
     private FluffImageHelper() { }
 
     private static String getImagePath(final ASCardDisplayable element) {
+        if ((element.getASUnitType() == ASUnitType.SV) && (element.hasMovementMode("a"))) {
+            return DIR_NAME_CONVFIGHTER;
+        }
```
Windchild292, line 189: I'm really not a fan of this following the below. Can it instead be done as a check on case SV with a fall through to case CV?

Windchild292 (commented):
One thing first

megamek/src/megamek/client/ui/swing/util/FluffImageHelper.java
SJuliez: Yeah that works

Windchild292 (approved):

PR #4624, opened 2023-07-14
CodeUtilities: Add Unit Testing

Windchild292:
Commit 1 adds unit tests, commit 2 expands Javadocs and fixes a rounding bug in MathUtility's long lerp.

### DIFF
--- a/megamek/src/megamek/codeUtilities/MathUtility.java
+++ b/megamek/src/megamek/codeUtilities/MathUtility.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 - The MegaMek Team. All Rights Reserved.
+ * Copyright (c) 2022-2023 - The MegaMek Team. All Rights Reserved.
  *
  * This file is part of MegaMek.
  *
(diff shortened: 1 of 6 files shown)


SJuliez (approved):
All good. Is there a difference between compareNullable and Objects.compare?

Windchild292:
Objects.compare doesn't handle one parameter being null and the other being not null... just both having values or both being null.

megamek/src/megamek/codeUtilities/MathUtility.java
```diff
@@ -21,57 +21,117 @@
...
+     * a f of 0d will return the minimum, a f of 1d will return the maximum. Otherwise, this will
+     * return the rounded long between the two points
+     * @note we assume this is used correctly, and thus do not check if the min and max are handled
+     * incorrectly
      */
-    public static long lerp(final long min, final long max, final long f) {
+    public static long lerp(final long min, final long max, final double f) {
         // The order of operations is important here, to not lose precision
-        return min * (1L - f) + max * f;
+        return Math.round(min * (1L - f) + max * f);
```
NickAragua, line 78: Every time I see a change to a "deep library" function like this, I get scared. Do we have any idea where it's used? And if so, can we get a short write-up of stuff that can be tested from the user side to see if the change "affects" anything?

SJuliez: According to my IDE the lerps are not currently used at all. I guess it's safe to merge.

PR #4647, opened 2023-07-19
clear calcDmgPerHitReport once it is added to vPhaseReport

### LINKED ISSUE #3654: [49.8 Testing Branch] Bug or RFE?  Confusing Infantry/BA damage message.
21:05:30,385 INFO  [megamek.MegaMek] {main}
megamek.MegaMek.initializeLogging(MegaMek.java:115) - Starting MegaMek v0.49.8-SNAPSHOT
	Build Date: 2022-05-16T01:41:50.033873500
	Release Date: 2022-02-18
	Today: 2022-05-24
	Origin Project: MekHQ
	Java Vendor: AdoptOpenJDK
	Java Version: 11.0.10
	System Locale: en_US

I dont know if this is version specific or just always like this.  I rarely use infantry or BA so its new to me regardless.  

Basically, the damage output for BA (or perhaps anything) vs. Infantry is confusing.  I will post a couple of screens as examples below. 
 It seemed like a bug to me at first but Mishra talked me through it and we both referenced TW and it appears the calculation may be correct, but we both agreed that how it is presented is very confusing.  Its confusing, at least to me the user, because:

1.  The damage calculation is hidden.  Here, in the example, it just shows "damage changed from 2 to 6" and doesnt tell you how it arrives at the 6.  Whether it is 1d6/2 or 1d6 or 1d6*2, etc.  
2. There is a "repeat" damage message that has no obvious purpose.  On the 2nd damage roll it shows the previous one, on the 3rd it shows the previous two, etc.  It may actually serve a purpose.  But we could not tell what it was exactly.  We thought it might be an attempt at arriving at a cumulative number?  Whatever it is doing, it is confusing how it is presented.
3. He also mentioned under the burst fire heading that it appears to suggest having a "running total" of damage.  We could not tell if that was happening or even needed to happen, but just wanted to point it out.

I can provide a save if needed, but but I dont have one directly related to these screenshots.
[megamek.log](https://github.com/MegaMek/megamek/files/8767738/megamek.log)
[mekhq.log](https://github.com/MegaMek/megamek/files/8767740/mekhq.log)
Screenshot 2022-05-24 162324
Screenshot 2022-05-24 165825
Screenshot 2022-05-23 231123

kuronekochomusuke:
- clear calcDmgPerHitReport, once it is added to vPhaseReport

image

fixes #3654

### DIFF
--- a/megamek/src/megamek/common/weapons/ACAPHandler.java
+++ b/megamek/src/megamek/common/weapons/ACAPHandler.java
@@ -75,6 +75,7 @@ public class ACAPHandler extends ACWeaponHandler {
         // Report calcDmgPerHitReports here
         if (!calcDmgPerHitReport.isEmpty()) {
             vPhaseReport.addAll(calcDmgPerHitReport);
+            calcDmgPerHitReport.clear();
         }
 
         // if the target was in partial cover, then we already handled
--- a/megamek/src/megamek/common/weapons/MGAWeaponHandler.java
+++ b/megamek/src/megamek/common/weapons/MGAWeaponHandler.java
@@ -171,6 +171,7 @@ public class MGAWeaponHandler extends MGHandler {
         // Report calcDmgPerHitReports here
         if (!calcDmgPerHitReport.isEmpty()) {
             vPhaseReport.addAll(calcDmgPerHitReport);
+            calcDmgPerHitReport.clear();
         }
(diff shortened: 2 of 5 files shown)


HammerGS (approved):
Tested the PR and seems to work no issues.

PR #4651, opened 2023-07-21
Small log cleanup

SJuliez:
This replaces the everpresent exception that is logged when MegaMek.jar isnt there with a simple warning; also loaded game options are now logged as one block instead of a load of individual entries

image
image


HammerGS (approved):
Tested the PR and seems to work for me.

PR #4658, opened 2023-07-22
allow key binds to work on report phases

### LINKED ISSUE #4656: [0.49.13] WASD does not scroll map during Round Report.
MegaMek standalone 0.49.13
Java 11
Windows 10

With the new 'separate window' Round Reports, while it is possible to use the mouse scroll (if you have a mouse) to move the map around (e.g. to see which of your units has been reported as having failed a PSR and has parked in a building _again_), WASD does not appear to scroll the map while the Round Report window is up, even if the main window has the focus.

kuronekochomusuke:
- allow game board key binds to work on report phases, now that these phases have the game board visible.

fixes #4656


HammerGS (approved):
Tested and works for me.

NickAragua (approved):

PR #4660, opened 2023-07-22
TestMech: Correction to internal structure validation

### LINKED ISSUE #4657: Awesome 11H invalid
In 0.49.14, with validation updated to check the number of crit slots for internal structure, the Awesome 11H registers as invalid. It clearly has Clan Endo Steel, but seems to think it's using IS Endo Steel after being loaded. My guess is that this is due to it being a Mixed Base IS unit. So we can maybe rework the unit to Base Clan (if that doesnt add other issues) or (probably better) change the loader to correctly load the structure type.

image

image

SJuliez:
Fixes #4657 
The structure validation was not quite complicated enough yet. But, seemingly no update to the loader is necessary. It is possible to get the correct internal structure from the Entity ... if you know how.


HammerGS (approved):
Tested and works for me.

mhjacks:
Tested and validated here too.  Thanks for the fix!

PR #4661, opened 2023-07-23
Feature - Add audio tab and master volume setting

sensualcoder:
This PR adds a Master Volume setting to the application to control the sound output, and also moves the audio settings into a new tab in the Client Settings menu.

In addition, there is a new Sound class that encapsulated storing and playing sound files, and a SoundManager class to provide an interface for all of the current sounds effects as well as handling audio settings changes.


HammerGS:
Thanks for the PR. We are getting close to the 0.49.14 Dev release, so this will get tagged with a "For new dev cycle (49.15).

If you haven't joined our Discord I'd recommend joining it as its an easy way to get feedback.

NickAragua (approved):

sensualcoder:
Updated this PR with some small changes

megamek/src/megamek/client/ui/swing/audio/Sound.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+package megamek.client.ui.swing.audio;
+
+import megamek.common.annotations.Nullable;
+
+import javax.sound.sampled.Clip;
+import javax.sound.sampled.FloatControl;
+
+public class Sound
```
NickAragua, line 27: Can we get some comments in this class describing what it does?

NickAragua (changes requested):
Looks good, I'd like to see a few comments on some the new classes.

sensualcoder:
Added some comments explaining the Sound class and methods and the methods in SoundManager. Let me know if I should add more :)

PR #4662, opened 2023-07-23
BA validation: Add multiple equipment validation tests

SJuliez:
This adds various tests for BattleArmor validity (although certainly not all by far). It makes one unit invalid, the Clan Interface Armor, which uses two Mission equipment, where TM p262 seems to say that only one is allowed on a suit (Edit: The TRO entry uses two mission equipments, this is not an MM data issue).
Accidentally included the change to the FilteredUnitListTool, but it doesnt matter, it's only a tool.

Fixes MegaMek/megameklab#1265
Fixes MegaMek/megameklab#1173


HammerGS (approved):
Tested and seems to be working for me.

PR #4663, opened 2023-07-25
Allow free facing change only after aerospace maneuvers that allow it.

### LINKED ISSUE #4606: [49.13] Half-roll maneuver allows illegal facing change for aerospace fighter
### Environment
This bug is present in version 0.49.13, but has also been confirmed to exist as far back as version 0.48. 

Executing a half-roll maneuver at the start of movement for an aerospace fighter will allow it to make a facing change of one hexside, even if should not be able to do so. Per Total Warfare page 85, the half-roll maneuver should have no effect on ability to change facing. 

Reproduction steps: 
1. Start a Megamek game and add an aerospace fighter (I use a Seydlitz-21).
2. Select a ground map of reasonable size (I use 32x33 Grasslands 1).
3. Start the game and deploy your aerospace fighter. 
4. Accelerate once, then click on a hex a few hexes to the northeast or northwest of your unit . Part of the movement path should be grayed out as expected, since the unit has not moved forward enough hexes to make a facing change, powered or otherwise.
5. Hit Escape to cancel all movement, accelerate once, then click on the same hex. Movement will now be allowed with a facing change in the unit's initial hex. 

Facing change correctly not allowed:
image

Facing change allowed after half-roll maneuver was declared:
image

This effect also appears in the Atmospheric Map, and can be reproduced similarly, although you will need to click on a hex directly northwest or northeast of your unit immediately after deployment to trigger the "unit has not moved enough hexes forward for a facing change" effect.

neoancient:
Currently facing changes are allowed after any aerospace maneuver, but these should only apply to the immelmann, the hammerhead, and the split-s.

Fixes #4606.

### DIFF
--- a/megamek/src/megamek/common/MoveStep.java
+++ b/megamek/src/megamek/common/MoveStep.java
@@ -3895,9 +3895,10 @@ public class MoveStep implements Serializable {
         }
 
         // if its part of a maneuver then you can turn
-        if (isManeuver()) {
+        if (isFacingChangeManeuver()) {
             return true;
         }
+
 
         if (en instanceof ConvFighter) {
             // conventional fighters can only turn on free turns or maneuvers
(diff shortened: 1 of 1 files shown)


HammerGS (approved):

HammerGS (commented):
Tested and works.

gsparks3:
One question while we're looking at this - does the current code allow choosing any facing after a hammerhead maneuver? Should only be a 180° flip. Not familiar enough with the code to tell just by looking.

Other two maneuvers can indeed select any facing afterwards.

neoancient:
That's what I get for relying on my faulty memory. Or maybe I thought the hammerhead could end facing any direction because MM allows it.

PR #4664, opened 2023-07-25
Sideslip mod for using VTOL jet booster

### LINKED ISSUE #4652: [0.49.14] VTOL Jet Boosters Not Applying +3 Modifier To Piloting Rolls
### Environment
megamek.MegaMek.initializeLogging(MegaMek.java:116) - Starting MegaMek v0.49.14-SNAPSHOT
	Build Date: 2023-07-07T03:05:14.634885372
	Today: 2023-07-20
	Origin Project: MegaMekLab
	Java Vendor: Eclipse Adoptium
	Java Version: 17.0.6
	System Locale: en_US

### Description
VTOL Jet Booster equipment is not applying the +3 PSR to avoid accidental sideslips when in use, see attached screenshots. Savegame attached that has two units with VTOL jet booster: [Onuris Attack VTOL C-Ph Mk I ID:100] and [Cardinal Transport C-Ph TX Mk III ID:106]

In Round 2, the Onuris and Cardinal both flanked at Jet Booster speed (orange text) but did not receive any penalty to their rolls.

### Files
[Jet Booster PSR Modifier.zip](https://github.com/MegaMek/megamek/files/12130523/Jet.Booster.PSR.Modifier.zip)

jetbooster1

jetbooster2

neoancient:
The rules for the VTOL jet booster are found on p 160 of TO: AU&E.

It works like MASC/supercharger except there is no roll for failure. Instead there is a +3 mod to the PSR to avoid sideslipping. The code is there to add it, but the mascUsed field of Entity never gets set. Furthermore the nag popup cannot depend on Entity::isMascUsed because it doesn't get set until the move is processed, and must rely on the MovePath.

Fixes #4652.

### DIFF
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -11348,7 +11348,7 @@ public abstract class Entity extends TurnOrdered implements Transporter, Targeta
     public PilotingRollData checkSideSlip(EntityMovementType moveType,
                                           Hex prevHex, EntityMovementType overallMoveType,
                                           MoveStep prevStep, int prevFacing, int curFacing, Coords lastPos,
-                                          Coords curPos, int distance) {
+                                          Coords curPos, int distance, boolean usingJetBooster) {
         PilotingRollData roll = getBasePilotingRoll(overallMoveType);
 
         if ((moveType != EntityMovementType.MOVE_JUMP)
@@ -11365,8 +11365,7 @@ public abstract class Entity extends TurnOrdered implements Transporter, Targeta
             if (isUsingManAce()) {
                 roll.addModifier(-1, "Maneuvering Ace");
             }
-            if ((getMovementMode() == EntityMovementMode.VTOL) && isMASCUsed()
-                    && hasWorkingMisc(MiscType.F_JET_BOOSTER)) {
+            if (usingJetBooster) {
                 roll.addModifier(3, "used VTOL Jet Booster");
             }
         } else if (moveType != EntityMovementType.MOVE_JUMP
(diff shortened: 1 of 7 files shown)


HammerGS (approved):
Tested and works.

megamek/src/megamek/common/Entity.java
```diff
@@ -11365,8 +11365,7 @@ public PilotingRollData checkSideSlip(EntityMovementType moveType,
             if (isUsingManAce()) {
                 roll.addModifier(-1, "Maneuvering Ace");
             }
-            if ((getMovementMode() == EntityMovementMode.VTOL) && isMASCUsed()
-                    && hasWorkingMisc(MiscType.F_JET_BOOSTER)) {
+            if (usingJetBooster) {
```
NickAragua, line 11368: Is there a way to have this go into an override of the method in the VTOL class instead? Have it call the base class and then append the "if using jet booster" stuff to the resulting PilotingRollData?

NickAragua (commented):
Looks fine, one suggestion for not adding to the Entity class and keeping the parameter list for checkSideSlip from expanding.

megamek/src/megamek/common/Entity.java
neoancient: It would be possible, but it wouldn't prevent the parameter list from expanding. It wouldn't be an override if it had different parameters, and if the VTOL version had different parameters you would have to conditionally typecast to VTOL to call the right method. I think it's much simpler to leave this code where it is.

NickAragua (approved):

megamek/src/megamek/common/Entity.java
neoancient: I thought about it a bit more and came up with a way to simplify it just a bit, taking your suggestion to override the method in VTOL.

Issue #4457, opened 2023-05-26
Skipped Last Unit in Firing Phase


Membooo:
Fill out the fields below with as much pertinent information as possible to help the developers solve the issue.
Search for your issue inside open and closed issues first before reporting so as not to make duplicates.
(Delete the above lines)

### Environment
What version of MegaMek does your issue apply to?  Note whether it is the stand alone MegaMek or HQ's MegaMek.
What operating system are you using?  Windows, Mac, or Linux?
What java version are you using?  If unsure, it will state your java version at the top of the megamek.log file.
(Delete the above lines)

HQ´s MegaMek 0.49.13 
	Origin Project: MegaMek
	Java Vendor: Eclipse Adoptium
	Java Version: 11.0.15
	Platform: Windows 11 10.0 (amd64)
	System Locale: de_DE
	Total memory available to MegaMek: 4 GB

[autosave.sav.gz](https://github.com/MegaMek/megamek/files/11575141/autosave.sav.gz)
[megamek CLIENT MEMO.log](https://github.com/MegaMek/megamek/files/11575142/megamek.CLIENT.MEMO.log)
[megamek DEDICATED SERVER.log](https://github.com/MegaMek/megamek/files/11575143/megamek.DEDICATED.SERVER.log)

### Description
Skipped last Unit in firing Phase! Firing Phase ended, before last Mech declared his shot!

### Files
[megamek CLIENT MEMO.log](https://github.com/MegaMek/megamek/files/11575137/megamek.CLIENT.MEMO.log)
[megamek DEDICATED SERVER.log](https://github.com/MegaMek/megamek/files/11575138/megamek.DEDICATED.SERVER.log)
[autosave.sav.gz](https://github.com/MegaMek/megamek/files/11575139/autosave.sav.gz)

kuronekochomusuke:
#3304

PhoenixHeart512:
Save file attached that replicates the issue for me 100% of the time.

Unzip this whole folder on top of an existing MHQ installation. The folder structure is matched to the MHQ directory, so all subfolders should get placed in the correct locations all at once as long as the root folder is correctly lined up. Overwrite any file conflicts that pop up.

There will be 2 Cardinals still left to fire. 

Use Cardinal ID:106 and fire both LRM15's at the enemy Lineholder ID:127. (north-east adjacent hex)

If it works, the second cardinal should get skipped.

The zip file includes:
1. save game named "atb1" which is in the state described above
[...]

Issue #4326, opened 2023-04-15
Last eligible unit in fire phase being skipped to end the Weapons Fire phase.


FlameDraken84:
MegaMek with MekHQ 0.49.12
Windows
Java 11.0.11

I am having a recurring issue where the last unit in the weapons fire phase is being skipped and ending the turn when that unit is eligible to fire weapons.  The issue is inconsistent when it appears as it can happen several turns back to back but then not occur in the next turn.  It is also not consistent between games.  Some games I won't have the issue, then a game right after I will see the issue pop up again.

In detail, I will have my second to last unit fire its weapons and click on done.  The game will not switch over to the final unit in the fire phase but will end the turn instead.

The save file attached has my second to last unit ready to fire.  Firing either the Commando or Strider's weapons will skip the other mech's fire phase.  All other units have fired at this time.

The log is from a load of the save just before the error.

[megamek.log](https://github.com/MegaMek/megamek/files/11239254/megamek.log)

[Bug WF Phase.sav.gz](https://github.com/MegaMek/megamek/files/11239232/Bug.WF.Phase.sav.gz)

HammerGS:
Are you using any of Sim options like Sim Fire, Sim Physical?

If so we’ve know the issue is related to that and they shouldn’t be used.

FlameDraken84:
I am using sim targeting, fire and physical.  

I also use skip ineligible physical but not firing if that helps.

HammerGS:
As mentioned, the sim options shouldn't be used as they are unofficial and giving issues. We'll attempt to track things down with the save provided. But other saves have proven impossible to duplicate.

HammerGS:
Only error in log

```
megamek.client.ui.preferences.SuitePreferences.readPreferencesNode(SuitePreferences.java:184) - No class with name mekhq.gui.dialog.nagDialogs.UnresolvedStratConContactsNagDialog found
java.lang.ClassNotFoundException: mekhq.gui.dialog.nagDialogs.UnresolvedStratConContactsNagDialog
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
	at java.base/java.lang.Class.forName0(Native Method)
	at java.base/java.lang.Class.forName(Class.java:315)
[...]

Issue #3304, opened 2021-12-10
Last unit ignored


kernhe:
### Environment
Version: 0.49.6-SNAPSHOT
OS: Linux 64 (ubuntu 18.04)
Java: openjdk version "11.0.11" 2021-04-20

### Description
I came upon an issue that is happening from while to while, I cannot replicate it, but it happens at least once in every battle: when clicking "Done" the last unit (mech, VOTL, ...) in the battle (e.g. the fourth mech in a lance if only 4 mechs are on my side, the 12th when I have a company,...) is ignored in the movement and than later in the firing phase.

I can select the last unit manually, but if I use the "Done" button in movement phase, after the 3rd click on done, the firing phase is starting. The bot moves all units.

Krrast:
I've run across this problem my last couple drops with Nightly MekHQ #618 version of MM.  Can provide log if it helps, didnt want to double the submission.

HammerGS:
Logs and a save would be a huge help.

Windchild292:
To clarify, a before save

Krrast:
[Black Hand30760929.cpnx.gz](https://github.com/MegaMek/megamek/files/7776037/Black.Hand30760929.cpnx.gz)
[megamek.log](https://github.com/MegaMek/megamek/files/7776038/megamek.log)
i dont have a save of the drop, but the drop that is pending is the second one i've had the issue with, the previous drop was last contract.  Player name is Black Hand. It didnt skip any of my movements that I can recall, but about 1/2 the time the BattleAxe's turn came during firing phase it skipped it.

Windchild292:
I just ran through two games of 25 phases on the above campaign, trying to recreate. I could not do so from the provided information... but do know one way this can occur. If you rapidly hit done with the doing nothing nag disabled, it will complete the turn for both units and seem to not provide a turn for the unit.

I'll need a MegaMek save from the phase this issue occurs, and preferably with one from the phase before it occur to trace anything down.

sshagent:
[mmlogs.zip](https://github.com/MegaMek/megamek/files/8053591/mmlogs.zip)

As per conversation on discord in #nightlies-discussion I've included logs from server and client.  I've put some notes in a text file to try and explain what we've seen.  
After the game was reloaded back in, every opportunity it was spamming in chat that the bots (or everyone) had disconnected that wasn't the case.  it kept saying it, so if you see that...we weren't doing that.

PhoenixHeart512:
[...]

PR #4674, opened 2023-07-26
Fix potential race condition in Game::turnVector.

neoancient:
In a simultaneous phase it is possible for turnIndex to change and then be changed again in another thread before it is used to access turnVector. If this happens in changeToNextturn() just when a player's final turn should start, it would result in that turn being skipped. Adding a synchronized block it prevents another thread from changing the value until the new turn is retrieved. I added synchronized blocks to several other places involving the same to fields to make them thread safe.

Given the intermittent nature of the issue, the only way to determine whether this fixes the issue is extensive testing.

Cf. #4457, #4326, #3304

### DIFF
--- a/megamek/src/megamek/common/Game.java
+++ b/megamek/src/megamek/common/Game.java
@@ -89,7 +89,7 @@ public class Game extends AbstractGame implements Serializable {
     /**
      * The current turn list
      */
-    private Vector<GameTurn> turnVector = new Vector<>();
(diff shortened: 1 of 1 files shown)


HammerGS (approved):
I’ve played a few games with Sim Fire, Sim Physical on and had no issues.

HammerGS:
I think we should merge this for .14 and get it into the wild for testing and feedback.

NickAragua (approved):
Seems straightforward. But it's threads, so lol?

PR #4678, opened 2023-07-29
unit tooltip display tank crits in mini armor display

### LINKED ISSUE #4636: RFE: Show motive crits on unit display and/or tooltip for vehicles
Vehicles can suffer an arbitrary number and combination of MP-reducing motive crits (see TW page 193). While multiple crits of the same type stack for MP-reduction purposes, the Driving Skill Roll penalty from each type of crit is only applied once. As of version 49.13, the reduced MP from motive crits (or any other source) is displayed in the tooltip, but the number and type of motive crits affecting the unit are not, nor is it shown anywhere else in the unit display. 

Ideally, it would be possible to view the motive crits currently affecting a vehicle somewhere in the unit display and/or unit tooltip (the latter of which is currently part of the unit display anyway). Note that this is only relevant for vehicles. VTOL units may need a slightly adjusted display since they do not suffer the same motive crits as ground vehicles; they instead lose 1 cruise MP each time the rotor is hit, plus another 1 cruise MP per Rotor Damaged critical hit. Since those do not affect DSRs, displaying them is not necessarily as important.

kuronekochomusuke:
- unit tooltip - display tank crits in mini armor display
- show Stabilizers, Turret Locked, Engine, Sensors, Motive System Damage Level

image
image
image
image

fixes #4636


JonhyBurner:
It might be a matter of opinion, but when moderate and hevy motive damage occurs, maybe lower levels should be crossed out as well, to be more visually consistent.

gsparks3:
No, those crits happen separately, and need to be noted separately (it's basically noting which of them are currently being applied to Driving Skill Rolls; you can actually get multiple instances of each, but the DSR penalties do not stack).

SJuliez:
I suggest not displaying anything until damage has occurred.

kuronekochomusuke:
for me I prefer to see the undamaged ones, so you can see which hits the unit can take.

megamek/src/megamek/common/Entity.java
```diff
@@ -14202,10 +14202,10 @@ public static String getEntityMajorTypeName(long typeId) {
      */
     public static String getEntityTypeName(long typeId) {
 
-        if ((typeId & ETYPE_BIPED_MECH) == ETYPE_BIPED_MECH) {
-            return "Biped Mech";
-        } else if ((typeId & ETYPE_LAND_AIR_MECH) == ETYPE_LAND_AIR_MECH) {
+        if ((typeId & ETYPE_LAND_AIR_MECH) == ETYPE_LAND_AIR_MECH) {
```
NickAragua, line 14205: It doesn't look like there are any actual code changes here, can you point out what changed substantively here? If it's just random code re-arrangement, let's roll this file back.

NickAragua (commented):
Looks mostly good, just puzzled about one thing.

megamek/src/megamek/common/Entity.java
kuronekochomusuke: the unit typeId is additive.   So the order of the checks matter to get the correct name.

So a Landair Mech is a Biped Mech and a Landair Mech.
Orignial code a Landair Mech was returning as Biped Mech.
With the new code order Landair Mech returns as Landair Mech.

NickAragua: Good thing we don't use this method for determining anything important, it must have been this way for quite a while. [nervous laughter]

NickAragua (approved):

Issue #1230, opened 2019-01-23
[0.45.2] 0 damage to VTOL rotor destroying mast mount


Xeno426:
### Environment
0.45.2
Platform Windows 7 6.1 (x86)
Java version 1.8.0_151

### Description
VTOLs that use Ferro-Lamellor armor are able to take 0 damage to their rotors. This should prevent the destruction of mast mounts, but this isn't the game in MM.
![rotor damage](https://user-images.githubusercontent.com/3220418/51576671-cd42dd00-1e7c-11e9-9c2d-ebfa2a6069fa.PNG)
0 damage isn't damage, so nothing should happen to the mast mount.
![mast mount dead](https://user-images.githubusercontent.com/3220418/51576677-d0d66400-1e7c-11e9-901a-ff644648d70c.PNG)
But it does.

wildj79:
As far as I can tell, this is working RAW.

From TO, p. 280.

>**Rules Level:** Experimental 
**Available To:** BM, IM, CV, SV, AF, CF 
**Tech Base (Ratings):** Clan (F/X-X-F) 
**Game Rules:** A location protected by Ferro-Lamellor Armor reduces all damage by 1 point for every 5 points (or fraction thereof) delivered per hit (to a minimum of 0 points of damage per hit). Excess damage will affect internal structure or structural integrity per normal rules for the location's structure type. Weapons reduced to zero damage effects by Ferro-Lamellor Armor (such as LB-X cluster munitions), may not inflict pilot injury in the event of a cockpit-location hit, nor may they deliver a penetrating or "floating" critical from special hit location rolls. **However, against Combat and Support Vehicles, such weapons may inflict Motive System damage effects normally (because such units feature more exposed movement mechanisms).** In addition, Ferro-Lamellor Armor negates the bonus armor-penetrating effects of Armor-Piercing Ammo (see p. 140, TW), Tandem-Charge Missiles (see p. 372) and 'Mech Tasers (see p. 345). If these weapons are fired against a location protected by Ferro-Lamellor, they deliver their standard damage to the armor only (Armor-Piercing Ammo hits as a standard AC round; Tandem-Charge Missiles hit as standard SRMs; 'Mech Tasers inflict only 1 point of armor damage).

Xeno426:
![xotl 0 damage is not damage](https://user-images.githubusercontent.com/3220418/51643553-7d265200-1f32-11e9-886c-d18bac601373.PNG)

Sleet01:
[...]

PR #4691, opened 2023-08-04
Prep work for MML #1230

Sleet01:
This change adds a minDoors field for all bays, with only Quarters and Infantry Bays (including BA) allowed to not designate any doors.

Also fixes a typo where Jump Ships had two max door count definitions but Space Stations had none.

NOTE: needs to be pulled prior to the MML PR or that one will break.

### DIFF
--- a/megamek/src/megamek/common/verifier/TestAero.java
+++ b/megamek/src/megamek/common/verifier/TestAero.java
@@ -35,23 +35,23 @@ import java.util.function.Function;
 /**
  * Class for testing and validating instantiations for Conventional Fighters and
  * Aerospace Fighters.
- * 
+ *
  * @author arlith
  * @author Reinhard Vicinus
  */
 public class TestAero extends TestEntity {
     private Aero aero = null;
-    
+
     /**
-     * An enumeration that keeps track of the legal armors for Aerospace and 
-     * Conventional fighters.  Each entry consists of the type, which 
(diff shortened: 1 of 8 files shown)


megamek/src/megamek/common/verifier/TestAero.java
```diff
@@ -170,38 +170,40 @@ public static int slotsPerArc(Aero aero) {
     public static int maxBayDoors(Aero aero) {
         if (aero.hasETypeFlag(Entity.ETYPE_WARSHIP)) {
             return 8 + (int) Math.ceil(aero.getWeight() / 100000);
-        } else if (aero.hasETypeFlag(Entity.ETYPE_JUMPSHIP)) {
+        } else if (aero.hasETypeFlag(Entity.ETYPE_SPACE_STATION)) {
```
NickAragua, line 173: Question: is there a rulebook reference for this? Obviously it's a bug, but just want to make sure we're using the right formula for the right stuff.

Sleet01: `ADVANCED AEROSPACE UNIT 
MAXIMUM BAY DOORS TABLE
Advanced Unit Type Maximum Bay Doors
JumpShips 7 + [Unit Weight ÷ 50,000]*
Space Stations 8 + [Unit Weight ÷ 75,000]*
WarShips 8 + [Unit Weight ÷ 100,000]*`

- `Strategic Operations` p. 155

PR #4699, opened 2023-08-11
Fix homing arty splash

### LINKED ISSUE #4697: Homing Artillery Splash Damage Being applied to flying fighters
Fill out the fields below with as much pertinent information as possible to help the developers solve the issue.
Search for your issue inside open and closed issues first before reporting so as not to make duplicates.
(Delete the above lines)

### Environment
Megamek 49.14

### Description
Using Aerospace units on a ground map sheet rules.
If an aerospace fighter (At altitude 1, 3, 5, ,7, and 9 in one test) happens to be in the hex where Homing artillery is designated to hit a TAG'd target, The 5 points of AOE splash damage is hitting the fighter.  It also hit a VTOL hovering 5 levels above the target mech.

Normal Artillery fire does not have this problem, its just the 5 point splash from Homing artillery. (Both arrow IV and thumper in testing.)

Incidentally I discovered that The VTOL was still being counted for stacking purposes despite its elevation when I couldn't move a WIGE into the same hex as the target mech.

### Files

MMimage

NickAragua:
Switched the homing artillery handler to use the standard "damage entity" logic rather than its own version. Fixes #4697.

### DIFF
--- a/megamek/src/megamek/common/weapons/ArtilleryWeaponIndirectHomingHandler.java
+++ b/megamek/src/megamek/common/weapons/ArtilleryWeaponIndirectHomingHandler.java
@@ -147,6 +147,7 @@ public class ArtilleryWeaponIndirectHomingHandler extends ArtilleryWeaponIndirec
         r.add(roll);
         vPhaseReport.addElement(r);
 
+        roll = 12;
         // do we hit?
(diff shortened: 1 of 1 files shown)


HammerGS (approved):
Can't speak to code quality. But tested and works for me.

PR #4700, opened 2023-08-11
[Draft] TestBot Removal

rjhancock:
Removal of the TestBot Client and related classes and tests that are long deprecated and not in use anymore.


NickAragua (changes requested):
A couple of logic errors in the ClientGUI class. Please take another pass and make sure you aren't eliminating code that's run when "!testbot".

Also, really not a fan of the arbitrary whitespace changes, as, even with "whitespace off", there's still a lot of them and it makes picking the actual changes out very difficult.

rjhancock:
You're right about the logic issues. Will get those fixed. Did this late last night and wasn't entirely focused (thanks for the catch).

It's also why I left this as draft as I still need to compile and test.

SJuliez:
I updated this; therefore closing this one.

PR #4716, opened 2023-08-22
Camo rotation and scaling

SJuliez:
This PR adds rotation and scaling to unit camos. This can be used to differentiate forces without using different camos and to have the camo actually be visible on smaller units. Or just to play around.

Example: 
Normal camo use:
image

Individual camo settings for rotation and scale
image

The settings are only applied to units, the camo itself is unchanged:
image

The settings are changed in the camo chooser. When ignored, things work as they did before.
image


SJuliez:
Yes as part of setting an individual camo. In the screenshot the smaller units use a different scale

neoancient (approved):

Thom293:
This looks really cool.  I have the nightly from today, but I cant find where to adjust these settings?  I tried individual camo in MHQ and Camo settings in MM.  Could I trouble you for where it is located?

SJuliez:
@Thom293 sorry never saw your comment. I suppose you found it by now but the settings are in the camo chooser dialog right above the OK/Cancel buttons

LumberingTroll:
Not sure what has happened but I cannot find these settings anywhere, when I go to to the individual camo settings for a unit there are no additional options like this. Tried using the normal MegaMek 49.14 version and the MegaMek version that is bundled with MekHQ 49.14

kuronekochomusuke:
this is in version 49.15 which has not been released yet.  can try it in the nightly or wait for the 49.15 release.

LumberingTroll:
> this is in version 49.15 which has not been released yet. can try it in the nightly or wait for the 49.15 release.

awesome thanks, I was not aware!

PR #4717, opened 2023-08-22
Treat FTL ammo as standard LRM ammo when affected by enemy ECM

### LINKED ISSUE #4704: Follow the Leader LRM Ammunition Unaffected by Hostile ECM
48.0 stable, standard Megamek client, windows 10, probable rules issue.

image
Simply put, according to Tactical Operations Advance Equipment, Page 180 (lowest section) Follow-The-Leader Missiles are supposed to act as regular LRMs when travelling into or through a hostile ECM field. As a couple rounds of testing seems to confirm, the rounds do not lose either their cluster bonus, or their single-location-grouping property when fired into an ECM field, at an ECM source. Whether they lose their to hit malus if outside of the ECM or not I do not know, but it appears ECM does nothing to this munition in Megamek at this time.

Save game attached. Steps to reproduce are simply load a catapult with FTL LRMs and shoot them at a raven 3L which is standing still, with the catapult outside the ECM field.
[ECM FTL Interaction.sav.gz](https://github.com/MegaMek/megamek/files/12328177/ECM.FTL.Interaction.sav.gz)

neoancient:
Per TO: AU&E p. 180, follow-the-leader LRM ammo is treated as regular ammo when it passes through an enemy ECM field. This removes the following affects:
1. Remove the +2 to-hit mod.
2. Remove the +1 cluster mod.
3. Apply damage in clusters of 5 rather than to a single location.

Fixes #4704


Sleet01 (approved):
Looks good to me.

NickAragua (approved):

PR #4721, opened 2023-08-23
Ryoken date fixes

IanBellomy:
Was tipped off to some discrepancies with the master unit list. 
Note the Ryoken P's date goes from 3150>>>3049, which smells fishy, but it's onboard tech is available then, so...


HammerGS:
Thanks, legacy of the Record Sheet Development

PR #4740, opened 2023-09-02
Fix 4739 custom protomech bay does not load

### LINKED ISSUE #4739: MegaMek does not populate ProtoMech Transport bays in custom dropships.
### Environment
MekHQ 49.14
Windows 10
Java 11.0.14

### Description
Loaded the Aurora dropship into MM Lab.  Removed 1 Mech Bay in favor of a protomech Bay with a 9 PM capacity and 3 doors.  Altered escape pods to meet new evacuation requirements and adjusted cargo to reflect remaining free tonnage.  Gave a distinct Model type  (Society) and saved into a custom folder in the MechFiles directory.  Exported to PDF (enclosed.)

Exported PDF shows the PM bay.  When selecting and loading the custom dropship into MegaMek for play, the Protomech Bay does not populate, either in the selection screen or in the lobby environment.

However, the following cannon dropships DO display their protomech transport bays properly:
Outpost (3070)
Arcadia (3066)
Arcadia (3080)

### Files

This is the .blk produced by MM Lab for the custom unit.
[Aurora (Society).zip](https://github.com/MegaMek/megamek/files/12502615/Aurora.Society.zip)

This is the PDF Export
[Aurora (Society).pdf](https://github.com/MegaMek/megamek/files/12502473/Aurora.Society.pdf)

This is a screenshot of the unit selection window
DS PM Bay Aurora

This is a screenshot of the lobby, attempting to load a protomech into the dropship (the load is grayed out due to 0 viable bays)
DS PM Bay Aurora Lobby

Sleet01:
It looks like there was a change, or series of changes, to ProtoMechBay.java back in March of 2022 that broke compatibility between _saving_ ProtoMech bays (or "ProtoMekBays") and _loading_ them.

The fix I've proposed here just reads both lines and treats them equally.  It's quick and easy and I've tested it with the original file from #4739 successfully.

An alternative fix would be to revert the March 2022 change completely; I feel this warrants consideration as there are several places in the MML UI (not to mention _in the filename itself_) where we use "ProtoMech".

Close #4739

### DIFF
--- a/megamek/src/megamek/client/ui/swing/lobby/LobbyMekPopup.java
+++ b/megamek/src/megamek/client/ui/swing/lobby/LobbyMekPopup.java
@@ -210,7 +210,7 @@ class LobbyMekPopup {
             popup.add(squadronMenu(clientGui, true, listener, joinedEntities));
         }
 
-        if (accessibleProtomeks) {
+        if (allProtomeks) {
             popup.add(protoMenu(clientGui, allProtomeks, listener, joinedEntities));
         }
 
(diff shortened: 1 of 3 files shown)


HammerGS:
We should move to Mek. We try to avoid Mech.

Sleet01:
> We should move to Mek. We try to avoid Mech.

The MML UI still uses "mech", and extant BLK files do as well.  Additionally, the reporter of the bug found that there's a further mismatch between the bay type "ProtoMek" and the unit type itself, so we'd probably need to update a lot more than just BLKFile.java and the MML UI.

I propose this:

1) roll back the partial change to "ProtoMekBay" that is breaking stuff, keeping the BLKFile.java fix that loads both formats for interim compatibly.
2) I'll implement (either just for this type, or for all bays and units) the save/load validation unit tests that can confirm saving, loading, and, I guess, that putting units into the bays works correctly.
3) we do a single big change to replace all instances of "mech" in the UI and code base for all 3 main programs at a later date.

@HammerGS what say you?

HammerGS:
Sounds good to me. The move to Mek needs to happen, but there is no set timeline for it. But we should make sure there are no impacts to MekHQ as well.

NickAragua (approved):

Sleet01:
> Sounds good to me. The move to Mek needs to happen, but there is no set timeline for it. But we should make sure there are no impacts to MekHQ as well.

Well, digging into the code more, I'm finding that all three projects sport an almost-even mix of both forms ("ProtoMech" vs "ProtoMek") so maybe I'll just throw up my hands in despair and try to fix the issue with loading ProtoMechs into ProtoMek bays (or vice-versa) in MekHQ.

I'd estimate that normalizing all remaining references using "ProtoMech" -> "ProtoMek" to be at least a week of full-time work but it looks like about that much work has been done already, so it might be good to set up a longer-running branch to track that.

Sorry for all the hassle; this turned out to be a can full of _other_ cans of worms.

HammerGS:
> > Sounds good to me. The move to Mek needs to happen, but there is no set timeline for it. But we should make sure there are no impacts to MekHQ as well.
> 
> Well, digging into the code more, I'm finding that all three projects sport an almost-even mix of both forms ("ProtoMech" vs "ProtoMek") so maybe I'll just throw up my hands in despair and try to fix the issue with loading ProtoMechs into ProtoMek bays (or vice-versa) in MekHQ.
> 
> I'd estimate that normalizing all remaining references using "ProtoMech" -> "ProtoMek" to be at least a week of full-time work but it looks like about that much work has been done already, so it might be good to set up a longer-running branch to track that.
> 
> Sorry for all the hassle; this turned out to be a can full of _other_ cans of worms.

Welcome to MegaMek! :)

PR #4758, opened 2023-09-07
Non-Meks in fire hexes

### LINKED ISSUE #4734: Grounded Aerospace Fighter Burned down.
### Environment
MM49.14

### Description
So because reasons I had a grounded aerospace fighter in a hex that was on fire.
Since aerospace fighters are heat tracking units it should of gained 5 heat.  (pg 43 Tacops)
Instead the "all other units" rule of needing an 8+ to not die was used.  (Which I failed prompting me to take notice of what killed me.)

SJuliez:
The way I understand TO:AR pg.41-43, all heat-tracking unit types (Mek, ASF, SC) are treated equally when standing in fire. All of those are not "All Other units" of pg.43 and don't get destroyed on a bad roll. They just suffer heat. This PR implements this.

Fixes #4734

### DIFF
--- a/megamek/src/megamek/server/GameManager.java
+++ b/megamek/src/megamek/server/GameManager.java
@@ -7607,21 +7607,15 @@ public class GameManager implements IGameManager {
 
             // check to see if we are a mech and we've moved OUT of fire
             Hex lastHex = game.getBoard().getHex(lastPos);
-            if (entity instanceof Mech) {
+            if (entity.tracksHeat() && !entity.isAirborne()) {
                 if (!lastPos.equals(curPos) && (prevStep != null)
-                        && ((lastHex.containsTerrain(Terrains.FIRE)
-                        && (prevStep.getElevation() <= 1))
-                        || (lastHex.containsTerrain(Terrains.MAGMA)
-                        && (prevStep.getElevation() == 0)))
+                        && ((lastHex.containsTerrain(Terrains.FIRE) && (prevStep.getElevation() <= 1))
+                        || (lastHex.containsTerrain(Terrains.MAGMA) && (prevStep.getElevation() == 0)))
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/server/GameManager.java
```diff
@@ -19505,11 +19508,6 @@ private void doFlamingDamage(final Entity entity, final Coords coordinates) {
             return;
         }
 
-        // mechs shouldn't be here, but just in case
-        if (entity instanceof Mech) {
-            return;
-        }
-
         // fire has no effect on dropships
         if (entity instanceof Dropship) {
```
kuronekochomusuke, line 19512: is this still needed with the check above?

NickAragua (approved):
Approved subject to removal of unnecessary dropship check on line 19512.

megamek/src/megamek/server/GameManager.java
```diff
@@ -19481,6 +19479,11 @@ private void resolveHarJelRepairs() {
      * @param coordinates the coordinate location of the fire
      */
     private void doFlamingDamage(final Entity entity, final Coords coordinates) {
+        // TO:AR p.41,p.43
+        if (entity.tracksHeat() || entity.isDropShip()) {
```
NickAragua, line 19483: Don't all dropships track heat?

gsparks3: DS technically are not heat-tracking units, since they never use the Heat Scale (even though their weapons usage is limited by heat sink capacity). They are explicitly called out as unaffected by fire in the TO:AR fire rules, however.

SJuliez: tracksHeat() is only true for Meks, ASF and Small Craft. Dropships have an override to return false.

PR #4760, opened 2023-09-08
Fix 4754 vtols and cvs missing troop space bays

### LINKED ISSUE #4754: [0.49.15-SNAPSHOT] Vehicles seem not to be registering troopspace correctly including DI Multipurpose VTOL
### Environment
Megamek integrated with MekHQ, megamek commit  dd008f7a14 on master on Fedora Linux, JVM 11

### Description
Cannot load battle armor onto VTOLs; can still load onto, e.g. the R10 IVC

### Files
image

Sleet01:
Fixes:
1) An off-by-one error I introduced, which prevents units with single bays (usually troop/infantry) from populating.
2) 55 or so BLK files with multiple `<transporters>` blocks that wouldn't have loaded correctly anyhow - only the first can be read.
3) Corrected several vehicles where multiple bays had been added, or the top bay was a correction for the lower bay, or (especially in the case of the Ajax D) bays, weapons, and equipment were incorrect.
4) Reverted the change that is causing MML to crash in the cache view, so that I could use it to correct 2 and 3 above.

Testing:
- loaded all of the vehicles with corrected bays in MML, with a sampling also loaded in MM, to verify all loadouts are now valid.
- loaded vehicles in MM and populated troop bays / BA handholds with various infantry and BA units:

image

close #4754

### DIFF
--- a/megamek/data/mechfiles/vehicles/3039u/Coolant Truck (Hover).blk
+++ b/megamek/data/mechfiles/vehicles/3039u/Coolant Truck (Hover).blk
@@ -3,7 +3,7 @@
 1
 </BlockVersion>
 
-# Write the version number just in case...
+##Write the version number just in case...
 <Version>
 MAM0
 </Version>
@@ -28,6 +28,10 @@ Coolant Truck
 2602
 </year>
 
+<originalBuildYear>
+2602
+</originalBuildYear>
+
 <type>
 IS Level 1
 </type>
@@ -37,8 +41,7 @@ Hover
 </motion_type>
 
 <transporters>
-insulatedcargobay:3.48:1:1
-insulatedcargobay:1.74:1:1
+insulatedcargobay:4.35:1:1::-1:0
 </transporters>
 
 <cruiseMP>
@@ -75,8 +78,8 @@ IS Vehicle Flamer Ammo
 </Rear Equipment>
 
 <Turret Equipment>
-Vehicle Flamer
-Vehicle Flamer
+Flamer (Vehicle)
+Flamer (Vehicle)
 </Turret Equipment>
 
 <source>
@@ -86,3 +89,8 @@ TRO: 3039
 <tonnage>
 30.0
 </tonnage>
+
+<fuelType>
+PETROCHEMICALS
+</fuelType>
+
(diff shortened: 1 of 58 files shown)


megamek/src/megamek/client/ui/swing/dialog/AbstractUnitSelectorDialog.java
Sleet01, line 1: Original dev is aware that this needs fixing, but I need MML for other work, so I'd like to revert this change until the permanent fix is ready.

megamek/src/megamek/common/util/BuildingBlock.java
Sleet01, line 1: I'd initially read these functions as finding the start and end _tags_ for a given block; instead they find the line index of the first _text_ line, and the line index after the _last_ text line, so e.g.:
```
<transporters>
troopspace:4.0
</transporters>
```
would give an index difference of `1`, and
```
<transporters>
</transporters>
```
would give a difference of `0`, rather than `1` as I had initially assumed.  Mea culpa.

megamek/data/mechfiles/vehicles/3039u/Coolant Truck (Hover).blk
```diff
@@ -37,11 +41,7 @@ Hover
 </motion_type>
 
 <transporters>
```
Sleet01, line 43: I read this as a botched correction, not as an attempt to create two separate bays.  Updating to a single bay with the new size produced entries that match the TRO.

megamek/data/mechfiles/vehicles/3058Uu/Bandit Hovercraft H.blk
```diff
@@ -41,8 +41,9 @@ Hover
 </motion_type>
 
 <transporters>
-troopspace:8.0
 BattleArmorHandles - troopers:-1
```
Sleet01, line 44: I only updated items to "OMNI" when I could confirm them via the TRO or sheet.

Sleet01:
Sorry, I just spotted the merge conflict; if the new version is working, my reversion is no longer necessary.

SJuliez (approved):

PR #4764, opened 2023-09-09
dont allow bot to deploy industrial mech in to water depth 2 or greater

### LINKED ISSUE #4550: Princess should not deploy units underwater that cannot survive there
From a 49.12 game:

End Phase
-------------------
*** Lumberjack LM4/C (Lyran Commonwealth Convoy) DESTROYED by being in water without environmental shielding! ***

IndustrialMech without sealing was deployed in depth 3 water when there was dry land available.

kuronekochomusuke:
- dont allow bot to deploy industrial mech in to water depth 2 or greater
- add isLocationDeadly() for locations that can be delployed or moved into, but being in them is deadly.  So it is a bad idea for the bot to delploy to them.
- highlight deadly deployment positions for players industrial mech in to water depth 2 or greater
- princess try to get out of water when flagged to die
- dont destroy unit if they get out of water

image
image

fixes #4550

### DIFF
--- a/megamek/src/megamek/client/bot/princess/BasicPathRanker.java
+++ b/megamek/src/megamek/client/bot/princess/BasicPathRanker.java
@@ -812,12 +812,12 @@ public class BasicPathRanker extends PathRanker implements IPathRanker {
                                                    jumpLanding, step, logMsg);
                     break;
                 case Terrains.ICE:
-                    hazardValue += calcIceHazard(movingUnit, hex, step,
+                    hazardValue += calcIceHazard(movingUnit, hex, step, movePath,
                                                  jumpLanding, logMsg);
                     break;
                 case Terrains.WATER:
                     if (!hazards.contains(Terrains.ICE)) {
-                        hazardValue += calcWaterHazard(movingUnit, hex, step,
+                        hazardValue += calcWaterHazard(movingUnit, hex, step, movePath,
                                                        logMsg);
                     }
                     break;
@@ -885,7 +885,7 @@ public class BasicPathRanker extends PathRanker implements IPathRanker {
         return 0;
     }
 
-    private double calcIceHazard(Entity movingUnit, Hex hex, MoveStep step, boolean jumpLanding,
+    private double calcIceHazard(Entity movingUnit, Hex hex, MoveStep step, MovePath movePath, boolean jumpLanding,
                                  StringBuilder logMsg) {
         logMsg.append("\n\tCalculating ice hazard:  ");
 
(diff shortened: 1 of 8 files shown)


megamek/src/megamek/client/bot/princess/BasicPathRanker.java
```diff
@@ -958,8 +958,14 @@ private double calcWaterHazard(Entity movingUnit, Hex hex, MoveStep step,
...
 
+        MoveStep lastStep = movePath.getLastStep();
         // Unsealed unit will drown.
-        if (movingUnit instanceof Mech && ((Mech) movingUnit).isIndustrial()) {
+        if (movingUnit instanceof Mech
+                && ((Mech) movingUnit).isIndustrial()
+                && !movingUnit.hasEnvironmentalSealing()
+                && (movingUnit.getEngine().getEngineType() == Engine.COMBUSTION_ENGINE)
+                && hex.depth() >= 2
+                && step.equals(lastStep)) {
```
NickAragua, line 968: So a mech can move through super deep water and not flood the engine as long as it ends the movement with the engine above water?

megamek/src/megamek/common/Entity.java
```diff
@@ -3018,6 +3018,14 @@ public boolean isBoardProhibited(int mapType) {
         return false;
     }
 
+    /**
+     * Returns true if the specified hex contains some sort of deadly
+     * terrain.
+     */
+    public boolean isLocationDeadly(Coords c) {
```
NickAragua, line 3025: I assume implementing this for other unit types is a "future development" type thing?

NickAragua (commented):
Looks good. A couple of questions.

megamek/src/megamek/client/bot/princess/BasicPathRanker.java
gsparks3: The full rules from the footnote on TW page 52 is interesting, and unusual compared to the rules for various other non-fully-sealed units winding up underwater. To stay underwater indefinitely, Industrialmechs need environmental sealing and an underwater-capable engine type. However, IMs that _don't_ meet these requirements can still enter a hex where they will be fully underwater, but they are destroyed if they _remain_ fully underwater in the end phase of the turn _after_ they initially entered the water. In other words, a non-sealed ICE IM could walk into depth 2 water and end its movement there on one turn and survive as long as it managed to climb out the next turn. They would be considered destroyed if they failed to leave the deep water after one turn.

NickAragua: Heh, so maybe then the criteria should be "if the move path both starts and ends under water", rather than just "if the move path ends under water".

kuronekochomusuke: I think going into depth 2 water is also dangerous, so it should be weighted also.    you many not be allowed to move out of it and end up dead earlier than if you were out of water.

I should also add a smaller weight for going into depth 1 water, since if you then go prone there for 2 rounds you will die.

megamek/src/megamek/common/Entity.java
kuronekochomusuke: yea, also to make it similar to isLocationProhibited()

NickAragua (approved):

PR #4765, opened 2023-09-10
safer movement for princess

### LINKED ISSUE #2698: [0.48] Princess scraps own units by overusing low/high gravity movement bonus
### Environment
MM 0.48   Java 11     Win 10/64

### Description
Even setting the Piloting caution to max, bot units will gladly use max MP in low-G conditions and fail PSR after PSR, comically destroying themselves in scattered piles around the battlefield.
Not really a "bug", but clearly an issue.

kuronekochomusuke:
- adjust fall tolerance range to 0.50 and 1.0 instead of 0.0 to 1.0 for PSRs.    
  Behavior -> Highway Menace ---- Never takes a risk. 

- when calculating fallback paths use lower of getRunMPwithoutMASC and getRunMP(MPCalculationSetting.NO_GRAVITY) 
  so that when there is low gravity fewer extended moves will be used.

- these should lower the chance for having excessive PSR movement failures when there is low gravity.

fixes #2698

### DIFF
--- a/megamek/src/megamek/client/bot/princess/Princess.java
+++ b/megamek/src/megamek/client/bot/princess/Princess.java
@@ -1239,8 +1239,8 @@ public class Princess extends BotClient {
 
             final long startTime = System.currentTimeMillis();
             getPathRanker(entity).initUnitTurn(entity, getGame());
-            final double fallTolerance =
-                    getBehaviorSettings().getFallShameIndex() / 10d;
+            // fall tolerance range between 0.50 and 1.0
+            final double fallTolerance = getBehaviorSettings().getFallShameIndex() / 20d + 0.50d;
                        
             final List<RankedPath> rankedpaths = getPathRanker(entity).rankPaths(paths,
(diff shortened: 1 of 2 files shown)


Sleet01 (approved):
Looks reasonable to me.

megamek/src/megamek/client/bot/princess/Princess.java
```diff
@@ -1239,8 +1239,8 @@ protected MovePath continueMovementFor(final Entity entity) {
 
             final long startTime = System.currentTimeMillis();
             getPathRanker(entity).initUnitTurn(entity, getGame());
-            final double fallTolerance =
-                    getBehaviorSettings().getFallShameIndex() / 10d;
+            // fall tolerance range between 0.50 and 1.0
```
NickAragua, line 1242: One of the problems with the path ranking algorithm is that it doesn't normalize the "how much damage can I possibly take from enemy fire" evaluation and the "how much damage can I take from falling" evaluation. The latter is always going to be considerably lower, because you only fall over once during a move, but get shot at by many things. This is an ok stopgap for smaller matches, but will have little effect during larger fights.

PR #4773, opened 2023-09-12
Initial fix #4737: planetary conditions overlay shown in map editor

### LINKED ISSUE #4737: [0.49.14] 'Planetary Conditions' overlay cannot be hidden in Map Editor
MM 0.49.14 (standalone and bundle)
Windows 10
Java 11

In the map editor, CTRL-P does not hide the new planetary conditions shortcut overlay (Which also appears to only render a single line, as obviously planetary conditions are not linked to the map.) That pop-up should probably be permanently hidden on the Map Editor render.

image

Sleet01:
This turned out to be the opposite issue from the Hotkey Overlay: the menu item is correctly being disabled, but the overlay itself reads its visibility setting initially from the GUIP setting, not from the menu item.

Fix is to explicitly set the overlay to invisible while instantiating the BoardEditor, after the menu and overlay are instantiated but prior to drawing the UI.

close #4737

### DIFF
--- a/megamek/src/megamek/client/ui/swing/BoardEditor.java
+++ b/megamek/src/megamek/client/ui/swing/BoardEditor.java
@@ -72,13 +72,13 @@ import static megamek.common.Terrains.*;
 // TODO: the board validation after a board load seems to be influenced by the former board...
 // TODO: copy/paste hexes
 public class BoardEditor extends JPanel
-        implements ItemListener, ListSelectionListener, ActionListener, 
+        implements ItemListener, ListSelectionListener, ActionListener,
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/client/ui/swing/BoardEditor.java
```diff
@@ -337,6 +337,7 @@ public BoardEditor(MegaMekController c) {
         controller = c;
         try {
             bv = new BoardView(game, controller, null);
```
Sleet01, line 339: Actual change is here.

SJuliez (approved):

PR #4787, opened 2023-09-16
aeros firing iGauss take +1 to hit penalty

### LINKED ISSUE #4768: iHeavy Gauss +1 to hit on fighters and small craft
### Environment
MM49.14

### Description
Improved heavy Gauss rifles should have a +1 penalty when being fired from fighters or small craft, and they currently do not

TO:AUE pg 126 "The Improved Heavy Gauss Rifle uses the same rules to mount, move with and fire as the standard Heavy Gauss Rifle (see p. 136, TW)"

TW pg 136: Fighters and Small Craft: Fighters and Small Craft apply a +1
to-hit modifier to any attack made with a heavy Gauss rifle.

NickAragua:
Fixes #4768 

Also, instead of checking data type, check ammo type.


Sleet01 (commented):
Looks good to me!

Sleet01 (approved):

PR #4804, opened 2023-09-25
Fix #4790: uncaught NPE when Sniper Turret fired on fleeing off-board target

### LINKED ISSUE #4790: NullPointerException at firing phase start
### Environment
MekHQ 0.49.14
Windows 10
Java 17.0.8

### Description
NullPointerException when starting firing phase - seems to be artillery related?

### Files
[megamek.log](https://github.com/MegaMek/mekhq/files/12517800/megamek.log)
[mekhq.log](https://github.com/MegaMek/mekhq/files/12517801/mekhq.log)

[Czechoslovak Foreign Legion30270919.cpnx.gz](https://github.com/MegaMek/mekhq/files/12517807/Czechoslovak.Foreign.Legion30270919.cpnx.gz)

Savegame: advance turn to firing phase to get NullPointerException. I had a Thumper artillery vehicle off screen, but I fled it to avoid sniper turret counterfire
[autosave.sav.gz](https://github.com/MegaMek/mekhq/files/12517810/autosave.sav.gz)

Custom mechs: Dervish and Archer are in this battle
[Czechoslovak Foreign Legion.zip](https://github.com/MegaMek/mekhq/files/12517859/Czechoslovak.Foreign.Legion.zip)

Sleet01:
Issue is that target is null by the time we get around to handling this indirect attack, because it has fled already.

Fix is to ignore this failure and just log the incident, as there doesn't seem to be any way to resolve damage from a missed counter-battery attack on an off-board target that has fled.

Tested by implementing this code in 0.49.14 and running OP's save file past the point that previously got hung.

close #4790

### DIFF
--- a/megamek/src/megamek/common/weapons/ArtilleryWeaponIndirectFireHandler.java
+++ b/megamek/src/megamek/common/weapons/ArtilleryWeaponIndirectFireHandler.java
@@ -104,11 +104,10 @@ public class ArtilleryWeaponIndirectFireHandler extends AmmoWeaponHandler {
 
         final Vector<Integer> spottersBefore = aaa.getSpotterIds();
 
+        Coords targetPos = target.getPosition();
+
         // Handle counter-battery on fleeing/fled off-board targets.
-        Coords targetPos;
-        try {
-            targetPos = target.getPosition();
-        } catch (Exception e) {
+        if (null == targetPos) {
             LogManager.getLogger().error(String.format("Artillery Target %s is missing; off-board target fled?", waa.getTargetId()));
             return false;
         }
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/weapons/ArtilleryWeaponIndirectFireHandler.java
```diff
@@ -103,7 +103,16 @@ public boolean handle(GamePhase phase, Vector<Report> vPhaseReport) {
         }
 
         final Vector<Integer> spottersBefore = aaa.getSpotterIds();
-        Coords targetPos = target.getPosition();
+
+        // Handle counter-battery on fleeing/fled off-board targets.
+        Coords targetPos;
+        try {
+            targetPos = target.getPosition();
```
NickAragua, line 110: If you know what the problem is (target == null), then you may as well check for that instead of using a try/catch block.

NickAragua (changes requested):
Right idea, but can be done without a try/catch block.

megamek/src/megamek/common/weapons/ArtilleryWeaponIndirectFireHandler.java
Sleet01: Yeah, I'd tried a more involved solution before paring it back to this - but had forgotten how much worse the try/catch performance is compared to just `if (null == X)`.

NickAragua (approved):

PR #4813, opened 2023-09-30
Aerospace landing fixes

### LINKED ISSUE #4743: Various Aerospace Issues - Aero V-Landing Does Not Apply Modifiers Correctly, Addtl issue with water.
Megamek 0.49.14, Win 10, Java

First Issue: V-LAND modifiers:

Per TW P87: "Units making vertical landings reduce by half (round down) the terrain modifiers..."

image shows full value being applied. This is consistent across rough, light trees, heavy trees, and water hexes.

Landing in rough additionally does not apply landing gear damage (P86 table foot note), which would apply on future land checks as well.

image

Water also has some fun things: this is not a clear hex as noted in the dialogue, and this modifier is not correct. (Should either be 3 or 1. 3 for a regular landing, 1 for a v-land)

Additionally: if the vehicle lands in water, it is unable to move, and falls into a stationary movement loop rather than being destroyed (as a fighter would be for entering water, see P82 which is referenced by 87 "for landing in water, see crash rules") Similarly, woods are undamaged by a fighter which lands in them, but the wording in TW is very unclear if this is what should take place.

Save file attached: to replicate: simply deploy the 0/0 fighter anywhere on board and attempt a V-Land or regular land as desired.

[Aerodyne.sav.gz](https://github.com/MegaMek/megamek/files/12505005/Aerodyne.sav.gz)

neoancient:
Fixes several issues related to aerospace units making landings:
1. Units making vertical landings only apply half the terrain modifier (TW, p. 87).
2. I added terrain modifiers for advanced terrain from TacOps.
3. When the landing path includes rough or rubble terrain, the landing unit takes damage (TW, p. 86, table).
4. A unit that lands in water of sufficient depth (2 for dropships, 1 for other units) is destroyed (TW p. 87, which refers to p. 82).
5. A dropship that lands in depth 1 water is immobile for the rest of the game (ibid).

I exempted LAMs and units with a flotation hull from water destruction.
Fixes #4743

### DIFF
--- a/megamek/i18n/megamek/client/messages.properties
+++ b/megamek/i18n/megamek/client/messages.properties
@@ -2465,6 +2465,7 @@ MovementDisplay.ConfirmUnJamRACDlg.title=Unjam?
 MovementDisplay.ConfirmPilotingRoll=You must make the following piloting\nskill check(s) for your movement:\n
 MovementDisplay.ConfirmSprint=Are you sure you want to sprint?
 MovementDisplay.ConfirmSuperchargerRoll=The movement you have selected will require a roll of {0} or higher\nto avoid Supercharger failure. Do you wish to proceed?
+MovementDisplay.ConfirmLandingGearDamage=Landing in a rough or rubble hex will damage the landing ger.\n\nAre you sure you want to land here?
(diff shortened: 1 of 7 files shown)


SJuliez (approved):
There's a typo in "landing gear" of the messages.properties. The footnote text in the table of TW p.86 suggests adding only the highest terrain modifier which would contradict the text of p.87. Seems to me to make sense doing it like the text says. So, all ok.

PR #4815, opened 2023-10-01
AmmoType: remove illegal Arty Cannon ammos (all TO ammo)

SJuliez:
TO special ammo is illegal for arty cannons, only IO Fuel-Air is allowed.


HammerGS (approved):
Tested and seems to work no issues.

PR #4820, opened 2023-10-06
Fix Immelman and Split-S maneuvers

### LINKED ISSUE #4791: [49.14] Immelmann and Split-S maneuvers not allowing selection of any facing
Per TW page 85, the Immelmann and Split-S maneuvers should allow the fighter to select any facing after executing the maneuver. As currently implemented in version 49.14, they allow for a facing change of exactly one hexside when a facing change would otherwise not be allowed, but no more.

Initial state: fighter has not moved far enough in a straight line to allow for a facing change (correct)
image

Split-S maneuver allows a single facing change (correct if that's all you wanted)
image

Even with Split-S, fighter cannot change facing more than one hexside (incorrect)
image

May be related to #4663.

Note that the automatic path calculator seems to be picky; even if only changing facing by a single hexside, clicking on a target hex after declaring a facing change maneuver will fail unless you shift-click to declare the facing change first.

The Hammerhead maneuver that forces a 180° facing change works correctly.

neoancient:
This fixes a bug I introduced in #4663. The change assumes that the maneuver type is set in the previous step, but it was only being set in the movement step and not the subsequent steps that implement the effects of the maneuver. I've changed it to include the maneuver type with all steps that are part of the maneuver, and this time I tested every maneuver type to make sure it's working.

Fixes #4791.

### DIFF
--- a/megamek/src/megamek/common/MovePath.java
+++ b/megamek/src/megamek/common/MovePath.java
@@ -567,10 +567,10 @@ public class MovePath implements Cloneable, Serializable {
                 step = new MoveStep(this, step.getType(), step.getBraceLocation());
             } else if (!step.getLaunched().isEmpty()) {
                 step = new MoveStep(this, step.getType(), step.getLaunched());
-            } else if (step.getManeuverType() != ManeuverType.MAN_NONE) {
-                step = new MoveStep(this, step.getType(), -1, -1, step.getManeuverType());
             } else if (step.isManeuver()) {
                 step = new MoveStep(this, step.getType(), step.hasNoCost(), step.isManeuver(), step.getManeuverType());
+            } else if (step.getManeuverType() != ManeuverType.MAN_NONE) {
+                step = new MoveStep(this, step.getType(), -1, -1, step.getManeuverType());
             } else if (step.hasNoCost()) {
                 step = new MoveStep(this, step.getType(), step.hasNoCost());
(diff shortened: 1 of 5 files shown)


megamek/src/megamek/common/MovePath.java
```diff
@@ -567,10 +567,10 @@ public void compile(final Game g, final Entity en, boolean clip) {
                 step = new MoveStep(this, step.getType(), step.getBraceLocation());
             } else if (!step.getLaunched().isEmpty()) {
                 step = new MoveStep(this, step.getType(), step.getLaunched());
+            } else if (step.isManeuver()) {
+                step = new MoveStep(this, step.getType(), step.hasNoCost(), step.isManeuver(), step.getManeuverType());
             } else if (step.getManeuverType() != ManeuverType.MAN_NONE) {
                 step = new MoveStep(this, step.getType(), -1, -1, step.getManeuverType());
-            } else if (step.isManeuver()) {
-                step = new MoveStep(this, step.getType(), step.hasNoCost(), step.isManeuver());
```
neoancient, line 573: This change in  order was required to preserve the noCost field of the additional steps of the maneuver. I thought the check for maneuverType could be removed along with the constructor it uses, but when I tried that all the maneuvers lost their thrust costs. I decided to leave well enough alone.

NickAragua (approved):
I had the same idea, but you beat me to it.

PR #4841, opened 2023-10-18
Fix machine gun array burst damage against conventional infantry

### LINKED ISSUE #4745: Machine gun arrays do not operate correctly against conventional infantry
### Environment
MegaMek Version: MekHQ v0.49.14 Development Snapshot
Operating System: Mac
Java Version: 20.0.1

### Description
Machine Gun Arrays[^1] currently appear to work against conventional infantry in the same manner as other targets. They roll on the cluster hits table, and they do not apply any additional damage.

image

This is incorrect:
> After a successful MGA attack against conventional infantry, the controlling player does not roll on the Cluster Hits Table. Instead, all machine guns grouped in the array hit the target; the controlling player rolls the appropriate number of damage dice for the type of MGA (light, standard and heavy) times the number of machine guns in the array. For example, the player rolls 2D6 x 4 for an MGA 4 to determine how much damage the infantry unit takes.

In the screenshot above, the MGA 3 should have rolled once to hit then dealt 6D6 damage.

[^1]: Page 137, Total Warfare, ninth printing

neoancient:
Against most units a machine gun array rolls on the cluster hits table to determine the number of weapons that hit. Against conventional infantry the number of burst damage dice is supposed to be multiplied by the number of linked weapons.

I've also modified the report to show the amount of damage done by any burst weapon.
1D6/2 is not an issue because it is only used by BA weapons, and cannot be part of a MGA.

Fixes #4745.


neoancient:
Closed because of cruft in the branch.

Issue #3594, opened 2022-04-24
Flashman 10E


WildKadabra:
All ER Medium Lasers are forward facing in the current version, one of the ER Medium Lasers in the left torso should be rear facing.

ilClan Rec. Guide 24

HammerGS:
Was fixed as part of https://github.com/MegaMek/megamek/commit/7904b26d10e19010212e4289edaac711c797a4de

PR #4853, opened 2023-10-25
Fixing fuel tons per burn day display for Jumpships

nutritiousemployee:
This is a fix for MekHQ issue #3594

While the actual fuel usage and cost calculations were correct, the display of TBD for Jumpships got reduced to 10% of the actual value bringing in a discrepancy.
Removing that part of the code solves it - after some testing it appears to also fix the "Burn Days (Station-keeping)" value in MegaMekLab (since the same chunk of code was used there).
Example: Monolith JS has 39.52 TBD, and carries 68 tons of fuel, so it's station keeping burn days should be 1.72 and not 17.21 as it displays now.

Edited, since i can't seem to figure out how to undo my commits: this also now fixes MekHQ #3773

In GameManager the unit being loaded will go through a try-catch. If there's an issue with the loading, a message will be sent into the server chat.

### DIFF
--- a/megamek/src/megamek/server/GameManager.java
+++ b/megamek/src/megamek/server/GameManager.java
@@ -3682,7 +3682,7 @@ public class GameManager implements IGameManager {
                 && !getGame().getPhase().isDeployment();
         try {
             loader.load(unit, checkElevation, bayNumber);
-        }catch(IllegalArgumentException e){
+        } catch (IllegalArgumentException e) {
             LogManager.getLogger().info(e.getMessage());
             sendServerChat(e.getMessage());
             return;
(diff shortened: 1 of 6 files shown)


NickAragua (approved):

megamek/src/megamek/server/GameManager.java
```diff
@@ -3680,8 +3680,13 @@ private void loadUnit(Entity loader, Entity unit, int bayNumber) {
         // Load the unit. Do not check for elevation during deployment
         boolean checkElevation = !getGame().getPhase().isLounge()
                 && !getGame().getPhase().isDeployment();
-        loader.load(unit, checkElevation, bayNumber);
-
+        try {
+            loader.load(unit, checkElevation, bayNumber);
+        }catch(IllegalArgumentException e){
```
NickAragua, line 3685: Hate to be that guy, but can we put spaces between the curly braces and the catch statement?

NickAragua (changes requested):

megamek/src/megamek/server/GameManager.java
nutritiousemployee: done!

PR #4862, opened 2023-10-29
check for natural aptitude piloting when doing physical attacks rolls

### LINKED ISSUE #4541: Natural Aptitude: Piloting not applying to melee attack rolls
### Environment
What version of MegaMek does your issue apply to?  using megamek 0.49.12 inside mekhq 0.49.12, but it should still apply to the current version (0.49.13) as well as many versions backwards.
What operating system are you using?  windows 10
What java version are you using?  19.0.1

### Description
Summary: melee attack preview odds when pilot has Natural Aptitude, Piloting calculate with a 3d6-L roll, but the actual roll is using a basic 2d6.

Rule: Special Pilot Ability `Natural Aptitude ` from AToW p121.  This is currently implemented in megamek as `Natural Aptitude, Gunnery` and `Natural Aptitude, Piloting` (which I will call `NA:P`).  The 3d6-L rolls would apply to all piloting checks, and melee attacks would fall into the category of piloting check.  However, based on the below examples, it seems to be using just a basic 2d6 for melee attacks.

When selecting a melee attack with a unit that has NA:P, the preview odds match the math for a 3d6-L, but it appears that the actual roll being used for the check is only a basic 2d6.  Either the previous odds are incorrectly applying the 3d6-L to the preview odds, or the 3d6-L is incorrectly not being applied to the actual final check.

Screenshot someone else posted of melee attack rolls on a unit with NA:P, with the option enabled to display the individual dice.  It appears to be rolling only 2d6 instead of 3d6-L:
https://cdn.discordapp.com/attachments/458706493659807801/1119777789416046672/image.png

Screenshot of a regular piloting check where the 3d6-L is working: https://cdn.discordapp.com/attachments/458706493659807801/1119777321520484372/image.png

Sample melee attack rolls from another test game where all pilots had NA:P: 7,8,9, 8,8,7, 8,5,4,5, 7,8,10, 10,3,9, 5,7,5,8, 4,6,3,8.  This is a small sample (24 rolls), but is far more likely to be 2d6 than 3d6-L considering the distribution.

### Files
Here is a save where all pilots have NA:P and some are about to perform melee attacks (Player names: Ishara's Grenadiers, Princess): 
[orion nap phys.gz](https://github.com/MegaMek/megamek/files/11780499/orion.nap.phys.gz)

kuronekochomusuke:
- check for natural aptitude piloting when doing physical attacks rolls

fixes #4541

### DIFF
--- a/megamek/src/megamek/server/GameManager.java
+++ b/megamek/src/megamek/server/GameManager.java
@@ -32047,7 +32047,11 @@ public class GameManager implements IGameManager {
         int damage = 0;
         PhysicalResult pr = new PhysicalResult();
         ToHitData toHit = new ToHitData();
-        pr.roll = Compute.d6(2);
+        if (ae.getCrew() != null) {
+            pr.roll = ae.getCrew().rollPilotingSkill();
+        } else {
+            pr.roll = Compute.d6(2);
+        }
         pr.aaa = aaa;
         if (aaa instanceof BrushOffAttackAction) {
             BrushOffAttackAction baa = (BrushOffAttackAction) aaa;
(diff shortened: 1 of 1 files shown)


NickAragua (commented):
I'm seeing a couple of conditional blocks in this large if/else area where the attack isn't strictly a physical attack where NA:P would apply. Specifically, I'm looking at the "Telemissile" and "Lay Explosives". Let's double check those.

NickAragua (approved):

PR #4867, opened 2023-11-02
Fixes for flame based infantry bugs

### LINKED ISSUE #4213: RFE: Add CI indirect fire for mortars
This was asked about on discord so I am bringing it here in case someone other than myself has any interest in coding it.

If anyone who has not contributed code to MegaMek, and has some experience fooling around with the MM codebase this could make for a good first PR. Don't be shy, just try. If nothing else it will give you a greater insight into the code and make your next attempt even easier.

### LINKED ISSUE #4793: Flame Based Infantry Bugs
### Environment
MM49.14

### Description
A:  Flamer (Man portable) (Used by most Flamer Infantry platoons or points), despite having a mode function to toggle between damage or heat, only does damage.  As does Mortar (heavy) Inferno.  Which implies there is a bigger problem.  Wonder if its secondary weapon based.  I know that a "Rifle (Federated-Barrett M42B)(Inferno)" did apply heat instead of damage when it was the weapon used to determine range.

B: If you turn on the game  option "Flamers per Battlemech manual" the option to toggle between damage and heat modes on infantry flame based weapons is removed, leaving only damage.

C: Infantry Inferno LRM and Inferno Mortar need additional mode options of heat, indirect, and indirect heat. (If CI mortars even work with IF yet.  Don't think they do.)

D: If an infantry unit's primary or secondary weapon have the F Flame-based special ability, the unit receives the F Flame-based ability.

Currently megamek seems to be deciding to only use the weapon that supplies the range brackets  Which is true for the N Non penetrating weapon special ability, But Flame based weapons have this foot note: "*Platoon special feature applies if one or more weapons assigned"  (TM pg 148, or pg 9 of the TM errata document.)

neoancient:
This fixes a number of related bugs with flame-based infantry weapons.
The most extensive change is the introduction of a subclass of Mounted that tracks both primary and secondary weapons and the movement of mode queries into Mounted. This was done because the flame-based effects are in play regardless of which weapon has the F attribute and the Mounted for the attack only had access to the weapon that is used to calculate the range.

For flame-based weapons that are capable of indirect fire, I added a new Indirect/Heat mode (and added indirect fire to the non-inferno mortar while I was at it).

InfantryHeatWeaponHandler needed significant revisions. As it is now, infantry would do heat but no damage to mechs if the BMM flamer option is checked (which should do heat + damage) and damage but no heat in all other situations. My changes have it doing heat + damage to all heat-tracking units (mechs, ASF, small craft) when the BMM option is checked, heat only to those units when not checked, and damage to all other units. If the weapon is in non-heat mode the InfantryWeaponHandler is used instead to do damage only.

Update: the heat+damage mode is now an unofficial option separate from BMM flamers.

Fixes #4793
Fixes #4213


megamek/src/megamek/common/Entity.java
```diff
@@ -11887,9 +11887,7 @@ public void setGameOptions() {
         }
 
         for (Mounted mounted : getWeaponList()) {
-            if (mounted.getType() instanceof Weapon) {
-                ((Weapon) mounted.getType()).adaptToGameOptions(game.getOptions());
-            }
+            mounted.adaptToGameOptions(game.getOptions());
```
NickAragua, line 11890: I assume it's not possible for a mounted returned here to be anything other than a weapon?

megamek/src/megamek/common/Mounted.java
```diff
@@ -254,13 +255,35 @@ public EquipmentType getType() {
...
+    public int getModesCount() {
+        return type.getModesCount();
+    }
+
+    protected EquipmentMode getMode(int mode) {
+        return type.getMode(mode);
+    }
+
+    public boolean hasModes() {
+        return type.hasModes();
```
NickAragua, line 267: I wonder if these three "type" should a) be turned into getType() for consistency?

And also I wonder if for safety we want to check for null in here? Maybe not because it's used in a lot of places and hasn't had a problem before, but just thinking out loud.

megamek/src/megamek/common/actions/WeaponAttackAction.java
```diff
@@ -366,7 +366,8 @@ private static ToHitData toHitCalc(Game game, int attackerId, Targetable target,
                         || atype.countsAsFlak()
         );
 
-        boolean isIndirect = (wtype.hasModes() && weapon.curMode().equals(Weapon.MODE_MISSILE_INDIRECT));
+        boolean isIndirect = (weapon.hasModes() && (weapon.curMode().equals(Weapon.MODE_MISSILE_INDIRECT)
```
NickAragua, line 369: Do we want to encapsulate this into a convenience function? I'm seeing it in at least two places.

NickAragua (commented):
Looks good, some minor thoughts.

megamek/src/megamek/common/Entity.java
neoancient: After the change it's calling adaptToGameOptions on mounted instead of mounted.getType(). Mounted now handles the weapon check, while the InfantryWeaponMounted subclass calls adaptToGameOptions on both weapons.

megamek/src/megamek/common/Mounted.java
neoancient: Agreed. type is doesn't get serialized and getType() will reinitialize it from the equipment lookup name, so it should be accessed through getType() throughout the class.

megamek/src/megamek/common/actions/WeaponAttackAction.java
neoancient: I thought about doing that after I had written it, but decided to leave it as is instead of going back and finding those occurrences. Since you've commented on it as well I've gone back and made the change.

neoancient:
Converting to draft pending implementing ruling about BMM heat option and non-Mech units.

HammerGS:
Curious if Princess will be able to use IDF with infantry/BA once this is done?

PR #4872, opened 2023-11-03
Add fluff info to MechCacheCSVTool output

SJuliez:
As the title says. Requested by @HammerGS 

Also allows to easily change the column delimiter. It is no longer strictly an output of the MechSummaryCache as the fluff info needs to be loaded from each unit's file.


HammerGS:
Tested and does exactly what I was looking for! Thank you!.

PR #4873, opened 2023-11-04
add report roll tooltip 

kuronekochomusuke:
- add report roll tooltip for piloting and gunnery skill checks
- add report roll tooltip to more of the reported rolls

image

image

image

image

image


HammerGS:
Love this.

HoneySkull (commented):
Design question:  Would it be beneficial to overload the `Report::add()` function to take a `Roll` parameter so that it can automatically `addDataWithToolTip(roll.getIntValue(), roll.getReport())` - in such a way that by default tooltips get generated?   It goes against Demitre's law a little because Report would need to know about Roll - but is a LOT of reporting code intermingled within the game logic -and tool tips are part of reporting.

kuronekochomusuke:
I think it would make things a little cleaner in game logic parts of the code.  just need to see what impacts returning roll in some of the edge cases, but in general it seems an easy change.

SJuliez (approved):
Looks good to me

PR #4883, opened 2023-11-10
Beast mounted infantry

neoancient:
Implements beast-mounted infantry rules from TO: AU&E, pp. 107-8, with construction rules on 106.The table of creatures is presented as sample types, so I've implemented it to allow custom stats, but the creatures in the table are included as standard presets. Some adjustments were needed to submarine movement to allow creature to move onto land when the ability is indicated. I also made a change to deployment that has such units deploy at a depth equal to the height, so that when at the surface they extend one level above the surface. This has not been an issue before because all submarine units are only one level high, but very large and monstrous mounts are taller, as are the yet-unimplemented large naval support vehicles. There is also now a tracker for how long an entity has been entirely submerged, used to track air reserves. I've added a line to the unit tooltip that shows the number of rounds of air remaining when relevant.


megamek/src/megamek/client/ui/swing/DeploymentDisplay.java
```diff
@@ -530,10 +530,12 @@ public void hexMoused(BoardViewEvent b) {
             } else if (!isAero && !isWiGE) {
                 // hovers and naval units go on the surface
                 if ((ce().getMovementMode() == EntityMovementMode.NAVAL)
-                        || (ce().getMovementMode() == EntityMovementMode.SUBMARINE)
                         || (ce().getMovementMode() == EntityMovementMode.HYDROFOIL)
                         || (ce().getMovementMode() == EntityMovementMode.HOVER)) {
                     ce().setElevation(0);
+                } else if (ce().getMovementMode().isSubmarine()) {
+                    // submarines have one level above the surface
```
NickAragua, line 537: below?

HammerGS:
I'm starting to get sprites in for the units I made. Once I get them I'll look at committing to the branch before merging if that works.

megamek/src/megamek/common/verifier/TestInfantry.java
```diff
@@ -197,52 +208,64 @@ public static int maxSecondaryWeapons(Infantry inf) {
...
+                    return 7;
+                case INF_UMU:
+                    return alt ? 6 : 10;
+                case VTOL:
+                    return alt ? 2 : 4;
+                default:
+                    return 10;
+            }
+        } else if (mount.getSize().troopsPerCreature == 1) {
+            return 10; // use foot infantry limit
```
NickAragua, line 232: Can we define some of these magic numbers? I'm seeing at least three 10s here.

megamek/src/megamek/server/GameManager.java
```diff
@@ -11630,11 +11643,24 @@ public Vector<Report> doSetLocationsExposure(Entity entity, Hex hex,
...
                     entity.setLocationStatus(loop, ILocationExposureStatus.WET);
                     vPhaseReport.addAll(breachCheck(entity, loop, hex));
                 }
+            } else {
+                int status = ILocationExposureStatus.WET;
+                if (entity.isConventionalInfantry() && (entity.relHeight() >= 0)) {
+                    status = game.getPlanetaryConditions().isVacuum() ?
+                            ILocationExposureStatus.VACUUM : ILocationExposureStatus.NORMAL;
+                }
+                for (int loop = 0; loop < entity.locations(); loop++) {
```
NickAragua, line 11658: Is this something we actually want applied to non-infantry units?

NickAragua (commented):
A couple minor questions/possible cleanup issues.

megamek/src/megamek/client/ui/swing/DeploymentDisplay.java
neoancient: No, above. Those that are multiple levels in height have one level above the surface (when deployed on top of the water) and the rest below.

megamek/src/megamek/common/verifier/TestInfantry.java
neoancient: Those are the maximum squad size defined by the rules (TM, p. 145, TO: AU&E, p. 136, 153). Each motive type has its own entry in the table. The fact that some of them are the same doesn't make them a universal value.

megamek/src/megamek/server/GameManager.java
neoancient: I was keeping the current behavior for other units, but simplifying it would fix superheavy vehicles and dropships that landing in level 1 water by putting their weapons above the waterline. A dropship that lands in depth 2+ water is destroyed, so there's no need to figure out which arcs are above the waterline in deeper water.

NickAragua (approved):

PR #4898, opened 2023-11-18
Fix test for aerospace units that don't require fuel

### LINKED ISSUE #4821: Fixed-Wing Support Vehicles with propellers and fusion engines should not require fuel
### Environment
MegaMek Version: MekHQ v0.49.14 Development Snapshot
Operating System: Mac
Java Version: 20.0.1

### Description
Unlike other aerospace units, Fixed-Wing Support Vehicles with propellers and certain types of engine have functionally infinite fuel, and don't explicitly allocate tonnage for that.

The specific rule may be found on page 129 of the sixth printing of _TechManual_:
> Fixed-Wing Support Vehicles that use the Prop-Driven chassis modification and are powered by Solar, Fission, or Fusion engine types do not require Fuel Points at all.

While this is implemented in MegaMekLab, MegaMek doesn't seem to be aware of it. As a result, all such units are treated as "out of fuel" and are effectively unusable as a result.

neoancient:
Currently the check for an aerospace unit that has run out of fuel exempts those with solar engines. The rules in TM, p. 129 state that this only applies to fixed wing and airship support vehicles with the prop chassis mod, and it also includes fission and fusion engines.

Fixes #4821

### DIFF
--- a/megamek/src/megamek/client/ui/swing/MovementDisplay.java
+++ b/megamek/src/megamek/client/ui/swing/MovementDisplay.java
@@ -4110,7 +4110,7 @@ public class MovementDisplay extends ActionPhaseDisplay {
         }
 
         IAero aero = (IAero) ce;
-        if ((aero.getCurrentFuel() < 1) && !(ce.hasEngine() && ce.getEngine().isSolar())) {
+        if ((aero.getCurrentFuel() < 1) && aero.requiresFuel()) {
             disableButtons();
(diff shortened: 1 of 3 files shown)


NickAragua (approved):

Saklad5:
It's actually illegal to have electric generators, including solar, without the Prop-Driven modification. You can't use electricity as reaction mass, after all.

PR #4903, opened 2023-11-19
Move Unit Roles to Unit Files

SJuliez:
As the title says, this moves the unit roles into the mtf/blk files and removes the way they are currently handled as a single file. On the MML side there is an accompanying editor update.

I notice this is difficult to review because of all the unit files. If necessary I can try uploading a draft that only has the code changes.


HammerGS (approved):
Tested and seems to work,

neoancient (approved):
I looked at the individual commits that had the code changes.

HammerGS:
I spotted check a bunch of units that they seemed right.

HammerGS:
Going to merge this, as I'm working on some infantry stuff and don't want to create too many conflicts.

PR #4923, opened 2023-11-25
Allow jdialog to be a parent rather than just jframe

gcoopercos:
Motivation:
In MekHq, the campaign creation user interface wasn't showing up.   After tracking this down, the modal campaign dialog was coming up behind the original dialog that spawned it.

Without rearranging things too much, a straight-forward fix appeared to be to make sure that the modal dialog had the owner of the JDialog that created it, not the owning JFrame.

I am guessing that this works ok on some platforms so it's not an issue for some. I'm also guessing that this is the case because which JDialog is rendered in front is arbitrary if they have the same JFrame. Regardless, this feels like a fairly benign change.

Change Description:
The 'megamek' changes are just to allow a JDialog to be the parent of another JDialog in the abstract classes... Something Swing already supports. The real functional change will be in mekhq.


NickAragua (approved):
This seems reasonable, especially since it's purely additive.

PR #4924, opened 2023-11-25
weapon panel factor in combat computer with displaying heat buildup

kuronekochomusuke:
- weapon panel - factor in combat computer quirk when displaying heat buildup


NickAragua (approved):

SuperStucco:
Should probably check to see if quirks are enabled in Options as well.  They can be enabled/disabled with the check box in Options => Advanced Rules => StratOps Quirks.

kuronekochomusuke:
hasQuirk() checks that

PR #4939, opened 2023-12-02
#4876 Fix Unit Display General tab rendered blank when switching units.

### LINKED ISSUE #4444: issue on movement phase, unit display and done button are not displaying correcrtly
Starting MegaMek v0.49.13-SNAPSHOT
	Build Date: N/A
	Today: 2023-05-20
	Origin Project: MegaMek
	Java Vendor: Eclipse Adoptium
	Java Version: 17.0.6
	System Locale: en_US

I have been getting this issue on the movement phase, the unit display and done button are not displaying correctly.

if i hide the unit display, it shows correctly once i show the unit display again.   also works fine if I reload the autosave. 

this seems to be something with html text in a label, it happening to the summary tab, sensor info on extra tab, and done buttons.  all these use html text

image
image

[megamek.log](https://github.com/MegaMek/megamek/files/11522439/megamek.log)

### LINKED ISSUE #4876: [0.49.15] Unit display general tab rendering blank when new unit selected.
0.49.15 standalone MM
Windows 10
Java11

Each time it is my turn to move (or when I switch which mech I wish to move), the 'general' tab of the unit display becomes completely blank (see screenshot). This is the case whether the display is docked or floating.

Using 'switch location' will force a refresh, as does 'switch view', but selecting a different mech to use will cause the error again.

image

Savegame: user 'SimonLandmine'
[Woodland_2023-11-06_03-49.sav.gz](https://github.com/MegaMek/megamek/files/13267208/Woodland_2023-11-06_03-49.sav.gz)

HoneySkull:
This is a proposed fix for #4876.  This code change moves the unit display select entity logic before BoardView setup and rendering methods.  I cannot 100% explain why this order of operations works, but it seems to be race condition in the Java Swing component repainting that occurs as methods like `BoardView.setHighlightColor()` trigger a `repaint()`.

It doesn't appear there are any side-effects to moving this call up in the order of operations as it appears all the other tabs update as expected as new units are selected in the movement phase.

Fixes #4876
Fixed #4444

### DIFF
--- a/megamek/src/megamek/client/ui/swing/MovementDisplay.java
+++ b/megamek/src/megamek/client/ui/swing/MovementDisplay.java
@@ -758,7 +758,8 @@ public class MovementDisplay extends ActionPhaseDisplay {
         clientgui.getBoardView().setHighlightColor(GUIP.getMoveDefaultColor());
         clear();
 
-        updateButtons();
+        updateButtonsLater();
+
         clientgui.getBoardView().highlight(ce.getPosition());
         clientgui.getBoardView().select(null);
         clientgui.getBoardView().cursor(null);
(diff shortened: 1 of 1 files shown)


HoneySkull:
> this fixes the unit summary for me, but I am still getting blank move and skip movement buttons when zoomed out to the max
> 
> image

I've been able to reproduce the blank move menu only by setting the skin to "plainSkin.xml" in the client settings as described in the earlier issue #4444.    This fix doesn't fix both the issues, but I have a hunch they both have a similar root cause.

kuronekochomusuke:
it now fixes both issues for me

PR #4948, opened 2023-12-10
Quirks in Unit Files

SJuliez:
This is the code part of moving quirks to unit files. The unit files will be in a separate PR to make this reviewable.

After some serious problems with migrating the quirks and discussion with Hammer and further thinking I dropped my original plan to have the ability to inherit quirks in unit files by naming another unit as a "base unit". Quirks are now simply per model, much like unit roles. 

(One problem of the base unit thing was that it needed to give a file path rather than just a unit name. This would have made renaming files or directories and simply moving units around a source for bugs. Using this in custom units would not have been straightforward.)

TODO:
- [ ] See where those unit changes come from


Thom293:
*Excited Noises*

megamek/src/megamek/common/Entity.java
```diff
@@ -15473,4 +15460,28 @@ public void setUnitRole(UnitRole role) {
...
+     *
+     * @param mounted the equipment to look for
+     * @return the (first) slot number that holds the mounted or -1 if none can be found
+     */
+    public int slotNumber(Mounted mounted) {
+        int location = mounted.getLocation();
+        if (location == Entity.LOC_NONE) {
+            return -1;
+        }
+        for (int slot = 0; slot < getNumberOfCriticals(location); slot++) {
```
NickAragua, line 15476: Man, the number of times I see this "loop through all critical slots until we find what we need" code just makes me sad. Wonder if we can include a reverse slot lookup type of thing mapping a Mounted to the list of occupied slots in a particular location. Definitely a project for another PR, though.

NickAragua (approved):

PR #4950, opened 2023-12-11
Quirks in Unit Files: unit files

SJuliez:
This adds quirks to unit files according to the former quirks xml file.


HammerGS (approved):
Data wise looks good to me

PR #4960, opened 2023-12-16
return impossible instead of raising an error

### LINKED ISSUE #4949: java.lang.IllegalArgumentException: Attacker or target not valid at megamek.common.actions.ThrashAttackAction.toHit
### Environment
0.49.15
Windows 10
Java Vendor: Eclipse Adoptium
Java Version: 11.0.13

### Description
image
I get the above error after clicking the 'Done' button at the end of the Round 15 Firing Phase. 

Exact reproduction steps with the below campaign and MegaMek save file:
1. Load the below campaign save file.
2. Click Load Saved Game from the Briefing Room for the TankBase - Hostile - Prevent Evac scenario.
3. Select OK, load the BA_01.sav file below.
4. Game > Edit Bots > Replace all with Princess > OK
5. Have Booster fire autorifle at pilot north of her.
6. Have Voltage fire autorifle at crew north of him.
7. Have Jester fire all weapons at Cataphract closest to him.
8. Have Calamity fire all weapons at Cataphract closest to her (same one as Jester).
9. Select Done.
10. After Fire Phase resolution log is produced, click done.
11. Error.

### Files
Provide the .cpnx campaign file that this issue occurs in.
[Poznan Heavy Lancers02.cpnx.gz](https://github.com/MegaMek/mekhq/files/13636685/Poznan.Heavy.Lancers02.cpnx.gz)

Also here is the MegaMek in-mission save file it occurs in.
[BA_01.sav.gz](https://github.com/MegaMek/mekhq/files/13636780/BA_01.sav.gz)

Provide any custom units that you are using in this campaign so that it will load on the developers machine.
[Black Knight BL-7-KNT Hussar.zip](https://github.com/MegaMek/mekhq/files/13636708/Black.Knight.BL-7-KNT.Hussar.zip)

Provide any other files that you have customized in this campaign like the factions.xml for example.
I'm using the customized "Heavy" scenarios from Discord here: https://discord.com/channels/458705327911731231/1125131384835555491/1125151393515327498
[scenariotemplates.zip](https://github.com/MegaMek/mekhq/files/13636750/scenariotemplates.zip)

And provide the megamek.log and mekhq.log files for any NPEs or errors that MekHQ might be reporting.
[mekhq.log](https://github.com/MegaMek/mekhq/files/13636773/mekhq.log)
[megamek.log](https://github.com/MegaMek/mekhq/files/13636774/megamek.log)

kuronekochomusuke:
- return impossible instead of raising an error

fixes #4949

### DIFF
--- a/megamek/src/megamek/common/actions/BAVibroClawAttackAction.java
+++ b/megamek/src/megamek/common/actions/BAVibroClawAttackAction.java
@@ -49,7 +49,7 @@ public class BAVibroClawAttackAction extends AbstractAttackAction {
         Entity te = null;
         // arguments legal?
         if ((ae == null) || (target == null)) {
-            throw new IllegalArgumentException("Attacker or target not valid");
+            return new ToHitData(TargetRoll.IMPOSSIBLE, "Attacker or target not valid");
         }
 
         if (target.getTargetType() == Targetable.TYPE_ENTITY) {
--- a/megamek/src/megamek/common/actions/BAVibroClawAttackAction.java
+++ b/megamek/src/megamek/common/actions/BAVibroClawAttackAction.java
@@ -15,6 +15,7 @@ package megamek.common.actions;
 
 import megamek.common.*;
 import megamek.common.options.OptionsConstants;
+import org.apache.logging.log4j.LogManager;
 
 /**
  * A BattleArmor uses its vibroclaws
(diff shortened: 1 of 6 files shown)


HoneySkull:
Has potential for masking the underlying issue.

kuronekochomusuke:
To me it is better to set an appropriate return value  here, vs throwing an unhandled exception that crashes the game.

HoneySkull:
I agree it's better to not crash the game with the unhandled exception.  I think, in addition, a quiet write to the error log indicating a null entity or target object was observed needs to be there because some other broken condition exists that causes the illegal parameters when these parameters should never be null. 

For example, in the future if someone reports that the Thrash Attack isn't working correctly, or the physical attack phase is skipped when the Thrash Attack should have been allowed - it will have returned an "Impossible" when it wasn't expected - if we also write a quick error to the log saying we hit a null parameter condition (in addition to returning the 'impossible' result), we will have some clue of what went wrong.

HoneySkull (approved):

SJuliez (approved):

PR #4961, opened 2023-12-16
User file directory

SJuliez:
We have sort of a fixed user file directory within MM, but it seemed to me that having a freely settable directory outside of any MM install (wherever the user wants to set it, like D:/myMMStuff) where units, camos, fonts etc. can be kept and never need to be moved for a new MM install or between MHQ/MML/MM would be a better solution. This is what this PR adds. The directory can be set in the Client settings and is part of the ClientPreferences. There will be accompanying PRs for MML where the user dir can be set in the Options and MHQ where it can be set in the MekHQ options->misc options.

As things are now, fonts and units can be placed anywhere within the user dir as they can be identified by their extension. Images must be placed in subdirectories as they are in MM itself, i.e. in <userdir>/data/images/camo, data/images/portraits, data/images/fluff/<unittype>.

Typically, the user dir setting will only take effect after a restart.

### DIFF
--- a/megamek/docs/UserDirHelp.html
+++ b/megamek/docs/UserDirHelp.html
@@ -2,6 +2,8 @@
 
 <P>Use this directory for resources you want to share between different installs or versions of MegaMek, MegaMekLab and MekHQ. The files listed below will also be loaded from this directory (in addition to what is loaded from MegaMek's own data). The directory should be an absolute path such as D:/MyBTStuff (in other words, not relative to your MegaMek directory).</P>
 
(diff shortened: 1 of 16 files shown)


Thom293:
Roles and Quirks and now this?

Superhero.

sldfgunslinger2766:
This, by far, might be one of the handiest things I've seen added to the suite in a long, long time 💪

HammerGS:
Feedback from Discord:

Will files put into the custom directory mentioned in ⁠announcements take priority over files of the same name in the base directory?

If so, I think a cool idea would be if you have the suite's whole image directory be supported, that way we could replace not only unit images and camos, but even things like map and Stratcon hex graphics if we really wanted to.

neoancient (approved):

PR #4976, opened 2023-12-21
ProtoMek max armor values

SJuliez:
Corrects ProtoMek maximum armor values. The original IO errata v1.21 note the error in the expanded protomek armor table, but the newer IO:AE errata seem to have forgotten. Looking at the table it still looks very much like an error. This PR corrects it.

@HammerGS could you give this a look; I feel the new IO:AE errata v3.01 should list this (taken from the old IO v1.21 errata):

image

Fixes MegaMek/megameklab#1345
Fixes MegaMek/megameklab#1154


HammerGS:
@SJuliez All indications are that errata stands. It's missed when the books got split

neoancient (approved):

PR #4981, opened 2023-12-22
add unit readout button that displays the unit summary, tro, and as card

### LINKED ISSUE #4242: RFE: Show TRO Tab in Game in the Unit Information.
Not sure how hard this would be, but the ability to view the TRO tab from the Unit Selector would be a useful addition. It would serve as a "Warbook" and let players see and read fluff and TRO presented data in game. 

This would help new players learn about other units information versus strictly seeing a Record sheet type information.

kuronekochomusuke:
- add unit readout button that displays the unit summary, tro, and as card
- add board view menu item for view readout

image
image

fixes #4242


HammerGS:
This is so cool to see!

SJuliez (approved):

PR #4990, opened 2023-12-31
FixedWingSupport: automatically set SI when walk MPs are set

SJuliez:
In FixedWingS, the SI value is always equal to the walk MP (safe thrust). This adds an override that sets SI automatically in FWS. This affects unit loading everywhere as well as unit building in MML.

Fixes MegaMek/megameklab#1320


HammerGS (approved):
Quick test and it works for me.

PR #4992, opened 2024-01-01
Don't display engine rating for SV engines

SJuliez:
Removes the engine rating (but not the engine type) from the mech summary and the record sheet for support vehicles as they don't use engine ratings.

old:
image

new:
image

### DIFF
--- a/megamek/src/megamek/common/Engine.java
+++ b/megamek/src/megamek/common/Engine.java
@@ -450,7 +450,11 @@ public class Engine implements Serializable, ITechnology {
      */
     public String getShortEngineName() {
         if (engineType < TYPE_KEYS.length) {
-            return String.format("%d%s", engineRating, Messages.getString("Engine." + TYPE_KEYS[engineType]));
+            if (hasFlag(SUPPORT_VEE_ENGINE)) {
+                return String.format("%s", Messages.getString("Engine." + TYPE_KEYS[engineType])).trim();
+            } else {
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/Engine.java
```diff
@@ -450,7 +450,11 @@ public int integralHeatSinkCapacity(boolean compact) {
      */
     public String getShortEngineName() {
         if (engineType < TYPE_KEYS.length) {
-            return String.format("%d%s", engineRating, Messages.getString("Engine." + TYPE_KEYS[engineType]));
+            if (hasFlag(SUPPORT_VEE_ENGINE)) {
+                return String.format("%s", Messages.getString("Engine." + TYPE_KEYS[engineType])).trim();
```
neoancient, line 454: No need to call String.format.
```suggestion
                return Messages.getString("Engine." + TYPE_KEYS[engineType]).trim();
```

neoancient (changes requested):

HammerGS:
Errata issued by CGL that could provide a better option. 

https://bg.battletech.com/forums/index.php/topic,27319.msg1984570.html#msg1984570

neoancient (approved):

neoancient:
We could change it to show the effective rating for the stackpole rule, but that's easily calculated from the information on the sheet and it only applies to fusion engines. I think it would be fine to leave it off.

HammerGS:
It was originally a CGL request to add the number. They created the errata specifically to show the rating for SV's. So we should go ahead and show the new errata value and engine type for them.

PR #5003, opened 2024-01-02
Implement 4397 internal bomb bay redux

### LINKED ISSUE #4397: RFE: Implement Internal Bomb Bay quirk
The Internal Bomb Bay quirk, described on page 227 of *Campaign Operations*, allows units, including fighters, to load additional bombs inside their cargo bays. However, dropping those bombs incurs a risk of explosion from ground fire on a 2d6 roll of 10+.

---

As a stopgap, the most simple implementation of this could simply increase the bomb capacity of each fighter by the number of tons of cargo they have (rounded down).

A more comprehensive implementation would require specifying which bombs are loaded where, and which ones are being dropped during combat, to determine whether the explosion risk is present that turn.

Sleet01:
(Note: this is a re-submission of PR 4974, due to merge issues I caused by fixing things out of order)

This has some pretty big changes:

### New gameplay features:

1. Aerospace fighters, Conventional Fighters, Fixed Wing Support aircraft, Small Craft, and Dropships with the "Internal Bomb Bay" quirk can now mount bomb-type munitions internally.  The Bombs section of the configuration dialog will indicate Internal and/or External bomb options as appropriate.  When bombing, the user will be prompted separately for Internal and External bomb stores to utilize (this is necessary for accounting reasons, and to keep the UI simple).
2. Units that utilize their internal stores to make attacks (or to TAG, if a TAG pod is mounted) chance ground fire touching off any other stores in their internal bays: on a 10+ roll at the end of a round in which a unit has taken damage from ground fire, all remaining internal bombs explode, dealing their full damage directly to SI!

### Changes/Additions

1. Re-organized Aero class hierarchy slightly:
```
|Entity
|- Aero
    |- TeleMissile
    |- AeroSpaceFighter                 // Inserted between Aero and subclasses
    |   |- ConventionalFighter
    |   |   |- FixedWingSupport
    |   |- FighterSquadron             // Now made up of ASFs, not base Aeros
    |- SmallCraft
    |   |- Dropship
    |   |- EscapePods                     // This should probably go under Aero, but out of scope for this PR.
    |- Jumpship
         |- Warship
         |- SpaceStation   
```
in service of reducing the number of classes that needed updates for IBB.  This should also allow reducing the code in Aero considerably and moving ASF-only functionality into that class, if desired.

2. Updated the Internal Bomb Bay quirk tooltip and text.
3. Added handling for Internal and External Bombs, including updates to configuration and bomb targeting dialogs.
4. Added IBB-related End Phase report, with relevant messages and explosion check / damage application.
5. Added ephemeral Cargo Bay generation for Aeros that mount the Cargo Bay miscellaneous equipment, as opposed to actual bays.  These are not saved, but are used when calculating bomb space for IBB (and, perhaps later, for cargo?).
6. Fixed "Invalid Design" checks for A) extra weight on ASFs with internal bombs and ephemeral bays, and B) "BA ECM" on Small Craft (MekHQ Issue [#1970](https://github.com/MegaMek/mekhq/issues/1970#issuecomment-1861759314)
7. Added accounting for ground-fire-only damage per entity; this is required for 4) above.
8. Added separate listing of Internal and External munitions for summary view and MUL saving/loading.
9. Added ability for VTOLs to bomb their current hex if a "bomb" hex was not added during the Move phase - this appears to have caused some confusion, and VTOLs should be able to bomb their final location in the Attack phase.
10. Fixed a timeout error that would cause Princess to hang when attempting to move ground units in the presence of a large number of bombs.  This may be the root of some of the recent Princess hangs we have not been able to track down; I will investigate further.

### TO-DO / Not Yet Implemented

1. Separately accounting for each bomb in each disparate cargo bay: Cargo Bays need to be upgraded to a real Class with internal logic and accounting for this to be feasible.  Currently we offer no way to account for location of bombs until they are instantiated after leaving the lobby; everything just gets thrown into "NOS".  This unfortunately means that cheesing bombers with multiple 6-bomb bays won't protect IBB users.
2. Counting cargo space filled with bombs as "used": see 1).
3. Effect of a Cargo crit on aerospace units mounting internal bomb stores and using IBB: forum question [here](https://bg.battletech.com/forums/index.php/topic,83510.msg1981761.html#msg1981761) has been answered and is provisionally implemented the same way as standard external bombs.  It looks like adding user choice of bombs to destroy will require wiring up some special event handling, which I feel is out of scope.
4. Verification of Fighter Squadron bombing capabilities; this came up as a question during development and is apparently not yet settled.

### Testing

1. Ran MM tests in Idea
2. Ran MHQ tests in Idea
3. Tested configuration, saving, loading, and MUL reinforcement with all valid Internal Bomb Bay unit types.
4. Verified operation of all valid bomber types, without and (where allowed) with Internal Bomb Bay.
5. Verified Princess operations with Internal, External, and mixed munition loads.
6. Verified MHQ compatibility (after concomitant MHQ patches)

### Note:

1. I updated the JVM options in build.gradle to use 4GB of RAM and disable AWT grabbing (source of a lot of freezing while debugging AWT windows).  I can remove this if folks don't feel it's necessary for them.

2. Pulling this PR will also require pulling the matching MekHQ PR [here](https://github.com/MegaMek/mekhq/pull/3805)

Close #4397


megamek/src/megamek/common/Aero.java
```diff
@@ -154,8 +154,12 @@ public String[] getLocationNames() {
     // fixed and pod-mounted.
     private int podHeatSinks;
 
-    protected int maxBombPoints = 0;
-    protected int[] bombChoices = new int[BombType.B_NUM];
+    protected int maxIntBombPoints = 0;
```
SJuliez, line 157: (not about this line) 

Since fighters are no longer directly Aero, do we ever instantiate Aero? It seems that the Aero class could be abstract.

megamek/src/megamek/common/AeroSpaceFighter.java
```diff
@@ -0,0 +1,179 @@
+/*
```
SJuliez, line 1: We have a standard license notice in https://github.com/MegaMek/megamek/wiki/MegaMek-Coding-Style-Guide

megamek/src/megamek/common/AeroSpaceFighter.java
```diff
@@ -0,0 +1,179 @@
...
+     * Other Aerospace units with cargo bays and the Internal Bomb Bay quirk may carry bombs internally.
+     * @return boolean
+     */
+    @Override
+    public boolean isBomber() {
+        return true;
+    }
+
+    @Override
+    /**
```
SJuliez, line 159: I suggest not repeating any comments from superclasses; hovering the method will show the superclass javadoc

megamek/src/megamek/common/Entity.java
```diff
@@ -4183,6 +4184,11 @@ protected void resetBombAttacks() {
             if (eq.getType().equals(spaceBomb) || eq.getType().equals(altBomb)
                     || eq.getType().equals(diveBomb)) {
                 bombAttacksToRemove.add(eq);
+            } else if (eq.getLinked() != null && eq.getLinked().isInternalBomb()){ // Does not do what's intended!
```
SJuliez, line 4187: "Does not do what's intended?"
?

megamek/src/megamek/common/IBomber.java
```diff
@@ -20,48 +20,79 @@
...
  * @author Neoancient
  */
 public interface IBomber {
-    
+
     String SPACE_BOMB_ATTACK = "SpaceBombAttack";
     String DIVE_BOMB_ATTACK = "DiveBombAttack";
     String ALT_BOMB_ATTACK = "AltBombAttack";
 
+    void setUsedInternalBombs(int b);
```
SJuliez, line 39: I'd love to have a javadoc comment on methods in an interface (as comments are visible in the implementors by hovering) unless it's totally obvious what a method does from the name.

megamek/src/megamek/common/battlevalue/BVCalculator.java
```diff
@@ -711,7 +711,7 @@ protected double arcFactor(Mounted equipment) {
     }
 
     /**
-     * Forwards to {@link #processWeapon(Mounted, boolean, boolean, int)} with a weaponCount
+     * Forwards to {@link #maintainUsedSearchlight(Mounted, boolean, boolean, int)} with a weaponCount
```
SJuliez, line 714: ???

SJuliez (changes requested):
A few comments.
Did you also check compatibility with MML? 
I noticed some code that could be omitted when there are no units left that are pure Aero class.

megamek/src/megamek/common/Aero.java
Sleet01: > (not about this line)
> 
> Since fighters are no longer directly Aero, do we ever instantiate Aero? It seems that the Aero class could be abstract.

It could be, although I wasn't planning to at this juncture.  It would let us move a lot of stuff out of Aero.

megamek/src/megamek/common/AeroSpaceFighter.java
Sleet01: Done.

megamek/src/megamek/common/IBomber.java
Sleet01: Updated.

megamek/src/megamek/common/battlevalue/BVCalculator.java
Sleet01: No idea; it appears to be an errant s-n-r from a prior patch.
Reverted.

Sleet01:
> A few comments. Did you also check compatibility with MML? I noticed some code that could be omitted when there are no units left that are pure Aero class.

I did check compatibility in MML, although there should be no impact there: bombs are only added in MM / MHQ, and the added "virtual" cargo space for ASFs (used to calculate internal bomb capacity for ASFs with "Cargo" equipment)  is not utilized by MML.

SJuliez (approved):

PR #5009, opened 2024-01-03
Fix Searchlights auto-enabled during daytime

Sleet01:
Currently the client setting to enable searchlights by default _should_ be skipping the pre-deployment Searchlight activation during daytime, but it is not.

This change disables starting with Searchlights on, even if the user has this option set, when doing so would not improve visibility.

### DIFF
--- a/megamek/src/megamek/server/GameManager.java
+++ b/megamek/src/megamek/server/GameManager.java
@@ -1090,7 +1090,8 @@ public class GameManager implements IGameManager {
             // reset spotlights
             // If deployment phase, set Searchlight state based on startSearchLightsOn;
             if (phase.isDeployment()) {
-                boolean startSLOn = PreferenceManager.getClientPreferences().getStartSearchlightsOn();
+                boolean startSLOn = PreferenceManager.getClientPreferences().getStartSearchlightsOn()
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/server/GameManager.java
```diff
@@ -1090,7 +1090,8 @@ private void resetEntityPhase(GamePhase phase) {
             // reset spotlights
             // If deployment phase, set Searchlight state based on startSearchLightsOn;
             if (phase.isDeployment()) {
-                boolean startSLOn = PreferenceManager.getClientPreferences().getStartSearchlightsOn();
+                boolean startSLOn = PreferenceManager.getClientPreferences().getStartSearchlightsOn()
+                        && game.getPlanetaryConditions().isIlluminationEffective();;
```
NickAragua, line 1094: Think we only need one semi colon at the end of a line. :)

NickAragua (changes requested):

megamek/src/megamek/server/GameManager.java
Sleet01: Sometimes Idea gets Ideas of its own... -_-;

Fixed.

NickAragua (approved):

PR #5012, opened 2024-01-04
QuirkEntry: remove maximum slot number check

SJuliez:
Weapon quirks are allowed on large craft and the slot numbers there can be higher than 11. The slot number test in the QuirkEntry constructor was obstructing this.

Fixes MegaMek/megameklab#1364

### DIFF
--- a/megamek/src/megamek/common/QuirkEntry.java
+++ b/megamek/src/megamek/common/QuirkEntry.java
@@ -90,7 +90,7 @@ public class QuirkEntry {
             throw new IllegalArgumentException("No location for " + code + " : " + unitId);
         } else if (StringUtility.isNullOrBlank(weaponName)) {
             throw new IllegalArgumentException("No weapon for " + code + " : " + unitId);
-        } else if ((slot < 0) || (slot > 11)) {
+        } else if (slot < 0) {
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/QuirkEntry.java
```diff
@@ -90,7 +90,7 @@ public QuirkEntry(String code, String location, int slot, String weaponName, Str
             throw new IllegalArgumentException("No location for " + code + " : " + unitId);
         } else if (StringUtility.isNullOrBlank(weaponName)) {
             throw new IllegalArgumentException("No weapon for " + code + " : " + unitId);
-        } else if ((slot < 0) || (slot > 11)) {
+        } else if (slot < 0) {
```
NickAragua, line 93: Maybe "slot > maximum number of slots in the location"? Or does it not really matter?

SJuliez: I would need the entity to do that. QuirkEntry is just a helper for these unit file lines: weaponquirk:stable_weapon:RT:3:ISGaussRifle
I'd say it's not such a big deal. (And I think it's more likely that the slot number is ok but there's no weapon there :D)

NickAragua: Fair enough

NickAragua (approved):

PR #5014, opened 2024-01-05
Fix CF warning sprite to work with isometric view.

HoneySkull:
This PR fixes the CF Warning rendering for isometric view.  Specifically when toggling back and forth between isometric and normal rendering.  The warning signs were not being updated to the correct elevation on toggle.

image

should be:

image


HoneySkull:
This is a minor gui cleanup.  The feature works in isometric view and non-isometric view - it just didn't update when toggling between the two.  (If one would select a different entity in deploy or movement phase, it would then update correctly.)

NickAragua (approved):

PR #5025, opened 2024-01-08
Fix #1021 - Add CTRL-Backspace to remove all trailing illegal move steps.

### LINKED ISSUE #1021: Allow Evade To Apply To Aero Unit Even If Player Overshoots Movement
### Environment
Megamek 0.44.0
Java 8
Windows 8.1

### Description
If the player overshoots the aero's movement capability and gets all the grey hexes and then clicks [Evade] megamek will draw the "Evade" text is grey over the ending hex but, will not apply it to the unit if the player then clicks done.  So, player must press escape and redo the movement so his/her cursor ends on the last hex to then enable evade.  Am just asking if megamek could be a little more forgiving to the player and still apply evade even if they overshoot it.

### Files
image

HoneySkull:
This PR is to support #1021 to add some convenience for Evading when Aero units have extended move paths beyond their movement points.  This is doesn't allow the Evade to work automatically like the issue suggests, however, it makes it easier to get back to the last legal move with a single keystroke instead of having to keep hitting the backspace key for each move step.  This works with any unit type, not just aerospace.

image

CTRL-Backspace

image


NickAragua (approved):

SJuliez:
It is my experience that we need to update the defaultKeyBinds.xml file when key binds are added or they won't work properly in a release (they do work in the IDE). Please add this.

HoneySkull:
> It is my experience that we need to update the defaultKeyBinds.xml file when key binds are added or they won't work properly in a release (they do work in the IDE). Please add this.

Good catch - will do.

HoneySkull:
Tested in IDE and clean install directory.  Ready for merge.

PR #5037, opened 2024-01-10
Ice Fix:  Princess Movement Cryophobia

Thom293:
Ice Fix: This PR changes one line of code to give Princess some Cryophobia in her movement planning.  Currently, On a map with approximately 50/50 Ice/Land, she would jump and walk mechs onto the ice every turn, and lose many to Ice before she ever got to  and during combat.  With this change she jumped mechs into the ice Zero times in 10 games.  And she will still walk on the ice when forced because she is landlocked.  Otherwise she will take a land path around.  

So in summary, it strongly discourages her from using ice as a path, but does not prevent it when necessary.

I simply changed a calculated hazard value (which has never been enough) to UNIT_DESTRUCTION_FACTOR and that solved it.  

I tested it over 10 bot v bot games and could see no drawbacks.  A sample savegame is attached below.

[ICET1.sav.gz](https://github.com/MegaMek/megamek/files/13890784/ICET1.sav.gz)

I attach a screen below that show a patch over ice that Locusts would still take when they deployed landlocked.  So it doesnt prevent use of ice entirely.

Screenshot 2024-01-10 101852

### DIFF
--- a/megamek/src/megamek/client/bot/princess/BasicPathRanker.java
+++ b/megamek/src/megamek/client/bot/princess/BasicPathRanker.java
@@ -33,11 +33,11 @@ public class BasicPathRanker extends PathRanker implements IPathRanker {
 
     // this is a value used to indicate how much we value the unit being at its destination
     private final int ARRIVED_AT_DESTINATION_FACTOR = 250;
-    
+
     // this is a value used to indicate how much we dis-value the unit being destroyed as a result of
     // what it's doing
     private final int UNIT_DESTRUCTION_FACTOR = 1000;
-    
+
     protected final DecimalFormat LOG_DECIMAL =
             new DecimalFormat("0.00", DecimalFormatSymbols.getInstance());
     private final NumberFormat LOG_INT = NumberFormat.getIntegerInstance();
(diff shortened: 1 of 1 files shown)


Thom293:
FYI - I only changed line 919. Not sure why it is showing those other changes.  I did not make them.

HoneySkull:
The checks that failed here are due to a failed jUnit test:

image

HoneySkull:
image

The expected hazard value was 166.7 but was actually 1000 with your changes.   What would you say the new expected value should be?  @Thom293

gsparks3:
Hmm ... at first I was going to say that infantry without UMU MP walking onto an ice hex over water is a 1 in 6 chance of instant destruction (ice breaks and they fall in), but reading the TO:AUE rules on page 48 more closely actually seems to imply that infantry do not check for breaking ice when entering a hex - although they can fall in if someone _else_ breaks the ice that they are standing on.

HoneySkull:
> image
> 
> The expected hazard value was 166.7 but was actually 1000 with your changes. What would you say the new expected value should be? @Thom293

@Thom293 - I opened a pull request for your Github fork with jUnit updates if that helps.

Thom293:
Thank you.  Yes I was just looking at this.  Its infantry related.  1000 is the pre-assigned value of UNIT_DESTRUCTION_FACTOR.  

Ill send you a message on Discord.  Infantry has its own path ranker, so I didnt test it, but if its lumped in here, I need to think about it.

Thom293:
Ok, so I think this PR has highlighted a vanilla bug with infantry hazard analysis.  They are checking for ice breakage but they dont break ice. That is why this is failiing an infantry test.  Im going to withdraw this one and see if I can fix the infantry bug along with this PR and re-submit.  If I cant fix the infantry bug, will ask someone else to help and maybe resubmit this.

PR #5040, opened 2024-01-11
First pass at adding TAG-phase Torso Twist

Sleet01:
This PR updates the Offboard / TAG phase to allow turret/torso twisting for units with TAG weapons in accordance with the errata ruling here:

https://bg.battletech.com/forums/index.php?topic=81851.0

, mainly by copying existing code in from `FiringDisplay.java`.

This also adds a separate message indicating that the Offboard/TAG phase is distinct from the Indirect Fire Targeting phase.

Testing:
- Ran full MM unit tests.
- Tested with player-controlled TAG units.

Note: Given how much shared code there is between `TargetingPhaseDisplay.java` and `FiringDisplay.java`, I'd like to take a crack at creating an abstract parent class for these two displays.  It should reduce total code by a couple thousand lines.


Sleet01:
Don't pull this one yet, I just realized I started on the code consolidation work before making units that twisted in the TAG phase _unable_ to twist again later that turn.

I'll try to get that done tomorrow.  While I think the `TargetingPhaseDisplay.java`/`FiringDisplay.java` code consolidation is worth looking at, it's too much work to complete right now.

Sleet01:
Okay, this should be ready to go.

megamek/src/megamek/common/Entity.java
```diff
@@ -249,6 +249,7 @@ public abstract class Entity extends TurnOrdered implements Transporter, Targeta
 
     protected int facing = 0;
     protected int sec_facing = 0;
+    protected GamePhase twistedPhase = null;
```
NickAragua, line 252: We probably want to make sure this is cleared out when resetting an entity as well.

NickAragua (changes requested):
One minor adjustment, looks good otherwise.

megamek/src/megamek/common/Entity.java
Sleet01: `twistedPhase` gets reset at the start of every Round, when `newRound()` is called:

```
        setAlreadyTwisted(false);
```
We don't want to reset it every phase as it has to persist between Targeting and Firing phases.

SJuliez (approved):
With the change request resolved I'll go and assume that this is good to go

PR #5041, opened 2024-01-11
Princess Ice & Infantry On Ice Fix

Thom293:
This PR gives Princess some Cryophobia for Mechs and non-hover land vehicles in her movement planning.  Currently, On a map with approximately 50/50 Ice/Land, she would jump and walk mechs onto the ice every turn, and lose many before she ever got to combat.  With this change she jumped mechs into the ice Zero times in 10 games.  And she will still walk on the ice when forced because she is landlocked.  Otherwise she will take a land path around.  There is simply no reason to be on Ice as a mech or non-hover ground vehicle except as an absolute last resort.

Also fixes a previously unreported bug related to Infantry and Ice where infantry would not willingly traverse Ice because they were grouped with the "I break the ice" crowd.  Moved them to their own "I dont break the ice" group and they now will move across ice.  Edited infantry test to reflect new expected value.

Sample savegame attached.  

[ICET1INFDeployed.sav.gz](https://github.com/MegaMek/megamek/files/13906476/ICET1INFDeployed.sav.gz)

Screen 1:  Shows post-fix path over ice that Locusts would still take when they deployed landlocked. So it doesnt prevent use of ice entirely.

Screenshot 2024-01-10 101852

Screen 2: Vanilla pre-fix MM screen that shows that infantry that deployed landlocked will remain landlocked the entire game.

Screenshot 2024-01-11 104527

Screen 3:  Shows post-fix infantry moving freely over ice from landlocked positions.

Screenshot 2024-01-11 110211

### DIFF
--- a/megamek/src/megamek/client/bot/princess/BasicPathRanker.java
+++ b/megamek/src/megamek/client/bot/princess/BasicPathRanker.java
@@ -33,11 +33,11 @@ public class BasicPathRanker extends PathRanker implements IPathRanker {
 
     // this is a value used to indicate how much we value the unit being at its destination
     private final int ARRIVED_AT_DESTINATION_FACTOR = 250;
-    
+
     // this is a value used to indicate how much we dis-value the unit being destroyed as a result of
     // what it's doing
     private final int UNIT_DESTRUCTION_FACTOR = 1000;
-    
+
     protected final DecimalFormat LOG_DECIMAL =
             new DecimalFormat("0.00", DecimalFormatSymbols.getInstance());
     private final NumberFormat LOG_INT = NumberFormat.getIntegerInstance();
@@ -45,19 +45,19 @@ public class BasicPathRanker extends PathRanker implements IPathRanker {
 
     private PathEnumerator pathEnumerator;
 
-    // the best damage enemies could expect were I not here. Used to determine 
+    // the best damage enemies could expect were I not here. Used to determine
     // whether they will target me.
     private Map<Integer, Double> bestDamageByEnemies;
-    
+
     public BasicPathRanker(Princess owningPrincess) {
         super(owningPrincess);
-        
+
         bestDamageByEnemies = new TreeMap<>();
-        
+
         LogManager.getLogger().debug("Using " + getOwner().getBehaviorSettings().getDescription()
                         + " behavior");
     }
-    
+
     FireControl getFireControl(Entity entity) {
         return getOwner().getFireControl(entity);
     }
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/client/bot/princess/BasicPathRanker.java
```diff
@@ -912,7 +921,9 @@ private double calcIceHazard(Entity movingUnit, Hex hex, MoveStep step, MovePath
                         breakthroughMod;
         logMsg.append("\n\t\tHazard value (")
               .append(LOG_DECIMAL.format(hazard)).append(").");
-        return hazard;
+        // Changed this to UNIT_DESTRUCTION_FACTOR because she suicided too often.
+        // No reason to be on the ice at all except as an absolute last resort.
+        return UNIT_DESTRUCTION_FACTOR;
```
HoneySkull, line 926: Be aware that returning a constant value here will essentially nullify the logic from line 915 through the end of the function.  The `hazard` return value can be 0 as assigned by `calcWaterHazard()` above it.  (i.e. for Amphibious types)

Consider maybe instead of returning a constant - maybe multiply `hazard` by some kind of "cryo-fear" factor - say 10 or something.  That way if `hazard` is 0, it will remain zero.

Thom293: Thank you.  Will look at tomorrow.

Thom293: Ok - so If I understand what you are saying - and I may not - yes this sets it to a fixed 1000 for everything remaining in this "might move on ice" group.  I did try to use incrementally larger numbers on ln 916, and no combination of numbers that I found prevented her from using the ice.  I also set a multiplier, and even increased those numbers to 999.5 and 999.1667. Nothing stopped her from using the ice until I set it to unit destruction factor.  (Although if you have another multiplier suggestion, I welcome it.  I was unable to code a working one).

I can move an exception for Aphibious units into this check, but that doesnt prevent the surface/breakage checks, and those are what are usually fatal, so I dont think even an amphibous vehicle should risk ice except as a last resort.

It would be different if this check prevented units from using water without ice, but it does not in my tests.  She moved mechs and hovers over and into water just fine even with this change.  (see screen below).

If I have misunderstood, or you have a better way to code it, I am happy to change.  I personally exhausted many non-working attempts before the final version I submitted, so improvement may be beyond my abilities.  Please let me what you would like for me to do.

image

Thom293: Talked to HS and answered question.  This is ready.

HammerGS:
Played around with this today, and seems to improve things. Will merge unless anyone has code objections.

PR #5050, opened 2024-01-13
Advanced search mul

### LINKED ISSUE #5046: RFE: Search by MUL ID in advanced search.
Would it be possible to add the ability to search by MUL ID.

There are times when I'm trying to chase down a mismatch in the MUL ID's and the only good option is to export the units and search that way.

kuronekochomusuke:
- add search for MUL id

image
image

fixes #5046


HammerGS (approved):
Tested and works great.

PR #5054, opened 2024-01-14
Fixes #5053 Update Aero movement envelope when accelerating or decelerating.

### LINKED ISSUE #5053: Aerospace movement envelope doesn't update when accelerating or decelerating in movement phase.
### Environment
Starting MegaMek v0.49.18-SNAPSHOT
	Build Date: N/A
	Today: 2024-01-13
	Origin Project: MegaMek
	Java Vendor: Eclipse Adoptium
	Java Version: 11.0.20.1
	System Locale: en_US

### Description
During the movement phase, when accelerating or decelerating with Aerospace units, the movement envelope is static to the initial crafts velocity at the beginning of the turn.  If a fighter starts the movement phase with velocity=1, a movement envelope shows the possible movement.  However, if the craft starts the movement phase with velocity=0 and you hit the `accelerate` button, the move envelope stays on velocity zero.

This is the behavior in v0.49.18-SNAPSHOT, as well as in the stable 0.48.0 - not sure if this ever worked.

Initial velocity of zero: 
image

Accelerate to velocity = 1:
image

When the fighter starts the movement phase with velocity 1:
image

NOTE: Not sure what the weirdness is with the left side of the movement envelope, but that's a different issue.

HoneySkull:
This PR fixes #5053 - When in the movement phase while moving an Aerospace entity, a "delta-v" is maintained during the selected unit's movement so that an Aero movement envelope can be dynamically updated when the pilot accelerates or decelerates at the beginning of the turn.

As it is, the movement envelope is static to the initial velocity of the craft upon entering the movement phase.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/MovementDisplay.java
+++ b/megamek/src/megamek/client/ui/swing/MovementDisplay.java
@@ -342,12 +342,6 @@ public class MovementDisplay extends ActionPhaseDisplay {
     public static final int GEAR_STRAFE = 12;
     public static final String turnDetailsFormat = "%s%-3s %-14s %1s %2dMP%s";
 
-    // Keep track of acceleration for Aero units to dynamically update their move envelope.
-    private static final int ZERO_DELTA_V = 0;
-    private static final int INC_DELTA_V = 1;
-    private static final int DEC_DELTA_V = -1;
-    private int deltaV = 0;
-
     /**
      * Creates and lays out a new movement phase display for the specified
      * clientgui.getClient().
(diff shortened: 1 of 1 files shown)


NickAragua (approved):

HoneySkull:
I found a better way to implement this.  Instead of maintaining a delta-v, the MovePath object already has a function that determines the 'final velocity'.   I am going to re-do this so that it uses that instead.  The major advantage is that it's already using existing code, AND it will automatically be up-to-date when cycling through the backspace key. (Say, when a user accelerates multiple MP and they backspace and unwind... the first draft wasn't handling that case well.

HoneySkull:
Yeah, this works much better with backtracking through the move path and is all around simpler code.  Less to go wrong.

PR #5061, opened 2024-01-16
Infantry Weight Breakdown 

SJuliez:
This PR:
- moves infantry weight calculation to TestInfantry 
- adds a report for weight calculation and updates TestInfantry to show the weight calculation when asked
- adds a specialized weight display dialog
- adds the getEntityVerifier() static method from MML's UnitUtil to TestEntity. It can later be removed in MML


kuronekochomusuke:
should the getTestEntity copies of the function be replaced with a call to TestEntity.getEntityVerifier ?

SJuliez:
Yes, that would make sense.

neoancient (approved):

PR #5070, opened 2024-01-21
CI Anti-Mek Gear

### LINKED ISSUE #5063: Infantry should store the fact that it has Anti-Mek gear
Currently, Infantry does not store the fact that it has anti-mek gear; it is assumed that an anti-mek skill of 8 means no training, no gear and anything below means gear; the skill itself is stored. But the skill with gear can still be 8 and the skill is not fixed anyway, so instead of the skill the presence of anti-mek gear should be stored.
TM p.155 and 315
This is tied to #5058

SJuliez:
This is the MM side of adding anti-mek gear for conv. infantry, replacing the approach of only storing anti-mek in the skill value.
- adds anti-mek gear as a MiscType (as its own class because why not)
- isAntiMekTrained() is renamed to hasAntiMekGear() which pops up in a few classes
- removes the TechAdvancement for AM gear from Infantry since it's now part of its gear, also removes a few unused methods (checked MHQ)
- adding or removing anti-mek attacks has been moved to TestInfantry; this is used in MML and in the MechFileParser upon loading; we might want to update this further to look at the game's year as these attacks aren't always available
- updates the BV calculator to (hopefully correctly) assign 5 or 8 or other anti-mek depending on gear and crew
- updates the BLKFile load and save for conv inf; I changed it to write the LOC_INFANTRY location where armor kits and anti-mek gear and prim or sec weapon are stored; this used to be blocked in favor of writing the info individually; it seems to work rather easily like this, but I had to explicitly block this location from writing the infantry prim or sec weapon as these still need to be written individually. Also, I removed the block version; it will no longer be written. It was ignored during load anyway.
- and, finally, updates TestInfantry to catch anti-mek gear on mechanized as invalid; I wanted to add general equipment checks but this makes many infantry units invalid as their intro dates are too early, so I commented that out for now, but it should be added at some point. Also tweaked the weight calculation report a bit.
-
Fixes #5063 
Also happens to ... Fixes MegaMek/megameklab#1333 as BA wrongly wrote InfantryAttacks into the blk file which it shouldn't

### DIFF
--- a/megamek/src/megamek/client/generator/skillGenerators/AbstractSkillGenerator.java
+++ b/megamek/src/megamek/client/generator/skillGenerators/AbstractSkillGenerator.java
@@ -144,7 +144,7 @@ public abstract class AbstractSkillGenerator implements Serializable {
         // based on whether the unit has anti-mek training, which gets set in the BLK file.
         // We therefore check if they are anti-mek trained before setting
         if (entity.isConventionalInfantry() && !((Infantry) entity).hasAntiMekGear()) {
-            skills[1] = Infantry.ANTI_MECH_SKILL_UNTRAINED;
+            skills[1] = Infantry.ANTI_MECH_SKILL_NO_GEAR;
         } else if (isForceClose()) {
             skills[1] = skills[0] + 1;
         }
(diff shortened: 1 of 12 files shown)


HammerGS:
About to start into a big infantry project, been testing and things seem to work. Going to merge this so I can continue to test.

PR #5071, opened 2024-01-21
MUL parser updates

kuronekochomusuke:
- use constants instead of string values
- add prefix to constants based on usage
- add new constants

https://github.com/MegaMek/mekhq/pull/3818  need to be committed with this one


SJuliez (approved):

HammerGS:
I see Juliez approved but had a fix after.  Let me know when its done for the merge.

kuronekochomusuke:
It is done.  it also needs the mekhq one merged with it.

PR #5095, opened 2024-01-28
Fixes #5088 Missing holes in Aero Move envelope.

### LINKED ISSUE #5088: [0.49.18-SNAPSHOT] Inaccurate Movement Envelope for Aerospace Fighters on Ground Maps
### Environment
Starting MegaMek v0.49.18-SNAPSHOT
	Build Date: N/A
	Today: 2024-01-26
	Origin Project: MegaMek
	Java Vendor: Eclipse Adoptium
	Java Version: 11.0.20.1
	System Locale: en_US

### Description
With the Aerospace move envelope, the resulting envelope for Aero units appears with gaps and doesn't represent all the movements possible.  This can be seen as far back as 0.48.0 stable when the initial velocity at the beginning of a turn is 2 or above (depending upon the turning limit).  However, it's more prominent with the enablement of updated move envelopes in 0.49.18-SNAPSHOT.  
 
image

It should look something like this: 

image

I found a simple code-tweak to fix this, but it is NOT performant at velocity 3 or greater and will cripple a machine at velocity 4 or higher.  To be honest, a very large move envelope isn't very useful for fighters past 2 velocity.  (I'm starting to wonder if this is why move envelope was disabled for aero int the first place.)

HoneySkull:
Fixes #5088 

This PR improves the aerospace movement envelope.  Using a path length comparator instead of a movement point comparator, the strange holes and missing edges are filled in more symmetrically.  

The move envelope for Aero here is not 100% accurate as some of the inside radius hexes are not accounted for, but it has a uniform envelope without holes or missing wedges and good performance past 8 velocity.

This PR also adds a fix for showing the current velocity move envelope when toggling "View Possible Moves" instead of reverting back to the initial move velocity when toggling.

This image shows the improved Aero move envelope without the edge gaps and wedges, but shows that there are still some inside radius hexes that are not represented.
image

This image shows the same envelope before this fix:
image

### DIFF
--- a/megamek/src/megamek/client/ui/swing/ClientGUI.java
+++ b/megamek/src/megamek/client/ui/swing/ClientGUI.java
@@ -975,7 +975,12 @@ public class ClientGUI extends JPanel implements BoardViewListener,
             case VIEW_MOVE_ENV:
                 if (curPanel instanceof MovementDisplay) {
                     GUIP.setMoveEnvelope(!GUIP.getMoveEnvelope());
-                    ((MovementDisplay) curPanel).computeMovementEnvelope(getUnitDisplay().getCurrentEntity());
+                    Entity entity = getUnitDisplay().getCurrentEntity();
+                    if (!(entity instanceof Aero)) {
+                        ((MovementDisplay) curPanel).computeMovementEnvelope(entity);
+                    } else {
+                        ((MovementDisplay) curPanel).computeAeroMovementEnvelope(entity);
+                    }
                 }
                 break;
             case VIEW_MOVE_MOD_ENV:
--- a/megamek/src/megamek/client/ui/swing/ClientGUI.java
+++ b/megamek/src/megamek/client/ui/swing/ClientGUI.java
@@ -976,7 +976,7 @@ public class ClientGUI extends JPanel implements BoardViewListener,
                 if (curPanel instanceof MovementDisplay) {
                     GUIP.setMoveEnvelope(!GUIP.getMoveEnvelope());
                     Entity entity = getUnitDisplay().getCurrentEntity();
-                    if (!(entity instanceof Aero)) {
+                    if (!entity.isAero()) {
                         ((MovementDisplay) curPanel).computeMovementEnvelope(entity);
                     } else {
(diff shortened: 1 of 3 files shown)


megamek/src/megamek/client/ui/swing/ClientGUI.java
```diff
@@ -975,7 +975,12 @@ public void actionPerformed(ActionEvent event) {
             case VIEW_MOVE_ENV:
                 if (curPanel instanceof MovementDisplay) {
                     GUIP.setMoveEnvelope(!GUIP.getMoveEnvelope());
-                    ((MovementDisplay) curPanel).computeMovementEnvelope(getUnitDisplay().getCurrentEntity());
+                    Entity entity = getUnitDisplay().getCurrentEntity();
+                    if (!(entity instanceof Aero)) {
```
neoancient, line 979: `entity instanceof Aero` is false for LAMs in fighter mode.
```suggestion
                    if (!entity.isAero()) {
```

megamek/src/megamek/client/ui/swing/MovementDisplay.java
```diff
@@ -4569,7 +4576,7 @@ public void computeMovementEnvelope(Entity suggestion) {
      * @return - This method will do nothing if the Entity passed in is null or
      *           is not an Aero based unity.
      */
-    private void computeAeroMovementEnvelope(Entity entity) {
+    public void computeAeroMovementEnvelope(Entity entity) {
         if ((entity == null) || !(entity instanceof Aero) || (cmd == null)) {
```
neoancient, line 4580: To handle LAMs correctly this line should also change to entity.isAero(), and a few lines below this should case to IAero instead of Aero.

neoancient (changes requested):
Looks good, just needs a couple minor changes to work correctly for LAMs.

megamek/src/megamek/client/ui/swing/MovementDisplay.java
HoneySkull: okay - I may need to go back and check to see if I've used `instanceof Aero` in other places as well.

neoancient (approved):

PR #5101, opened 2024-01-30
Armor rework, part 1a

neoancient:
This is the first half of the first step of reworking armor `EquipmentType`s to consolidate the armor statistics into one place and streamline their handling. The goal of the first part is to eliminate the static methods in EquipmentType and the construction data in Entity and TestEntity. This PR does so for armor names and weights/points per ton. I left it off here to keep the change from getting even bigger, but the second half of this step involves costs, removal of the remaining static methods from EquipmentType  (many of which now simply pass the work on to ArmorType), and addition of support vehicle armors.


NickAragua (approved):
Looks fine code-wise.

Issue #5103, opened 2024-01-30
[0.49.17] To-hit calculation bugs re dusk/dawn searchlights and TacOps BAP option


Gribbly1:
Bundled Megamek with MekHQ 0.49.17
Windows 10, Java Version: 11.0.13

Searchlights should not counteract to hit penalties for dusk/dawn. Units without BAP should not have reduced TN vs units in woods. This WHM-6R does not actually have BAP.

Player name: test

[save1.sav.gz](https://github.com/MegaMek/megamek/files/14104770/save1.sav.gz)

![image](https://github.com/MegaMek/megamek/assets/55139229/c6ec8aa6-ee04-4ca1-90af-5ae0badc922a)

kuronekochomusuke:
similar BAP issues
#5072
#5030

RaozSpaz:
![image](https://github.com/MegaMek/megamek/assets/132851640/7923a7a7-5b55-4205-b68f-0cf1138649b2)
So just to be certain I went in and turned off the BAP TOps rule and still got the modifier. Also it does effect enemies too
![image](https://github.com/MegaMek/megamek/assets/132851640/88188c98-1b70-4409-8a25-069443af1020)

PR #5108, opened 2024-01-31
correct issues with BAP targeting to hit bonus

### LINKED ISSUE #5030: [HQ49.17] Displayed ToHit# and Tool-Tip in the Round Report are bugged when effected by active probe advanced rule.
The image illustrates what happens every time an attack is made while effected by the "TacOps Beagle Active Probes" advanced rule.  It seems part of the Tool-Tip is displayed in front of the ToHit# while the Tool-Tip itself is incomplete.

probe

Also, unrelated to this issue, is there a way to change the color of the ToHit# and ToHitRolls text?  I'm having a really hard time seeing them and I don't see a way to change them in the client settings.

### LINKED ISSUE #5072: [0.49.17] Modifier for BAP shooting through woods shows up in wrong places
### Environment
MegaMek version 0.49.17, packaged with MekHQ
Windows 11
Java 11.0.14.1

### Description
During the shooting phase, in the section of the unit window where accuracy modifiers are displayed, mechs firing through woods were given the "target behind woods and attacker has BAP" modifier despite none of the mechs having BAPs or pilots with SPAs. I am not yet sure whether this was actually affecting accuracy or if it was just a display bug.

That modifier also appeared, clearly erroneously, in the firing results list. Where it should just state a to-hit number it instead includes "behind woods and attacker has BAP)'>" before that number.

To reproduce, just load into the game and find a position where you'll be shooting through woods. (This should be easy.)

### Files
Screenshot 2024-01-21 172153
firingreportBAPbug
[megamek.log](https://github.com/MegaMek/megamek/files/14003576/megamek.log)
[BAPbugsave.sav.gz](https://github.com/MegaMek/megamek/files/14003577/BAPbugsave.sav.gz)
Player name: Evil Atlas

kuronekochomusuke:
- correct issues with BAP targeting always providing bonus to hit when shooting through woods
- correct issue with C3 member providing bonus to any target not just targets within the C3 member's BAP range
- show a updated message when receiving the bonus from a C3 member
- remove / from message that seems to cause display issues in the report
- show active probe range on unit tooltip

image
image

image

#5103
fixes #5030
fixes #5072


megamek/src/megamek/common/LosEffects.java
```diff
@@ -237,6 +237,10 @@ public int getUltraWoods() {
         return ultraWoods;
     }
 
+    public boolean thruWoods() {
+        return (lightWoods > 0) || (heavyWoods > 0) || (ultraWoods > 0);
```
NickAragua, line 241: lol no jungle?

NickAragua (approved):

PR #5116, opened 2024-02-02
Armor rework, part 1b

neoancient:
Continuation of #5101.
Removes armor names array from EquipmentType.
Moves armor cost calculations from EquipmentType to ArmorType.
Splits protomech armor from standard armor due to differences in cost.


megamek/src/megamek/common/equipment/ArmorType.java
```diff
@@ -293,7 +307,7 @@ private static ArmorType createClanFerroFibrous() {
         armor.setInternalName("Clan Ferro-Fibrous");
         armor.addLookupName("Clan Ferro-Fibrous Armor");
         armor.addLookupName("Clan Ferro Fibre");
-        armor.tonnage = TONNAGE_VARIABLE;
```
NickAragua, line 296: Is there a reason we're removing setting the tonnage?

NickAragua (approved):
Just one question, more likely for my own education than anything else.

megamek/src/megamek/common/equipment/ArmorType.java
neoancient: It's not used. Armor tonnage is determined by the number of points of armor on the unit and calculated by points per ton or weight per point depending on unit type.

Issue #5113, opened 2024-02-01
Exceptions upon starting a new game 


SJuliez:
When starting the lobby, a bunch of exceptions show up pointing to Entity.getArmorWeight, l.10837. They seem to have no further effect, units load fine after that.
Edit: It's mostly Gun Emplacements; these can't be added in the lobby

HammerGS:
![image](https://github.com/MegaMek/megamek/assets/13752452/cce9b735-3bc0-42f2-b82c-5f25dc3cbe34)

[megamek.log](https://github.com/MegaMek/megamek/files/14141813/megamek.log)

```
megamek.common.MechFileParser.<init>(MechFileParser.java:78) - 
java.lang.NullPointerException: Cannot invoke "megamek.common.equipment.ArmorType.getPointsPerTon(megamek.common.Entity)" because the return value of "megamek.common.equipment.ArmorType.forEntity(megamek.common.Entity, int)" is null
	at megamek.common.Entity.getArmorWeight(Entity.java:10838)
	at megamek.common.Entity.getArmorWeight(Entity.java:10853)
	at megamek.common.loaders.BLKAeroSpaceFighterFile.getEntity(BLKAeroSpaceFighterFile.java:181)
	at megamek.common.MechFileParser.parse(MechFileParser.java:157)
	at megamek.common.MechFileParser.<init>(MechFileParser.java:73)
	at megamek.common.MechSummaryCache.loadMechsFromZipFile(MechSummaryCache.java:850)
	at megamek.common.MechSummaryCache.loadMechsFromDirectory(MechSummaryCache.java:758)
	at megamek.common.MechSummaryCache.checkForChanges(MechSummaryCache.java:257)
	at megamek.common.MechSummaryCache.loadMechData(MechSummaryCache.java:246)
	at megamek.common.MechSummaryCache.lambda$getInstance$0(MechSummaryCache.java:85)
	at java.base/java.lang.Thread.run(Thread.java:840)

```
```
23:13:30,099 ERROR [megamek.common.MechFileParser] {Mech Cache Loader}
megamek.common.MechFileParser.<init>(MechFileParser.java:78) - 
java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
	at megamek.common.Entity.getArmorTechLevel(Entity.java:10682)
	at megamek.common.equipment.ArmorType.forEntity(ArmorType.java:40)
[...]

PR #5119, opened 2024-02-02
ArmorType fixes

neoancient:
Fixes for errors introduced by #5101
1. Added a "no armor" virtual armor for gun emplacements.
2. Override getArmorTechLevel(int) in GunEmplacement to avoid an IndexOutOfBounds exception.
3. If the armor type is not found using the armor type constant and the clan flag, look for it with the Clan flag reversed. This is a workaround for some mixed tech vehicles that have the unit tech level where the armor tech level should be, resulting in checks for Clan vehicular stealth, Clan heavy ferro-fibrous, etc.
4. Remove the space at the end of the lookup names for industrial armor. I left the space at the end of the name because there was a comment that it is that way on purpose (though I don't know what the reason is), but the mtf file loader is looking for it without the trailing space.


SJuliez (approved):
Fixes #5113

PR #5126, opened 2024-02-04
turn timer updates

kuronekochomusuke:
- add game options for turn timer for targeting, movement, firing, and physical phases
- add game option to allow turn timer extensions
  - the extensions can be used to allow more time for newer players or complex units like aerospace
  - add key bind CTRL-F4 to activate a turn timer extension, can be activated only when timer is active
  - when turn timer gets to 2 seconds, resets turn timer if extension is active
  - when turn timer extension is activated show a ⭐ after the time in the timer bar
  - when turn timer is reset, send chat message to other clients
- start turn timer bar with caution color, switch to warning color when 10 seconds left
- adjust timer display with the GUI Scale

image
image
image
image
image


sldfgunslinger2766:
🤘

BLR-IIC:
I like this!  Being able to specify a timer for individual phases makes it possible to just limit the time for movement, which is helpful for the simultaneous movement option.

neoancient (approved):

PR #5130, opened 2024-02-05
Fix remaining armor test failures

Sleet01:
This should allow the remaining failing unit tests to pass.

The issue is that the ArmorType types are not guaranteed to be initialized, because `ArmorType.initializeTypes()` is only called from `EquipmentType.initializeTypes()`, and _that_ is only called in a few specific paths that rely on actually starting up the game itself.

I've added a check in `ArmorType.of()` that should avoid _some_ infinite recursions as well.

These fixes will allow tests to pass but are pretty fragile.  It would be better to ensure that ArmorTypes are always initialized no matter what, but I'm not sure where would be best so I will leave that to others.

### DIFF
--- a/megamek/src/megamek/common/equipment/ArmorType.java
+++ b/megamek/src/megamek/common/equipment/ArmorType.java
@@ -34,6 +34,9 @@ public class ArmorType extends MiscType {
     private static final List<ArmorType> allTypes = new ArrayList<>();
 
     public static ArmorType of(int type, boolean clan) {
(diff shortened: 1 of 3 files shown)


megamek/src/megamek/common/equipment/ArmorType.java
```diff
@@ -251,6 +255,29 @@ public int getBAR() {
         return bar;
     }
 
+    private static ArmorType createUnknownArmor() {
```
Sleet01, line 258: We used to have an "Unknown" armor type, which is still used in some tests / files.  This should cover that option.

neoancient (approved):

PR #5134, opened 2024-02-06
Initialize EquipmentType for individual unit tests.

HoneySkull:
This PR ensures that specific unit test classes that mock BattleArmor.class can be run from an IDE as individual unit tests by initializing the static collections of ArmorType and EquipmentType as they may not have been initialized when running these test classes as jUnit tests in isolation.

FireControlTest.java
PrincessTest.java
EntityTest.java
PlanetaryConditionsTest.java

Without these changes, these tests depend upon some other test to run before them that initialize these static collections.


neoancient (changes requested):
EquipmentType.initializeTypes() calls ArmorType.initializeTypes(). There's no need to call it a second time.

neoancient (approved):

Issue #4827, opened 2023-10-12
Null pointer exception when conducting counter-battery fire with Inferno-IVs


Saklad5:
### Environment
MegaMek Version: MekHQ v0.49.14 Development Snapshot
Operating System: Mac
Java Version: 20.0.1

### Description

I was using Inferno Arrow-IVs for counterbattery fire, reasoning that a direct hit would be likely to crit important components in the enemy artillery platform. Unfortunately, because the current implementation does not actually model the hex that off-board artillery resides in, this causes a null pointer exception to occur when the attack is resolved.

There are two ways this could be resolved: either skip setting fires with null hexes, which causes only the "Inferno SRM" effect of Inferno-IVs to be applied, or actually set fires on the off-board hexes, causing the artillery to remain on fire until it disengages.

Since the rules themselves aren't quite clear on this, I'm going to ask on the rules forum as well. In the interim, the first approach is probably sufficient to fix the issue for now.
```
java.lang.NullPointerException: Cannot invoke "megamek.common.Hex.terrainLevel(int)" because "h" is null
	at megamek.server.GameManager.deliverArtilleryInferno(GameManager.java:10064)
	at megamek.common.weapons.ArtilleryWeaponIndirectFireHandler.handle(ArtilleryWeaponIndirectFireHandler.java:330)
	at megamek.server.GameManager.handleAttacks(GameManager.java:34325)
	at megamek.server.GameManager.handleAttacks(GameManager.java:34261)
	at megamek.server.GameManager.endCurrentPhase(GameManager.java:2471)
	at megamek.server.GameManager.changeToNextTurn(GameManager.java:2708)
	at megamek.server.GameManager.endCurrentTurn(GameManager.java:1562)
	at megamek.server.GameManager.receiveAttack(GameManager.java:13045)
	at megamek.server.GameManager.handlePacket(GameManager.java:763)
	at megamek.server.Server.handle(Server.java:1265)
	at megamek.server.Server$PacketPump.run(Server.java:111)
	at java.base/java.lang.Thread.run(Thread.java:1623)
```

Saklad5:
I've submitted a [question](https://bg.battletech.com/forums/index.php/topic,82870.0.html) about this on the rules forum.

Sleet01:
Is this fixed by the fix for #4803 ?

Saklad5:
[...]

PR #5138, opened 2024-02-07
Fix 4952: Princess does not select alt ammo

### LINKED ISSUE #4952: [49.16 - J17] Princess Issues with Advanced Ammo Selection (MMLs and LACs at least).
Dont know how hard this is to take on, but, would be a good boost to her if fixed.

TLDR:  My guess is she is not doing whatever the "what is the best ammo for this round" check until after her first shot with MMLs.  She may not do it ever with normal AC ammo (appears to just default to first in list).  She does appear to make an ammo decision with LBX and swap ammo prior to first firiring and as apprpriate during match.

MMLs: She will not change from starting ammo until shots are fired.  Once she shoots any ammo, she will swap them as needed, but she will not swap from starting ammo until she has shot it - which if a problem is SRM ammo is initially loaded.

Scenario - load two identical MML units in opposition.  Both with SRM and LRM ammo.  SRM ammo initially loaded.  

Neither unit will fire until in SRM range.

Change one to LRM ammo initially loaded, that one will fire in LRM range and then swap to SRM ammo as it closes.  Other unit will still only fire in SRM range.

Add reinforcements 8 rounds later.  After initial battle with both starting units using SRMs, survivor will swap to LRMs and shoot at reinforcements.

So my guess is she is not doing whatever the "what is the best ammo for this round" check until after her first shot.  

Screen 1 shows both in LRM range but not shooting. 
image

LAC:  Testing shows LAC-5 equipped mech has never switched from initially loaded basic ammo to either Precision or AP (has both) ammo (with a TN of 4 might want AP ammo, although it didnt matter here).   While approaching should have deffo switched to Precision. It LOOKs like with LAC she will fire whatever is in the first ammo slot, regardless of what is stated to be loaded at the bottom (here they match but did not in most other cases and it it defaulted to first in list).

image

LBX:  Screen 3 shows that LBX5 loaded with cluster swithed to solid ammo before first shots.  Also when enemy mech was down with TN of 3, she switched back to cluster even when she still had solid ammo to try to "finish it off."  So, looks like LBX checks for something and checks it pretty well, but MML and LAC do not.

image

I dont have a screen and dont have time to test right now, but I seem to recall seeing this behavior with hot loaded LRMS before too.  I would assume AtMs and anything with alternate ammos may need to be looked at.  And although not exactly on point, If it matters, if the optional "rapid fire AC" rule is checked, she does seem to make some decision there, as she doesnt always shoot in rapid mode, but will sometimes.

Here are sample saves for both MML and LAC scenarios, game logs and customs for the LAC issue.

[Ammo.zip](https://github.com/MegaMek/megamek/files/13653067/Ammo.zip)

Sleet01:
This PR adds alternate ammo selection for Princess, based on the existing toHit guessing and calculations, by allowing her to iterate over every weapon _and_ all its ammo, if applicable, when calculating utility of various actions.

Currently, Princess has no conception of what ammunition is available to units other than whatever ammo is currently loaded.  She neither knows about nor accounts for switching ammo, either by her own units or by enemy units, with two small exceptions:

1. When calling `getPreferredAmmo()` for attacks she is _already planning to execute_, Princess can switch ammunition.
2. When an enemy switches ammunition for its weapons, she can incorporate that info into planning her _next_ turn's movement - but she cannot foresee it.

In case 1, the weapon in question must have already had sufficient range to attack the target; currently Princess can't switch ammos to extend a weapon's range, so the number of situations in which Princess can _currently_ take advantage of ammo switching is limited.  Also, this function must explicitly define a situation in which one ammunition should be taken over another, and how to effect that switch; it does not cover all available options.  The new code, however, will utilize the existing attack utility calculations and should not need to be expanded even if new weapons or munition types are added - it will just work.

In case 2, it is currently possible, although rare, that two opposing Princesses will both be able to take advantage of ammo switching via `getPreferredAmmo()` and switch their loaded ammo for some of their weapons; in such a case, the new ammo will be considered in future calculations but both Princesses will effectively "forget" that the _original_ ammo type ever existed, so while being able to adapt to each others' choices, neither will be able to switch back very easily.

The work here allows Princess to iterate over every valid, available ammo type for each ammunition-reliant weapon when:
- assessing possible moves for her forces,
- predicting incoming enemy fire while moving,
- deciding how to actually allocate fire _after_ movement has occurred.

It also adds:
- More consideration for expected damage when calculating shots (as opposed to quick guessing, for pathfinding)
- Higher aversion to friendly fire with artillery / AE weapons (doubling the negative weight of damage to friendlies in assessing utility)
- More accurate assessment of ADA Missile utility during the Indirect phase, when an Arrow IV unit may wish to hold off on firing in order to make ADA Missile direct-fire attacks later in the turn.
- A `getAmmo()` call that returns only the ammo valid for a passed-in weapon
- Additional info in the mouse-over display for declared attacks; these now include the ammo utilized.
- Some additional logging, specifically covering `try/except` clauses that contained pernicious bugs.

These improvements will affect all ammunition-fed weapons, including artillery, if the unit that will fire has more than one ammunition type loaded for any of its weapons.  Otherwise, the logic will not be affected.

While this requires a measurable amount of additional calculation (now iterating over all weapons _and_ all available ammo for those weapons multiple times every turn for every unit on the field), I have made attempts to pare down the total amount of calculations, and so far speed does not seem to be adversely affected (in my limited testing).  I've added several early escape paths in calculations, skipping of already-fired weapons (such as Artillery or TAG), and changed some locations where we repeatedly extract objects from the `game` via ID lookup to just pass in the objects themselves.

**Testing done:**
1) ran all MM and MHQ unit tests,
2) added a few new unit tests to show that the specific cases from the original issue are covered properly,
3) ran a wide mix of unit types in human-bot and bot-bot matches utilizing multiple alternate munitions for every class of weapon:
- Missiles: SRMs, LRMs, MMLs, ATMs, MRMs, Thunderbolts, Streaks
- Ballistics: ACs, LACs, U/ACs, Gausses, HAGs, Rifles, LB-Xs, MGs, Artillery Cannons
- Energy: PPCs, Lasers, Flamers, TAG
- Bombs: AE, Arrow IV, TAG pod
- Artillery (on- and off-board): HE, Copperhead, AIV, AIV Homing

**Future work:**
- There is an opportunity to combine at least part of the various versions of *Plan() functions to use the same inner loop, so that future work will require less boilerplate.  I didn't do this as the differences between each are subtle and it is not immediately necessary.
- The entirety of code called by or contained within `calculateFiringTurn()` could be put into a class, with various subclasses, and Princess would then attach an instance of one of these to herself.  In this way we could make Princess's ability to calculate toHit and expected damage data a tunable knob for adjusting her difficulty level.
- TAG planning really should be separate from Indirect Fire planning, especially now that the TAG sub-phase can now include torso twisting.  I've marked the appropriate location with a TODO.
- Princess still won't have much idea what to do with support munitions like Smoke or Thunder/ FASCAM, as she lacks any way to assess their utility.  This could likely be implemented in a low-weight fashion by caching projected incoming fire results or enemy movent paths, and assigning a utility score for blocking those hexes off, since Princess already runs those calculations every turn to guess what opponents will do.

Close #4952

### DIFF
--- a/megamek/src/megamek/client/bot/princess/AeroPathUtil.java
+++ b/megamek/src/megamek/client/bot/princess/AeroPathUtil.java
@@ -4,10 +4,9 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 
-import megamek.common.IAero;
-import megamek.common.MovePath;
-import megamek.common.UnitType;
+import megamek.common.*;
 import megamek.common.MovePath.MoveStepType;
+import megamek.common.annotations.Nullable;
 
 /**
  * Helper class that contains functionality relating mostly to aero unit paths.
@@ -15,32 +14,32 @@ import megamek.common.MovePath.MoveStepType;
  */
 public class AeroPathUtil {
     public static List<List<MoveStepType>> TURNS;
-    
+
     static {
         // put together a pre-defined array of turns. Indexes correspond to the directional values found in Coords.java
         TURNS = new ArrayList<>();
         TURNS.add(new ArrayList<>()); // "no turns"
-        
+
         TURNS.add(new ArrayList<>());
         TURNS.get(1).add(MoveStepType.TURN_RIGHT);
-        
+
         TURNS.add(new ArrayList<>());
         TURNS.get(2).add(MoveStepType.TURN_RIGHT);
         TURNS.get(2).add(MoveStepType.TURN_RIGHT);
-        
+
         TURNS.add(new ArrayList<>());
         TURNS.get(3).add(MoveStepType.TURN_RIGHT);
         TURNS.get(3).add(MoveStepType.TURN_RIGHT);
         TURNS.get(3).add(MoveStepType.TURN_RIGHT);
-        
+
         TURNS.add(new ArrayList<>());
         TURNS.get(4).add(MoveStepType.TURN_LEFT);
         TURNS.get(4).add(MoveStepType.TURN_LEFT);
-        
+
         TURNS.add(new ArrayList<>());
         TURNS.get(5).add(MoveStepType.TURN_LEFT);
     }
-    
+
     /**
      * Determines if the aircraft undertaking the given path will stall at the end of the turn.
      * Only relevant for aerodyne units
@@ -63,13 +62,13 @@ public class AeroPathUtil {
         if ((movePath.getFinalVelocity() == 0) && isAirborne && !isSpheroid) {
             return true;
         }
-        
-        if (isSpheroid && (movePath.getFinalNDown() == 0) 
+
+        if (isSpheroid && (movePath.getFinalNDown() == 0)
                 && (movePath.getMpUsed() == 0)
                 && !movePath.contains(MoveStepType.VLAND)) {
             return true;
         }
-        
+
         return false;
     }
 
@@ -110,14 +109,14 @@ public class AeroPathUtil {
      */
     public static Collection<MovePath> generateValidAccelerations(MovePath startingPath, int lowerBound, int upperBound) {
         Collection<MovePath> paths = new ArrayList<>();
-        
+
         // sanity check: if we've already done something else with the path, there's no acceleration to be done
         if (startingPath.length() > 0) {
             return paths;
         }
-        
+
         int currentVelocity = startingPath.getFinalVelocity();
-        
+
         // we go from the lower bound to the current velocity and generate paths with the required number of DECs to get to
         // the desired velocity
         for (int desiredVelocity = lowerBound; desiredVelocity < currentVelocity; desiredVelocity++) {
@@ -125,16 +124,16 @@ public class AeroPathUtil {
             for (int deltaVelocity = 0; deltaVelocity < currentVelocity - desiredVelocity; deltaVelocity++) {
                 path.addStep(MoveStepType.DEC);
             }
-            
+
             paths.add(path);
         }
-        
+
         // If the unaltered starting path is within acceptable velocity bounds, it's also a valid "acceleration".
         if (startingPath.getFinalVelocity() <= upperBound &&
            startingPath.getFinalVelocity() >= lowerBound) {
             paths.add(startingPath.clone());
         }
-        
+
         // we go from the current velocity to the upper bound and generate paths with the required number of DECs to get to
         // the desired velocity
         for (int desiredVelocity = currentVelocity; desiredVelocity < upperBound; desiredVelocity++) {
@@ -142,13 +141,13 @@ public class AeroPathUtil {
             for (int deltaVelocity = 0; deltaVelocity < upperBound - desiredVelocity; deltaVelocity++) {
                 path.addStep(MoveStepType.ACC);
             }
-            
+
             paths.add(path);
         }
-        
+
         return paths;
     }
-    
+
     /**
      * Helper function to calculate the maximum thrust we should use for a particular aircraft
      * We limit ourselves to the lowest of "safe thrust" and "structural integrity", as anything further is unsafe, meaning it requires a PSR.
@@ -159,7 +158,7 @@ public class AeroPathUtil {
         int maxThrust = Math.min(aero.getCurrentThrust(), aero.getSI());    // we should only thrust up to our SI
         return maxThrust;
     }
-    
+
     /**
      * Given a move path, generate all possible increases and decreases in elevation.
      * @param path The move path to process.
@@ -167,75 +166,100 @@ public class AeroPathUtil {
      */
     public static List<MovePath> generateValidAltitudeChanges(MovePath path) {
         List<MovePath> paths = new ArrayList<>();
-        
+
         // clone path add UP
         // if path uses more MP than entity has available or altitude higher than 10, stop
         for (int altChange = 0; ; altChange++) {
             int altChangeCost = altChange * 2;
-            
+
             // if we are going to attempt to change altitude but won't actually be able to, break out.
             if ((path.getFinalAltitude() + altChange > 10) ||
                     path.getMpUsed() + altChangeCost > path.getEntity().getRunMP()) {
                 break;
             }
-            
+
             MovePath childPath = path.clone();
-            
+
             for (int numSteps = 0; numSteps < altChange; numSteps++) {
                 childPath.addStep(MoveStepType.UP);
             }
-            
+
             if ((childPath.getFinalAltitude() > 10) ||
                     childPath.getMpUsed() > path.getEntity().getRunMP()) {
                 break;
             }
-            
+
             paths.add(childPath);
         }
-        
+
         // clone path add DOWN
         // if the path is already at minimum altitude, skip this
         // if path uses more MP than entity has available or altitude lower than 1, stop
         if (path.getFinalAltitude() > 1) {
             for (int altChange = 1; ; altChange++) {
                 MovePath childPath = path.clone();
-                
+
                 for (int numSteps = 0; numSteps < altChange; numSteps++) {
                     childPath.addStep(MoveStepType.DOWN);
                 }
-                
-                // going down doesn't use MP, but if we drop down more than 2 altitude it causes a massive 
+
+                // going down doesn't use MP, but if we drop down more than 2 altitude it causes a massive
                 // difficulty PSR, which is just not worth it.
                 if ((childPath.getFinalAltitude() < 1) ||
                         childPath.length() > 2) {
                     break;
                 }
-                
+
                 paths.add(childPath);
             }
         }
-        
+
         return paths;
     }
 
     /**
      * Given a move path, generates all possible rotations from it,
-     * without any regard to legality. Mostly because it's intended for spheroid 
+     * without any regard to legality. Mostly because it's intended for spheroid
      * dropships in atmosphere, which can rotate as much as they want.
      */
     public static List<MovePath> generateValidRotations(MovePath path) {
         List<MovePath> childPaths = new ArrayList<>();
-        
+
         for (int x = 1; x < TURNS.size(); x++) {
             MovePath childPath = path.clone();
-            
+
             for (MoveStepType turn : TURNS.get(x)) {
                 childPath.addStep(turn);
             }
-            
+
             childPaths.add(childPath);
         }
-        
+
         return childPaths;
+    }
+
+    public static List<MovePath> generateValidRotation(MovePath path, @Nullable Coords poi) {
+        List<MovePath> childPaths = new ArrayList<>();
+        if (null != poi) {
+            MovePath childPath = path.clone();
+            Coords curCoords = childPath.getFinalCoords();
+
+            int dir = curCoords.direction(poi);
+
+            // Only turn if not facing correct direction
+            if (dir > 0) {
+                for (MoveStepType turn : TURNS.get(dir)) {
+                    childPath.addStep(turn);
+                }
+                childPaths.add(childPath);
+            }
+        }
+
+        return childPaths;
+
+    }
+
+    public static Coords getSpheroidPOI(Game game, Entity mover) {
+        return game.getBoard().getCenter();
     }
 }
--- a/megamek/src/megamek/client/bot/princess/AeroPathUtil.java
+++ b/megamek/src/megamek/client/bot/princess/AeroPathUtil.java
@@ -241,12 +241,17 @@ public class AeroPathUtil {
     }
 
     public static int getSpheroidDir(Game game, Entity mover) {
-        final StringBuilder msg = new StringBuilder("Deciding where to point ")
-                .append(mover.getDisplayName()).append("...");
+        boolean debug = LogManager.getLogger().isDebugEnabled();
+        final StringBuilder msg = (debug)
+                ? new StringBuilder("Deciding where to point ")
+                .append(mover.getDisplayName()).append("...")
+                : null;
 
         // Face the center of the board
         int dir = mover.getPosition().direction(game.getBoard().getCenter());
-        msg.append("\nMap center is to the ").append(ClientCommand.getDirection(dir));
+        if (debug) {
+            msg.append("\nMap center is to the ").append(ClientCommand.getDirection(dir));
+        }
 
         int enemyDir = dir;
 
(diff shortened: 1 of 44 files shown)


megamek/src/megamek/common/Entity.java
```diff
@@ -4202,6 +4202,13 @@
         return ammoList;
     }
 
+    public ArrayList<Mounted> getAmmo(Mounted weapon) {
+        return (ArrayList<Mounted>) ammoList.stream()
+                .filter(
+                    ammo -> AmmoType.isAmmoValid(ammo, (WeaponType) weapon.getType())
+                )
+                .collect(Collectors.toList());
```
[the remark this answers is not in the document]
Sleet01: This is, in fact, what I meant to do.

Thom293:
Legendary!

megamek/src/megamek/common/Entity.java
```diff
@@ -4202,6 +4202,13 @@
...
+    }
```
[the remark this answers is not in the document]
Sleet01: This is a long-standing issue that I don't feel needs to be resolved now.

megamek/src/megamek/common/weapons/ArtilleryWeaponIndirectFireHandler.java
```diff
@@ -69,7 +69,7 @@
                 r.indent();
                 r.newlines = 0;
                 r.subject = subjectId;
-                r.add(wtype.getName());
+                r.add(wtype.getName() + " (" + atype.getShortName() + ")");
```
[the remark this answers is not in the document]
Sleet01: atype should never be `null` for ArtilleryWeaponIndirectFireHandlers now.

Sleet01:
I'm going to continue running stress tests and trying to patch major exceptions where I can, but the actual ammo selection code has been working well and should be ready to go.

kuronekochomusuke:
#4827

Sleet01:
> #4827

Yeah, I think I finally fixed it.  Turns out we had a ton of "do artillery to X" functions that didn't actually check if they were on-board or not.

Sleet01:
I think I've patched enough stuff to do end-to-end runs on new and old code for speed comparisons.  @Thom293 is going to run tests on (from what I recall) their monster rig as well.

Thom293:
First, thank you very much for tackling this.  

Ok - a few notes after testing.   I havent looked at the code and probably would not understand it if I did.  These are just in-game observations:
Tested with Sleets mul, Dropport map, Bot v Bot, 2 identical teams, fixed to 15 rounds.  Timed them just to get a baseline on how much processing time is increased.  

1.  Pre-ammo changes, 2 games were 9 minutes each.
2. Post-ammo changes, 2 games were 17 minutes each.  Just based on visual observation, the ASF move calculation phase, the Rifleman with 4 LAC-5s and alternate ammos, and the Vulture with ADA and Arty Ammos had a very noticeable lag before each of them moved.  The Rifleman took a good 10 seconds to calculate each move.  The vulture 4-5.  Just based on personal feel, although the delay was noticeable, it wasnt long enough to be annoying or detrimental to the experience, IMHO.

Functional notes after reviewing combat logs of 6 games with ammo changes active:
1.  She switches ammo very often.  It isnt perfect, but even if just shipped as-is its a huge improvement.  Great job.
2. She does not seem to swith to cluster ammo at long range for the bonus to hit with LBX.  For instance, she was shooting at 11s with solid ammo instead of using cluster.  She can and does switch LBX ammo back and forth at close range, as needed.
3. The Torrent Heavy Bombers had 6 AAA missiles, a rocket pod, a TAG pod, and some bombs.  one of the Kirgiz-D had 2 AAA missiles.  There were also active Pentagon Dropships in the Air.  Despite a target rich battlefield, I didnt see any AAA missiles fired at all.  Example:  Turn 1.  Torrent closes to MG range on Kirgiz-D and fires MGs and Rocket Pod.  Turn 2, fires TAG.  Turn 3, closes to MG Range on Kirgiz D and fires MGs.  She never fired her missies from either Torrent or the Kirgiz.  These should have been fired first round at hostile air.
4. Everyone Hates Lacie Woo.  For some reason, the Paladins and Arrow Vultures would waste entire arty salvos on a single ejected pilot.  Poor Lacie was killed in two different games with an entire arty salvo directed just at her - and there were many many other much more high valure targets on the board.  So perhaps the "kill the pilots" priority needs to be looked at.  If it matters, my AI was on Striker behavior.  
5. Smoke ammo - not a bug, But wanted to put here for future refrence.  This was discussed and just quoting Sleet so I dont get it wrong:  Yeah, it would be "easy" to record all of the prospective path endpoints as "smokable" hexes, and then target those with smoke if no other target affords itself.  We do something similar with actions right now; Princess's last resort is looking for a murder club laying around ^_^

Thank you very much for these changes.  Seems such a huge improvment so far. 

Here is the .mul and save if anyone needs it.  
[AmmoMulSave.zip](https://github.com/MegaMek/megamek/files/14217608/AmmoMulSave.zip)

megamek/src/megamek/common/actions/WeaponAttackAction.java
```diff
@@ -4340,7 +4347,7 @@
      */
     private static ToHitData compileTerrainAndLosToHitMods(Game game, Entity ae, Targetable target, int ttype, int aElev, int tElev,
                 int targEl, int distance, LosEffects los, ToHitData toHit, ToHitData losMods, int toSubtract, int eistatus,
-                WeaponType wtype, Mounted weapon, int weaponId, AmmoType atype, EnumSet<AmmoType.Munitions> munition, boolean isAttackerInfantry,
+                WeaponType wtype, Mounted weapon, int weaponId, AmmoType atype, Mounted ammo, EnumSet<AmmoType.Munitions> munition, boolean isAttackerInfantry,
```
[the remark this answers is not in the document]
Sleet01: This is not getting refactored, sorry.

Thom293:
Edited my notes to correct my mistake about missiles and add the Dropship info.

megamek/src/megamek/client/bot/princess/AeroPathUtil.java
```diff
@@ -159,83 +159,132 @@
...
         }
-        
+
         return childPaths;
     }
+
+    public static List<MovePath> generateValidRotation(MovePath path, int dir) {
+        List<MovePath> childPaths = new ArrayList<>();
+        MovePath childPath = path.clone();
+        Coords curCoords = childPath.getFinalCoords();
```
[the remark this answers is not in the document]
Sleet01: Still cleaning this up for Spheroid Dropship protective turns; will be cleaned up tonight.

megamek/src/megamek/client/bot/princess/AeroPathUtil.java
```diff
@@ -159,83 +159,132 @@
...
+
+    public static int getSpheroidDir(Game game, Entity mover) {
+        // Face the center of the board
+        int dir = mover.getPosition().direction(game.getBoard().getCenter());
+
+        // Then determine if we need to protect part of the ship
+        if (mover.getDamageLevel() != Entity.DMG_NONE) {
+            // TODO: finish this
+            int leastArmor = 9999999;
+            int leastLoc = Dropship.LOC_NONE;
```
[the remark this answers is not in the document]
Sleet01: Still cleaning this up for Spheroid Dropship protective turns; will be cleaned up tonight.

megamek/src/megamek/common/pathfinder/SpheroidPathFinder.java
```diff
@@ -35,12 +37,31 @@
...
+    }
+
+    private SpheroidPathFinder(Game game, int direction) {
         this.game = game;
+        this.direction = direction;
+    }
+
+    /**
+     * We want to be able to set a point that the entity should turn to face
+     * @param target
```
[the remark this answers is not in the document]
Sleet01: Still cleaning this up for Spheroid Dropship protective turns; will be cleaned up tonight.

megamek/src/megamek/client/bot/princess/FireControl.java
```diff
@@ -2700,10 +2777,11 @@ Mounted getPreferredAmmo(final Entity shooter,
                 return validAmmo.get(0);
             }
 
+            // Todo: delete
```
Sleet01, line 2780: The existing getAtmAmmo code was interfering with toHit-based ammo iteration and selection, so I've disabled it here for testing and will most likely remove it.  Same with the tests for this function.

Going forward we should be able to rely more on toHitData and weaponFireInfo for determining missile ammo to use, based on actual calculated expected damage and to-hit chances.

megamek/src/megamek/client/bot/princess/FireControl.java
```diff
@@ -3186,16 +3264,17 @@ Mounted getAntiAirAmmo(final List<Mounted> ammoList,
             }
         }
 
+        // Todo: delete
```
Sleet01, line 3267: Ditto.

megamek/src/megamek/client/bot/princess/Princess.java
```diff
@@ -767,8 +767,13 @@ private Map<Mounted, Double> calcAmmoConservation(final Entity shooter) {
                     ammoCount += ammoCounts.get(ammoType);
                 }
                 msg.append(" has ").append(ammoCount).append(" shots left");
+                // Desired behavior:
```
Sleet01, line 770: This can be tuned further by adding terms, but this should give us ammo weapon to-hit thresholds of:

- >= 12% (hit on 11+) when ammo is abundant; default Princess will take iffy shots but not if they require 12s.
- >= ~20% (hit on 9+) when ammo drops to 3 shots total; default Princess will try to close or hold fire at range.
- >= ~50% (hit on 7+) when ammo is down to 1 shot left.

Raising the aggroFactor (aggressiveness slider in Bot settings) or increasing ammo per weapon type lowers the computed thresholds.  Energy weapons are not affected.

megamek/unittests/megamek/client/bot/princess/FireControlTest.java
```diff
@@ -920,7 +920,9 @@ public void testGetPreferredAmmo() {
         when(mockWeaponMML5.getType()).thenReturn(mockMML5);
 
 
+        // Todo: delete
```
Sleet01, line 923: See above re: ATM code.

megamek/src/megamek/client/bot/princess/AeroPathUtil.java
```diff
@@ -238,28 +239,52 @@ public static List<MovePath> generateValidRotations(MovePath path) {
...
+
+        } else {
+            // Get all enemies, find centroid, face that.
+            final Coords centroid;
+            ArrayList<Entity> enemies = new ArrayList<>();
+            Iterator<Entity> eIt = game.getAllEnemyEntities(mover);
+            while (eIt.hasNext()) {
+                enemies.add(eIt.next());
+            }
+            centroid = PathRanker.calcAllyCenter(mover.getId(), enemies, game);
```
Sleet01, line 283: Oops, this should be enemies[0].getId(), will update.

megamek/src/megamek/common/Compute.java
```diff
@@ -1859,6 +1858,93 @@ public static Entity exposed_findC3Spotter(Game game, Entity attacker,
...
 
+    /**
+     * Function that attempts to find one TAG spotter, or the best TAG spotter,
+     * @param game
+     * @param attacker should be an artillery unit
+     * @param target (if null, do not return a spotter!)
+     * @param stopAtFirst if we only want to know that there is one, not which is best
+     * @return The Spotter.
+     */
+    public static Entity findTAGSpotter(Game game, Entity attacker, Targetable target, boolean stopAtFirst) {
```
NickAragua, line 1869: When you're calculating movement or firing phase, TAG attacks for the round have already been declared and processed, so I don't really see why there's a need for this expensive calculation. You should be able to go to the "game" object and look through the list of active TAG attacks instead.

megamek/src/megamek/client/bot/princess/WeaponFireInfo.java
```diff
@@ -434,13 +459,36 @@ private WeaponAttackAction buildBombAttackAction(final HashMap<String, int[]> bo
...
+            msg.append("\nThe unit will be giving up ")
+                    .append(myWeaponsDamage)
+                    .append(" damage from other weapons");
+        }
 
         // Get a list of incoming or potentially incoming guidable weapons from the relevant Princess and compute damage.
         int incomingAttacksDamage = Compute.computeTotalDamage(owner.computeGuidedWeapons(shooter, target.getPosition()));
+        int utility = incomingAttacksDamage - myWeaponsDamage;
+        msg.append("\n\tUtility: ").append(utility).append(" damage (Max, estimated)");
+        LogManager.getLogger().debug(msg.toString());
```
NickAragua, line 489: Let's make this conditional or eliminate it, as .debug is a very expensive call relatively speaking

megamek/src/megamek/client/bot/princess/FireControl.java
```diff
@@ -1538,16 +1615,38 @@ FiringPlan guessFullFiringPlan(final Entity shooter,
...
+
+            // Energy / ammo-independent weapons
+            if (AmmoType.T_NA == wtype.getAmmoType()) {
+                bestShoot = buildWeaponFireInfo(shooter, target, weapon, null, game, false);
+            } else {
+                ArrayList<Mounted> ammos;
+                ammos = shooter.getAmmo(weapon);
+
+                WeaponFireInfo shoot;
+                for (Mounted ammo: ammos) {
```
NickAragua, line 1630: This looks like it could get expensive. Since we're not picking ammo type during the movement phase (and these "guess" methods are intended to be a cheap way to evaluate expected firepower for a move), is there really an ammo type that would heavily influence what move we make? I suppose for stuff like MML launchers it might make a difference, but in most cases it won't. Do we want to spend some time cutting down unnecessary computations here even at the expense of accuracy?

The reason I ask is that move path processing is very performance-sensitive, so any additional load here is multiplied many-fold, especially for faster units with more MP/jump MP.

NickAragua (changes requested):
A few thoughts regarding performance, which is absolutely critical when modifying functions that evaluate move paths for the bot.

SJuliez (approved):

HammerGS:
Changes look complete and Juliez added approval.

PR #5140, opened 2024-02-08
Ba tech advancement

neoancient:
Adds tech advancement data for BA exoskeleton, which was invented earlier than the PA(L) suit.

Partial fix for MegaMek/megameklab#1203


HammerGS (commented):
Looks good

PR #5165, opened 2024-02-14
Reduce connections between Player/Team and Game/Entity

SJuliez:
This is mostly a refactor to reduce connection between Player/Team and the TW classes Entity/Game to facilitate other game types (in some distant future). 
- Some methods were not used and were deleted. 
- Some methods have been moved into a ServerHelper class as they were only used by GameManager.
- Player is changed to use IGame instead of Game; for the BV sum I added countForStrengthSum() to InGameObject, see its comment

One change is worth noting (and I hope I'm not making an error): I had to remove the firePlayerChangeEvent from Player.setDone() and move it to the callers. Most calls were by the GameManager. I don't think there are any game-listeners server-side, nor can I imagine a good use for them there and so I did not add the event fire there. Also, I removed a bit of code that re-added listeners to a new Game that is set in the GameManager, the reason being they're transient. I could not think of a situation where old game listeners should be re-added to a new game server-side (set by a new scenario or game load or new game start; the game object is never replaced while the game's running). Nor are there any game listeners present in the test game I started. MekHQ adds a game listener only to the client's game.

### DIFF
--- a/megamek/src/megamek/client/Client.java
+++ b/megamek/src/megamek/client/Client.java
@@ -1432,6 +1432,7 @@ public class Client implements IClientCommandHandler {
 
                     if (player != null) {
                         player.setDone(c.getBooleanValue(1));
+                        game.processGameEvent(new GamePlayerChangeEvent(player, player));
                     }
                     break;
                 case PRINCESS_SETTINGS:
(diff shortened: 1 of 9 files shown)


SJuliez:
Test fails. Will need to update.

PR #3500, opened 2022-03-08
3498: Adding Clanner Tracking to MegaMek

Windchild292:
This handles the bug raised in #3498, namely that any melee attacks are based on the 'Mech's tech base and not the pilot / commander's status.

PR #5166, opened 2024-02-14
Adding Clan Personnel Tracking to MegaMek

### LINKED ISSUE #3498: Dezgra-Mode
Inner Sphere piloting Clan Mechs should be able to perform physical attacks. Currently the behavior is dictated by the mech type and should be depending on the pilot.

This needs an additional flag for each pilot and could be handled via the "Method for Rolling Pilot Skills": 
if  Pilot Type == Clan AND rule "No Physical Attacks for Clan" AND Dezgra-Flag for Pilot not set ==> that mech pilot skips physical attacks

This flag should also be maintained in MHQ. 

In addition, this could be a punishment in MHQ, like Imprison (and it should be reversable, but logged in personal log)

Discussion: 2022-03-07, discord channel "megamek" https://discord.com/channels/458705327911731231/458706493659807801/950349175772549120

SJuliez:
This is a replacement for #3500 
This is Windchild's work, I just replaced references to clanner with clan pilot
Hopefully Fixes #3498
Accompanied by a MHQ PR.


neoancient (approved):
It looks ok, but there is a conflict with the code that added the ability to set pilot hits.

PR #5169, opened 2024-02-15
Code cleanup

SJuliez:
This is only a cleanup, no intended other changes. Removes unused methods and methods that are identical their super (as determined by the IDE in a combined MHQ/MML/MM setting), unnecessary/duplicated comments, a duplicate method to determine if a fighter is in a squadron; also makes Aero abstract as it no longer represents ASF and should not be instantiated.


Sleet01 (approved):
LGTM!

PR #5170, opened 2024-02-15
Use IGame in TurnOrdered

SJuliez:
Replaces Game with IGame in and around TurnOrdered where possible and moves a method up to IGame.


Sleet01 (approved):
LGTM!

PR #5172, opened 2024-02-15
make emi description constants public for mekhq

kuronekochomusuke:
- make emi description constants public for mekhq

need for https://github.com/MegaMek/mekhq/pull/3834


Sleet01:
I don't see any issues, but I wonder if there's a better place for public strings.  That, or keep the strings private and make a public getter?

SJuliez (approved):

Issue #1429, opened 2019-08-04
Bug Unable to move on map edges Intermittent 


MarcJC:
Playing on the new Stable 46.0  

A bug occurs sometimes when a unit (seen it happen with Mech and Areo units)  It does not always occur, and often will fix itself after a unit has spent a turn stuck.  (This behavior may have something to do with the unit turning)

As luck would have it the auto save from a crash happens to have the bug on it.  

Not sure what is causing the error so this is gonna be a bit long to give the exact steps to make sure it comes up.  I would bet most of the steps are irrelevant and you should try the short version first.

Short Version the Jenner IIC on the north edge is stuck.   It can turn or back up but attempting to walk normally or jump refuses to do anything.  

Long version.  Load the attached HQ file (custom file included if needed) Load the autosave from the chase mission with 8 units (4 mech 4 areo)  Replace player to get rid of ghost.  Set retreat edge to south.  

Deploy the planes to hex 0358 facing North, and hex 1828 facing South West.  
Move the Archer directly North.  
The Jenner should come up and be stuck.  

Let me know if you need any extra help detecting the issue.  

I should mention that I had a crash earlier while playing due to the MegaMek window staying on top of the other windows, refusing to respect ALT-TAB, and having to be minimized to get to other windows.  When I declared victory this turn the game froze when trying to load the HQ windows to determine the results to send back to HQ.  I think the windows were stuck under the MegaMek window and could not get focus.  I could not pull up any other windows or minimize MegaMek and could not close the window in any way.  All attempts to close the window were ignored.  
I ended up having to reboot to fix.  On the reload of the saves the window is acting normally.  
I would chalk this up to a weird JAVA issue and not worry too much about it, but if it becomes replicateable or others notice the issue, we will make sure to let you know.  

[autosave.sav.zip](https://github.com/MegaMek/megamek/files/3464557/autosave.sav.zip)

neoancient:
Note that stable 0.46.0 has not been released. This is 0.46.0rc.1, which is the first candidate for the new stable.
I am able to reproduce this reliably in the stable branch but it does not occur in the master branch. I have traced it to a missed commit that left debugging code in place.

MarcJC:
Good catch then.  
You all do great work.

PR #5176, opened 2024-02-18
Fix MML #1429: MegaMek side of fix for MML NPE in Aero Equipment tab

Sleet01:
This is the MM fix for MegaMek/megameklab#1429; what I thought was a guard check against weapon being `null` was actually a form of overloading that's used exclusively by the megameklab UI to select AV range bands when displaying equipment.

I've created a bare getter that returns the value that UI builder needs, and added a check to call it if the weapon passed to `wtype.getRanges()` is null.

Testing:
- Re-ran existing MM unit tests.
- Tested all Aerospace subclasses in MML

I'd like to add a unit test to MegaMekLab to make sure that this doesn't regress but I need some help on that, so for now this should be sufficient.

Close MegaMek/megameklab#1429

### DIFF
--- a/megamek/src/megamek/common/WeaponType.java
+++ b/megamek/src/megamek/common/WeaponType.java
@@ -608,28 +608,33 @@ public class WeaponType extends EquipmentType {
     }
 
     public int getMaxRange(Mounted weapon) {
+        if (weapon == null) {
+            return getMaxRange();
+        }
         return getMaxRange(weapon, weapon.getLinked());
     }
 
+    public int getMaxRange() {
+        return maxRange;
+    }
+
     public int getMaxRange(Mounted weapon, Mounted ammo) {
-        if (null != weapon) {
-            if (getAmmoType() == AmmoType.T_ATM) {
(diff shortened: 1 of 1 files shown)


SJuliez (approved):
There is also the possibility that weapon.getlinked() is null. I assume the method isn't called in that case.

Sleet01:
> There is also the possibility that weapon.getlinked() is null. I assume the method isn't called in that case.

In that case the base maxRange() is returned.

SJuliez (approved):

metalconscript:
I have the NPE but I am unsure how to apply the fix for the dropship/small craft equipment tab. Please advise, thank you.

Sleet01:
> I have the NPE but I am unsure how to apply the fix for the dropship/small craft equipment tab. Please advise, thank you.

The fix is checked in to the development branch, so your options as of right now are:
1. download the latest nightly build and play on the bleeding edge of alpha code,
2. wait until 0.49.19 is released.

If you decide to go with option 1, please come over to the MegaMek discord so we can help get you up and running; it's not quite as easy as playing with the current release version (from what I recall).

PR #5186, opened 2024-02-20
SV armor weight and space calculations

neoancient:
Fixes two issues with SV armor:
1. I left out the multiplier that accounts for the number of points when calculating the armor weight, resulting in the weight always being calculated for a single point.
2. The svslots field in EquipmentType is initialized to -1, which is supposed to indicate that it uses the mech slots value. ArmorType sets all the values explicitly when they are not zero, so I added initialization to zero in the constructor.

Fixes MegaMek/megameklab#1430.


Sleet01 (approved):
Looks good to me!

SJuliez (approved):

PR #5188, opened 2024-02-21
Unit shadow

SJuliez:
This adds a programmatic drop shadow to icons when not airborne. This is meant to replace the static shadow we have around many icons. The drop shadow reflects a unit's height and is oriented like the terrain shadows.

image

### DIFF
--- a/megamek/src/megamek/client/ui/swing/tileset/EntityImage.java
+++ b/megamek/src/megamek/client/ui/swing/tileset/EntityImage.java
@@ -72,8 +72,16 @@ public class EntityImage {
     private static final int IMG_HEIGHT = HexTileset.HEX_H;
     private static final int IMG_SIZE = IMG_WIDTH * IMG_HEIGHT;
 
+    private static final float SHADOW_INTENSITY = 0.7f; // 0 = no shadow, 1 = black shadow
+    private static final float SHADOW_OFFSET = 5f; // due to unit height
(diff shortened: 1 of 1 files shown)


HammerGS:
Should I be planning to switch the sprites back to the original unshaded versions?

SJuliez:
ground units yes.

HammerGS:
Another thought on this, I know in the past I've seen comments that players don't like shadow effects. With the move back to original sprites, should this be made an option? Defaulted to on.

SJuliez:
Yes I can add that. I could make the unit shadows depend on the terrain shadow setting to avoid yet another checkbox

HammerGS:
> Yes I can add that. I could make the unit shadows depend on the terrain shadow setting to avoid yet another checkbox

I think that would be a great idea, but I would set the option to on by default.

NickAragua (approved):
Looks fine. Do we want to have this as an option for people using potato computers? Some of the computations in EntityImage look a little gnarly and may be not great for performance?

edit: lol looks like Hammer had a similar idea.

SJuliez:
Yes, it's now tied to the terrain shadows. Actually, the icons are cached and during a game (after the lobby) I found it difficult to get them to do the EntityImage thing at all. But in the lobby, they constantly redraw so I guess, better that way.

Thom293:
Hey -  I just used this for the first time.  Looks really damn good.  Thank you.

PR #5196, opened 2024-02-23
Update BA armor weight calculations for MML 1439 fix

Sleet01:
This fixes some armor math for BAs.

Testing:
- Ran all unit tests (3 failing in MekHQ, unrelated)
- Confirmed various suits came within <10% of TRO weights (some equipment like Modular Weapon Mounts are not implemented)

Close MegaMek/megameklab#1439

### DIFF
--- a/megamek/src/megamek/common/verifier/TestBattleArmor.java
+++ b/megamek/src/megamek/common/verifier/TestBattleArmor.java
@@ -36,7 +36,7 @@ public class TestBattleArmor extends TestEntity {
      * Keeps track of the number of free MP a Quad BA gets.
      */
     public static int QUAD_FREE_MP = 2;
-    
+
     /**
      * Base chassis weight by suit weight class for Inner Sphere suits
      */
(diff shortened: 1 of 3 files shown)


megamek/src/megamek/common/verifier/TestBattleArmor.java
```diff
@@ -478,9 +478,13 @@ public double getWeightStructure() {
...
     }
 
+    public double getWeightArmor(int trooper) {
+        return ArmorType.forEntity(ba).getWeightPerPoint() * ba.getOArmor(trooper);
+    }
+
     @Override
     public double getWeightArmor() {
-        return ArmorType.forEntity(ba).getWeightPerPoint();
+        return getWeightArmor(1);
```
NickAragua, line 487: Did you mean to use a magic number here, or do we want to use "firstArmorIndex"?

Sleet01: Looks like that is what I intended -_-;

neoancient (approved):
It works as is, but in my opinion it would be cleaner to override BattleArmor::getArmorWeight. I'll leave that to your judgment. This is probably also a problem for protomechs.

PR #5203, opened 2024-02-24
Planetary Conditions Enums

kuronekochomusuke:
https://github.com/MegaMek/mekhq/pull/3834 requires this to be merged

convert the int constants to enums 
- light enum
- weather enum
- wind enum
- wind direction enum
- atmosphere enum
- fog enum
- blowing sand enum
- emi enum


megamek/src/megamek/common/Compute.java
```diff
@@ -7391,8 +7388,9 @@ public static boolean useSpheroidAtmosphere(Game game, Entity en) {
             return false;
         }
         // aerodyne's will operate like spheroids in vacuum
+        PlanetaryConditions conditions = game.getPlanetaryConditions();
         if (!((IAero) en).isSpheroid()
-                && !game.getPlanetaryConditions().isVacuum()) {
+                && !conditions.getAtmosphere().isLighterThan(Atmosphere.THIN)) {
```
SJuliez, line 7393: Trace is lighter than thin. Is this intended?

megamek/src/megamek/common/Dropship.java
```diff
@@ -750,6 +753,9 @@ public boolean canLandVertically() {
      */
     @Override
     public boolean canTakeOffVertically() {
-        return (isSpheroid() || game.getPlanetaryConditions().isVacuum()) && (getCurrentThrust() > 2);
+        PlanetaryConditions conditions = game.getPlanetaryConditions();
+        boolean spheroidOrLessThanThin = isSpheroid()
+                || conditions.getAtmosphere().isLighterThan(Atmosphere.THIN);
```
SJuliez, line 758: Trace is lighter than thin but not vacuum. Is this intended?

megamek/i18n/megamek/common/messages.properties
```diff
@@ -228,10 +228,10 @@ PlanetaryConditions.DisplayableName.Weather.Moderate\ Snowfall=Moderate Snowfall
 PlanetaryConditions.DisplayableName.Weather.Heavy\ Snowfall=Heavy Snowfall
 PlanetaryConditions.DisplayableName.Weather.Sleet=Sleet
 PlanetaryConditions.DisplayableName.Weather.Snow\ Flurries=Snow Flurries
-PlanetaryConditions.DisplayableName.Weather.Blizzard=Blizzard
 PlanetaryConditions.DisplayableName.Weather.Ice\ Storm=Ice Storm
 PlanetaryConditions.DisplayableName.Weather.Light\ Hail=Light Hail
 PlanetaryConditions.DisplayableName.Weather.Heavy\ Hail=Heavy Hail
+PlanetaryConditions.DisplayableName.Weather.Lightning\ Storm=Lightning Storm
```
SJuliez, line 234: I suggest not using spaces in the name of  the resource if that is not too much effort

SJuliez: I see that it is a legacy issue, just a suggestion

megamek/src/megamek/common/PlanetaryConditions.java
```diff
@@ -273,80 +93,123 @@ public Object clone() {
...
 
-    public String getAtmosphereDisplayableName() {
-        return getAtmosphereDisplayableName(atmosphere);
+    public void setFog(Fog fog) {
+        this.fog = fog;
     }
 
-    public String getFogDisplayableName() {
-        return getFogDisplayableName(fog);
+    public BlowingSand getBlowingSand() {
```
SJuliez, line 187: I suggest adding isBlowingSand() to avoid the verbose getBlowingSand().isBlowingSandNone()

SJuliez: I saw that used somewhere but now I can't find it anymore

SJuliez (changes requested):
I have a few comments. Also, I suggest placing features in a common package, i.e. to use a package called planetaryconditions in common and having PlanetaryConditions.java and the enums in there. I know that we have all manner of groupings but this seems a sensible way to do it with the (possible) advantage of allowing a low visibility of classes

megamek/src/megamek/common/Compute.java
kuronekochomusuke: The previous function had a confusing name 

    public boolean isVacuum() {
        return (atmosphere == ATMO_VACUUM) || (atmosphere == ATMO_TRACE);
    }

megamek/src/megamek/common/Dropship.java
kuronekochomusuke: see above

SJuliez (approved):

SJuliez:
The CI fails at cache rebuilding. Did you give this a test? (delete the unit cache and rebuild). Might be the test itself has a problem

kuronekochomusuke:
there was an issue with one of the units that has been fixed a few days ago.

PR #5208, opened 2024-02-26
Close loopholes that allow off-board arty to attempt Flak attacks at …

Sleet01:
…on-board Aeros.

Noticed this while testing another fix with a pair of large combined-arms forces run by Princess: off-board artillery units firing munitions that get the Flak bonus were targeting Aerospace units during the Indirect phase.

This is wrong for two reasons (see TacOps: AR):

1. Flak attacks are direct fire, not indirect.
2. An artillery unit has to be on the same map or board as the target Aerospace unit to use direct Flak attacks.

This closes two avenues for targeting Aerospace with Indirect attacks from off-board (but does not deter indirect attacks at, say, airborne VTOLs).

Testing:
- Tested with manual and bot-controlled off-board artillery units targeting on-board ASFs, Dropships, and VTOLs.

### DIFF
--- a/megamek/src/megamek/common/Compute.java
+++ b/megamek/src/megamek/common/Compute.java
@@ -7412,7 +7412,10 @@ public class Compute {
     }
 
     public static boolean isFlakAttack(Entity attacker, Entity target) {
-        boolean validLocation = !(attacker.isSpaceborne() || target.isSpaceborne());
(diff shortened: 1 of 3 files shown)


NickAragua (approved):

Sleet01:
Apologies, ran into a couple of other wrinkles when following up with the Aerospace player community.

PR #5226, opened 2024-03-07
Fix for infantry skipping non-ammo-fed infantry weapons during attack phase

### LINKED ISSUE #5224: [49.18 Nightly] - Infantry Firing issue
This was discussed in discord.  Very quick bug report.  Infantry (at least tested with Horse Cavalry AR and Laser versions) no longer fire their weapons.  They will spot and do leg and swarm attacks, but not shoot their guns.

Sort of related:  4 of them were spotting one unit.  Seems like there should be a "has been spotted" flag so multiple units dont waste their spot on a single target.  
Screens showing them not firing, MHQ Save, Customs,

EDIT:  Just to remind myself, I also need to post a UI and Turret not firing issue later.

image
image

[Crazed Profanity30690614INFBUG.cpnx.gz](https://github.com/MegaMek/megamek/files/14516620/Crazed.Profanity30690614INFBUG.cpnx.gz)
[Mine.zip](https://github.com/MegaMek/megamek/files/14516634/Mine.zip)

Sleet01:
This was caused by the assumption that Infantry weapons would use AmmoType `T_NA`, like energy weapons.
Instead they use `T_INFANTRY`, because of course.

Fix is to check for both types when skipping ammo selection in the attack planning portion of the Attack phase.  The guess code, used during movement planning, appears to be working correctly.

Testing:
- Let Princess annihilate some 'mechs with beast-mounted infantry.

Close #5224

### DIFF
--- a/megamek/src/megamek/client/bot/princess/FireControl.java
+++ b/megamek/src/megamek/client/bot/princess/FireControl.java
@@ -1202,7 +1202,7 @@ public class FireControl {
                 String shootingCheck;
 
                 // Energy / ammo-independent weapons
-                if (Arrays.asList(AmmoType.T_NA, AmmoType.T_INFANTRY).contains(wtype.getAmmoType())) {
+                if (effectivelyAmmoless(wtype)) {
                     shootingCheck = checkGuess(shooter, enemy, weapon, null, game);
                     if (null != shootingCheck) {
                         ret.append(shootingCheck);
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/client/bot/princess/FireControl.java
```diff
@@ -1202,7 +1202,7 @@ private String checkGuess(final Entity shooter,
                 String shootingCheck;
 
                 // Energy / ammo-independent weapons
-                if (AmmoType.T_NA == wtype.getAmmoType()) {
+                if (Arrays.asList(AmmoType.T_NA, AmmoType.T_INFANTRY).contains(wtype.getAmmoType())) {
```
NickAragua, line 1205: Should we split this logic into a function for a little more readability?

NickAragua (commented):
One suggestion, looks fine otherwise.

megamek/src/megamek/client/bot/princess/FireControl.java
Sleet01: Couldn't hurt, and we can make future updates less painful.  One sec...

NickAragua (approved):

PR #5230, opened 2024-03-07
Check for superheavy when determining IS crits for Clan structure

neoancient:
The reduced crit slot requirement for superheavies is only being applied to IS structure. Mixed tech superheavies can use Clan structure, so it needs to be taken into account. I also included prototype endo steel, because somebody's going to do it.

I removed the `entity instanceof Mech` check because these equipment types can only be used on mechs.

Fixes MegaMek/megameklab#1303


NickAragua (approved):

HammerGS:
Update to this.

https://bg.battletech.com/forums/index.php?topic=70587

The clan types are illegal.

PR #5240, opened 2024-03-11
Fix MML 1365: Sponson Turrets not being included in TechLevel calcs

Sleet01:
This is an MM fix for an MML  bug that affected MHQ, but leaving that aside...

There were several issues here:

1. The entry for Sponson Turrets was out of date, and missing more recent info from the latest TRO/source erratas.
2. Tank BLK file loading does its final tech level calculations _prior_ to loading any equipment.  There may be another refresh in Entity or Tank after all weapons are added, but _only_ for weapons.
3. Sponson turrets (and pintle mounts) were not ever considered for calculating TechLevel in the first place.

This PR fixes those issues, at least for Sponson Turrets.  Pintle mounts are just always available from PS onward, so should not need composite Tech Advancement handling.

NOTE: if, in the future, a piece of equipment is found that should - but does not - affect the Tech Level or availability eras, check all the *TechAdvancement() functions and _ensure_ that it is actually handled!  There is probably a lot of equipment that should be, but isn't.

Testing:

1. Verified that ~3012-to-3025-era vehicles mounting all-Standard equipment with Sponson Turrets appear under various tech and era limitations that should allow them (per the original bug), both in MM and MHQ.
2. Ran all unit tests for all three projects.
3. Sampled existing canon vehicles with Sponson Turrets to confirm that the updates have not changed their availability.

Close MegaMek/megameklab#1365

### DIFF
--- a/megamek/src/megamek/common/MiscType.java
+++ b/megamek/src/megamek/common/MiscType.java
@@ -8617,8 +8617,8 @@ public class MiscType extends EquipmentType {
         //Tech Progression tweaked to combine IntOps with TRO Prototypes/3145 NTNU RS
         misc.techAdvancement.setTechBase(TECH_BASE_ALL)
                 .setTechRating(RATING_B).setAvailability(RATING_F, RATING_F, RATING_F, RATING_D)
-                .setAdvancement(DATE_PS, DATE_NONE, 3079, DATE_NONE, DATE_NONE)
-                .setApproximate(false, false, true, false, false)
+                .setAdvancement(DATE_PS, 3079, 3080, DATE_NONE, DATE_NONE)
+                .setApproximate(false, true, false, false, false)
                 .setStaticTechLevel(SimpleTechLevel.STANDARD);
         return misc;
     }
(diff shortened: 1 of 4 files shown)


megamek/src/megamek/common/MiscType.java
HammerGS, line 1: Please leave the Tech Progression. You’ll notice a note about this having CGL unpublished errata.

Sleet01: @HammerGS Actually, reverting the advancement and approximate settings breaks the summary output:
With code changes and updated advancement matching IntOps 2016 pdf (this does not appear to have been updated anywhere else):
image


With code changes but current advancement:
image

With the code changes, both seem to provide the correct _behavior_ (Sponson Turrets available as experimental in years prior to 3074, Advanced to 3079, Common from 3080 onward).
If the current Advancement values are correct, there's another bug in the Summary code somewhere.

HammerGS: Part of the issue and makes things very complicated around Tech Progression. Is the Tables in Int Ops don't take into account the tables in TRO:Prototypes and RS:NTNU which CGL considers the correct progression.

So to keep with the official record sheets inline with those books, the RS Team for CGL had to combine and blend with the Int Ops tables (which are very problematic). 

Makes the correct advancement any unit with Sponson should be Experimental from PS to around 3074 to 3079 should be Experimental ONLY. Then after 3079 should become Common. Basically there is no Advanced period for them they go straight from Experimental to Common.

We have a lot of weapons and equipment that suffer from blended data. CGL has access to the changes but it's their call when and if they drop the errata for the update IntOps book.

Sleet01: @HammerGS I've pushed the reversion of the tech advancement, along with an update to the composite tech advancement code: it turns out we did not handle the case where a tech was experimental right up until it was common.

HammerGS: Awesome!  There is some techs that do that.

NickAragua (approved):

Issue #5229, opened 2024-03-07
[49.18 Nightly]  Princess: Weapon Selection - choosing ammo weap over energy weap at same range brackets against high TMM enemy


Thom293:
FYI This build is from 2.22. Ill upgrade tomorrow.

A Thunderbolt shooting at an enemy vtol with 4 or 5 TMM.  TBolt ended previous turn at 0 heat.  

Tbolt shoots its SRM2 and LRM15 at the Warrior, but not its LL or ML.  That seems...off.  Something to do with targeting logic maybe?

Screen below, and log and save and custom attached.  

![image](https://github.com/MegaMek/megamek/assets/23645569/bcb10d41-e48a-4bbf-810a-5c49acb0a187)

[TBOLTnofirelaser.sav.gz](https://github.com/MegaMek/megamek/files/14519212/TBOLTnofirelaser.sav.gz)

[Mine.zip](https://github.com/MegaMek/megamek/files/14519235/Mine.zip)

Thom293:
fyi played another 6 rounds.  He never fired his lasers again.  No weather conditions.

If it matters, he is landlocked and cant advance further toward the battle, but the enemies are in LL and ML range for certain.

Sleet01:
It's more likely to do with the heat-to-damage calculation; if it wasn't cool enough to fire all weapons, it would probably just choose the lowest-heat ones.
The targeting code should actually favor energy weapons _over_ ammo-based weapons when targeting high-TMM enemies, as ammo attacks have a higher bar to clear due to the ammo conservation code.

  I'll confirm, though.

Thom293:
Ok.  I checked its heat though and even turns where it had zero heat it wasnt firing the lasers.  Just odd .

I didnt report it because I dont have a save or screen but I saw a turret not shoot when it could easily hit me over multiple turns.  Its almost as if it gets "stuck" deciding not to use a particular weapon for some reason.

Sleet01:
It's a good report, just not sure if it's a bug-bug or a weird emergent behavior stemming from various new code acting "correctly" but producing the wrong choices.

Sleet01:
[...]

PR #5248, opened 2024-03-14
Fix 5229: Multi-Trac units not firing energy weapons

Sleet01:
There were a couple issues in the MultiTargetFireControl planning track:

1. Energy weapons, or any weapon that doesn't actually track ammunition, would be skipped due to the way iterating over ammo was set up.
2. All firing plans were being assessed as if the shooter were an Aerospace unit, which include much stricter heat penalties.  As a result, Princess would be much less likely to fire multiple high-heat weapons in the same turn, even for 'Mechs.

TODO: since MultiTargetFireControl iterates over every target, for every weapon, for every _possible arc_, there is a good chance that we are calculating the same to-hit chances several times for given target-weapon pairings.  We could speed this up by caching results in a map, similar to TAGging is now being tracked.

This should not affect Energy-only Bay Weapons.

Testing:
- Ran all 3 projects' unit tests,
- Ran version of fix on provided save game,
- Created new game for full fix with multiple Multi-Trac units.

Close: #5229

### DIFF
--- a/megamek/src/megamek/client/bot/princess/FireControl.java
+++ b/megamek/src/megamek/client/bot/princess/FireControl.java
@@ -2598,7 +2598,8 @@ public class FireControl {
                                                                              ammoConservation);
             final FiringPlan plan = determineBestFiringPlan(parameters);
 
-            LogManager.getLogger().info(shooter.getDisplayName() + " at " + enemy
+            // Debug logging; Princess does her own info logging
+            LogManager.getLogger().debug(shooter.getDisplayName() + " at " + enemy
                     .getDisplayName() + " - Best Firing Plan: " + plan.getDebugDescription(true));
             if ((null == bestPlan) || (plan.getUtility() > bestPlan.getUtility())) {
                 bestPlan = plan;
--- a/megamek/src/megamek/client/bot/princess/FireControl.java
+++ b/megamek/src/megamek/client/bot/princess/FireControl.java
@@ -2598,9 +2598,6 @@ public class FireControl {
                                                                              ammoConservation);
             final FiringPlan plan = determineBestFiringPlan(parameters);
 
-            // Debug logging; Princess does her own info logging
-            LogManager.getLogger().debug(shooter.getDisplayName() + " at " + enemy
-                    .getDisplayName() + " - Best Firing Plan: " + plan.getDebugDescription(true));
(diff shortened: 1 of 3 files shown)


megamek/src/megamek/client/bot/princess/FireControl.java
```diff
@@ -2598,7 +2598,8 @@ FiringPlan getBestFiringPlan(final Entity shooter,
                                                                              ammoConservation);
             final FiringPlan plan = determineBestFiringPlan(parameters);
 
-            LogManager.getLogger().info(shooter.getDisplayName() + " at " + enemy
+            // Debug logging; Princess does her own info logging
+            LogManager.getLogger().debug(shooter.getDisplayName() + " at " + enemy
```
NickAragua, line 2602: General optimization suggestion: for methods that are called many times, calling logging methods (especially within a loop) leads to a performance decrease, even if the logging level doesn't result in the message being logged.  Let's make the adjustment to comment this out as we don't really need the info outside of direct debugging.

megamek/src/megamek/client/bot/princess/MultiTargetFireControl.java
```diff
@@ -78,6 +79,10 @@ public FiringPlan getBestFiringPlan(final Entity shooter,
 
             if (currentPlan.getUtility() > bestPlan.getUtility()) {
                 bestPlan = currentPlan;
+
+                // Debug logging; Princess does her own info-level logging
+                LogManager.getLogger().debug(shooter.getDisplayName() + " - Best Firing Plan: " +
```
NickAragua, line 84: As per the other logging comment - this is in a loop so we want to eliminate it.

NickAragua (changes requested):

megamek/src/megamek/client/bot/princess/FireControl.java
Sleet01: I think we can just do the debug logging alongside the Princess-level .info() logging, which is once per unit per round.

NickAragua: Sure. This logging is is once per unit per enemy per round. Granted, it's during the firing phase which doesn't really have performance issues usually, but still.

NickAragua (approved):

PR #5265, opened 2024-03-19
Load data from story arcs

AaronGullickson:
This PR is related to MegaMek/mekhq#2997, but needs to be handled separately because it involves megamek rather than mekhq code. Creators will in many cases want to load custom data such as portraits, camos, force logos, and customized mechs to be used in the story arc, but these will need to be loaded by megamek to be accessible. That is the purpose of this PR. It adds code to loop through subdirectories in the `data/storyarcs` directory to look for the following things and loads them:

* portraits (must be in subdirectory `data/images/portraits`, relative to the story arc top-level directory).
* camo (must be in subdirectory `data/images/camo`, relative to the story arc top-level directory). An additional complexity here is that the images must actually be in subfolders of `data/images/camo` for this to work. This seems to be a general limitation of camo images. If you try to put images directly in the main `data/images/camo` and not in a subdirectory they will not work either. Might be a bug?
* custom units (must be in subdirectory `data/mechfiles` relative to the story arc top-level directory). 

I was also thinking about doing boards, but these appear to be much more complicated, so I am gong to hold off on that for the moment. 

I also want to allow for custom force logos, but that is a mekhq issue so it will be handled from there.


HammerGS:
How will this interact with the new userfolder?

AaronGullickson:
Well the code for merging directories is what is enabling it to happen, but other than that it should not interfere.

HammerGS:
Awesome, I just know the "can the user directory support Story Arc question will come early in the testing" :)

AaronGullickson:
Well, users can still use the stuff in there, but creators have to properly put them in their story arc packages, which should be fine.

AaronGullickson:
Something is borked from the long hibernation of my megamek fork and it has these annoying merge commits that I cannot seem to get out. I am going to create a fresh fork and try this PR again.

PR #5281, opened 2024-03-25
Game types preparatory work

SJuliez:
This PR does some preparatory work to incorporate other types of game (mainly targeted at SBF) into MM. In order to (for now) bypass the lobby I added YAML scenario support (far more readable and writable than the previous way of writing scenarios; this is about MM scenarios, not MHQ); these may now be of the type SBF. For the following classes I added interface/abstract/SBF implementation where they were missing:
- Game
- ClientGUI
- Client
- GameManager

This builds on Neo's separation of Server and GameManager which I was very happy to have.
All of the implementations are stubs. I refrained from copying in the TW contents of the old classes; I'd rather build them slowly and maybe not copy old mistakes too much. The functionality with this PR includes: read an SBF scenario, build the game, start a Server+SBFGameManager with the game, start an SBFClient and SBFClientGUI, connect client to server and show an "under construction" page in the GUI. So, while the resulting GUI doesnt yet know to show anything, the MM game system is in place and connected and the game has players and units (formations for SBF).
The new scenario reader (ScenarioV2) is, for now, also incomplete, i.e. it does not read all options the old one (ScenarioV1) did.
My recent experience is that when working on bigger changes it is wise to have parts merged at some point to avoid getting bogged down with merge conflicts even if those parts are incomplete.

Most of the changes cannot impact standard games. However the Client has been divided up into AbstractClient and Client. This could affect standard games.


SJuliez:
Please don't merge, It's got a bug I have to correct.

BLR-IIC:
> I'm thinking we'll have a new release in a couple of weeks. I'm hoping that it will achieve milestone status. Thinking we hold off merging this for the next dev cycle.

Isn't the whole point of the milestone process to allow these things to be implemented and find the milestones in previous releases as the project moves forward?  That's how I remember you describing the milestone process on the BattleTech forum.  This line of reasoning seems like a step backward to me.

HammerGS:
> > I'm thinking we'll have a new release in a couple of weeks. I'm hoping that it will achieve milestone status. Thinking we hold off merging this for the next dev cycle.
> 
> Isn't the whole point of the milestone process to allow these things to be implemented and find the milestones in previous releases as the project moves forward? That's how I remember you describing the milestone process on the BattleTech forum. This line of reasoning seems like a step backward to me.

It is not. The current MILESTONE is still is two years old and while better than 0.48.0 we still need to get newer. We aren't saying this sits for months we are talking a couple of weeks. The last few dev releases have all had issues in them, sometimes it better to try to get a good dev out, which can become a MILESTONE than continuously push code. Especially when a PR like this touches some very important parts of Megamek. We've been largely focused on this release to get more fixes in. 

Criteria and Frequency of Milestone Releases

**The decision to designate a release as a ‘Milestone’ depends on several factors:**
- Stability and Reliability: A release must demonstrate a high level of stability and reliability during the observation period.
- Nature of Updates: Releases focused primarily on bug fixes are more likely to achieve ‘Milestone’ status than those introducing major new features or significant changes, like Large Naval Craft, Mobile Structures, or Space to Surface play (not saying we are implementing those anytime soon!).
- Frequency of Milestones: We anticipate MILESTONES will come up every 6-9 months. However, this could be longer or shorter depending on what is happening with development.

SJuliez:
> > I'm thinking we'll have a new release in a couple of weeks. I'm hoping that it will achieve milestone status. Thinking we hold off merging this for the next dev cycle.
> 
> Isn't the whole point of the milestone process to allow these things to be implemented and find the milestones in previous releases as the project moves forward? That's how I remember you describing the milestone process on the BattleTech forum. This line of reasoning seems like a step backward to me.

@BLR-IIC 
Btw, this is really only preparatory work. In terms of the LotR movies, this is the point where Galadriel says "I feel it in the earth" :)
The (almost) only visible thing this PR produces is this:
image

SJuliez:
Found the bug, should be good to merge.

megamek/src/megamek/common/planetaryconditions/PlanetaryConditions.java
```diff
@@ -988,4 +990,23 @@ public String getTemperatureIndicator() {
 
         return MSG_INDICATOR_TEMPERATURE_NORMAL;
     }
+
+    @Override
+    public String toString() {
+        return "[PlanetaryConditions]"
+                + ((temperature != 25) ? "; Temperature: " + temperature : "")
+                + ((gravity != 1) ? "; Gravity: " + gravity : "")
+                + (!atmosphere.isStandard() ? "; Pressure:" + atmosphere : "")
```
kuronekochomusuke, line 999: Is this used in serialize it to a file?  

if so, it will break when the language is changed, since atmosphere.toString() returns the value from the resource file for the selected language.  Same with the others.

SJuliez: I will never intentionally use toString() for serialization. I add toString() implementations so I see useful info in the IDE when debugging.

PR #5285, opened 2024-03-26
For crew calculations, consider pintle weapons in the same location as in the same pintle

pavelbraginskiy:
Fixes MegaMek/megameklab#1461

The rules for pintle mounts make it clear more than one weapon can be placed in the same pintle mount, and the option to add pintles only ever adds a single pintle mount per location anyway.


Sleet01 (approved):
Looks good to me.

PR #5288, opened 2024-03-27
[Typo] Replaced Instances of Incindiary with Incendiary

IllianiCBT:
### Current Implementation
There are around 7 instances of 'incindiary' used across mm.

### Problem
This appears to be a typo. It looks like this is a known issue, as there is a case included for it in `megamek/src/megamek/client/ratgenerator/MissionRole.java`

image

This appears to be isolated to mm, as I found no instances of its use in mml or mhq.

### Solution
All instances of 'incindiary' have been replaced by 'incendiary', maintaining capitalization, where applicable.

### DIFF
--- a/megamek/src/megamek/client/ratgenerator/MissionRole.java
+++ b/megamek/src/megamek/client/ratgenerator/MissionRole.java
@@ -20,7 +20,7 @@ import megamek.common.UnitType;
 
 /**
  * Used to adjust availability to conform to a particular mission role.
- * 
+ *
  * @author Neoancient
  */
 public enum MissionRole {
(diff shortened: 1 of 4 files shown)


megamek/src/megamek/client/ratgenerator/MissionRole.java
```diff
@@ -627,7 +627,6 @@ public static MissionRole parseRole(String role) {
             case "raider":
                 return RAIDER;
             case "incendiary":
-            case "incindiary":
```
NickAragua, line 630: Let's keep this in case there's some misspelled data somewhere.

NickAragua (changes requested):

megamek/src/megamek/client/ratgenerator/MissionRole.java
IllianiCBT: Sounds good :)

IllianiCBT:
Closed due to non-issue.

PR #5289, opened 2024-03-27
Improved MissionRole Switch Cases & Role List

### LINKED ISSUE #5267: Role typo in forcegenerator parse code
Should be an easy one for newcomers...

In the forcegenerator XMLs there is a role element, that includes the option for designating infantry as Anti-Mech equipped/specialized/whatever - 'anti_mek'.  In MissionRole.java the processor uses two alternatives:

https://github.com/MegaMek/megamek/blob/5808376bf3f0f0e3f2ebc383ea9b73f4e6b60660/megamek/src/megamek/client/ratgenerator/MissionRole.java#L706-L708

Note that it accepts a space or a dash, but not an underline.  As the underline version is already used in the forcegenerator XML files (and I have continued this with my own personal customizations), I would recommend replacing the two unused entries with the version in use.  This will also match the text in the docs/RAT Stuff/rat-generator.txt help file.

Edit: jumped the gun a bit, looks like there are multiple instances of this.  Here is a quick summary, based on the XMLs and the rat-generator.txt help file:

- fire_support (XML/help) vs. fire support (parser)
- sr_fire_support (XML/help) vs. sr fire support (parser)
- inf_support (XML) vs. infantry_support (help) vs. "infantry support" or "inf support" (parser)
- ew_support (XML/help) vs. ew support (parser)
- anti_aircraft (XML/help) vs. "anti aircraft" or "anti-aircraft" (parser)
- anti_infantry (XML/help) vs. "anti infantry" or "anti-infantry" (parser)
- specops (XML/help) vs. "spec ops" or "specops" (parser) * optional, for clarity
- ground_support (XML/help) vs. ground support (parser)
- mech_carrier (XML/help) vs. mech carrier (parser)
- asf_carrier (XML/help) vs. asf carrier (parser)
- vee_carrier (XML/help) vs. vee carrier (parser)
- protomech carrier (parser only) * should update forcegenerator files and help file, recommend using underscore for consistency
- troop_carrier (XML/help) vs. troop carrier (parser)
- pocket ws (help only) vs "pocket ws" or "pocket warship" (parser) * optional, recommend using underscore for consistency, updating help file, and applying to XMLs as appropriate
- field_gun (XML/help) vs. field gun (parser)

IllianiCBT:
### Problem
Currently there are a number of unused roles listed as viable cases in `megamek/src/megamek/client/ratgenerator/MissionRole.java`.

There are also missing or incorrect roles listed in `megamek/docs/RAT Stuff/rat-generator.txt`.

### MissionRole.java
I have gone through and removed any unused cases.

### rat-generator.txt
I have clarified that either `infantry_support` or `inf_support` are accepted cases for the infantry support role. Both are in use and, while it might be advantageous to unify everything to use the same case, that is beyond the scope of this PR.

I have replaced 'incindiary' with 'incendiary' as the suggested case for units that can set fires. In #5288 Nick suggested we leave uses of 'incindiary' intact, however replacement in this instance leaves other uses intact while ensuring that future uses of the case are using the correct spelling. As this is user facing, I also wanted to ensure that the spelling was correct.

I have added the missing `mechanized_ba` role and assigned it the definition `eligible to be carried by Omni units as mechanized BA`.

I changed `pocket ws` to `pocket_warship`, as while `pocket ws` would have been accepted, its only use is in `rat-generator.txt` so by changing it to `pocket_warship` I was able to remove that case from `MissionRole.java`.

### Note
This initially started out as an attempt to investigate #5267. However, upon investigation, I found that existing functions relied upon the cases (when used) as currently implemented. As I was unable to find which functions were reliant on the current role cases, I opted instead to just leave the class intact and to only tidy things up so that we weren't checking for unused cases.

### DIFF
--- a/megamek/docs/RAT Stuff/rat-generator.txt
+++ b/megamek/docs/RAT Stuff/rat-generator.txt
@@ -183,7 +183,7 @@ Warship
         infantry_support or inf_support - commonly found supporting infantry formations
         ew_support - carries additional electronics for ECM, scanning, or command functions
         spotter - carries TAG or equivalent, such as C3 master
-        incindiary - can easily set fires
+        incendiary - can easily set fires
         mag_clamp - battle armor or ProtoMek with magnetic clamps
         artillery - carries tube artillery or artillery cannon
         missile_artillery - carries missile artillery, typically Arrow IV
--- a/megamek/docs/RAT Stuff/rat-generator.txt
+++ b/megamek/docs/RAT Stuff/rat-generator.txt
@@ -216,7 +216,7 @@ Warship
         ba_carrier - has battle armor bays
         mechanized_ba - eligible to be carried by Omni units as mechanized BA
         tug - has tug adaptor
-        pocket warship - pocket warship DropShip
+        pocket_warship - pocket warship DropShip
         corvette - WarShip class
         destroyer - WarShip class
         frigate - WarShip class
--- a/megamek/docs/RAT Stuff/rat-generator.txt
+++ b/megamek/docs/RAT Stuff/rat-generator.txt
@@ -180,7 +180,7 @@ Warship
         comma-separated list of roles the variant was designed for. Supported values are:
         fire_support - focus on long range firepower
         sr_fire_support - focus on short range firepower
-        infantry_support - commonly found supporting infantry formations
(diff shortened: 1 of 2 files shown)


Sleet01 (approved):
Looks good to me.

PR #5293, opened 2024-03-29
Fix 3882 (MM part): ASFs not tracked correctly in MHQ

Sleet01:
This patch fixes a couple of issues by continuing the replacement of "AERO" values with "AEROSPACEFIGHTER":

1. ASFs being counted as non-ASF units when TOE is read to create StratCon scenario "protect" requirements,
2. Air support Scenario Modifiers not generating any friendly units even when applied,
3. RAT Generator tag in Random Unit Generator dialog finding zero ASF unit types for any faction or subfaction (regardless of era),
4. RAT Generator throwing an IndexOutOfBounds exception whenever zero-unit ASF RATs are generated.

These fixes mainly involve removing UnitType.AERO values in favor of UnitType.AEROSPACEFIGHTER.

Testing:
1. Ran all three projects' unit tests,
2. Verified functionality of RAT Generator with specific known-existing era:faction:class:weight requirements.
3. Verifed function of Air Support scenario modifier in MHQ.

Note: this fix is independent of MegaMek/mekhq#3882 and can be applied before or after the PR for that issue.

### DIFF
--- a/megamek/data/forcegenerator/2398.xml
+++ b/megamek/data/forcegenerator/2398.xml
@@ -28,7 +28,7 @@
 		<salvage pct='0'></salvage>
 		<weightDistribution era='2398' unitType='Mek'>3,4,2,1</weightDistribution>
 		<weightDistribution era='2398' unitType='Tank'>3,4,3,1</weightDistribution>
-		<weightDistribution era='2398' unitType='Aero'>1,1,1</weightDistribution>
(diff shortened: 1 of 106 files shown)


Sleet01:
Similar to the MHQ fix, this now looks huge due to replacing 'Aero' with 'AeroSpaceFighter' in the `data/forcegenerator/` files, but in fact is not that many actual code changes.

SJuliez (approved):

PR #5301, opened 2024-03-30
Flight path indicator

HoneySkull:
### Description
This PR introduces a graphic enhancement called Flight Path Indicator to help visualize flight path planning with fighters on ground maps.  When units have remaining velocity in the move phase, the Flight Path Indicator will project the remaining path in front of the flying unit to show the path trajectory the unit will take for the remaining velocity.  The color (and shape) of the indicator tells the pilot where the fighter can turn along its path, where the path ends, and that a fighter will fly off the map.

image

image image
Empty red circles or flags indicate the fighter cannot turn on that hex along its flight path.

image image
A yellow solid circle or flag indicates the fighter can make a turn for a cost.

image image
A yellow empty circle or flag indicates the fighter could make a turn for a cost but doesn't have enough movement to do so.

image image
A green solid circle or flag indicates the fighter can make a free turn at that location along its flight path.

image
A yellow diamond indicates the fighter has remaining velocity at the edge of the map and will fly off the map.

image
A flag indicates the hex that will end the flight path on this trajectory at the current velocity.

image
When the fighter turns, the Flight Path Indicator adjusts to the new trajectory and turn restrictions.

image
Fighter used up its MP turning.  If it had enough thrust points, it could turn on the empty yellow.  Therefore, solid indicators are 'can turn' while empty indicators are 'no turn'.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/MovementDisplay.java
+++ b/megamek/src/megamek/client/ui/swing/MovementDisplay.java
@@ -572,6 +572,7 @@ public class MovementDisplay extends ActionPhaseDisplay {
                     public void performAction() {
                         clear();
                         computeMovementEnvelope(ce());
+                        updateMove();
                     }
                 });
 
@@ -4654,6 +4655,7 @@ public class MovementDisplay extends ActionPhaseDisplay {
         } else if (actionCmd.equals(MoveCommand.MOVE_CANCEL.getCmd())) {
             clear();
             computeMovementEnvelope(ce);
+            updateMove();
         } else if (ev.getSource().equals(getBtn(MoveCommand.MOVE_MORE))) {
             currentButtonGroup++;
             currentButtonGroup %= numButtonGroups;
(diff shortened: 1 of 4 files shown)


HoneySkull:
I'm going to make the final hex flag consistent with the visual language and represent turn status on the final hex.  Green flag will mean free turn, red empty flag will mean no turn possible, yellow flag will mean turn with cost, and yellow empty flag will mean 'could turn with cost if you had MP left to do so'.

HoneySkull:
Note to self:  Infinite loop trying to calculate Flight Path Indicator on space map with Tac Ops Aerospace advanced movement turned on:

image

HoneySkull:
Added an enhancement to let the player know how many remaining hexes are required to travel when the flight path indicator has them going off the map.  This is useful in some cases because it's not discernable how long the flight path is once it leaves the board.  

image

SJuliez (approved):

PR #5319, opened 2024-04-06
Add Symbol Font Google Material Symbols

SJuliez:
This adds a font with a ton of icons, google's Material Symbols together with its license (Apache 2.0). These icons seem useful e.g. for marking hexes or other uses. Because of the way thy are centered, they are less useful for inlining with other text. Also adds a method in FontHandler to get that symbol font without having to remember its name.


SJuliez:
@HammerGS You alright with this? It has a different license, Apache 2.0 (not OpenFont). Still seems ok to me.

rjhancock:
Must provide appropriate credit but otherwise the licensing does not prevent its inclusion. The Apache License is actually more permissible than the OpenFont license in regards to fonts it seems.

PR #5331, opened 2024-04-08
Fix #3978 [MekHQ]: NPE When Attempting to Save (Manual or Auto)

### LINKED ISSUE #3978: issue on Extras tab Current Sensors issue when really long
0.49.10

issue on Extras tab Current Sensors issue when really long.  it causes the info to be shifted off screen.  needs to word wrap when the text is really long.

currrentsensorstoolongScreenshot 2022-11-12 170959

AaronGullickson:
This PR will fix issue MegaMek/mekhq#3978, which was causing an NPE when BotForce tried to write entities to XML. The immediate reason is that EntityListFile was calling `entity.getGame` in a couple of places without checking for it being null, although it was checking in a variety of places. This PR fixes the problem by adding in checks for null in every place where game is being used.

The more long term solution would be to make EntityListFile more robust to this, but I didn't want to get into bigger coding changes with the goal to release 49.19 shortly, so I will make a separate issue for that.

### DIFF
--- a/megamek/src/megamek/common/EntityListFile.java
+++ b/megamek/src/megamek/common/EntityListFile.java
@@ -746,7 +746,7 @@ public class EntityListFile {
                 output.write("\" " + MULParser.ATTR_QUIRKS + "=\"");
                 output.write(String.valueOf(entity.getQuirkList("::")));
             }
-            if ((entity.getC3Master() != null) && (entity.getGame() != null)) {
+            if ((entity.getGame() != null) && (entity.getC3Master() != null)) {
                 output.write("\" " + MULParser.ATTR_C3MASTERIS + "=\"");
                 output.write(entity.getGame()
                         .getEntity(entity.getC3Master().getId())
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/EntityListFile.java
```diff
@@ -746,7 +746,7 @@ public static void writeEntityList(Writer output, ArrayList<Entity> list) throws
                 output.write("\" " + MULParser.ATTR_QUIRKS + "=\"");
                 output.write(String.valueOf(entity.getQuirkList("::")));
             }
-            if (entity.getC3Master() != null) {
+            if ((entity.getC3Master() != null) && (entity.getGame() != null)) {
```
SJuliez, line 749: looking at getC3Master() it can already fail when game == null, so these checks should be reversed, I think.

SJuliez (changes requested):

megamek/src/megamek/common/EntityListFile.java
AaronGullickson: Good catch

AaronGullickson: Fixed now

SJuliez (approved):

PR #5338, opened 2024-04-09
Fix MekHQ #3986: MHQ wont save, sometimes

AaronGullickson:
This should fix Megamek/mekhq#3986. The issue is that Entity.c3UUID can be null in BotForces, but that is never checked in EntityListFile. The original XML writer in MHQ for BotForces avoided this problem by ... not checking C3 at all. Ultimately, we need to correct all of this by consolidating XML writers, but that is a bigger project. 

I fixed it simply by checking for null status before writing. I looked for other cases where we might be writing a null object in the same method and I could not find any. I tested it on the campaign file by generating random opponents when advancing day and saving. I was able to do it in 20 cases without problem including many cases where I know I had a C3 entity.

### DIFF
--- a/megamek/src/megamek/common/EntityListFile.java
+++ b/megamek/src/megamek/common/EntityListFile.java
@@ -753,8 +753,10 @@ public class EntityListFile {
                         .getC3UUIDAsString());
             }
             if (entity.hasC3() || entity.hasC3i() || entity.hasNavalC3()) {
-                output.write("\" " + MULParser.ATTR_C3UUID + "=\"");
-                output.write(entity.getC3UUIDAsString());
+                if(entity.getC3UUIDAsString() != null) {
+                    output.write("\" " + MULParser.ATTR_C3UUID + "=\"");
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/EntityListFile.java
```diff
@@ -753,8 +753,10 @@ public static void writeEntityList(Writer output, ArrayList<Entity> list) throws
                         .getC3UUIDAsString());
             }
             if (entity.hasC3() || entity.hasC3i() || entity.hasNavalC3()) {
-                output.write("\" " + MULParser.ATTR_C3UUID + "=\"");
-                output.write(entity.getC3UUIDAsString());
+                if(entity.getC3UUIDAsString() != null) {
```
kuronekochomusuke, line 756:
```suggestion
                if (entity.getC3UUIDAsString() != null) {
```

missing space

kuronekochomusuke (approved):

PR #5346, opened 2024-04-10
Validate that all armor is allocated

pavelbraginskiy:
Since having equipment that's unallocated causes a unit to be invalid, this does the same for armor points. 

Implemented for Meks, Tanks, Suppot Vees, and AeroFighters.

Not implemented for ProtoMechs and Small Craft because the method here doesn't work and sometimes reports a unit as valid when not all points have been allocated, and I have no idea why.

Not implemented for Advanced Aerospace because I have no idea what's going in with advanced aerospace.

### DIFF
--- a/megamek/src/megamek/common/verifier/TestAero.java
+++ b/megamek/src/megamek/common/verifier/TestAero.java
@@ -519,6 +519,12 @@ public class TestAero extends TestEntity {
             correct = false;
         }
 
+        if (!getEntity().hasPatchworkArmor()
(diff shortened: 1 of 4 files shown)


gsparks3:
Issue: I am pretty sure that there are multiple canon units with some points of unallocated armor as built. This would cause them to be treated as invalid, when AFAIK it is allowed by the rules, or at least not forbidden.

pavelbraginskiy:
An MML nag was the entire point, I forget to allocate armor _all the time_. 
I'll think of a cleaner and more correct way to implement it later, but it'll surely be an MML PR if it happens.

PR #5351, opened 2024-04-12
Fix null pointer when turning with shortcut keys in movement phase.

### LINKED ISSUE #5350: [0.49.19-SNAPSHOT #1461] NPE When turning Aerospace with Shift-D or Shift-A Shortcut keys in Movement Phase
### Environment
Starting MegaMek v0.49.19-SNAPSHOT
	Build Date: N/A
	Today: 2024-04-11
	Origin Project: MegaMek
	Java Vendor: Eclipse Adoptium
	Java Version: 11.0.20.1
	System Locale: en_US

### Description
Null Pointer Exception thrown when attempting to make a left or right turn in the movement phase with Aerospace fighters using the shortcut keys "Shift-A" or "Shift-D" - The null pointer exception seems to be thrown from the FiringPhase event handler but occurs while in the movement phase.

Steps to re-produce:
1.  Create a new game with an Aerospace fighter.  (Edit: Doesn't need to be an aerospace fighter)
2. Complete a full turn.
3. During the movement phase on subsequent turns, accelerate the fighter move forward 8 hexes or until the fighter has a "free turn"
4. Turn the Aerospace fighter left or right by using "Shift-A" or "Shift-D".

image

image
### Files
image

[megamek.log](https://github.com/MegaMek/megamek/files/14952769/megamek.log)

HoneySkull:
Fixes #5350 

During the movement phase that follows a previous firing phase - when using the short cut keys "Shift-A" or "Shift-D to turn left or right results in a null pointer exception in the `FiringDisplay` class when it is checking for torso twist left or right.

The expression was updated to promote the check for firing display visibility and ignore event status BEFORE checking the selected entity which can be null in the `FiringDisplay` during the movement phase.

This bug could also cause a null pointer when trying to turn other units - not just aerospace fighters like the bug report suggests.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/FiringDisplay.java
+++ b/megamek/src/megamek/client/ui/swing/FiringDisplay.java
@@ -281,11 +281,13 @@ public class FiringDisplay extends AttackPhaseDisplay implements ItemListener, L
 
                     @Override
                     public boolean shouldPerformAction() {
+                        // can get called outside of the firing phase, need to check 'isVisible' first in expression.
                         if (!clientgui.getClient().isMyTurn()
-                                || ce().getAlreadyTwisted()
-                                || clientgui.getBoardView().getChatterBoxActive()
                                 || !display.isVisible()
-                                || display.isIgnoringEvents()) {
+                                || display.isIgnoringEvents()
+                                || ce() == null
+                                || ce().getAlreadyTwisted()
+                                || clientgui.getBoardView().getChatterBoxActive()) {
                             return false;
                         } else {
                             return true;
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/client/ui/swing/FiringDisplay.java
```diff
@@ -281,11 +281,13 @@ public void performAction() {
...
                     @Override
                     public boolean shouldPerformAction() {
+                        // can get called outside of the firing phase, need to check 'isVisible' first in expression.
                         if (!clientgui.getClient().isMyTurn()
-                                || ce().getAlreadyTwisted()
-                                || clientgui.getBoardView().getChatterBoxActive()
                                 || !display.isVisible()
-                                || display.isIgnoringEvents()) {
+                                || display.isIgnoringEvents()
+                                || ce() == null
```
HoneySkull, line 288: If cd() returns null, this will exit out of the expression and wont evaluate the latter conditions in the expression.

Sleet01 (approved):
Looks good!  I think I added the getAlreadyTwisted() functionality for the TAG-phase torso twisting, but I didn't realize it was possible for the entity to be `null` later on -_-;

PR #5352, opened 2024-04-12
#5348 stop adding jump start step when deleting movement

### LINKED ISSUE #5348: Jump-mode backspace past beginning of move queue locks game
### Environment
Megamek 0.49.18

Tested under Adoptium Java 11 and 17 on Windows 10, and OpenJDK 11 and Adoptum 17 on current Fedora Linux, problem persists in all cases.

### Description
When you're in movement phase with a mek and in jump mode and backspace past the beginning of the movement list, the game completely freezes and needs to be task killed to close. This can be done just be pressing the 'jump' button and backspace once. 

### Files
[megamek.log](https://github.com/MegaMek/megamek/files/14938673/megamek.log) - doesn't appear to have anything from this problem
[jump-backspace-to-crash-0.49.18.sav.gz](https://github.com/MegaMek/megamek/files/14938674/jump-backspace-to-crash-0.49.18.sav.gz) - playername Weaver - this is just a pxh-1 by itself ready to move on turn one, so select 'jump' and hit backspace to crash it.

Lu9us:
Fixes #5348
This pr removes some checks in the movement control class that was stopping the movement control system leaving jump mode when you delete the last movement node, this was causing an infinte loop where the system constantly tried to add an illegal jump start node to the start of the move path only to have it subsquently removed and then re-added again this was locking up the app.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/MovementDisplay.java
+++ b/megamek/src/megamek/client/ui/swing/MovementDisplay.java
@@ -1360,13 +1360,10 @@ public class MovementDisplay extends ActionPhaseDisplay {
         clientgui.getBoardView().setSensorRange(ce, cmd.getFinalCoords());
 
         // set to "walk," or the equivalent
-        if (gear != MovementDisplay.GEAR_JUMP) {
-            gear = MovementDisplay.GEAR_LAND;
-            Color walkColor = GUIP.getMoveDefaultColor();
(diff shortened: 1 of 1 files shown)


HoneySkull (approved):

HoneySkull:
Verified and tested.  As this condition can cause the client to hard-freeze, I recommend merging this PR in 0.49.19.

PR #5359, opened 2024-04-13
Implements #5261: Simple rotating per-round save file system.

### LINKED ISSUE #5261: [RFE] Have Default be 3 Autosaves, just like MHQ, to avoid losing progress. This would solve many many issues.
Over last 2 days, but honestly all of the time, people lose progress because the single autosave gets overwritten.  For me personally as a tester, I am often using nightlies and have an error or something I am testing and it would be very useful to go back more than 1 turn.  Currently there is an intermittant bug where the game freezes on the last turn before exporting a .mul and people lose all of their progress.  Being able to go back 2 or 3 turns would solve this.  One of the players lost an 18 round Company + sized battle this morning.

image

And this is from a new player 2 days ago that didnt understand what I meant by backing up their autosave and they had to restart a 40+ round battle too.

image

We do have incremental autosaves with timestamps, but then you get hundreds of Autosaves.  Also, they are an optional setting that you have to know enough to have to enable.  

I propose a happy medium.  Have the default setting be 3 autosaves just like MHQ does for the last 3 rounds. Not something that needs to be enabled.  Just on by default.  A new user wont always know what options to turn on and I imagine someone new that plays an 18 round battle and then loses all of their progress might not have a favorable opinion of the suite and may not come back   Just name them Autosave A, B, C or 1, 2, 3, and they overwrite each other constantly.  We can tell the timestamp from the file details and know which one was most recent.

image

This would solve a  great many issues and reduce one of the largest sources of frustration with MM - losing all of your progress.  It would also help the QA team greatly.

Thank you.

Sleet01:
This patch adds user-configurable per-round rotating save files, based on the system used by MekHQ, to MegaMek.

These saves are separate from the other autosaves and default to 3, so the last 3 rounds of the most recent game will be saved.  Critically, these save files will not be overwritten by starting MegaMek; a game must actually be entered and the start of a new round must occur for this save system to be triggered.

File name format is "Round-<current round#>-autosave_<datetime stamp>.sav.gz", e.g. `Round-3-autosave_2024-04-13_03-37-08.sav.gz`.  Note that this means only 3 rotating save files _total_ will be saved, not 3 per separate game.  If we want that functionality, more work is required.

Testing:
- Ran small game over several rounds, confirming file rotation works and files can be loaded and resumed.

Close #5261


Sleet01:
Based on comments from the QA team, I would like to strongly advocate for including this functionality in the .19 release in order to facilitate better testing results.

Sleet01:
Files from a short Princess v Princess game:

image

Sleet01:
New Game Option location:
image

HoneySkull (approved):
reviewed the code - looks good.

HoneySkull:
Can you test this with:
1. Maximum Rotating Saves set to 0.
2. Change the configured save path in the properties set to something other than the default.
3. Using the default megamek config file (as if from a new fresh install)

Thom293:
I can test this in about 10 hours if no one gets to it first.

Sleet01:
> Can you test this with:
> 
> 1. Maximum Rotating Saves set to 0.
> 2. Change the configured save path in the properties set to something other than the default.
> 3. Using the default megamek config file (as if from a new fresh install)

@HoneySkull @Thom293 Tested and confirmed:
1. Setting Max to 0 prevents new rotating saves (and prevents deleting old rotating saves).
2. Changing the configured save path in properties causes saves, including rotating saves, to appear in the new dir.
3. Default config file sets Maximum Rotating Saves to 3.

Given that, I'm going to pull this.

PR #5367, opened 2024-04-16
correct issue when using multiple boards and restricting deployment a…

kuronekochomusuke:
- correct issue when using multiple boards and restricting deployment any to a rectangle for units
- fixes issue where it limits rectangle to only 1st board


HammerGS (approved):
Tested and seems to work.

PR #5368, opened 2024-04-17
Fixes #5215 - Secondary turret firing arc display to reflect turret rotation. (tanks)

### LINKED ISSUE #5215: [49.18] Secondary turret firing arc display does not reflect turret rotation
On a multi-turret vehicle (such as the Gulltoppr OmniMonitor), rotation of the FT (front turret) using the turret twist mini-menu does not visually update the firing arcs for the weapon. The primary turret updates as normal. 

In this image, the turret was rotated to point northeast prior to re-opening the FT rotation menu:
image

While the visual display is incorrect, the actual turret rotation appears to be correct.
image

This is likely related to #2659 , as the FT here uses the same rotation menu as the mech turrets there.

HoneySkull:
This PR fixes issue #5215 by checking for dual turret rotation and updating the selected weapon to be the dual turret upon rotation.  This doesn't specifically address the Mek turret rotation firing display issue as this logic is specific to tanks and super heavy vehicles.


HammerGS (approved):
Tested and seems to be working with no issues.

PR #5376, opened 2024-04-19
Fix #5375: quirks and SPAs applying twice to ADA missiles only

### LINKED ISSUE #5375: 49.19 - AA Targeting Quirk applying twice
### Environment
49.19
Windows
Java 11

### Description
The issue was presented by a player in the Aerospace Academy discord from version 49.18. However I was immediately able to replicate the issue in 49.19 with a standard Longbow. 
image

And just for confirmation I did also run the test against a standard VTOL unit [not just ASF] and it also has the duplication.
image

One other final test was just to confirm that the bug was happening with all weapons and not just ADAs
image

### Files
This can be replicated in any MegaMek with any unit using AA Targeting Quirk.

Sleet01:
This was caused by changing ADA (and only ADA) to-hit calculations to continue in the standard To-Hit mod path after calculating ADA's unusual range mod; all other artillery direct and indirect fire to-hit mods were unaffected.  Since the main-line mod calculation path calcs Quirks and SPAs last, but the Artillery path calcs them first, ADA was double-dipping.

Fix is to move the Artillery-specific quirk and SPA calculations to after the ADA mod path returns to the main path, so ADA only gets these calculated once (but other artillery still gets theirs calculated separately from the main-line path).

Testing:
- Ran a variety of attack calculations using Anti-Air quirk units and every kind of weapon system versus various airborne units
- Ran all projects' unit tests.

### DIFF
--- a/megamek/src/megamek/common/actions/WeaponAttackAction.java
+++ b/megamek/src/megamek/common/actions/WeaponAttackAction.java
@@ -4937,10 +4937,19 @@ public class WeaponAttackAction extends AbstractAttackAction implements Serializ
             // Return without SRT set so that regular to-hit mods get applied.
             return toHit;
         }
+
+        // ADA has its to-hit mods calculated separately; handle other direct artillery quirk and SPA mods here:
+        // Quirks
+        processAttackerQuirks(toHit, ae, te, weapon);
+
+        // SPAs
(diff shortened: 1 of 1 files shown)


HoneySkull (approved):
The code change looks good based on the description of the fix.

AaronGullickson (approved):
LGTM

PR #5373, opened 2024-04-18
Fix: #2659 - Mek Quad turret firing arc reflects turret rotation changes.

HoneySkull:
Fixes #2659 - This PR fixes firing arcs updates for mek mounted turrets such as that on a QUAD mek.  This is similar to the fix for vehicle turret firing arcs.  When changing the turret direction, the selected turret weapon will be updated if selected, or the first weapon on the turret if a non-turret weapon is the current selected weapon.

image

Issue #2659, opened 2021-02-16
Battlemech Quad Turret firing arc does not reflect turret rotation changes


mjbroekman:
### Environment
 	MekHQ 0.47.17
	Java vendor N/A (openjdk)
	Java version 15.0.1
	Platform Mac OS X 10.15.7 (x86_64)

### Description
When a Quad mek is equipped with battlemech turrets, the firing arcs of the turret do not update to reflect turret rotation. Instead, the firing arcs remain fixed according to the firing arc of the location that the turret is in.
<img width="1237" alt="Screen Shot 2021-02-16 at 9 01 58 AM" src="https://user-images.githubusercontent.com/5141448/108073817-7a758880-7036-11eb-9f58-ab03a0415fcb.png">
<img width="878" alt="Screen Shot 2021-02-16 at 9 02 19 AM" src="https://user-images.githubusercontent.com/5141448/108073848-85301d80-7036-11eb-879b-9fdacaa5e697.png">
<img width="860" alt="Screen Shot 2021-02-16 at 9 02 29 AM" src="https://user-images.githubusercontent.com/5141448/108073864-89f4d180-7036-11eb-852f-c1f581708cbf.png">

### Files
[firingarc.sav.gz](https://github.com/MegaMek/megamek/files/5989146/firingarc.sav.gz)
[megameklog.txt](https://github.com/MegaMek/megamek/files/5989151/megameklog.txt)
[mekhqlog.txt](https://github.com/MegaMek/megamek/files/5989152/mekhqlog.txt)

HoneySkull:
![image](https://github.com/MegaMek/megamek/assets/83041327/c1df54bf-d8f0-4fba-bac4-b1d2181cbadd)

PR #5377, opened 2024-04-19
Correct mek arc facing when mek is turned.

HoneySkull:
This is a correction PR #5373.  

The Fix for #2659 didn't account for mek facing.  This PR ensures the mek turret facing is correct for selected turret facing and mek facing.


Sleet01 (approved):
LGTM!

PR #5406, opened 2024-04-23
Mounted rework

neoancient:
This overhaul of Mounted is the next step in my equipment overhaul. The vast majority of the changes involve type checking, and errors would be caught by the compiler.

What it does:
1. Creates subclasses of Mounted and moves functionality that only applies to one subclass of EquipmentType into the appropriate Mounted subclass.
2. Adds a generic type parameter to Mounted to improve type safety working with mounted equipment.
3. Adds a few convenience methods.

Benefits:
1. Reduces the size of Mounted by removing code that's only used for some type of equipment.
2. Removes much of the need for type checking and casting of equipment. For example, Entity::getWeapons returns List<WeaponMounted>, and WeaponMounted::getType returns WeaponType. No need for wtype = (WeaponType) mount.getType().
4. Prepares the way for moving code from Entity to a Mounted subclass. A step in the near future is an ArmorMounted, which will handle the armor fields and weight calculations, and deal with patchwork more elegantly.

Note that the calculation of ammo shots is still in the base class rather than AmmoMounted because it is also used by mines, which are MiscType. It can probably be simplified but I left that for the future, as this is already very extensive.

Also I handled the movement of bay weapons and ammo to WeaponMounted in a separate PR that builds on this one, since that change also touches a lot of code.


SJuliez (approved):
Considering how central Mounted is, the amount of change is rather lower than I had expected. We have some equipment with non-obvious type, e.g. C3M is a weapon and (iirc) Radical Heat Sinks are an AmmoType but I imagine if it worked before then it'll work now.

PR #5409, opened 2024-04-24
Remove code for using a background image on boards

SJuliez:
This removes Board and BoardView code that allowed using a large background image for the board instead of the usual hex tileset. This was unused (although we did have one such background image in the distribution) and probably completely unknown. It is also rather inflexible and it is hard to see how to integrate it into other systems. Therefore it seems better to relieve Board and BoardView of the code parts.
The deleted board file was probably suitable for the deleted image file but the board file did not use the necessary keyword and using a standard tileset it doesn't look like a well-made map.


Sleet01 (approved):
LGTM

PR #5416, opened 2024-04-25
Weapon bay rework

neoancient:
This is the second half of #5406. It moves bay weapons and ammo to WeaponMounted and makes them easier to work with. Instead of dealing with a list of equipment ids and having to retrieve the weapon and ammo and typecast it to the correct type, WeaponMounted::getBayWeapons and WeaponMounted::getBayAmmo return a list of WeaponMounted and AmmoMounted, respectively.

I also moved some point defense system code from Mounted to WeaponMounted.


megamek/src/megamek/common/Compute.java
```diff
@@ -254,6 +254,20 @@ public static float randomFloat() {
         return random.randomFloat();
     }
 
+    /**
+     * Selects a random element from a list
+     * @param list The list of items to select from
+     * @return     An element in the list
+     * @param <T>  The list type
+     */
+    public static<T> T randomListElement(List<T> list) {
```
SJuliez, line 263: I was sure we had that somewhere but couldnt find it. But it might be placed in CollectionUtil

megamek/src/megamek/common/actions/TeleMissileAttackAction.java
```diff
@@ -61,15 +63,15 @@ public boolean getPDOverheated() {
      * Returns the list of Counter Equipment used against this physical attack
      * This is for AMS assignment to manual tele-operated missiles
      */
-    public ArrayList<Mounted> getCounterEquipment() {
+    public List<WeaponMounted> getCounterEquipment() {
```
SJuliez, line 66: We might want to do a null check here and initialize the list if it is null like in addCounterEquipment() to avoid having to null check the return value. Similarly in WeaponAttackAction. But it is a legacy issue, another time is just as good.

SJuliez (approved):
looks good to me.

megamek/src/megamek/common/Compute.java
neoancient: I was pretty sure it was somewhere but I couldn't find it either.

megamek/src/megamek/client/bot/princess/FireControl.java
```diff
@@ -704,7 +704,7 @@ ToHitData guessToHitModifierForWeapon(final Entity shooter,
         }
         // Bays compute arc differently
         final boolean inArc = (bayWeapon)
-                ? Compute.isInArc(game, shooter.getId(), (int) weapon.getBayWeapons().get(0), target)
+                ? Compute.isInArc(game, shooter.getId(), weapon.getBayWeapons().get(0).getEquipmentNum(), target)
```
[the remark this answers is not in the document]
neoancient: getBayWeapons now return WeaponMounted instead of the equipment number.

PR #5430, opened 2024-04-28
Candidate fix for #5417 NPE Freeze bot moves invalid or nonsense.

### LINKED ISSUE #5417: 49.20 MM Standalone Build - NPE Freeze.  Kicking and Replacing Bots does not restart action.
Using this specific build https://github.com/MegaMek/megamek/actions/runs/8825771894

was doing bot v bot tests to test the new ammo code.  Twice got an NPE that effectively ended the game.  Froze action.  Tried to kick and replace bots.  No joy.  Had to close.  Was able to load autosave and proceed for a few more turns, but it froze again.  The map is 2x3 Rolling Hills 2. No water on the map at all. 

image

A save and log attached.  Was the same error message both times in log. 
 
[megamek - Copy.log](https://github.com/MegaMek/megamek/files/15104412/megamek.-.Copy.log)
[Round-7-autosave_2024-04-24_22-47-29.sav.gz](https://github.com/MegaMek/megamek/files/15104414/Round-7-autosave_2024-04-24_22-47-29.sav.gz)
[Customs41424.zip](https://github.com/MegaMek/megamek/files/15104422/Customs41424.zip)

HoneySkull:
Fixes #5417 

Moved `processGameEvent()` back into `setPhase()` where it was.  `setPhase()` gets called from places currently that `receivePhase()` doesn't get called from.  It looks like this was pulled out as a refactor but has un-intended side effects for the bots processing.

image

Before making this tweak, I was able to easily reproduce the NPE by running a bot vs bot as described in #5417.   After this change, the bot games end naturally after 15 to 25 rounds.   Before this tweak, the same bot games would either end in a NPE or infinite mindless bot moves that would go into over 100 rounds.

### DIFF
--- a/megamek/src/megamek/common/Game.java
+++ b/megamek/src/megamek/common/Game.java
@@ -674,11 +674,11 @@ public class Game extends AbstractGame implements Serializable, PlanetaryConditi
     public void receivePhase(GamePhase phase) {
         final GamePhase oldPhase = this.phase;
         setPhase(phase);
-        processGameEvent(new GamePhaseChangeEvent(this, oldPhase, phase));
     }
 
     @Override
     public void setPhase(GamePhase phase) {
+        final GamePhase oldPhase = this.phase;
         this.phase = phase;
         // Handle phase-specific items.
         switch (phase) {
@@ -707,6 +707,8 @@ public class Game extends AbstractGame implements Serializable, PlanetaryConditi
                 break;
             default:
         }
+
+        processGameEvent(new GamePhaseChangeEvent(this, oldPhase, phase));
     }
 
     public void processGameEvent(GameEvent event) {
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/Game.java
```diff
@@ -674,11 +674,11 @@ public GamePhase getPhase() {
     public void receivePhase(GamePhase phase) {
         final GamePhase oldPhase = this.phase;
```
kuronekochomusuke, line 675: should this be remove since oldPhase is no longer used

HoneySkull: Yes that should be removed. - I saw that after I committed it.

kuronekochomusuke (approved):

PR #5431, opened 2024-04-28
Generic BV 

AaronGullickson:
This PR creates a `getGenericBV()` abstract method on entity that is then implemented in each inheriting class. The idea of "generic BV" is to calculate the average BV of each unit by unit type and weight. This generic BV can then be used to balance forces when generating opposing forces. It essentially provides a generic score for each unit by unit type and weight that does not include information about unit quality or pilot quality, allowing for more diversity in the types of units that players can face in various game modes and avoiding problems like being spammed by regiments of Wasps and Stingers. The first implementation will be to have AtB/Stratcon and Story Arcs use this as a (default) balancing method. 

I sought a statistical solution to figure out what the average BV should be, since only some unit classes are in certain weight increments. To do that, I used a statistical model to predict BV by weight for each unit type based on our database of units. The results indicated that an elasticity model worked well for all units. The elasticity model is given by:

$$\log(\hat{bv}_i) = \beta_0+\beta_1*log(weight_i)$$

In this model, we take the natural log of both the dependent and independent variable. The slope ($\beta_1$) in this model can be directly interpreted as the percent change in BV for a one percent increase in tonnage. In short, the model estimates relative change on relative change rather than absolute change on absolute change. More information about the procedure can be found at https://gleeful-croquembouche-32c070.netlify.app/. 

Each elasticity model produces a slope and intercept value. To convert these values back into predicted BV, we have to use a transformation since this is a non-linear model. In each case, the approach is:

```java
 Math.round(Math.exp(B0 + B1*Math.log(getWeight())))
```

Where B0 and B1 are replaced by the appropriate intercept and slope for that uni type, respectively.

The one exception to this approach is gun emplacements which do not have a weight. I therefore just assigned them the mean BV across all instances of gun emplacements in our database.

### DIFF
--- a/megamek/src/megamek/common/AeroSpaceFighter.java
+++ b/megamek/src/megamek/common/AeroSpaceFighter.java
@@ -100,7 +100,7 @@ public class AeroSpaceFighter extends Aero {
     }
 
     @Override
-    protected int getGenericBattleValue() {
+    public int getGenericBattleValue() {
         return (int) Math.round(Math.exp(3.729 + 0.898*Math.log(getWeight())));
     }
 }
--- a/megamek/src/megamek/common/AeroSpaceFighter.java
+++ b/megamek/src/megamek/common/AeroSpaceFighter.java
@@ -98,4 +98,9 @@ public class AeroSpaceFighter extends Aero {
     public long getEntityType() {
         return super.getEntityType() | Entity.ETYPE_AEROSPACEFIGHTER;
     }
+
+    @Override
+    protected int getGenericBattleValue() {
+        return (int) Math.round(Math.exp(3.729 + 0.898*Math.log(getWeight())));
+    }
 }
--- a/megamek/src/megamek/common/BattleArmor.java
+++ b/megamek/src/megamek/common/BattleArmor.java
@@ -1977,4 +1977,9 @@ public class BattleArmor extends Infantry {
     public int firstArmorIndex() {
         return 1;
     }
+
+    @Override
+    protected int getGenericBattleValue() {
+        return (int) Math.round(Math.exp(3.157 + 1.514*Math.log(getWeight())));
+    }
 }
--- a/megamek/src/megamek/common/BattleArmor.java
+++ b/megamek/src/megamek/common/BattleArmor.java
@@ -1979,7 +1979,7 @@ public class BattleArmor extends Infantry {
     }
 
     @Override
-    protected int getGenericBattleValue() {
+    public int getGenericBattleValue() {
         return (int) Math.round(Math.exp(3.157 + 1.514*Math.log(getWeight())));
     }
 }
--- a/megamek/src/megamek/common/ConvFighter.java
+++ b/megamek/src/megamek/common/ConvFighter.java
@@ -133,7 +133,7 @@ public class ConvFighter extends AeroSpaceFighter {
     }
 
     @Override
-    protected int getGenericBattleValue() {
+    public int getGenericBattleValue() {
         return (int) Math.round(Math.exp(2.943 + 0.795*Math.log(getWeight())));
     }
 }
--- a/megamek/src/megamek/common/ConvFighter.java
+++ b/megamek/src/megamek/common/ConvFighter.java
@@ -131,4 +131,9 @@ public class ConvFighter extends AeroSpaceFighter {
     public long getEntityType() {
         return Entity.ETYPE_AERO | Entity.ETYPE_CONV_FIGHTER;
     }
+
+    @Override
+    protected int getGenericBattleValue() {
+        return (int) Math.round(Math.exp(2.943 + 0.795*Math.log(getWeight())));
+    }
 }
--- a/megamek/src/megamek/common/Dropship.java
+++ b/megamek/src/megamek/common/Dropship.java
@@ -761,7 +761,7 @@ public class Dropship extends SmallCraft {
     }
 
     @Override
-    protected int getGenericBattleValue() {
+    public int getGenericBattleValue() {
         return (int) Math.round(Math.exp(6.5266 + 0.2497*Math.log(getWeight())));
     }
 }
--- a/megamek/src/megamek/common/Dropship.java
+++ b/megamek/src/megamek/common/Dropship.java
@@ -759,4 +759,9 @@ public class Dropship extends SmallCraft {
                 || conditions.getAtmosphere().isLighterThan(Atmosphere.THIN);
         return spheroidOrLessThanThin && (getCurrentThrust() > 2);
     }
+
+    @Override
+    protected int getGenericBattleValue() {
+        return (int) Math.round(Math.exp(6.5266 + 0.2497*Math.log(getWeight())));
+    }
 }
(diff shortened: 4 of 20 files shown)


AaronGullickson:
Forgot to add that the R code for the analysis is now also a repo in this organization at https://github.com/MegaMek/bv-analysis. That should make it easy to update in the future as new units are added.

SJuliez (approved):

SJuliez:
Code is ok. But it does mean you have to create or have an Entity to get at the generic BV. The values could, alternatively, be returned from 

```
public class GenericBV {

  getGenericBV(long eType, double weight) { 
    switch (eType) ... 
  }

  getGenericBV(Entity entity) { 
    return getGenericBV(entity.getEntityType(), entity.getWeight);
  }
}
```

Then, Entity would just have to `return GenericBV.getGenericBV(this);`

This would pool the values in one place, which I could imagine to be better for maintenance. But it would upend most of your code changes and be a bit of effort getting the etypes right, so I'm leaving this as a suggestion.

SJuliez:
Also, probably subjective, but the formula is simpler as

$$genericBV = A * weight^\{beta_1}$$
i.e. `return (int) Math.round(41.6 * Math.pow(getWeight(), 0.898);` for Meks

AaronGullickson:
> Code is ok. But it does mean you have to create or have an Entity to get at the generic BV. The values could, alternatively, be returned from
> 
> ```
> public class GenericBV {
> 
>   public int getGenericBV(long eType, double weight) { 
>     switch (eType) ... 
>   }
> 
>   public int getGenericBV(Entity entity) { 
>     return getGenericBV(entity.getEntityType(), entity.getWeight());
>   }
> }
> ```
> 
> Then, Entity would just have to `return GenericBV.getGenericBV(this);`
> 
> This would pool the values in one place, which I could imagine to be better for maintenance. But it would upend most of your code changes and be a bit of effort getting the etypes right, so I'm leaving this as a suggestion.

My general preference is for an object oriented approach over static classes, although I get the point about maintenance.  I don't think we will need to update this very frequently. The more likely change will come if we create new entity types or split entity types (e.g. IndustrialMechs), in which cases the need to implement the abstract method will be a good reminder. 

> Also, probably subjective, but the formula is simpler as
> 
> genericBV=A∗weightbeta1 i.e. `return (int) Math.round(41.6 * Math.pow(getWeight(), 0.898);` for Meks

Thats a good point. I think I prefer being able to just enter the intercept and slope values directly without having to first exponentiate the intercept - its less likely to lead to a transcription error.

Thom293:
So, I dont know full plans for this, but if I could comment on one thing.  There has been great concern about difficulty recently and what some consider unreasonable others consider reasonable.  So, to whatever extent this is implemented, I woud request that the caps/ranges be customizable so all types can enjoy how they wish. 

Thank you for your consideration.  Will explain in more detail below.

Some people claim they dont like to be over bv or outnumbered.  Another described it with words to the effect of "Its like a day job now.  I havent lost a mech in ages."  

I tried to take both camps into account when adding modifiers and just flat out made them easier, but with more variety.  I cut chance for the most difficult modifers by 50% and reduced the chance of a "multi-unit" negative modifier by 20%.  In exchange, the variety of modifiers has increased and the number of "mid-tier" modifiers has increased.  

But even with that, some people dont like the enemy to be even 1 bv over or one unit more than their force.  They want to curbstomp the OpFor.  This is a valid pursuit and desire.  But just as equally, some people (like me) think Princess needs a 50% advantage to pose a real challenge.  Personally, I dont even fight near equal bv battles anymore - I just give control of my units to princess and the whole battle is done in 3 mins.  Even without uber elite pilots, an intelligently designed player lance with half-decent mechs under princess control can beat a larger bv randomly generated Princess unit easily.  And once you get into the age of case there is almost no chance of losing a mech.

Another Example from Joker's stream - where they do this very often.  Here is an OpFor that has more than double the units and 14k more BV and they won easily.  

image

IllianiCBT:
> The one exception to this approach is gun emplacements which do not have a weight. I therefore just assigned them the mean BV across all instances of gun emplacements in our database.

Will this need to be updated once Raoz's turret update gets merged?

Regarding Thom's comments. This is more of a mhq thing, but an easy solution could be to include a Campaign Option that increases Princess' GBV allowance by x%. Potentially tying x% to the AtB Difficulty combobox.

HammerGS:
> Another Example from Joker's stream - where they do this very often. Here is an OpFor that has more than double the units and 14k more BV and they won easily.

A big part of the this comes from heavily SPA equipped pilots that don't have a BV cost, if you take 2 identical lances and give one Elite Pilots with lots of SPA's and the other Regulars the outcome is a given. Without a penalty to SPA's it's easy to rack up huge modifiers that make things a cake walk. 

One critical weakness is Princess herself, once you learn her min-max approach you can easily ambush her turn over turn. One day she will be able to plan turn over turn (Duchess) and that will improve things. 

A few thoughts:
1. CGL is aware of the no BV for SPA's or Quirks and has a long term plan to address this. 
2. We need to get away from masses of 5P 6G or 5P 5G Opfors and bring them into line skill wise with what the player has.
3. We need to evolve to having an Opfor the challenges the player but fits the narrative of the campaign. 

I agree we need ranges to create difficulties but in the word of BattleTech combat should always be about 1 to 1 ratio combat, it's war and sometime you're outgunned other times under gunned.

RaozSpaz:
> The one exception to this approach is gun emplacements which do not have a weight. I therefore just assigned them the mean BV across all instances of gun emplacements in our database.

This is just to determine allotments right? Like how many a place can have at most? If so then I think a mean value should be fine. Otherwise yeah, my newer turrets I do have them split into categories that you could use as a rough guide for what their score could be. Could discuss it with you.

@Thom293 Im not sure the problem is so much as being over BVed by just 1. Ive never heard people complain only about small margins. The complaints are typically people seeing BVs upwards of 20+% over their BV. And that isn't even including the modifiers being tacked onto it. This can be especially bad in cases of players who are new to HQ and aren't readily aware of how regular it can be to call in reinforcements and the like. 

And then much like Hammer pointed out, a lot of Joker's crew are very skilled members. Veteran or Elite Pilots stacked with SPAs. They shoot a target and it typically dies. Just look how many of them are melee mechs equipped with Melee Master. The only time they have issues is when someone gets focused down or a lucky headshot happens. Thats why I do think Thark has a good train of thought here where the difficulty can be more tied to the difficulty of the mission.

Thom293:
The reason is irrelevant. Otherwise you get into the "yeah buts" he is playing with only tanks and flimsy VTOLs, or he is playing in an early timeline, or he has 7 unit lances with 0/1 pilots, or he is new, or he is a vet, or he doesnt use SPAs, or He uses Edge, Or he uses Direct Blow, FABs, Rapid Fire AC, Skin of the Teeth, etc., etc., etc...


Allowing people to customize removes all of that.  Allowing the merge of the already coded should-already-be-working difficulty setting and the was-always-intended customizable modifier # would solve 95% of these issues at least in the short term and probably in the long term. And it would not impede rollout of generic BV or any other system. I Still don't understand why an already existing solution to this debate was halted.  Better to keep having the argument for perpetuity, I guess.

PR #5436, opened 2024-04-29
lightning storm updates

kuronekochomusuke:
- update lightning storm effects 
- resolve end turn lightning storm damage
- option to target units

image


kuronekochomusuke:
what do you think about a option to randomly hit a spot containing a unit?  I implemented it as a random hex on the board, and with a large map it is not likely to hit anything.  the rule says using any manner the players agree on for the locations.

HammerGS:
I like it. Mechs as lightning rods.

megamek/src/megamek/server/totalwarfare/TWGameManager.java
```diff
@@ -31253,6 +31268,98 @@ public List<SmokeCloud> getSmokeCloudList() {
...
+        vFullReport.addAll(newReports);
+
+        Building bldg = game.getBoard().getBuildingAt(coords);
+        if (bldg != null) {
+            Vector<Report> buildingReport = damageBuilding(bldg, damage, coords);
+            vFullReport.addAll(buildingReport);
+        }
+
+        List<Entity> hitEntities = game.getEntitiesVector().stream()
+                .filter(e -> e.getPosition().equals(coords)
```
SJuliez, line 31350: getPosition can return null, better turn this comparison around

SJuliez: also, should the lightning storm disregard offboard units?

megamek/src/megamek/server/totalwarfare/TWGameManager.java
```diff
@@ -31253,6 +31268,98 @@ public List<SmokeCloud> getSmokeCloudList() {
...
+                        && !(e instanceof GunEmplacement))
+                .collect(Collectors.toList());
```
SJuliez, line 31352: 31352 can be replaced with .toList()

megamek/src/megamek/server/totalwarfare/TWGameManager.java
SJuliez: why not gun emplacements?

SJuliez (changes requested):
please see the comments

megamek/src/megamek/server/totalwarfare/TWGameManager.java
kuronekochomusuke: by default it only targets a on board hexes.  
adding off board hexes would make hitting anything very rare, since you can have units sheets away from the main board.

if you use the option to target units, it will target off board units also if they have a posiition.

for gun emplacements I was thinking the building takes the damage.  that is what it looks like it does for other damage types.

image
image

HammerGS:
Let me know if the changes are done. Aiming to get some merged in for the nightly.

kuronekochomusuke:
it is done

PR #5440, opened 2024-05-01
MM fix for MML issue where Partial Wing not accounted for

### LINKED ISSUE #5439: BA Jump MP bonus from Partial Wing only visible in-game with suitable atmosphere
### Environment
MHQ combined build 0.49.19 and .20
Kubuntu Linux 22.04.4
Java Vendor: Eclipse Adoptium
Java Version: 17.0.8.1

### Description
After some recent refactoring, the Partial Wing bonus to Jump MP for BAs is only shown when A) in a game, with B) a viable atmosphere.

This is correct behavior _for in-game BAs_, but breaks the MML UI by hiding the Jump MP bonus (or rather, not adding it at all) outside of actual games.

### Files
image
image

Sleet01:
Adds PW Jump MP accounting outside of games so MML stats are correct.

Testing:
- Ran MM and MML unit tests (MHQ currently broken)
- Tested with Kage suits in MML and verified correct Jump MP with and without Partial Wing
- Tested with Kage suits in MM and verified correct Jump MP with atmosphere and without.

Close #5439

### DIFF
--- a/megamek/src/megamek/common/BattleArmor.java
+++ b/megamek/src/megamek/common/BattleArmor.java
@@ -528,6 +528,10 @@ public class BattleArmor extends Infantry {
                     && ignoreGameLessThanThin) {
                 mp++;
             }
+        } else {
+            if ((mp > 0) && hasWorkingMisc(MiscType.F_PARTIAL_WING)) {
+                mp++;
+            }
         }
 
         if ((mp > 0)
--- a/megamek/src/megamek/common/BattleArmor.java
+++ b/megamek/src/megamek/common/BattleArmor.java
@@ -519,6 +519,7 @@ public class BattleArmor extends Infantry {
             mp++;
         }
 
+        // MM is concerned with in-game conditions that impact Partial Wing mp
         if ((game != null)) {
             PlanetaryConditions conditions = game.getPlanetaryConditions();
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/BattleArmor.java
```diff
@@ -528,6 +528,10 @@ && hasWorkingMisc(MiscType.F_PARTIAL_WING)
                     && ignoreGameLessThanThin) {
                 mp++;
             }
+        } else {
+            if ((mp > 0) && hasWorkingMisc(MiscType.F_PARTIAL_WING)) {
+                mp++;
+            }
         }
```
SJuliez, line 535: can't that whole block starting at 522 be written as 
```
if ((mp > 0) && hasWorkingMisc(MiscType.F_PARTIAL_WING) && 
  (mpCalculationSetting.ignoreWeather || (game==null) || game.getPlanetaryConditions().getAtmosphere().isLighterThan(Atmosphere.THIN))) {
  mp++;
}
```
?

kuronekochomusuke: I think that is harder to read and easier for some one to break things when updating something.

Sleet01: I tend to agree with @kuronekochomusuke on this one, although I think adding a comment to clearly state which branch is needed for which project would be a good move.

PR #5441, opened 2024-05-01
MM's symbol font for some map indicators

SJuliez:
I noticed that on my selected move font the symbols for CF warning and flight path indicators were tofu and I couldnt find any font they worked with except the java logical serif. So this PR changes the symbols to some of the symbol font that is part of MM. This way they'll work for everyone and are independent from the move font. Additionally StringDrawer is upgraded to allow absolute Y centering without regard for ascent/descent which performs better for this use case (single symbols).
For the flight path symbols I thought I'd use arrows that are slightly more self-explanatory. Also the symbols are larger and less transparent so they show up well over complicated terrain or when zoomed out.

image

image

### DIFF
--- a/megamek/i18n/megamek/client/messages.properties
+++ b/megamek/i18n/megamek/client/messages.properties
@@ -1166,6 +1166,7 @@ CommonSettingsDialog.colors.MoveMASCColor=Move MASC Color
 CommonSettingsDialog.colors.MoveRunColor=Move Run Color
 CommonSettingsDialog.colors.MoveSprintColor=Move Sprint Color
 CommonSettingsDialog.colors.PrecautionColor=Precaution Color
+CommonSettingsDialog.colors.OkColor="OK" Color
 CommonSettingsDialog.colors.PlanetaryConditionsBackgroundTransparency=Planetary Conditions / Keybinds Background Transparency
 CommonSettingsDialog.colors.PlanetaryConditionsBackgroundTransparency.tooltip=0-256
 CommonSettingsDialog.colors.PlanetaryConditionsColorBackground=Planetary Conditions / Keybinds Background Color
--- a/megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
+++ b/megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
@@ -167,6 +167,7 @@ public class CommonSettingsDialog extends AbstractButtonDialog implements ItemLi
     private ColourSelectorButton csbWarningColor;
     private ColourSelectorButton csbCautionColor;
     private ColourSelectorButton csbPrecautionColor;
+    private ColourSelectorButton csbOkColor;
     private ColourSelectorButton csbMyUnitColor;
     private ColourSelectorButton csbAllyUnitColor;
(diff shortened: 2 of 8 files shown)


HoneySkull:
I kind of wondered if changing the font for the movement text would impact the special characters. But when I did my testing, changing the font didn't seem to change my rendering for these indicators.

At first glance the arrows appear visually noisy, but then again, I've been staring at the dots for a few months now.  (When I originally had this idea my thought was to use arrows, but used directionally neutral circles to prototype to see how it worked and found that it felt clean ((but less intuitive))).

HammerGS:
Just my opinion, but I really like the arrows. I feel it makes it clear you can do something. Building of the coloring in the original.

HoneySkull:
Maybe render the flag always pointing north.
image

HoneySkull:
I like the grey arrow for flying off the map - but can we render the excess movement always facing north so it's more readable?

image

HoneySkull:
I initially made the no-turn (red open circles) hollow so that it was easily discernable from the green for red/green color deficiency.  I think this work, because there will never be a green single arrow.

HoneySkull (approved):
I like the StringDrawer improvements. :)

gsparks3:
> I initially made the no-turn (red open circles) hollow so that it was easily discernable from the green for red/green color deficiency. I think this work, because there will never be a green single arrow.

Is there any marker that could be added to distinguish "powered turn" vs "free turn" as well? Just got confirmation from a friend with such issues that the yellow and green remain hard to distinguish. Not sure what would fit without cluttering things up too much / being intuitively understandable, though (the arrows are absolutely _perfect_ for "can turn" vs "can't turn").

For the final-position flag and excess movement text, I'd agree with having them both rendered so they always face north if possible (especially the latter, for obvious reasons).

SJuliez:
OK appreciate the comments. 
- Thought it cluttersome too. I have a new proposal
- Not rotating flag: Absolutely
- The fly-off arrow being gray was not intentional, forgot to add the color. It is always green (I think it was that way before)
- There are the Warning Color and Caution Color options in the client settings. I added an "OK Color" and made use of these colors. So, players should now be able to change the colors to those with enough contrast (see screenshots) if green/yellow/red doesn't work
- The lone arrow is no longer reserved for "must go straight". I think this is better and the color config should make it work.


image
image
image

image

image

An example where it may immediately turn:
image

Seems to work in space:
image

gsparks3:
Hmm. Technically I think the "free turn" indicator shouldn't show up at all in space, since you don't get free turns, but otherwise that looks great!

HoneySkull:
@SJuliez Thank you for taking such great care of my baby!  You've turned my very cool feature into an even cooler feature.  Great work!  I can't wait for people to love this!

HoneySkull (approved):

SJuliez:
The dueFreeTurn() method did not consider if the unit is in space, this is done outside that method for path finding, but not for the flight path indicators. I couldnt see a reason why a unit in space should ever report that it could have a free turn and added that restriction. Seemed to work fine and now it doesnt show a free turn indicator in space.

PR #5444, opened 2024-05-02
BoardView Sprite separation

SJuliez:
This continues my "quest" to relieve the BoardView of parts that are specific to TW game play and overall make it lighter and be able to display a board for other game types.

This PR:
- moves the generation of some types of sprites out of BoardView to separate handlers 
- the handlers are kept by the ClientGUI as the GUI central. The *Displays forward their demands for sprites not to the boardview but the ClientGUI which then addresses its handlers
- I stopped at this point so that the PR doesnt become any bigger and to avoid picking up merge conflicts; it doesnt address all sprites yet
- I renamed ConstructionFactorWarning to CollapseWarning only to shorten it
- As there can sometimes be a high number of sprites and sprites are no longer by default kept in their own lists some special treatment is necessary as can be seen in BoardView's behindTerrainSprites etc.; if additional performance problems show up, there'll be a solution
- Note that the changes from the other PR for the flight path indicator symbols take precedence over this PR's changes; I decided to make those their own PR; improvements are in the other PR
- ideally this PR should have no visible effect.


kuronekochomusuke:
should be sprite classes be moved into a sprite package?

SJuliez:
> should be sprite classes be moved into a sprite package?

There's a lot of package-private access used in the sprites. I just tried moving them all into their own package. The IDE told me there'd be 176 access errors and, unhelpfully, it did not offer to auto-correct this. So I guess this will have to be its own PR :)

It is also my experience that renames and moves are actually best kept separate from other stuff. So, another time.

kuronekochomusuke (approved):

PR #5452, opened 2024-05-06
AS advanced search and rear weapon conversion

### LINKED ISSUE #5447: Advanced Search By Alpha Strike ability ART
### Environment
49.19

### Description
Attempting an advanced search to filter by Artillery Special Abilities.  Was looking for Dropships with artillery.  Such as the Fortress.

Systematically going down the list of ART options, No results were returned for Small Craft or Dropships which I know have Artillery Systems on them, plus all the ones I didn't know about which I was looking for.

I suspect the problem is related to these units using a different format to display damage in different firing arcs.

PS.  The Planet Lifter Tactical Support conventional Fighter has something going on with it.  It has an Arrow IV (Mounted in the rear? is that the problem?), but its not showing up on the Card automatically.

SJuliez:
Fixes #5447 
This changes conversion to count specials from rear-facing weapons such as the Arrow IV on the Planetlifter Tactical Support for the standard specials list (instead of trying to place them in the non-existent REAR specials). Also affects NARC, C3M and other weapon-based specials.
Also the AS advanced search now also searches for specials in the arcs of large aero units, so a search for ARTLT finds the Fortress Dropship.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/unitSelector/ASAdvancedSearchPanel.java
+++ b/megamek/src/megamek/client/ui/swing/unitSelector/ASAdvancedSearchPanel.java
@@ -32,7 +32,6 @@ import megamek.common.alphaStrike.BattleForceSUA;
 import javax.swing.*;
 import javax.swing.border.EmptyBorder;
 import java.awt.*;
-import java.awt.event.ActionListener;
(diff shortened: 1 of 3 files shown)


Sleet01 (approved):
LGTM

PR #5454, opened 2024-05-07
Experimental option

SJuliez:
I find myself sometimes wishing that MM not automatically select a unit for me whenever a turn for me comes around as this centers the map on that unit and makes it difficult to see e.g. Princess movements. This adds an option in the Client Settings to turn this behavior off (on by default).


Sleet01 (approved):
LGTM

PR #5458, opened 2024-05-09
Ignore armortype slots in other-equipment cost calculation

SJuliez:
Fixes MegaMek/megameklab#1502


HammerGS (approved):
Tested and working.

PR #5465, opened 2024-05-12
Bay ammo correction

SJuliez:
- Fixes an obvious error that made probably all large aeros invalid by not finding their bay ammo. 
- Fixes (I guess, I don't see how this could have worked before) a test class. I moved around stuff in there but the main change is calling EquipmentType.initializeTypes() before the tests.


Sleet01 (approved):
LGTM
It probably worked previously because the Dropship BLK strings would load even without the equipment being initialized, because cargo bays weren't "equipment" per se.
I imagine that things are stricter now, after all the updates in .20.

PR #5467, opened 2024-05-13
Modernize board validity checking code

SJuliez:
This changes board/hex/terrain validation to use a list of error Strings instead of a StringBuffer. This way the resulting error list can be formatted as needed (usually with a String.join) at the receiver instead of being preformatted. Line breaks are unnecessary. Also shortens code as a separate validity boolean is not needed.
The list would also lend itself to using a report line class with more functionality than the simple String report line if we wanted that.
Also corrects a board with errors.
Also compacts terrain checking a bit and imports the Terrains.* constants in Terrain to simplify code.


Sleet01 (approved):
Looks good to me!

PR #5481, opened 2024-05-17
Name Changes Updates

SJuliez:
Adds a further test to the NameChangesValidator (checks to see if an entry has equal left and right side making it useless) and adds some comments. Also cleans up name-changes.txt, removing all problems found with the validator.


Sleet01 (approved):
LGTM

PR #5487, opened 2024-05-19
Accessibility window cleanup and safety

SJuliez:
As the title says. Seeing that we got reports of exceptions in AccessibilityWindow (even though it possibly wasnt even used) I tried to make it as safe as possible. The main point is, probably, to surround creating the description of an (attack) action with a try-catch. 
@Sleet01 I remember you making changes and I believe I talked you out of the try catch. I disgree with myself now and retroactively agree fully to having a try catch to *really* prevent any exceptions from creating a text line. I placed it in AccessibilityWindow itself so the burden isnt on the code in the Actions and it cant be forgotten. An argument can probably be made to do the same in EntityActionLog for that other description method in EntityAction. Another argument might be made for creating the AccessibilityWindow only when the player calls it up, to not have a game listener in place when no one needs it. Another time.

Also updates the code in EntityAction and AbstractEntityAction with defaults and comments.


Sleet01 (approved):
Looks good to me!

PR #5510, opened 2024-05-25
Add test for invalid canon units

SJuliez:
Updates the Cache test to also test for invalid canon units. Currently fails pretty hard.


SJuliez:
The CI failures are due to this test failing.

HammerGS (approved):
Seems to work as I understand it.

PR #5513, opened 2024-05-26
DS loading, issue 5495

### LINKED ISSUE #5495: [0.49.19.1] Bug: Dropships with full mech bays become invalid when deployed on the ground due to passengers
Bundled Megamek with MekHQ 0.49.19.1

When I load a dropship with mechs and deploy it on the ground map at zero elevation the design becomes invalid due to passengers, presumably corresponding to the number of mechwarriors. The invalid status persists when the mechs are removed and the Dropship deployed airborne.

image

image

image

image

SJuliez:
This removes code from GameManager that added small unit (mek) crew to an aero craft's passenger list when loading the small unit. This has been present for > 6 years but seems wrong as such crew doesn't require passenger quarters, as the bay has accomodations for such crew. Real passengers on the other hand do require accomodation, so the crew quarters validation code should indeed count them.

Also adds a test that has the GameManager load an Atlas onto a Leopard. Fails with the code previous to this PR.

Fix #5495

### DIFF
--- a/megamek/src/megamek/server/GameManager.java
+++ b/megamek/src/megamek/server/GameManager.java
@@ -3687,14 +3687,6 @@ public class GameManager extends AbstractGameManager {
         // set deployment round of the loadee to equal that of the loader
         unit.setDeployRound(loader.getDeployRound());
 
-        // Update the loading unit's passenger count, if it's a large craft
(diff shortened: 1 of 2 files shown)


Sleet01 (approved):
LGTM!

Issue #5509, opened 2024-05-25
Invalid units (not data)


SJuliez:
49.20 (in IDE)

The following units currently register as invalid when rebuilding the cache (meaning: not in the lobby or anything. This is in addition to reports of DS being invalid when having loaded meks)

Oni Battle Armor [MRR](Sqd4)
Oni Battle Armor [MRR](Sqd5)
Oni Battle Armor [MRR](Sqd6)
Oni Battle Armor [PPC](Sqd4)
Oni Battle Armor [PPC](Sqd5)
Oni Battle Armor [PPC](Sqd6)
Clan Interface Armor (Sqd1)
Delta Air Cruiser MA-80
Tsuru VIP Aircraft
Drone M-3
Nekohono'o (HQ)
Czar Dropship
Scytha A
Centurion CNT-1A
Morgenstern MR-1SE
Tatsu MIK-OF
Aquilla Transport Jumpship
Cudgel CDG-1B
Zeus-X ZEU-X
Wasp LAM WSP-110
Uni ATAE-70 ArtilleryMech
Svartalfa 3
Escape Pod
Life Boat
Intrepid Assault Craft (2331)
Intrepid Assault Craft (2478)
Air Car
Ground Car
Heavy Transport B1
Jeep
Jet Sled
Speeder
Bandit (C) Hovercraft G
Kestrel VTOL (ML)
Peregrine Attack VTOL (Cargo)
Saladin Assault Hover Tank (Clan Cargo)
Flatbed Truck (Armor)
Flatbed Truck (SRM)
Ajax Assault Tank C
Glaive Medium Tank (MFB)
Aeron Strike VTOL (BAP)
Balac Strike VTOL (LRM)
J-37 Ordnance Transport
Prowler Multi-Terrain Vehicle (ECM)
Prowler Multi-Terrain Vehicle (Sealed)
Prowler Multi-Terrain Vehicle (Succession Wars)
Prowler Multi-Terrain Vehicle
Trireme Infantry Transport (CS)
Zugvogel Omni Support Aircraft A
Zugvogel Omni Support Aircraft C
Zugvogel Omni Support Aircraft D 'Raubvogel'
Zugvogel Omni Support Aircraft E
Zugvogel Omni Support Aircraft Prime
Drillson Heavy Hover Tank (Sealed)
Moray Heavy Attack Submarine (Original)
Bishop Transport VTOL
Zugvogel Omni Support Aircraft F
Shackleton AESV
[...]

PR #5520, opened 2024-05-28
Cargo loading on SV

### LINKED ISSUE #5515: [0.49.19.1] Bug: Megamek incorrectly loading support vehicles with Cargo Containers as invalid
Bundled Megamek with MekHQ 0.49.19.1

Reviewing invalid cargo units in Megamek I corrected and resaved valid versions. After refreshing the cache the corrected support vehicles with cargo containers still loaded as invalid and overweight with erroneous cargo transport bays.

[Test Cargo Container.zip](https://github.com/MegaMek/megamek/files/15447590/Test.Cargo.Container.zip)

This fresh unit was saved as valid with cargo containers. When loaded from cache or the file Megamek adds extra equipment transport bays, although the actual unit data file is still correct.

image

image

image

SJuliez:
Fixes #5515 

This removes code that tried to add cargo bays for F_CARGO equipment. I assume this was added to facilitate the internal bomb bay quirk. Cargo bays count towards weight and cannot be added in this way. I played around with this quite a bit and found that it was somewhat ineffective on units that can possibly have IBB (DS, SC, FWS, AF and CF). DS and SC do not use the that part of the code. F_CARGO is only on Cargo Containers (10t) and variable size cargo bays. For cargo bays, adding more cargo bays seems redundant (and didnt even work for a reason I couldn't see). For cargo containers, they seem not accessible to IBB with or without this code. I suppose we have to find another way to allow IBB in containers.

This makes a handful of SV canon units valid again, see #5509

Also (helpfully) adds transports to the output of DebugEntity.

### DIFF
--- a/megamek/src/megamek/common/loaders/BLKFile.java
+++ b/megamek/src/megamek/common/loaders/BLKFile.java
@@ -320,11 +320,6 @@ public class BLKFile {
                                 * ((InfantryWeapon) mount.getType()).getShots());
                             mount.getLinked().setShotsLeft(mount.getLinked().getOriginalShots());
                         }
-                        if (etype.hasFlag(MiscType.F_CARGO)) {
-                            // Treat F_CARGO equipment as cargo bays with 1 door, e.g. for ASF with IBB.
-                            int idx = t.getTransportBays().size();
-                            t.addTransporter(new CargoBay(mount.getSize(), 1, idx), isOmniMounted);
-                        }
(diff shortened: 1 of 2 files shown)


Sleet01 (approved):
LGTM.  I'd guess that the F_CARGO block in `BLKFile.java` was a remnant I should have deleted after the ASF BLK file loader got it added.

PR #5527, opened 2024-05-30
RATGen/force generator role handling updates

### LINKED ISSUE #5286: RAT generator is not adjusting relative frequencies for conventional infantry roles
The RAT generator has the capability of using roles (not the standard lance-based roles like Striker/Fire/Command/etc. from Campaign Operations and other sources) to help define what units do based on data from the era availability XMLs.  MissionRole.Java has a number of functions that deal with these, including one to help adjust relative frequencies based on roles provided in the XML data files - adjustAvailabilityByRole

https://github.com/MegaMek/megamek/blob/cf71c486c33c6790392aa7f8917b56023ea845ad/megamek/src/megamek/client/ratgenerator/MissionRole.java#L155-L157

This function is missing code to handle requests for many roles associated with conventional infantry, specifically:

- marine
- xct
- mountaineer
- paratrooper
- anti-mek (anti-Mech infantry)

The adjustAvailabilityByRole function changes the availability rating (relative frequency) of various units based on how they match a set of provided roles, or what roles they carry if no roles are provided.  What is missing, is anything to process the indicated roles in the same manner as ones it already does.  So when using the RAT generator for infantry it doesn't matter much of what role you select as it's the same either way.  This applies to the RAT Generator tab as well as the Force Generator tab (which is how I came to look for this - the faction rules XMLs were not reliably generating specialist infantry like paratroopers or anti-Mech platoons, and the RAT Generator was confirming that with the relative availabilities).

The solution will involve adding a few more case entries under the top-most switch statement at [line 170](https://github.com/MegaMek/megamek/blob/cf71c486c33c6790392aa7f8917b56023ea845ad/megamek/src/megamek/client/ratgenerator/MissionRole.java#L170), following the same pattern as the other main examples such as RECON or FIRE_SUPPORT.  These use the avAdj[] array to indicate a modifier to boost (add to) or reduce (subtract from) the standard availability, with each index of the array providing a sequentially larger modifier.  For example, handling the paratrooper role might look like this:

```
case PARATROOPER:
   if (mRec.getRoles().contains(PARATROOPER) {
      avRating += avAdj[2];
   } else {
      avRating -= avAdj[3];
   }
   break;
```

The second element of the array is consistent with the positive adjustment in other primary roles, and the third element is consistent with the fall-through which applies a more strongly negative adjustment for units which do not meet the role.

The ANTI_MEK block will need some extra handling, so that it returns null (means unit is unsuited for requested role) for certain infantry e.g. mechanized infantry type, or those equipped with encumbering advanced-rules armor (might deserve a new backing field such as 'canDoAntiMek' on ModelRecord object along with the others that are populated in the constructor).

SuperStucco:
The role handling for converting the force generator XMLs to RAT tables/generating random forces has a few holes that need patching, mostly around infantry roles and non-combat units.  There were also a few errors, such as limiting MINELAYER/MINESWEEPER to Mechs (rather than ground vehicles) and infantry.

closes #5286 

Role/unit type validation:
- adds missing SR_FIRE_SUPPORT and MIXED_ARTILLERY roles to unit type validations
- updates role/unit type validations to properly account for all unit type/role combinations
- organize and comment for readability (enum integer-equivalent less/greater than comparisons are pretty hard to follow at a glance unless you know all the values by heart)

Availability value modifications by role:
- adds missing roles for when specific roles are requested, mostly infantry but a few standard ones as well
- more extensive handling of non-combat roles (not needed now, but will be useful later)
- replace array index calls with named variables for readability
- relax application of role modifiers when no roles are called for, while still filtering out non-combat units
- organize and comment for readability

Non-combat/support unit identification:
- simplifies identification of non-combat support and civilian units

Role parser:
- adds missing MIXED_ARTILLERY role to string parser

Help file is updated to list the full set of roles, along with valid unit types and typical uses.

### DIFF
--- a/megamek/docs/RAT Stuff/rat-generator.txt
+++ b/megamek/docs/RAT Stuff/rat-generator.txt
@@ -161,12 +161,14 @@ ProtoMek
 VTOL
 Naval
 ConventionalFighter
-Aero
+AeroSpaceFighter
 Small Craft
 Dropship
 Jumpship
 Warship
- Mek, Tank, BattleArmor, Infantry, ProtoMek, VTOL, Naval, ConventionalFighter, Aero, Small Craft, Dropship, Jumpship, Warship. Needed to distinguish between different unit types that have the same name, such as Centurion or Eagle (both of which can apply to either a Mek or an ASF).
+Space Station
+
+This is required to distinguish between different unit types that have the same name, such as Centurion or Eagle (both of which can apply to either a Mek or an ASF).
 
 'omni' attribute: can be either 'Clan' or 'IS'. For non-omnis it is absent. Needed to distinguish between omni and fixed-configuration variants, since omnis taken as salvage/isorla use the operating faction to determine which configuration is used. Clan and IS must be stated to distinguish between the Clan Battle Cobra and the ComStar copy.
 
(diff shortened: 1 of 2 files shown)


IllianiCBT (approved):
I'm not able to see any issues here. Glad to see some of the missing roles added and our coverage expanded.

IllianiCBT:
Never mind, I got so excited I failed to notice the PR is still in Draft. >_>

SuperStucco:
Just making sure there isn't any last minute changes prompted by a round of data changes, now ready for review.
This should fix #5286

IllianiCBT (approved):
LGTM (for real this time)

IllianiCBT:
> Just making sure there isn't any last minute changes prompted by a round of data changes, now ready for review. This should fix #5286

If you add 'closes #5286' somewhere in your PR description, GitHub will link the two PRs. :)

PR #5528, opened 2024-05-30
Infantry attacks with secondary weapons

### LINKED ISSUE #5506: [49.20] Issue : Some infantry squads are not being generated with their primary/secondary weapon as an attack option
### Environment
49.20
Windows
Java 11

### Description
image
As shown in the above image some infantry squads do not have their primary/secondary weapon as an attack option. When actually in battle this leads to an inability to use said weapons as an attack and in several cases render infantry entirely incapable of attacking. 

### Files
No files needed. Load up any megamek game or just the megameklab and look up the infantry.

SJuliez:
InfantryWeaponMounted was not a subclass of WeaponMounted but of Mounted<something> instead. I assume this was done to keep the IWM class simple and not influenced by WeaponMounted stuff that it would never need. Unfortunately this meant that infantry with secondary weapons got no weapon attacks, as IWMs were not entered into a unit's weapon list (and could not without bigger changes - List<WeaponMounted>). So, I made IWM a subclass of WeaponMounted and in nothing short of a miracle, it immediately compiled without any error and I could not find any errors when I set up a game where I shot at a target mek with two infantry units. We may of course find some error somewhere later because of this but I feel that mangling Entity's weaponList would not be the right way to go.

Fixes #5506

### DIFF
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -4563,6 +4563,23 @@ public abstract class Entity extends TurnOrdered implements Transporter, Targeta
     }
 
     /**
+     * Returns the CriticalSlots in the given location as a list. The returned list can be empty
+     * depending on the unit and the chosen slot but not null. The entries are not filtered in
+     * any way (could be null although that is probably an error in the internal representation
+     * of the unit.)
+     *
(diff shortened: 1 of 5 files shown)


HammerGS:
Seems to be failing on the CodeQL

Sleet01 (approved):
LGTM!

PR #5531, opened 2024-05-30
Allow Harjel on support vehicles

pavelbraginskiy:
TO:AUE page 100.

### DIFF
--- a/megamek/src/megamek/common/MiscType.java
+++ b/megamek/src/megamek/common/MiscType.java
@@ -6977,7 +6977,7 @@ public class MiscType extends EquipmentType {
         misc.tonnage = 1;
         misc.criticals = 1;
         misc.cost = 120000;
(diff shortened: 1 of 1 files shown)


pavelbraginskiy:
Tested for construction, not tested for gameplay.

HammerGS:
We need to ensure we capture the validation limits on this before merging. 

_BattleMechs and IndustrialMechs may mount a HarJel system in any body location(s) desired, except for the location containing the cockpit._ 

_Vehicles may mount HarJel systems in any location(s) desired except the Body and the Rotor._

pavelbraginskiy:
MML will not insert Harjel into the body of a support vehicle, and the blk file for a support vehicle with Harjell in the body fails to load. 

Since SupportTanks are Tanks, the TestEntity routine runs the Tank equipment location checks on Support Tanks as well, so that should flag SVs with body harjel as invalid as well if something somehow slips through.

SJuliez (approved):

Issue #5328, opened 2024-04-07
RFE:  Enable Edge Triggers by default0


Thom293:
Title.
For characters under bot control,  If edge is enabled you want them enabled.  If it isn't, they don't appear.  

I think you want to exclude MHQ player units since they will have their own settings.  But for bot units or generated in MM should be on by default I think.  

Will do a companion MHQ issue.  

Thank you.

Thom293:
Related: https://github.com/MegaMek/mekhq/issues/3984

repligator:
As of latest master (https://github.com/MegaMek/megamek/pull/5944), units added to bot or player forces while edge is enabled will automatically have all edge triggers set to on (checked). Closing as completed.

Issue #4147, opened 2023-01-31
Mech selection window opens at last opened location on screen, even if that was on a second monitor that's no longer connected


mikerob:
Fill out the fields below with as much pertinent information as possible to help the developers solve the issue.
Search for your issue inside open and closed issues first before reporting so as not to make duplicates.
(Delete the above lines)

### Environment
MegaMek 0.49.11
Windows 11
megamek.MegaMek.initializeLogging(MegaMek.java:115) - Starting MegaMek v0.49.11
	Java Vendor: Eclipse Adoptium
	Java Version: 11.0.17
	Platform: Windows 11 10.0 (amd64)

### Description
When opening the unit selection window ("Add Combat Units ..." the window opens at the screen location it was last closed at. That's cool, unless that location was on an external monitor that's no longer connected.  The only way to get the window back is to re-connect the monitor and move it back to the primary display.  

Once that's done, it re-opens on the primary monitor after that.

To reproduce:
Connect second monitor (extend desktop)
Open unit selection window and drag to second monitor
Close MegaMek
Disconnect second monitor
Run MegaMek and open unit selector
Observe lack of window and inability to select anything on main window (since it's lost focus to the off-screen window)
Re-connect second monitor (extend desktop)
Observe presence of unit selector!

Is there a way to check the available screen resolution before spawning the window, and if it would spawn off-screen, instead spawn it at 0,0 or aligned with the top left corner of the main window?

Note that the same thing happens with MekLab.

Issue #3984, opened 2022-11-15
[49.10] Stuka-K5b2 data issues


gsparks3:
It seems there was a slight problem with the fix for issue #3862.

As of 49.10, both versions have the unit code "STU-K5b2". The version with an introduction date of 2751 should instead be "STU-K5b", while the version with an introduction date of 2709 should remain "STU-K5b2".

Also, the version with an introduction date of 2709 should have standard PPCs, while the version with an introduction date of 2751 should have ERPPCs. Currently this is backwards (earlier version has ERPPCs, later version has standard PPCs).

PR #5550, opened 2024-06-04
Set Default Edge States to True

IllianiCBT:
This PR sets the default state of Edge usage from _false_ to _true_. The logic is that if users have Edge enabled, they probably want to use it. It's my hope that this will reduce the amount of pilot editing required when Edge is enabled. This is particularly useful for mhq, which inherits these default states.

While I was at it, I also adjusted edge labels from 'MaxTech' to 'Destiny/ATOW'.

This was discussed with Hammer a little while ago and we chose to reference Destiny/ATOW as those are current rules; as opposed to MaxTech which hasn't been current for a number of years now. Also, (to my understanding) our implementation of Edge veers closer to the rules in Destiny/ATOW than MaxTech.

Finally, I removed a bunch of leading and trailing whitespaces

## Closes
Along with [#4147](https://github.com/MegaMek/mekhq/pull/4147), this PR closes [#3984](https://github.com/MegaMek/mekhq/issues/3984)


SJuliez (approved):

SJuliez:
Does this also fix #5328 ?

PR #5560, opened 2024-06-08
Changed Show Report Sprites in Report Log Default State to False

IllianiCBT:
This PR updates the default state of the 'Show Report Sprites in Report Log' option from `true` to `false`. I also went ahead and added a small note indicating that this option is resource intensive. This is only a bandaid fix, but helps address the long round and battle resolution load times users have been reporting.


HammerGS (approved):
Looks good to me.

PR #5565, opened 2024-06-11
Fix aerospace attack values for LBX autocannons

neoancient:
The AV for LBX autocannons (other than the LB2X) are incorrect. See TW, pp. 303-4.

### DIFF
--- a/megamek/src/megamek/common/weapons/autocannons/CLLB10XAC.java
+++ b/megamek/src/megamek/common/weapons/autocannons/CLLB10XAC.java
@@ -39,8 +39,8 @@ public class CLLB10XAC extends LBXACWeapon {
         this.criticals = 5;
         this.bv = 148;
         this.cost = 400000;
-        this.shortAV = 10;
-        this.medAV = 10;
+        this.shortAV = 6;
+        this.medAV = 6;
         this.maxRange = RANGE_MED;
         rulesRefs = "207, TM";
         techAdvancement.setTechBase(TECH_BASE_CLAN)
--- a/megamek/src/megamek/common/weapons/autocannons/CLLB20XAC.java
+++ b/megamek/src/megamek/common/weapons/autocannons/CLLB20XAC.java
@@ -39,8 +39,8 @@ public class CLLB20XAC extends LBXACWeapon {
         criticals = 9;
         bv = 237;
         cost = 600000;
-        shortAV = 20;
-        medAV = 20;
+        shortAV = 12;
+        medAV = 12;
         maxRange = RANGE_MED;
         rulesRefs = "207, TM";
         flags = flags.andNot(F_PROTO_WEAPON);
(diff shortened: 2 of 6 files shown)


IllianiCBT (approved):
LGTM

gsparks3:
I seem to recall this previously causing issues with aerospace units not doing the correct amount of damage with LBX autocannons. [LBXHandler.java](https://github.com/MegaMek/megamek/blob/89685d1ccab4b6586590f9f836475eec4e456910/megamek/src/megamek/common/weapons/LBXHandler.java#L67) in particular contains code (lines 67-74) that will reduce LBX attack values to 60% of their specified value. If it is preferred to have the correct values in the weapon definition files, removing that code might be best. Either way, this could use testing to confirm that aerospace units are doing the expected amount of damage with LBX autocannons after the changes.

gsparks3:
Tested the Gradle artifact for the PR; I can now confirm that this causes LBX autocannons to do less damage than intended. Does seem to be following the `Math.floor(0.6 * av)` behavior from that bit of code in LBXHandler.java (6 * 0.6 = 3.6, floor to 3 for LBX10, 12 *0.6 = 7.2, floor to 7 for LBX20). 
image


If you decide to remove the LBXHandler code and leave the adjusted weapon AVs (a perfectly reasonable choice), also fix the [Silver Bullet Gauss Rifle](https://github.com/MegaMek/megamek/blob/master/megamek/src/megamek/common/weapons/gaussrifles/ISSilverBulletGauss.java) AV, since it counts as an LBX weapon.

Sleet01:
Pretty sure this is the way it is because we have multiple consumers of the AV values, only some of which are actually Aerospace units.  Just changing the AV values won't produce correct behavior in all cases.

Sleet01:
> If you decide to remove the LBXHandler code and leave the adjusted weapon AVs (a perfectly reasonable choice), also fix the [Silver Bullet Gauss Rifle](https://github.com/MegaMek/megamek/blob/master/megamek/src/megamek/common/weapons/gaussrifles/ISSilverBulletGauss.java) AV, since it counts as an LBX weapon.

Also note that I changed the SB Gauss AV value _to_ 15 _from_ 9 specifically because of this code.  Which has been in place for 11 years.

Was this PR submitted due to an issue, or just because the values looked wrong?

gsparks3:
> Also note that I changed the SB Gauss AV value _to_ 15 _from_ 9 specifically because of this code. Which has been in place for 11 years.
That must have been where I remembered it from, I was looking for the issue report or change history for a LBX change and couldn't find anything, heh.

Sleet01 (changes requested):
I'd like us to hold off on this unless it also includes updating the LBX handler that already lowers the AV value, as well as any other weapons currently affected by it (such as the Silver Bullet Gauss).
It may be that we no longer need to have both a full and an Aero-specific AV value, but I don't think that question has been answered satisfactorily yet.

SJuliez:
It feels like we can close this, right?

IllianiCBT:
I opted to withhold the review requested from me, as I don't know enough about the rules to make a reliable judgement call. Also, this was before I opted to avoid reviewing MM content too much due to my unfamiliarity with the codebase.

SJuliez:
Closing this as per the above comments (seems unnecessary or at least, it would require the LBX handler to be updated)

PR #5572, opened 2024-06-14
Exclude units by role on random generation, force generator table calculation updates

### LINKED ISSUE #5564: RFE: add ability to exclude roles on random generation
Kinda-sorta a placeholder so I don't forget while working on something else, but if someone else gets to it first...

The current random generation system allows the use of 'mission roles' to selectively control what gets generated. For example, the CAVALRY role promotes generation of fast, well armed units while excluding slower ones.  What it doesn't have, is an option to also exclude based on additional roles. For example, requesting URBAN while also excluding all FIELD_GUN to generate conventional infantry suitable for urban combat while also excluding those flagged as having field guns.

This would be added through the random generation parameters system, in parallel and similar to how the current force generator roles are being handled.  Thankfully the implementation will be much simpler, pulling out table entries that match the exclusion list after they are generated by the existing inclusion list.

This doesn't need to be added to the RAT table generator UI tab under the Create a Random Army dialog, as it's more of a programmers utility. But if someone gets industrious...

SuperStucco:
The initial intention in this work was to change random selection parameters to allow excluding some units from random selection by role, to close #5564. Turns out the selection table math in the same place has a few kinks, so those fixes got rolled in.

Math fixes first:

- calculating the total weight for the weighted random selection was including units that would then be filtered out for various reasons.  This resulted in a much higher total weight, and lower than intended weight for all the models of that chassis.  The total weight is now calculated based on the same units that are processed into the random selection table.
- calculations that changed availability numbers to weights ( 2 ^ (av/2) ) returned doubles, which were being truncated to int while added to the total weight, and then the total weight used in other double-producing calculations.  Doubles are now being used throughout the calculations to preserve accuracy.
- year comparisons for pre-/early production availability modifiers have been adjusted to do the intended comparisons
- the method for adjusting availability with the dynamic +/- adjustment system has  small tweak to handle a 1-point difference
- interpolation of availability values for years between era data points have been tweaked

There are a few bit of weirdness showing up, such as RAC/LAC field gun platoons being processed in 3058 even though they have much later introduction date (3062 for the former, 3068 for the latter).  The LAC field gun models are not even present in the 3058.xml file.  These models may be getting processed based on tech availability date rather than the keyed-in introduction date, but I'm treating that as out-of-scope for this PR and the calculations are successfully filtering them out.

These should not result in many changes for units like Mechs with few models per chassis.  The largest changes will be seen with conventional infantry where there are regularly 10+ models per chassis, with the RATGen/force generator table numbers now more consistent with the availability numbers.

### DIFF
--- a/megamek/src/megamek/client/ratgenerator/AbstractUnitRecord.java
+++ b/megamek/src/megamek/client/ratgenerator/AbstractUnitRecord.java
@@ -42,9 +42,7 @@ public class AbstractUnitRecord {
     /**
      * Returns availability value modified for rating differential on +/- rating, and adjusted for
      * the first few years before introduction (field experiments, etc.) and immediately after
-     * introduction (production lines working out problems).
-     *
-     * rating differential
+     * introduction (low rate initial production).
      *
      * @param initialAv        AvailabilityRating for the chassis or model.
      * @param formationRating  force equipment rating.
--- a/megamek/src/megamek/client/ratgenerator/AbstractUnitRecord.java
+++ b/megamek/src/megamek/client/ratgenerator/AbstractUnitRecord.java
@@ -40,25 +40,33 @@ public class AbstractUnitRecord {
     }
 
     /**
-     * Adjusts availability rating for the first couple years after introduction.
+     * Returns availability value modified for rating differential on +/- rating, and adjusted for
+     * the first few years before introduction (field experiments, etc.) and immediately after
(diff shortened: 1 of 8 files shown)


HammerGS (approved):
Seems to work.

HammerGS:
@SuperStucco This breaks tests in MekHQ and will need a companion PR

PR #5586, opened 2024-06-17
Updated Game Options and Report Log Default Preferences

IllianiCBT:
This PR corrects a mistake in #5560, where the incorrect option was disabled by default. This error occurred due to confusion over which option was causing issues. Not being aware of both options, I didn't question the error and pushed the change.

Now, unit sprites are enabled by default in the round report and this option is no longer labeled as 'resource intensive'.

Unit tooltips are now disabled by default in the round report and this option is now labeled as 'resource intensive'.


SJuliez (approved):
Indeed! I was noticing in some test games that the real resouce hog are the tooltips.

PR #5591, opened 2024-06-18
Only consider MM internal dirs when creating the official unit list

SJuliez:
In other words, the MSC will ignore used dirs, the "unofficial" dir and story arc dirs when set to "ignoreUnofficial". Also removing an unused field.


HammerGS (approved):
Tested and seems to work

PR #5598, opened 2024-06-19
Fix 5596: Fix range updates and projections in space

### LINKED ISSUE #5596: [0.49.20] ASF in space can fire beyond weapon range
ASF in space can fire SRM4 at a target at range 7. The weapon display shows "out of range" but the fire button allows firing. The shot misses automatically (correctly) but still happens. Incidentally, the firing arcs show ground ranges beyond the allowed aero range on a weapon such as the SRM4

Sleet01:
A variety of different updates appear to have broken the part of our firing arc draw code that stops adding arcs for shorter-ranged weapons, specifically:

- ASF weapons with variable range (e.g. ATMs)
- Bay weapons

This fix returns that functionality so that only the in-range sections of a weapon or bay's arc will display, and the range will accurately reflect the selected ammo.

This patch also adds the ability to select variable-range ammo for _enemy_ units and see their range bands without changing the loaded ammo on the viewed unit, for better turn planning by human players.

Testing:
- Ran all 3 projects' existing unit tests
- Tested extensively with ATM rounds on Aerospace units of various types
- Also confirmed ground unit firing arc display has not been impacted

Close #5596

### DIFF
--- a/megamek/src/megamek/client/ui/swing/ClientGUI.java
+++ b/megamek/src/megamek/client/ui/swing/ClientGUI.java
@@ -2920,11 +2920,7 @@ public class ClientGUI extends AbstractClientGUI implements BoardViewListener,
     }
 
     public Optional<AmmoMounted> getDisplayedAmmo() {
-        AmmoMounted ammo = unitDisplay.wPan.getSelectedAmmo();
-        if (null == ammo) {
-            return Optional.empty();
-        }
-        return Optional.of(ammo);
+        return Optional.ofNullable(unitDisplay.wPan.getSelectedAmmo());
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


megamek/src/megamek/client/ui/swing/ClientGUI.java
```diff
@@ -2918,6 +2919,14 @@ public Optional<WeaponMounted> getDisplayedWeapon() {
         }
     }
 
+    public Optional<AmmoMounted> getDisplayedAmmo() {
+        AmmoMounted ammo = unitDisplay.wPan.getSelectedAmmo();
+        if (null == ammo) {
```
SJuliez, line 2924: I believe 
`return Optional.ofNullable(unitDisplay.wPan.getSelectedAmmo());`
can replace this method's whole code.

SJuliez: I believe I should have used that myself :)

SJuliez (approved):
I have one suggestion, although I should have thought of that for my own code... Not important, I can do that later, I'm constantly working in ClientGUI anyway

megamek/src/megamek/client/ui/swing/ClientGUI.java
Sleet01: Alrighty, I'll plug that in.  Thanks!

PR #5602, opened 2024-06-19
Princess - enhanced anti-Mech targeting

SuperStucco:
This PR enables Princess to take aimed shots and, if the option is enabled, use TacOps called shots.  This will mostly come into play late in a game as Princess starts to take advantage of Mechs that have reduced or no armor in some locations although skilled units that mount a targeting computer and powerful ranged weapons may start taking immediate advantage to 'thin the herd'.

- All ground units, including VTOLs and gun emplacements, will use this. Fixed wing aircraft and large craft are excluded (they have enough firepower as-is).  Blue water naval is also excluded as Mechs in littoral areas are frequently waist-deep in water/under partial cover or fully submerged.
- Only Mechs are considered as valid targets. Ground vehicles are vulnerable enough, VTOLs have high to-hit numbers under most conditions, and it's generally good enough to simply hit fixed wing aircraft.
- Enhanced targeting is not used against Mechs with partial cover.
- Immobile Mechs are targeted with aimed shots using weapons that can do so, and standard attacks for the rest.  Called shots are ignored to avoid turning long term campaign pilots into pudding by calling every shot high.
- Only conventional infantry and battle armor will take aimed shots at the head against immobile Mechs, for the same reason.
- Attackers with a targeting computer may attempt aimed shots at active Mechs provided the resulting target number is low enough and they can do enough damage to the selected body location.
- Aimed shots prioritize torso and leg locations that have low/no armor, plus arms if they have a major weapon.  When they cannot do enough damage to the target body location they may try for called shots instead.
- Attackers without a targeting computer, and those with weapons that cannot make aimed shots, may make use of TacOps called shots if the option is enabled. Shots from the sides will be adjusted to enter the rear arc, while shots from the front or rear will be called high or low, depending on armor conditions.
- Use of enhanced targeting is controlled by the Self Preservation setting for Princess.  The further to the left, the more it will take long-odds aimed or called shots; the further to the right, the more conservative it will be.  Despite the name, this setting is used for a number of similar aggression-related features, such as how much heat it will build up or how high a target number is before an attack is considered not effective.

There is one point to note in the code - the setting of a called shot is entirely set up for the single-button toggle-through system for human players.  A TODO has been left to indicate where it would be better to have a single send notification deliver the called shot setting rather than sending the change notification multiple times to toggle through.


Thom293:
ohmygodhellyes.

megamek/src/megamek/client/bot/princess/Princess.java
```diff
@@ -928,6 +1032,318 @@ protected Vector<Minefield> calculateMinefieldDeployment() {
...
+                        target.getInternal(Mech.LOC_LARM) > 0) {
+                    rankedLocations.add(Mech.LOC_LARM);
+                }
+                if (rankedLocations.contains(Mech.LOC_RARM) && rankedLocations.contains(Mech.LOC_LARM)) {
+                    break;
+                }
+
+            }
+
+            // Favor right torso over left, except if right torso is completely gone
```
NickAragua, line 1108: Why?

megamek/src/megamek/client/bot/princess/Princess.java
```diff
@@ -618,6 +621,99 @@ protected void calculateFiringTurn() {
                     logger.debug(shooter.getDisplayName() + " - Detailed Best Firing Plan: " +
                             plan.getDebugDescription(true));
 
+                    // If the shooter is a ground unit (including VTOLs and infantry) and only
+                    // shooting at a single Mech target with all weapons, consider making an
+                    // aimed shot if the target is shut down or the attacker has a targeting
+                    // computer. Alternatively, consider using the called shots optional rule to
+                    // adjust the hit table to something more favorable.
+
+                    boolean isShutdownShot = false;
```
NickAragua, line 630: Let's take this newly added logic and put it into a separate function to avoid bloating getFiringTurn()

megamek/src/megamek/client/bot/princess/Princess.java
```diff
@@ -618,6 +621,99 @@ protected void calculateFiringTurn() {
...
+                    boolean isAimedShot = false;
+                    boolean isCalledShot = false;
+                    int advancedTargetingThreshold = ToHitData.IMPOSSIBLE;
+                    int maxAdvancedTargetNumber = 10;
+                    int aimLocation = Mech.LOC_NONE;
+                    int locationDestruction = 999;
+                    int calledShotDirection = CalledShot.CALLED_NONE;
+
+                    int targetID = plan.get(0).getTarget().getId();
+                    if (shooter.getUnitType() <= UnitType.GUN_EMPLACEMENT &&
```
NickAragua, line 640: This condition will need an explanation.

megamek/src/megamek/client/bot/princess/Princess.java
```diff
@@ -618,6 +621,99 @@ protected void calculateFiringTurn() {
...
+                    boolean isShutdownShot = false;
+                    boolean isAimedShot = false;
+                    boolean isCalledShot = false;
+                    int advancedTargetingThreshold = ToHitData.IMPOSSIBLE;
+                    int maxAdvancedTargetNumber = 10;
+                    int aimLocation = Mech.LOC_NONE;
+                    int locationDestruction = 999;
+                    int calledShotDirection = CalledShot.CALLED_NONE;
+
+                    int targetID = plan.get(0).getTarget().getId();
```
NickAragua, line 639: Any time you call "plan.get(0).whatever" without checking the get(0) for null, you're risking a null pointer exception. Let's make sure to check for that prior to executing any of the added code since I see a lot of calls to it in there.

megamek/src/megamek/client/bot/princess/Princess.java
```diff
@@ -618,6 +621,99 @@ protected void calculateFiringTurn() {
...
+                        // Get location and maximum allowed target number for aimed shots
+                        if (isShutdownShot || isAimedShot) {
+
+                            boolean rearShot = plan.get(0).getToHit().getSideTable() == ToHitData.SIDE_REAR;
+
+                            if (isAimedShot) {
+                                advancedTargetingThreshold = Math.max(maxAdvancedTargetNumber -
+                                        getBehaviorSettings().getSelfPreservationIndex(), 2);
+                            } else {
+                                advancedTargetingThreshold = 12;
```
NickAragua, line 677: Let's go through all these magic numbers and make them constants.

NickAragua (changes requested):
Looks reasonable and probably won't impact performance much since it's done after the bulk of the work. Need some cleanup. Also, let's address the CodeQL scan issues.

megamek/src/megamek/client/bot/princess/Princess.java
SuperStucco: Most Mechs are 'right handed' i.e. if they have a big gun it's in the right arm or right torso.  By favoring the right torso, it's more likely that at least one main weapon will be damaged if not outright removed by the arm falling off.  The side torsos are usually lower armor than the CT, so they get checked first.

I'll see about adding something descriptive to cover that reasoning.

SuperStucco:
> Looks reasonable and probably won't impact performance much since it's done after the bulk of the work. Need some cleanup. Also, let's address the CodeQL scan issues.

Any recommendation for better string comparisons?  The default code replacements solve the problem but are syntactically ugly (to my eyes, at least).  Even better would be a more elegant way of testing for various anti-Mech attacks.

megamek/src/megamek/client/bot/princess/Princess.java
SuperStucco: I considered it but not entirely certain how to factor it out, given it has to return multiple settings - whether to test for an aimed shot and/or called shots, get an aimed shot location, get a called shot direction, and setting an appropriate maximum to-hit number.  Thoughts?

IllianiCBT:
This might have occurred to you already but uses of equalsIgnoreCase() adds in null protection in addition to being liked by CodeQL.

I'd use ignores case, just in case we ever change what cases we use (for some reason), and because I don't believe matching case is important here.

Sleet01:
Given that this is a fairly extensive update to Princess's behaviours, and contains a number of preconceptions about "best practice", I'd ask that this be gated behind a Princess config option that is set to "off" by default, at least for the first iteration.

SuperStucco:
> Given that this is a fairly extensive update to Princess's behaviours, and contains a number of preconceptions about aimed-/called-shot "best practice", I'd ask that this be gated behind a Princess config option that is set to "off" by default, at least for the first iteration.

I have zero idea how to add UI stuff.  Lets keep this to climbing one small Matterhorn at a time...

It's also not that far off what was originally put in for ye olde TestBot back in the day, just a little more elegant and less brute force.

Sleet01:
> > Given that this is a fairly extensive update to Princess's behaviours, and contains a number of preconceptions about aimed-/called-shot "best practice", I'd ask that this be gated behind a Princess config option that is set to "off" by default, at least for the first iteration.
> 
> I have zero idea how to add UI stuff. Lets keep this to climbing one small Matterhorn at a time...
> 
> It's also not that far off what was originally put in for ye olde TestBot back in the day, just a little more elegant and less brute force.

The Java UI stuff is embarrassingly simple and straightforward, it's mostly just a lot of stultifying boilerplate.
But I fully understand wanting to stay away from it ^_^

SuperStucco:
Big push from working on the last couple of requests.

Most of the code has been pulled out of the host method and into various other methods.  In the process the target/attacker type limits have been slightly rewired, while there isn't any new ability to handle selecting aim locations or called shot directions for non-Mech targets it is now much easier to add them later.

I've plumbed in a lot of system controls, including an overall on/off switch, with the defaults matching the original PR specs.  That should make it easier if someone chooses to wire them into a UI at some future point.

megamek/src/megamek/client/bot/princess/Princess.java
```diff
@@ -1154,29 +1154,43 @@ protected boolean checkForEnhancedTargeting (Entity shooter,
...
-     * @return              {@link Mech} constant for location to shoot, or {@code Mech.LOC_NONE}
-     *                      for none
+     * Determine which location to aim for on a general target. Returned location constant is
+     * relative to the provided target type.
+     * Currently only supports aimed shots against Mechs.
+     * @param target        Entity being shot at
+     * @param planOfAttack  Proposed attacks against {@code target}
+     * @param rearAttack    true if attacking from rear arc
+     * @param includeHead   true if the head is a valid location, ignored for non-Mech targets
+     * @return              location constant to aim for, with the {2code LOC_NONE} constant
```
Sleet01, line 1282: Should this be "@code" rather than "2code"?

Sleet01 (approved):
Just one minor typo in a comment, otherwise looks good.  Looking forward to trying it out!

PR #5617, opened 2024-06-21
Princess - improved AMS handling

### LINKED ISSUE #5084: Princess and using AMS as a manual weapon
When the option to switch AMS for manual use as a weapon is enabled, Princess will turn it off once it gets to within appropriate range.  However, it doesn't turn the system back to automatic - it leaves it on manual, depriving it of the primary use.

Princess should ensure AMS is turned back to automatic in the end phase of any round where they are sufficiently far from any targets, for example run/jump distance.

SuperStucco:
Currently, Princess will use AMS as a manual weapon without first switching it to the proper mode (this might warrant some checks inside ToHitData, similar to those which prevent combing called and aimed shots).  This PR ensures that Princess will only consider using it manually if the system has the correct mode set.

It also adds some code to the end phase to select an appropriate mode for the AMS systems each unit may be carrying.  Most of the time this will be on/automatic, but laser AMS may be shut off to assist with heat management on seriously overheating units and relatively undamaged units may turn off conventional AMS to conserve ammo when running low. Manual/use as a weapon mode is restricted to when the unit comes into close contact with infantry; once it breaks contact with the infantry and stops firing it the mode will reset to on/automatic.

Will close #5084 .

### DIFF
--- a/megamek/src/megamek/client/bot/princess/Princess.java
+++ b/megamek/src/megamek/client/bot/princess/Princess.java
@@ -90,6 +90,9 @@ public class Princess extends BotClient {
     private final Set<Integer> attackedWhileFleeing = Collections.newSetFromMap(new ConcurrentHashMap<>());
     private final Set<Integer> crippledUnits = new HashSet<>();
 
+    // Track entities that fired an AMS manually this round
+    private List<Integer> manualAMSIds;
+
     /**
      * Returns a new Princess Bot with the given behavior and name, configured for the given
      * host and port. The new Princess Bot outputs its settings to its own logger.
@@ -628,6 +631,13 @@ public class Princess extends BotClient {
                         double existingTargetDamage = damageMap.getOrDefault(targetId, 0.0);
                         double newDamage = existingTargetDamage + shot.getExpectedDamage();
                         damageMap.put(targetId, newDamage);
+
+                        // Track manual AMS use each round
+                        if (shot.getWeapon().getType().hasFlag(Weapon.F_AMS)) {
+                            if (shot.getWeapon().curMode().equals(Weapon.MODE_AMS_MANUAL)) {
+                                flagManualAMSUse(shooter.getId());
+                            }
+                        }
 
                         if (shot.getUpdatedFiringMode() != null) {
                             super.sendModeChange(shooter.getId(), shooter.getEquipmentNum(shot.getWeapon()), shot.getUpdatedFiringMode());
@@ -2207,7 +2217,6 @@ public class Princess extends BotClient {
 
                 // If there are enough nearby enemy infantry (only counted if the game option is
                 // set), choose manual fire
-                // FIXME: also set manual if this entity fired AMS manually this round
                 if (!enemyInfantry.isEmpty() && !curEntity.isAirborne()) {
                     int infantryRange = enemyInfantry.stream().mapToInt(e -> Compute.effectiveDistance(game, curEntity, e)).min().getAsInt();
                     if (infantryRange <= 3) {
@@ -2215,10 +2224,14 @@ public class Princess extends BotClient {
                     }
                 }
 
+                // If AMS was used manually this round, chances are it will be needed next round too
+                if (usedManualAMS(curEntity.getId())) {
+                    newAMSMode = EquipmentMode.getMode(Weapon.MODE_AMS_MANUAL);
+                }
+
                 for (WeaponMounted curAMS : activeAMS) {
 
                     EquipmentMode curMode = curAMS.curMode();
-
 
                     // Turn off laser AMS to help with overheating problems
                     if (curAMS.getType().hasFlag(WeaponType.F_ENERGY)) {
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/client/bot/princess/Princess.java
```diff
@@ -2157,6 +2170,172 @@ private void launchFighters(MovePath path) {
...
+                    stream().
+                    filter(w -> w.getType().hasFlag(AmmoWeapon.F_AMS) && w.hasModes()).
+                    collect(Collectors.toList());
+
+            if (!activeAMS.isEmpty()) {
+
+                // Set default to on/automatic and test to see if it should be off or manual instead
+                EquipmentMode newAMSMode = EquipmentMode.getMode(Weapon.MODE_AMS_ON);
+
+                boolean isOverheating = (curEntity instanceof Mech) && (curEntity.getHeat() >= 14);
```
Sleet01, line 2213: This heat threshold number should either be:
A) a constant declared at the top of the file, or
B) a value passed in to the function / read from some member, or
C) a reference to any other existing heat limit variable already used by Princess
so we don't have conflicting behavior.

megamek/src/megamek/client/bot/princess/Princess.java
```diff
@@ -2157,6 +2170,172 @@ private void launchFighters(MovePath path) {
...
+                    newAMSMode = EquipmentMode.getMode(Weapon.MODE_AMS_MANUAL);
+                }
+
+                for (WeaponMounted curAMS : activeAMS) {
+
+                    EquipmentMode curMode = curAMS.curMode();
+
+                    // Turn off laser AMS to help with overheating problems
+                    if (curAMS.getType().hasFlag(WeaponType.F_ENERGY)) {
+                        if (isOverheating) {
```
Sleet01, line 2235: As above.  Also, while we apply stricter overheating checks to ASF than to 'mechs in other sections of the Princess code, it looks like isOverheating would ignore ASFs, causing them to be less likely to disable their LAMS units.

megamek/src/megamek/client/bot/princess/Princess.java
```diff
@@ -2157,6 +2170,172 @@ private void launchFighters(MovePath path) {
...
+                            newAMSMode = EquipmentMode.getMode(Weapon.MODE_AMS_OFF);
+                        }
+                    } else {
+
+                        // Determine if ammo needs to be conserved
+                        boolean conserveAmmo = curAMS.getLinkedAmmo().getUsableShotsLeft() <= (int) Math.floor(curAMS.getOriginalShots() *
+                                behaviorSettings.getSelfPreservationValue() / 100.0);
```
Sleet01, line 2242: Could this be wrapped into the existing ammo conservation code?

megamek/src/megamek/client/bot/princess/Princess.java
```diff
@@ -2157,6 +2170,172 @@ private void launchFighters(MovePath path) {
...
+
+                        // Consider turning off AMS to conserve ammo unless it's needed for infantry
+                        if (conserveAmmo && !newAMSMode.equals(Weapon.MODE_AMS_MANUAL)) {
+
+                            int ammoTN = 12 - behaviorSettings.getBraveryIndex();
```
Sleet01, line 2247: Likewise; this could probably be incorporated into the existing conservation / threshold calculation code.

megamek/src/megamek/client/bot/princess/Princess.java
```diff
@@ -2157,6 +2170,172 @@ private void launchFighters(MovePath path) {
...
+                                    break;
+                                default:
+                                    break;
+                            }
+
+                            if (ammoTN < 10) {
+                                if (Compute.d6(2) >= ammoTN) {
+                                    newAMSMode = EquipmentMode.getMode(Weapon.MODE_AMS_OFF);
+                                }
+                            }
```
Sleet01, line 2283: Why is the choice to change mode randomized rather than deterministic?

Sleet01 (changes requested):
A couple of questions, and a couple nits I'd like picked, but otherwise looks very promising.

megamek/src/megamek/client/bot/princess/Princess.java
SuperStucco: Deliberate choice - fighters are more vulnerable to damage so having AMS active, even when overheating, is considered useful.  I considered included them at first but decided just to do Mechs and keep the complexity down.  Can look into adding something if warranted, though.

megamek/src/megamek/client/bot/princess/Princess.java
SuperStucco: Maybe, but there's a distinct difference.  The existing ammo conservation code is a shoot/no shoot determination.  For AMS it is actually a mode change that prevents automatic operation in the following turn.  Keeps the existing code about determining whether to fire, and this on whether to turn this specific system off or on.

megamek/src/megamek/client/bot/princess/Princess.java
SuperStucco: While it's good to have the bulk of the behavior calculated rather than random, I'm not entirely sold on a fully deterministic Princess; for one, it is too easy to predict.  There will be circumstances like this, where there is no clear 'good' outcome that can be easily calculated vs the impact the decision will have.  Sometimes it's best all 'round to have it to do a simple gut check and hope for the best.

megamek/src/megamek/client/bot/princess/Princess.java
SuperStucco: Yeah, ideally it would tie into the existing code that handles heat.  The big problem is the existing code is during the movement/weapons fire phases, and like the stealth armor control this is happening during the end phase (which is why it can have a look at the heat level rather than trying to predict it).  So all that fire control objects/information is no longer valid or available.

I'll put in a constant for now.  I would like something a little more... dynamic, I guess, for what constitutes 'overheating'.  But keeping the complexity down seems to be the better option to start with.

Sleet01: > Yeah, ideally it would tie into the existing code that handles heat. The big problem is the existing code is during the movement/weapons fire phases, and like the stealth armor control this is happening during the end phase (which is why it can have a look at the heat level rather than trying to predict it). So all that fire control objects/information is no longer valid or available.
> 
> I'll put in a constant for now. I would like something a little more... dynamic, I guess, for what constitutes 'overheating'. But keeping the complexity down seems to be the better option to start with.

I might be misremembering, but I think there's a heat limit value that's used in the "with/without overheating" firing plan code that might be usable.  IIRC we calculate something based on aggression level; if it's not immediately available, that might be nice to pull out to a method or instance member.

Sleet01 (approved):
LGTM

PR #5636, opened 2024-06-26
Fix for squadron ammo weapons

### LINKED ISSUE #5569: [49.20-SNAPSHOT] ClassCastException from squadrons firing external ordnance
### Environment
MekHQ Windows JDK 11 build artifact from: https://github.com/MegaMek/mekhq/actions/runs/9492053037
Standalone Megamek

### Description
There was a known issue with aerospace squadron ammo weapons in 0.49.19, which was since fixed (hence why I was testing this using the nightly build). However, external ordnance on squadrons (currently only tested with RL pods) seem to still have issues. Firing the RL pods followed by another ammunition-using weapon will result in a ClassCastException popup:
image

Firing _only_ the RL pods, or only the RL pods and non-ammo-using weapons, will work as expected. However, the next time you try to fire an ammo-using weapon that game, the error will occur.

Also tested on 49.19.1, where both ammo-using weapons and external ordnance can be declared for firing, and their ammunition is consumed with no errors being thrown, but the weapon attacks simply fail to appear in the Weapon Attack Phase report or to have any effect on the target. 

### Files

Savegame for reproduction: [Round-1-autosave_2024-06-12_22-57-34.sav.gz](https://github.com/user-attachments/files/15813212/Round-1-autosave_2024-06-12_22-57-34.sav.gz)
Log: [megamek.log](https://github.com/user-attachments/files/15813197/megamek.log) (no data in other logs)

neoancient:
When determining the to-hit number for squadrons, the ammo is determined using the squadron's equipment list, but the ammo is on the individual fighters. The ammoCarrier field holds the id of the actual Entity with the ammo.

Fixes #5569

### DIFF
--- a/megamek/src/megamek/client/ui/swing/FiringDisplay.java
+++ b/megamek/src/megamek/client/ui/swing/FiringDisplay.java
@@ -22,6 +22,7 @@ import megamek.client.ui.swing.widget.MegamekButton;
 import megamek.common.*;
 import megamek.common.actions.*;
 import megamek.common.enums.AimingMode;
+import megamek.common.equipment.AmmoMounted;
 import megamek.common.equipment.WeaponMounted;
 import megamek.common.event.GamePhaseChangeEvent;
 import megamek.common.event.GameTurnChangeEvent;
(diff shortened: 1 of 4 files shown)


Sleet01 (approved):
LGTM!

PR #5644, opened 2024-06-28
Changed all instances of the role "infantry_support" to "inf_support"

repligator:
Changed all instances of the role "infantry_support" to "inf_support". This doesn't actually change the function of anything, just makes troubleshooting easier. See PR #5289 and Issue #5267.

Ran 317 tests, 312 passed, 5 ignored. Also manually tested Force Generator to make sure the effected units and formations still appear.

### DIFF
--- a/megamek/data/forcegenerator/2398.xml
+++ b/megamek/data/forcegenerator/2398.xml
@@ -188,7 +188,7 @@
 			<availability>General:9</availability>
 		</model>
 		<model name='(Wheeled MG)'>
-			<roles>infantry_support,urban</roles>
+			<roles>inf_support,urban</roles>
 			<availability>PIR:3,General:2</availability>
 		</model>
 		<model name='(Wheeled Primitive)'>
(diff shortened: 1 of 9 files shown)


HammerGS:
Just a question on this if the issue is between code and notes, why don't we fix the notes as Infantry_Support in the code seems clearer?

repligator:
Either is fine with me.

IllianiCBT:
Do we have a precedent for uses of `inf_x` vs `infantry_x` in the roles? If so, we probably still want to switch to `inf_x` for consistency. If not, then just changing the readme is probably the cleanest approach.

HammerGS:
Can we change the supporting documentation rather than code. Going forward we can ensure X_support the X can be a full name to be very clear what we are looking at,.

HammerGS:
Closing as addressed with https://github.com/MegaMek/megamek/commit/02742357f211ec5dcfda4d838f019e0bc33e96a0

Issue #5648, opened 2024-06-29
Diacratics and/or name changes causing issues with Force/RAT Generator


repligator:
### Environment
49.20
Java 17
Linux

### Description
Select year 3100, faction Ghost Bear Dominion, unit type Battle Armor, check Marine, hit generate. Clan Medium Battle Armor (Naval) 'B**a**r' (Sqd5) and/or Clan Medium Battle Armor (Naval) 'B**ä**r' (Sqd5) will not appear. The log will show various errors relating to not being able to find one or both of the aforementioned squads. I've tried changing various ä's to a's and a's to ä's, without any luck.

### Files
[megamek.log](https://github.com/user-attachments/files/16041857/megamek.log)

HammerGS:
Going to leave this open for a bit.

repligator:
Same results as before.
[megamek.log](https://github.com/user-attachments/files/16044413/megamek.log)

HammerGS:
The underlying issue is the faction data isn't defined. If you open any of the Clan files in teh data\forcegenerator\faction_rules folder you'll see the difference. There isn't an editor to create these files and have to be hand made assuming we have canon data on the formations. 

As of the development of this (2018) we did have that. So this will have to get addressed at some point but won't be a quick of easy fix. 

<img width="637" alt="image" src="https://github.com/MegaMek/megamek/assets/13752452/a4f9f085-8d77-4658-b957-3097f07954a4">

repligator:
@HammerGS Was this meant for #5647 ?

HammerGS:
It was.

HammerGS:
Ok, two part issue the Diacratic was part of the problem. But the bigger issue was a complete lack of data for the unit. The roles underneath in the RAT Generator will only filter if that unit has that role. Currently most units don't have a role so you're better to avoid selecting them. 

Fixed in the nightlies.

Issue #5647, opened 2024-06-29
Rasalhague Dominion Force Generator Issues


repligator:
### Environment
Linux
Java 17

### Description
When using the Force Generator, and selecting Rasalhague Dominion, only one Unit Type is available; combined. Only one formation is available; Random. Only one Equipment Rating is available; Random. Hitting generate will always generate a cluster.

Log entries include

`13:58:51,230 WARN  [megamek.client.ui.swing.ForceGeneratorOptionsView] {AWT-EventQueue-0}
megamek.client.ui.swing.ForceGeneratorOptionsView.refreshUnitTypes(ForceGeneratorOptionsView.java:635) - No unit type node found.

13:58:51,231 WARN  [megamek.client.ui.swing.ForceGeneratorOptionsView] {AWT-EventQueue-0}
megamek.client.ui.swing.ForceGeneratorOptionsView.refreshRatings(ForceGeneratorOptionsView.java:754) - No rating found.`

### Files
[megamek.log](https://github.com/user-attachments/files/16041728/megamek.log)

HammerGS:
The underlying issue is the faction data isn't defined. If you open any of the Clan files in teh data\forcegenerator\faction_rules folder you'll see the difference. There isn't an editor to create these files and have to be hand made assuming we have canon data on the formations. 

As of the development of this (2018) we did have that. So this will have to get addressed at some point but won't be a quick of easy fix. 

<img width="637" alt="image" src="https://github.com/MegaMek/megamek/assets/13752452/a4f9f085-8d77-4658-b957-3097f07954a4">

repligator:
I've sent in a PR. Even though CGB is technically 50-ish years out of date, most of it is still accurate, and I'd argue that occasionally anachronistic force names are the lesser evil compared to not being able to generate forces at all.

PR #5649, opened 2024-06-29
Use UTF encoding for reading MTF files

SJuliez:
This adds a missing CharSet so that MTF files read diacritics correctly.

This might fix the issues #5648 and #5647


HammerGS (approved):
Tested and working across the programs.

PR #5652, opened 2024-06-30
Fix ammo carrier issue that prevents Princess from firing ammo-based weapons

Sleet01:
Noticed while testing the 0.49.20 MekHQ version of MegaMek: Princess does not fire any weapons that use ammo.

Solution: consider the current unit the "ammo carrier" if no ammo carrier ID is provided.

Testing:
- Ran several Princess v. Princess matches with ammo-based weapons, both Ground and Low Alt, including meks, BA, infantry, field artillery/guns, tanks, VTOLs, and ASF.


HammerGS (approved):
tested and seems to work

PR #5656, opened 2024-06-30
Clean up of 17 Actions and normalizing across all 3 projects.

rjhancock:
Brings all actions within same patterns and works to normalize all 3 repos with same actions.


SJuliez (approved):
Looking good, as far as I can read this stuff

PR #5660, opened 2024-07-01
Startup And Windows CI Fixes

rjhancock:


HoneySkull:
I pulled this and was able to build locally on Windows using Gradle and verified the resulting exe launched megamek successfully.

megamek/build.gradle
```diff
@@ -50,7 +56,12 @@ application {
 }
 
 ext {
-    mmJvmOptions = ['-Xmx4096m', '--add-opens', 'java.base/java.util=ALL-UNNAMED', '--add-opens', 'java.base/java.util.concurrent=ALL-UNNAMED', '-Dsun.awt.disablegrab=true']
+    mmJvmOptions = [
+        '-Xmx2048m',
```
HoneySkull, line 60: Should this default to 4096m?

megamek/build.gradle
```diff
@@ -243,21 +255,19 @@ distributions {
 createExe {
     description = 'Create Windows executable that wraps MM jar'
     mainClassName = application.mainClass
-    outfile = 'MegaMek.exe'
```
HoneySkull, line 246: This outfile assignment can be left in safely.   Was tested.

megamek/build.gradle
rjhancock: No, I upped it when I shouldn't have. It should be 2G and not 4G

megamek/build.gradle
rjhancock: Can be but not needed. It's default to project name plus exe.

HammerGS:
Should we poll the community and move to 4?

rjhancock:
4 what? Repos?

HammerGS:
Gig of memory

rjhancock:
Poll it. We can bump all 3 at the same time.

kuronekochomusuke:
are 2 copies of SubmitBug.html needed?

image

rjhancock:
No, but I can do cleanup of this stuff before the merge.

PR #5668, opened 2024-07-02
Fix gradle createExe task to build locally on Windows and launch.

HoneySkull:
Gradle was failing to create the Launch4J exe when building locally from Gradle.  

Updated embedded exe jvm arguments to prevent hidden errors that were stopping megamek to launch when double-clicking the exe file.

The default jvm arguments are preserved in the MegaMek.l4j.ini file.


HoneySkull:
The jvm argument part of this looks like it was fixed in a pending pull request #5660, however, to build locally using `./gradlew clean build`, I needed to add the `doNotTrack()` task method, otherwise gradle complains about having unreadable tracked values for parameter 'outfile' and doesn't create the MegaMek.exe.

rjhancock:
What was causing the issue as this was only occurring on windows and I never could figure out what was going on? And searching wasn't helping.

What was in the debug output when ran?

HoneySkull:
Addressed in PR #5660

Issue #5664, opened 2024-07-02
0.49.20 Cybernetic upgrades and BV


strategoslevel3:
I would like to request that MegaMek include the correct BV Costings for Cybernetics in its BV calculation for units (including pilots and conventional infantry). Currently the BV costs generated are not current with the official core rules in ISO:AE but reflect prior systems (Sword and Dragon/Jihad HitSpots).

pg 184 ISO:AE

SJuliez:
I'd like to clarify this. The BV calculation generally follows IO:AE p.184, meaning that the pilot skill is directly modified. However, the current calculation does not consider non-stacking bonuses (e.g. gunnery modifications of multiple cybernetics should not stack). Also, some cybernetics are not implemented. What errors did you see?

strategoslevel3:
It’s the complete wrong system. The cybernetic implants are based on the old Jihad supplements and not ISO-AE

strategoslevel3:
A MekWarrior with P5/G3 has the following cybernetic implants:

Artificial Pain Shit (Cost -1 Effective PS)
Multi-Modal Sensory Implant (Cost -1 Effect Gunnery. Benefit -1 Gunnery)
VDNI (Cost -1 Gunnery and Piloting.  Benefit -1 Gunnery and Piloting)

These modifiers do stack

The should have a cost of x1.68 for P3/G2.  Currently the vDNI is not being added on, and the BV Claculation window is only showing x1.32 for the Paint Shunt and Sensory Impants.
________________________________

strategoslevel3:
After further testing on 50.04

VDNI and Multi Modal Sensory implants should both give a -1 TN to gunnery but are not stacking. Only the VDNI is applied while the MMSI is not

Tripple Core Processor is only giving a +1 Initive on a C3i networked unit when it should give a +3 when part of an active C3i network <https://1drv.ms/i/c/0f54bc317d0efccb/EdebAQqyiMpMgBTC_5544MgBJsI2byGDMC0jc_NXnLfiZg>
[...]

PR #5672, opened 2024-07-03
Adds DNI Cockpits to Celestials

repligator:
Adding DNI cockpits to Celestial mechs, partially implementing issue https://github.com/MegaMek/megamek/issues/5664. This is accomplished by adding a piece of equipment, DNICockpit, to the head that does not use a slot and can not be hit. It will display correctly in MML, but can't be manually added to a design. It's not a very pretty solution, but it (mostly) works, and can be easily removed later if DNI is switched to a quirk.

This should correctly calculate BV for all implants on lines 1206 to 1247 of megamek/src/megamek/common/battlevalue/BVCalculator.java. That being said, some of the features of the implants (for example Prototype VDNI) are still not implemented.

Also contains some totally unrelated fixes to Crab descriptions.

Tests:
All unit tests for MM passed or skipped, loaded units in MML, compared BV cost between implanted pilots and pilots with the equivalent natural level of skills in MM, fought short battle to confirm that some of the implants are functioning during battle.

P.S.
I need to remember to stop working in the master branch...


IllianiCBT:
I would much prefer we come up with a permanent solution rather than a temporary fix. Do you think this is feasible, or is a temporary fix our best option without a significant amount of extra work?

repligator:
The way I see it, a permanent solution would require everyone to agree on how implants (and implant based hardware on mechs) are going to be handled, and then implementing that into the code-base. I don't know if this would mean something like quirks, or SPAs, or something totally different; but it would require time and effort regardless.
This PR is a shamelessly hacky temporary fix until that time and effort is available. It was simple to implement, and should be even simpler to rip out once a better solution is found.
That being said, I have zero emotional attachment to this code. If you and others want to avoid a temporary solution, please feel free to reject/close the PR.

repligator:
Closing. We can always implement it better, later.

Thom293:
IMHO in a project of this type there is a utility in using a working but imperfect fix.  The perfect fix may never come, and if it does, it's a bonus.  Just my opinion.

Saklad5:
The way I see it, modifications like VDNI are similar to an Armored Cockpit. Perhaps a similar approach could be used?

This also applies to Enhanced Imaging interfaces, Damage Interrupt Circuits, and SLDF Advanced Neurohelmets.

Issue #1950, opened 2020-05-15
RFE: Princess option - "Advance to contact"


SimonLandmine:
The "Destination Edge" addition is excellent for getting Princess to move across a battlefield, in simulation of a 'breakthrough' or 'convoy' mission (for Princess - a 'defence' or 'ambush' scenario for the player). However, I've noticed that, with "Destination Edge" set, Princess will depart the board (at least in 0.46.1) if the units make to the destination edge (potentially if outnumbered). Units with this setting will effectively 'flee' towards their destination, unless Crippled, in which case they will fall back to the "Retreat Edge".

However, if you want Princess to simulate an "advance to contact" setting in a Double Blind scenario, where Princess' units don't start with LOS to the player's units, then this isn't entirely ideal, and you will occasionally need to do something to get Princess' attention.

My suggestion - add an "Advance to contact" option checkbox to the Princess bot configuration window.

Suggested implementation - When checked, this option would set a flag on a Princess instance. While the flag is set, units will advance at Run (or Walk? but not Sprint - they need to be ready to react on contact) speed away from their starting edge* (or towards a destination edge?), until encountering forces from another Team, at which point the Flag will be cleared, and the standard assigned Princess behaviour for that bot will resume.

* Note - this could encounter an issue with CENTER start, but it might just result in units scattering entertainingly. That would probably require testing, but that could still be a viable outcome.

**EDIT** Currently, if "Destination Edge" is checked, but "Auto Flee Edge" is unchecked, Princess' behaviour is even odder - her units mill about at their destination edge, as if unsure what to do now they've got there, and unwilling to move away from it.

SimonLandmine:
An additional thought about this proposal - it would probably work better as a 'fallback option', so that if enemy are encountered, and then are no longer visible, Princess would then revert to this advancing behaviour.

HammerGS:
Related to https://github.com/MegaMek/megamek/issues/203

SimonLandmine:
Related, but would be an alternative - 'Single Blind' would give Princess 'god's eye view', while 'Advance to Contact' maintains the possibility of flanking, which Single Blind would negate.
'Advance to Contact' simply means that Princess does something if there's no enemy in sight.

Thom293:
[...]

PR #5675, opened 2024-07-04
Princess -  advance to the front, not the rear for double-blind

SuperStucco:
Removes the double-reverse on which board edge to move towards if there are no contacts in double-blind.  Now, when double-blind rules are in play and Princess has no active or sensor contacts it will try to move to the map edge opposite from where it deployed.  If deployed in the center or edge, it's units will move away from the edge nearest each unit's current position effectively milling around in the center of the map.  At some point in the process it should either get direct or sensor contact on hostile targets and move to engage.


Thom293:
Niiiiice.  Can't wait to try this out.

BLR-IIC:
I've been awaiting a change like this for awhile!  I'm just surprised it's just one line of code!  O.o

Thom293:
Yeah.  Nice someone finally looked at it.  

I believe this will close #1950 fyi.

NickAragua (approved):

PR #5689, opened 2024-07-05
Added Sentry to project

rjhancock:
Updated MegaMek.java with sentry and initial error tracking Updated CI to package zip/tar only
Updated Build scripts for single distribution package

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,6 +37,10 @@ jobs:
           build-scan-terms-of-use-url: "https://gradle.com/terms-of-service"
           build-scan-terms-of-use-agree: "yes"
 
+      - name: Add MMRevision file
+        run: |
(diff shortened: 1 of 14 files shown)


SJuliez (approved):
fwiw

SJuliez:
Can I ask you to maybe give your auto formatter a shutdown? I feel some of the things that are happening aren't really helpful. Screens these days have a width above 40 characters :) And writing every parameter may sometimes be good but not as a baseline from the auto formatter. Thanks.

rjhancock:
The each parameter on it's own line for longer functions is actually me, not the auto-formatter. Got into that habit from my work in Swift. If the entire call fits within 120 characters, I leave it, if it went over, I split each on its own line.

It became easier for me to read as the IDE also put the variable names from the function inline.

PR #5696, opened 2024-07-07
TargetRoll modernization

SJuliez:
This modernizes some very old code of TargetRoll and TargetRollModifier. TargetRollModifier is now immutable for which I had to adapt two places in Compute. Also updates the test for TargetRoll.

### DIFF
--- a/megamek/src/megamek/common/Compute.java
+++ b/megamek/src/megamek/common/Compute.java
@@ -2520,41 +2520,36 @@ public class Compute {
             return toHit;
         }
 
+        int dedicatedGunnerMod = ((entity instanceof Mech mek) && (mek.getCockpitType() == Mech.COCKPIT_DUAL)
+                && entity.getCrew().hasDedicatedGunner()) ? 2 : 1;
+
         if ((entity.getMovementMode() == EntityMovementMode.BIPED_SWIM)
             || (entity.getMovementMode() == EntityMovementMode.QUAD_SWIM)) {
-            toHit.addModifier(3, "attacker used UMUs");
+            toHit.addModifier(3 / dedicatedGunnerMod, "attacker used UMUs");
         } else if (entity instanceof LandAirMech && movement == EntityMovementType.MOVE_VTOL_WALK) {
-            toHit.addModifier(3, "attacker cruised");
+            toHit.addModifier(3 / dedicatedGunnerMod, "attacker cruised");
         } else if (entity instanceof LandAirMech && movement == EntityMovementType.MOVE_VTOL_RUN) {
-            toHit.addModifier(4, "attacker flanked");
+            toHit.addModifier(4 / dedicatedGunnerMod, "attacker flanked");
         } else if ((movement == EntityMovementType.MOVE_WALK) || (movement == EntityMovementType.MOVE_VTOL_WALK)
                 || (movement == EntityMovementType.MOVE_CAREFUL_STAND)) {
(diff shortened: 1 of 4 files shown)


megamek/src/megamek/common/Compute.java
```diff
@@ -2520,41 +2520,36 @@ public static ToHitData getAttackerMovementModifier(Game game, int entityId,
...
         } else if (movement == EntityMovementType.MOVE_JUMP) {
             if (entity.hasAbility(OptionsConstants.PILOT_JUMPING_JACK)) {
-                toHit.addModifier(1, "attacker jumped");
+                toHit.addModifier(1 / dedicatedGunnerMod, "attacker jumped");
             } else if (entity.hasAbility(OptionsConstants.PILOT_HOPPING_JACK)) {
-                toHit.addModifier(2, "attacker jumped");
+                toHit.addModifier(2 / dedicatedGunnerMod, "attacker jumped");
             } else {
-                toHit.addModifier(3, "attacker jumped");
+                toHit.addModifier(3 / dedicatedGunnerMod, "attacker jumped");
```
Sleet01, line 2546: It looks like we can get non-integer results here, which I believe Java will round down.
1. is that expected behavior?
2. do we check anywhere that this code produces the expected results for every combination?

megamek/src/megamek/common/TargetRoll.java
```diff
@@ -23,75 +29,72 @@
...
-            default:
-                return Integer.toString(total);
-        }
+        return switch (total) {
+            case IMPOSSIBLE -> "Impossible";
+            case AUTOMATIC_FAIL -> "Automatic Failure";
+            case AUTOMATIC_SUCCESS -> "Automatic Success";
+            case CHECK_FALSE -> "Did not need to roll";
+            default -> Integer.toString(total);
+        };
```
Sleet01, line 92: So fresh and so clean!

megamek/src/megamek/common/TargetRoll.java
```diff
@@ -198,79 +202,79 @@ public void append(TargetRoll other) {
...
+        } else {
+            modifiers.removeIf(this::isAutomatic);
         }
-        modifiers = toKeep;
         recalculate();
     }
 
+    private boolean isAutomaticOrImpossible(TargetRollModifier modifier) {
+        return modifier.getValue() == AUTOMATIC_FAIL || modifier.getValue() == AUTOMATIC_SUCCESS
+                || modifier.getValue() == IMPOSSIBLE;
```
Sleet01, line 245: This can also be accomplished fairly quickly and cleanly with:

    return Set.of(AUTOMATIC_FAIL, AUTOMATIC_SUCCESS, IMPOSSIBLE).contains(modifier.getValue());

Sleet01 (approved):
A couple of questions/comments but overall looks good.

megamek/src/megamek/common/Compute.java
SJuliez: 1. I think so as it was like that before
2. No. Currently there are no combinations, this method only returns a single modifier. I did not check the rules as - see 1. :)

megamek/src/megamek/common/TargetRoll.java
SJuliez: yeah :) although strictly speaking this should be localized...

Sleet01 (approved):
LGTM!

PR #5702, opened 2024-07-08
Initial Work on MMLogger with Sentry Tracking

rjhancock:
Initial work on adding a central MMLogging class to handle log writing, Sentry reporting, and Panel display.


megamek/src/megamek/logging/MMLogger.java
```diff
@@ -0,0 +1,69 @@
...
+
+public class MMLogger {
+  MMLogger() {
+    throw new IllegalStateException("MMLogger Utility Class");
+  }
+
+  public static void info(String message) {
+    Logger logger = LogManager.getLogger();
+
+    if (logger.isInfoEnabled()) {
```
SJuliez, line 33: this suggests that you plan to replace all current info() calls with this one, right? Because it doesnt make sense to check if info is enabled for some but not all info logs. In that case, would it make sense to really replace *all* logger calls? And if so, we'd need to have an error() call without an exception parameter because not all errors come from an exception.

megamek/src/megamek/logging/MMLogger.java
```diff
@@ -0,0 +1,69 @@
...
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the Free
+ * Software Foundation; either version 2 of the License, or (at your option)
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * for more details.
+ */
```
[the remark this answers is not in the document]
SJuliez: Also, indentation is supposed to be 4 spaces...

SJuliez (changes requested):
A few comments and requests. In addition (Im starting to feel bad), I'd like to ask for at least a bare minimum of javadoc on MMLogger as it is not self-evident what MMLogger.error(ex, msg, title) does.

megamek/src/megamek/logging/MMLogger.java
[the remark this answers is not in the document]
rjhancock: Noted. On both.

megamek/src/megamek/logging/MMLogger.java
```diff
@@ -0,0 +1,69 @@
...
+
+import javax.swing.JOptionPane;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import io.sentry.Sentry;
+
+public class MMLogger {
+  MMLogger() {
```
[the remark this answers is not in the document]
rjhancock: Didn't think about it at the time.

megamek/src/megamek/logging/MMLogger.java
rjhancock: That is the intention and I can add the other error method.

But the number of places that would need to be adjusted for this, I'm not going to do it all at once and only a few files at a time. A slow migration to this system.

rjhancock:
I"ll add the JavaDoc when I make the rest of the changes.

SJuliez (approved):

megamek/src/megamek/logging/MMLogger.java
SJuliez: I misphrased that. It's not that I expect you to actually do any replacing, I only meant how things would ideally be. Mass-replacing will only create merge conflicts. As there is no intolerable downside to keeping LogManager calls, we can leave it to whoever touches classes to switch to MMLogger.

rjhancock: Yea, my intention is to do 5-10 files at a time for Sentry and just replace the logs as I go to keep it uniform. It'll take a few months just for MM a few more for MML and a few more for MHQ. I figure by the time we reach the start of J21 testing I'll finally be done adding this in everywhere.

Issue #2577, opened 2021-01-15
Placable objects


musicman609:
### Environment
46.1

### Description
Can we get placable boxes or random objects that can be placed at random by one side or another during deployment? I've been going through and adding all the specialized maps for all the early supplements, and more than a few of those scenarios require crates or whatnot distributed at random either to be destroyed, collected, etc.

I know I could place markers on the map itself in the map editor that can be destroyed (buildings, fuel tanks, etc),  but this removes the randomization factor.

repligator:
See also #353

PR #5706, opened 2024-07-08
Carry objects

NickAragua:
Implements cargo carrier rules from TW:261 (mechs and protomechs only) and (partly) TOAR:90 (just the part where you can pick up a lighter object with one hand only).

Includes:
- picking up cargo (mechs and protomechs only)
- dropping/unloading cargo (drop if you keep moving, unload if it's the last step)
- can't fire if loaded/unloaded cargo this turn
- can't fire torso/arm weapons if carrying cargo (can use other arm when carrying really light cargo)
- cargo will get damaged if the carrying unit is shot at
- cargo may get destroyed if carrying unit drops it on the move (even more likely when flying/jumping)
- graphical indicators and reporting of cargo-related activity
- cargo can be 0-weight and invulnerable
- if you fall, you drop the cargo

Does not include
- TSM bonus
- Loading cargo for non-mech/protomech units

A story told in a few screenshots:

image

image

image

image

image

image


HammerGS:
Would there be any opportunity in this to handle the Lift Hoists being able to load and carry a unit and drop on the map?

NickAragua:
Definitely not units. I'll look into Lift Hoists/infantry being able to load cargo in a subsequent pull request.

NickAragua:
Not quite complete, but can be put through some initial testing/feedback.

megamek/src/megamek/common/actions/PushAttackAction.java
```diff
@@ -46,35 +46,27 @@ public ToHitData toHit(Game game) {
...
 
         if (ae.getGrappled() != Entity.NONE) {
             return "Unit Grappled";
         }
-
-        if (ae.isEvading()) {
-            return "attacker is evading.";
-        }
-
-        if (!game.getOptions().booleanOption(OptionsConstants.BASE_FRIENDLY_FIRE)) {
```
NickAragua, line 66: Removed this because it's being done in the base class already.

megamek/src/megamek/common/actions/PushAttackAction.java
```diff
@@ -46,35 +46,27 @@ public ToHitData toHit(Game game) {
      */
     protected static String toHitIsImpossible(Game game, Entity ae, Targetable target) {
         String physicalImpossible = PhysicalAttackAction.toHitIsImpossible(game, ae, target);
-        String extendedBladeImpossible = null;
```
NickAragua, line 49: Removed this because it's being done in the base class already.

repligator:
NPE when creating a 1,000 ton object.
[megamek.log](https://github.com/user-attachments/files/16313719/megamek.log)

repligator:
A "So-and-so" has fled the map, taking (Object) with it message would be nice.

repligator:
If certain long/complicated texts are pasted into the item name field (for example, an entire Wikipedia article) the player settings will go blank.
Screenshot_20240720_134253

The tonnage field will accept negative numbers, although this seems to have no affect on gameplay.

Everything else looks good.

NickAragua:
Heh, guess I better put a character limit on the names.

HammerGS:
Related: https://github.com/MegaMek/megamek/issues/2577

NickAragua:
Updated with testing results, will wait for approval/any other reviews.

repligator:
It looks like invincible objects are still getting destroyed under certain circumstances. In this case, the carrier was delimbed. It didn't generate any sort of "object destroyed" message either.
[gamelog.zip](https://github.com/user-attachments/files/16338950/gamelog.zip)
[megamek.log](https://github.com/user-attachments/files/16338952/megamek.log)
[where-is-the-ball.sav.gz](https://github.com/user-attachments/files/16338967/where-is-the-ball.sav.gz)

NickAragua:
Looks like the Dasher fell in round 4 due to leg loss which resulted in the ball being dropped but not put on the ground. I'll add some messaging and review why it didn't show up on the ground.

NickAragua:
Added a report for when cargo is dropped as a result of a fall.

repligator:
Another disappearing invincible object. Last seen being carried by `ID: 4` . Not sure if the disappearance was caused by the destruction of the mech, or the fact the pilot ejected.
[gamelog_1.zip](https://github.com/user-attachments/files/16363635/gamelog_1.zip)
[megamek.log](https://github.com/user-attachments/files/16363636/megamek.log)
[Uploading where_is_the_ball_2.sav.gz…]()

NickAragua:
Yep, missed that "workflow". Cargo is now dropped whenever an entity is destroyed for any reason (ejection, etc).

Note that cargo that winds up off the board (via push or skid or otherwise) isn't coming back.

megamek/src/megamek/common/Game.java
```diff
@@ -3273,8 +3278,77 @@ public Map<String, BehaviorSettings> getBotSettings() {
...
         this.botSettings = botSettings;
     }
-
-    /**
+    
+    /**
+     * Place a carryable object on the ground at the given coordinates
+     */
+    public void placeGroundObject(Coords coords, ICarryable carryable) {
+    	if (!groundObjects.containsKey(coords)) {
```
SJuliez, line 3286: The entire method code can be written as
`groundObjects.computeIfAbsent(coords, k -> new ArrayList<>()).add(carryable);`
(not tested)

megamek/src/megamek/common/ICarryable.java
```diff
@@ -0,0 +1,30 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with MegaMek. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package megamek.common;
+
+/**
+ * An interface defining all the required properties of a carryable object.
+ */
+public interface ICarryable {
```
SJuliez, line 25: I had the hope at some point for something like this to be placed under (i.e. extend) BTObject to be in a neat hierarchy (fwiw). It would then be the first place to add this override:
```
default boolean isCarryableObject() {
    return true;
}
```
It could add a few overrides to return false to the other checks else but that's not terribly important.
It would make sense to change getName() to generalName() (I was out of ideas for a better term) which makes it the equivalent of getChassis().

(Also... it could be named Carryable without the I as there's no old Carryable class in the way)

NickAragua: Didn't realize that there was a BTObject. I'll investigate how practical it'll be to update the interface to fit into the hierarchy.

The I in front of the name follows the general convention (and in MM as well to some extent) for interfaces; it's easier to tell at a glance that a piece of code is an interface if it has the I in front of it. Although I guess with the "default" functions nowadays the difference is a bit blurry.

SJuliez:
It is possible to create an EntityImage for Carryables and thus allow it to apply a Camo (the player color e.g. or a fallback brown) and even give them a facing (to rotate them). They could also have smoke :) (no, that doesnt make much sense)
It takes a bit of fluffy code to create it and get an image as a result, that would go into the GroundObjectSprite. I can figure that out if this sounds like a good idea.

NickAragua:
Refactored the ICarryable interface so it's part of the BTObject hierarchy.

I'm kind of ambivalent re: camo, as the cargo may not necessarily belong to the player (i.e. scenario rules call for "attacker nominates X squares that contain cargo"). The one thing that I do want to ideally do is have the cargo sprite appear "under" the units rather than over (it looks odd). Any advice on that?

SJuliez:
Ah ok, I didn't know if they should be player-colored. My intention was more to not keep them this black/white. Giving them a fixed brownish camo (we sure have one) is a simple option. But we can leave that for any point later.

Sprite ordering ... is not in an ideal place. There is Sprite.getSpritePriority() but it is only used with the xyzSpriteHandlers while the older code just draws the sprites according to type. Since the handler sprites are mostly the ones that go over everything (move envelope) they currently draw after the units and it wouldnt be helpful to swap that. I'll take a look if the unit sprites can be added to the allSprites treemap without a big hassle so they would respect draw order, then the sprite priority should help. Will be back on this.

repligator:
As of the latest push, the object names are displaying like this
Screenshot_20240726_085242

NickAragua:
Should be better now. Got a little too overzealous in my refactoring.

SJuliez:
I have created #5801
That one injects the entity sprites into the AbstractBoardView sprite drawing process and so allows the Carryable sprites to draw under units. You can pull it into this one, if you want. I merged them locally and it seemed to work fine.

NickAragua:
Ok, merged #5801 into this one. I think this one's ready to go now.

SJuliez:
The briefcase sprite doesnt scale with the board zoom.

Suggestion: replace the empty prepare() in GroundObjectSprite with this:

```
    @Override
    public void prepare() {
        updateBounds();
        image = createNewHexImage();
        Graphics2D graph = (Graphics2D) image.getGraphics();
        UIUtil.setHighQualityRendering(graph);
        graph.scale(bv.scale, bv.scale);
        graph.drawImage(CARGO_IMAGE, 0, 0, null);
    }
```

:)

repligator:
The report for a dropped object seems to get nested under the last physical attack. ID:95 was the one carrying the object.
Screenshot_20240728_064022

SJuliez:
I'm getting problems when trying to pick up stacked objects, i.e. more than 1 in a hex, specifically a light 2t and a heavy 30t. The game lets me pick up one, no questions asked (as the 30t is too heavy) but it does not really pick it up.

SJuliez:
Would it make sense to show carried objects in the unit tooltip? (or does it already?)

SJuliez:
Hoping I'm not missing something that's there already, I feel that it's likely we will at some point want to be able to identify specific carryables and maybe have them have an owner (thinking of capture the flag). May I suggest having ICarryable extend InGameObject instead of BTObject and adding the following to Briefcase.
As currently owner and id arent used anywhere, no other changes are necessary but this adds the infrastructure.


```
	private int id;
	private int ownerId;
	
	@Override
	public int getId() {
		return id;
	}

	@Override
	public void setId(int newId) {
		this.id = newId;
	}

	@Override
	public int getOwnerId() {
		return ownerId;
	}

	@Override
	public void setOwnerId(int newOwnerId) {
		this.ownerId = newOwnerId;
	}

	@Override
	public int getStrength() {
		return 0;
	}
```

NickAragua:
Added sprite scaling and unit tooltip; adjusted the ordering of "cargo dropped" messages during PSR resolution; adjusted object hierarchy; make sure to pick up the correct object when the "default" object being picked up isn't the first in the list.

SJuliez (approved):

PR #5708, opened 2024-07-08
Princess - using heat mapping for persistence/history awareness

SuperStucco:
Princess has a few weaknesses, among them the inability to only 'see' what's happening right now in the current game round.  It has no ability to forecast the future, or look to past rounds as a guide.  This PR addresses the latter using heat mapping, providing Princess the ability to continue 'moving to contact' in a somewhat intelligent manner even when it temporarily loses contact in double blind situations.

A heat map works much like a standard map, except instead of showing geographic/physical qualities such as elevation it tracks data.  Higher numbers (usually, but not exclusively) show more importance for a given position.  When the information is time sensitive, the values are modified for each time period so that older information becomes less important than more current information.

In this specific implementation, each heat map tracks two (or three, depending on settings) sets of information.  The first is general positions of combat forces, using unit BV with some adjustment by movement capability i.e. faster units are less likely to stay in the same location, so generate a lower weight than a slower unit with the same BV.  When units hang out/'camp' in the same locations over time this generates a spike, or hotspot, in the heat map that Princess will treat as a threat or source of support, depending on which team is being tracked.  The second set of information is a rough track on movement each round, so that when units move over the same ground this also generates hotspots similar to beating down a path in a field of grass.  The heat map also has the ability to track unit movement on an individual basis.  Princess will use detailed path tracking for units it controls, but will only estimate the path for other units using straight-line interpolation between last known position and current position.

Two initial applications of heat mapping have been implemented in Princess behavior.  The first, is the already-mentioned 'move to contact' behavior.  If Princess loses contact with enemy forces under double-blind rules it will first move towards any hotspots in an attempt to reacquire a target.  The longer Princess is out of contact the less important each hotspot gets, until it reverts to the default of advancing to the opposite map edge.  The second application is using heat mapping to track it's own units as well as it's allies.  It uses these hotspots of friendly activity as the basis for herding behavior, so that activity is anchored around high BV units or those that are on stationary overwatch.  If there is not enough information in the friendly heat map, such as widely dispersed low-BV units, then it reverts to the original behavior of calculating an average position of all friendly units.

There are several other potential applications for this which can be expanded on later, such as:

- artillery targeting
- delivery of FASCAM/in-game minefields
- damage avoidance
- predicting axis of attack
- persisting heat map data to file when fixed maps are used


megamek/src/megamek/client/bot/princess/HeatMap.java
```diff
@@ -0,0 +1,816 @@
...
+    }
+
+    /**
+     * Update the team activity tracker for a given position
+     * @param position    position to update
+     * @param adjustment  modifier to existing value
+     */
+    private void updateTeamActivityMap(Coords position, int adjustment) {
+        int mapValue = teamActivity.getOrDefault(position, 0) + adjustment;
+        teamActivity.put(position, mapValue);
```
Sleet01, line 536: This probably needs a valid position check, as it looks like the callers don't do any on-board/of-board validation.

megamek/src/megamek/client/bot/princess/HeatMap.java
```diff
@@ -0,0 +1,816 @@
...
+
+        // Get positions of tracked entities with known positions
+        List<Coords> activePositions = new ArrayList<>();
+        for (int curId : lastPositionCache.keySet()) {
+            Entity curEntity = game.getEntity(curId);
+            if (curEntity != null &&
+                    !curEntity.isDestroyed() &&
+                    curEntity.isDeployed() &&
+                    !curEntity.isCarcass() &&
+                    (isFriendlyTeam || curEntity.isVisibleToEnemy() || curEntity.isDetectedByEnemy())) {
```
Sleet01, line 574: This exact block appears several times; it could be replaced with a function, e.g.

    if (okayToTrack(curEntity)) {

with `okayToTrack` containing this long conditional.

Sleet01 (approved):
Some style comments, otherwise looks good!

We should schedule some performance tests with the test crew once this goes in, just to see if anything needs tweaking.

Sleet01:
> artillery targeting
> delivery of FASCAM/in-game minefields
> damage avoidance
> predicting axis of attack

I think we could use the same heatmap for all of these: a single "passability" map with each hex inversely weighted according to its MP consumption, side traversal costs, relative elevation change per side, internal and neighboring hazards, and number of PSRs it can generate.

On the one hand, this would require a large amount of processing to generate initially.  But on the other hand, we can "pre-bake" it at map generation time, or prior to deployment, and every Princess can be handed a copy.
We'd only need to update it when fires are started, or mines laid, stuff like that.

HammerGS:
Very excited to see how this works, and agree with Sleet01 on the future expansions.

SuperStucco:
Well, I made a few updates and tried to do a debug run to ensure everything is running as intended.  IDEA downloaded something and now it's not building, at all with this error:
`A problem occurred evaluating root project 'MegaMek'.
> Could not get unknown property 'mmBranch' for task ':cloneMMRepo' of type org.gradle.api.tasks.Exec.
`

HammerGS:
You might need to invalidate the caches, and let gradle refind everything

SuperStucco:
> You might need to invalidate the caches, and let gradle refind everything

How to go about doing that?

HammerGS:
In IDEA should be Under FILE - INVALIDATE CACHES, then should be the first two.

Hit us up on Discord, there have been some changes with the build scripts

PR #5717, opened 2024-07-09
Disable invalid unit test

pavelbraginskiy:
The test behavior turns out to not be deterministic, and it also flags several units which show as valid in MM and MML.
Until the problems with this test can be understood and resolved, it should be disabled.


Sleet01 (approved):
Regrettable but makes sense.
LGTM.

PR #5738, opened 2024-07-15
Munition autoconfig tweaks

Sleet01:
Thanks to some early QA work, a number of errors in the "final" munition autoconfig code have been identified.
This patch will:

1.  Fix Artemis munitions being selected for non-Artemis launchers;
2.  Update legality check on ammo types to always use year restrictions with the _option_ to use variable era tech levels;
3. Add tracking of enemy TSM units for consideration when weighting heat-producing munitions;
4. Split "missing munitionLoadoutSettings.xml" error into user warning and debug message with full exception;
5. Adjust "energy boat" unit accounting to only care about those that track heat;
6. Updated some `isAvailable()` methods to allow Clans access to IS equipment at the same availability from the Clan Invasion era onward, per IO pg 33.
7. Fix errors in getPrototypeDate and getProductionDate that would return no date for munitions missing the Common and/or Production entry (e.g. Air-Defence Arrow Missiles), causing some tech to be unavailable to certain combinations of faction and clan/mixed tech bases.
8. Updated unit tests

Testing:
- Tested with a number of factions, Clan/IS/Mixed tech levels, and eras, alternating Variable Tech off and on.

### DIFF
--- a/megamek/mmconf/munitionLoadoutSettings.xml
+++ b/megamek/mmconf/munitionLoadoutSettings.xml
@@ -75,6 +75,7 @@
     <entry key="defaultStandardMunitionWeight">2.0</entry>
     <entry key="defaultMissileStandardMunitionWeight">2.0</entry>
     <entry key="defaultDeadFireMunitionWeight">3.0</entry>
+    <entry key="defaultArtemisCapableMunitionWeight">0.0</entry>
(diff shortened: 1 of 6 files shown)


IllianiCBT (approved):
I took a read through and everything looks good to me :)

PR #5750, opened 2024-07-18
Setting up Utility classes with Gradle build

rjhancock:
Allow using Gradle to handle utilities classes for use.
Updated Gradle to 8.9
Removed a deprecation related to testing frameworks.
Removed build scripts no longer needed (replaced by build command)
Added output of mm-revision.txt file
Re-added Launch4J ini file.


gradlew
HoneySkull, line 1: Isn't this a generated file?  This change would get overwritten when generated right?

HoneySkull (approved):
This looks good, (I'm assuming the `gradlew` file only gets generated on project creation - so that should be good.)

gradlew
rjhancock: It gets updated when Gradle gets updated, no other time.

PR #5751, opened 2024-07-18
Add "Miss" marker and drift hits for better artillery visualization

Sleet01:
Hey kids, d'you like violence?  Do you want to fire N off-board artillery munitions through your opponents' eyelids?

Screenshot_20240718_000348
(This screenshot does not show the new "Missed" markers because I was using GM mode)

I noticed during some other testing that there are two major issues with our current artillery handling:
1. A miss is represented _graphically_ by a Hit marker at the original target location; a player has to scroll through round reports to find out what actually happened to their shots.  More often than not they fly off the map, or at least halfway across it - and without graphical representation of the final touchdown point, discerning where shots _did_ go requires painstaking reading and cross-referencing to map coordinates.  Not fun.
2. There is no graphical indication of where off-target shots _did_ land!  See above.

This patch fixes both issues by adding a "missed" marker at the original target point, and using the "hit" fountain of flame marker to designate hexes where artillery drifted to.

The "missed" markers, like "automatic hit" or "incoming" markers, are only visible to the player who fired the round, and their teammates.
The "gout of flame" drift markers, however, are visible to everyone but do not indicate whose shot caused them nor how far they drifted.

Screenshot_20240717_160819
Screenshot_20240717_160905
Screenshot_20240717_160923

Also fixes the map popup appearing immediately after pressing the right mouse button on Linux (and supposedly, Macs).  Now the popup should only appear after the right mouse button is released on any platform, and never while panning / scrolling the map around.

Testing:
- Ran several small and large games between human and bot, or human observing two bots, with off-board artillery


Sleet01:
^^^ rebased to get latest build scripts.

gsparks3:
One question: does the "miss" marker have to be teammate-only? In tabletop rules, the target hex is public information as soon as the shot lands. Nominally that's to ensure fairness/avoid cheating, which Megamek does anyway, but it does provide relevant information to the opposing side.

Sleet01:
It looks like the original target hex explicitly has to be declared so that opposing units can use shots to determine the artillery's approximate location for overrun; this is different from counter-battery fire, which only cares about where the shot _lands_.
I don't think we currently support overrun, but we might as well match the rules for it in case it becomes useful in the future.  I'll update the PR later today.

megamek/src/megamek/common/Board.java
```diff
@@ -1650,7 +1650,29 @@ public Hashtable<Coords, Collection<SpecialHexDisplay>> getSpecialHexDisplayTabl
     }
 
     public void setSpecialHexDisplayTable(Hashtable<Coords, Collection<SpecialHexDisplay>> shd) {
+        Hashtable<Coords, Collection<SpecialHexDisplay>> temp = new Hashtable<>();
```
NickAragua, line 1653: Can you go into some detail on what's happening in this method?

NickAragua (approved):
Looks fine, I'm just confused about the setSpecialHexDisplayTable method.

megamek/src/megamek/common/Board.java
Sleet01: It's a little complicated, but the reason is so that the Miss markers don't get wiped out by other players' specialHexes updates prior to being displayed.

For reasons I'm not entirely clear on, when a client (say, Princess) sends out an update to their own specialHexDisplay set, the entire `specialHexes` of the central board instance is just replaced wholesale by that packet's contents.  The server later sends a subset of its specialHexes HashTable out to each of the clients at various times, which is how _some_ icons get distributed - the game server also does some pruning so that hidden specialHexDisplays don't get sent to clients that shouldn't see them - but due to the overwriting, this means only the data of the last client to send its update in is then sent out.  I imagine this is to make sure all clients' states are synchronized, but it seems kind of sloppy.

This doesn't affect _some_ markers, such as the AutoHit, Incoming, or Adjusted markers, because they are actually drawn during the local client's board redraw based on known WAA instances, _not_ the specialHexes HashTable.
It also doesn't affect Notes, which are passed around using an entirely separate, non-destructive _addition_ packet (or deleted using a similarly targeted packet).

But because the Miss and Drifted Hit markers are generated during WAA _handling_, based on die rolls, they have to be inserted into the game server's copy of the board (where the handling takes place) and then re-distributed to the attached clients with the rest of the specialHexDisplay data (minus filtration if applicable) so that everyone can see them.  The method used by AutoHit et. al doesn't work because _these_ markers are created during the handling of the WAAs, and that handling isn't deterministic enough to re-run on all the clients in order to generate the markers _there_.  Plus we don't want to re-handle WAAs just to display some graphics.

Now, you might say, why not just copy all of the Miss or Drifted Hit SHDs from the current specialHexes instance to the passed-in `shd` prior to using it to overwrite `specialHexes`?  
In three words: `Concurrent Modification Exception`.  
I tried every possible combination of multithreaded data class, concurrency control, and weird iterator hack I could come up with; because copying the Miss data from `specialHexes` may require adding a new key to `shd`, or updating the linked list in an existing entry, there's no escaping the CME.

So what I do here is traverse `specialHexes` once, grab all the Miss entries with their coordinates, and then insert them into the `shd` version _after_ it's been assigned to the instance's variable.
This should only be O(N) time (where N is old `specialHexes` total number of entries) so I'm not concerned about speed.  In testing it's quite fast; the big slowdown is calculating AE damage from all the artillery shots.

I considered adding a new member (a la Infernos) to track just the Miss and Drifted Hit locations, but it seemed like a lot of extra work for little gain - for one thing, we don't need to do any of the later bookkeeping that Infernos require - and this approach works well enough.

If we end up adding more specialHexDisplays, especially with additional bookkeeping tasks, then we should probably consider adding a mergeSpecialHexes() method, or making it a more fully-featured class in its own right.

Sleet01:
Converting to draft so that I can add some UI controls and some bomb graphics.

megamek/src/megamek/common/Board.java
SJuliez: The middle of these loops 

```
                    if (!temp.containsKey(e.getKey())) {
                        temp.put(e.getKey(), new LinkedList<>());
                    }
                    temp.get(e.getKey()).add(special);
```

may be a case for replacing with
`temp.computeIfAbsent(e.getKey(), k -> new LinkedList<>()).add(special);`

Sleet01: @SJuliez That is so nice!  TYVM.

Sleet01:
Updates done.  New change: 
- bomb hits and misses, 
- Client board view options

Screenshot_20240719_110957

PR #5754, opened 2024-07-20
Fix #5287: Infernos do not ignite buildings when using "Ignite" target option

### LINKED ISSUE #5287: Inferno SRMs do not set buildings on fire when using collapse or ignite attack options
Seen in multiple MM versions, screenshot from 49.15.

Using "Clear" on a building with inferno SRMs will ignite it, using "Collapse" or "Ignite" will not, and only does damage. This is counterintuitive since the "Ignite" attack is used to set fire to hexes with most other weapons.

Additionally it seems like infernos do not deal double damage to terrain using the latter two attack types, unlike the "Clear" attack. All 4 attacks hit with 4 inferno missiles.

image

Sleet01:
Fixes the issue where Inferno SRMs would not ignite buildings when the "Ignite" target option.
Also laid groundwork for removing / refactoring the SRMInfernoHandler, as most of its code lives in other classes (or directly copies theirs) and some things _definitely_ need review if not complete rewrites.

Not touched:
- Changing collapse to ignite buildings - this would break TW handling of Infernos; Ignite is a TacOps option.
- Infernos not doing double damage to hexes when clearing - I could not find a rule reference for this.

Testing:
- Ran tests with all combos of SRM Infernos, Inferno Bombs, Inferno Artillery, and confirmed all work correctly.

Close #5287

### DIFF
--- a/megamek/src/megamek/common/WeaponType.java
+++ b/megamek/src/megamek/common/WeaponType.java
@@ -411,11 +411,11 @@ public class WeaponType extends EquipmentType {
         } else if (hasFlag(F_FLAMER)) {
             return 4;
         } else if (hasFlag(F_PLASMA)) {
-            return 2;
+            return TargetRoll.AUTOMATIC_SUCCESS;
         } else if (hasFlag(F_PLASMA_MFUK)) {
-            return 2;
+            return TargetRoll.AUTOMATIC_SUCCESS;
         } else if (hasFlag(F_INFERNO)) {
-            return 2;
+            return TargetRoll.AUTOMATIC_SUCCESS;
         } else if (hasFlag(F_INCENDIARY_NEEDLES)) {
             return 6;
         } else if (hasFlag(F_PPC) || hasFlag(F_LASER)) {
--- a/megamek/src/megamek/common/AmmoType.java
+++ b/megamek/src/megamek/common/AmmoType.java
@@ -485,6 +485,24 @@ public class AmmoType extends EquipmentType {
     }
 
     /**
+     * Analog to WeaponType.getFireTNRoll(), but based on munitions.
+     * See TO:AR pg 42
+     * @return
+     */
+    public int getFireTN() {
+        if (munitionType.contains(Munitions.M_INFERNO)) {
+            return TargetRoll.AUTOMATIC_SUCCESS;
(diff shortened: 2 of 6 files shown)


IllianiCBT (approved):
LGTM

IllianiCBT:
Holding off merging this till tomorrow so someone else has had a chance to take a look as I'm not as familiar with mm's inner-workings as I am mhq, but at least to my eyes, everything looks good.

megamek/src/megamek/common/WeaponType.java
```diff
@@ -411,11 +411,11 @@ public int getFireTN() {
         } else if (hasFlag(F_FLAMER)) {
             return 4;
         } else if (hasFlag(F_PLASMA)) {
-            return 2;
+            return TargetRoll.AUTOMATIC_SUCCESS;
         } else if (hasFlag(F_PLASMA_MFUK)) {
```
NickAragua, line 415: I know what it actually stands for, but I get a good chuckle every time I see this acronym

NickAragua (approved):

gsparks3:
> * Infernos not doing double damage to hexes when clearing - I could not find a rule reference for this.

Inferno SRMs should do 4 damage per missile to woods hexes when used to Clear them (Total Warfare, page 141). If this is happening, everything is fine; if not, that's a bug and should be fixed.

Interestingly, RAW they do not appear to have an equivalent effect on Jungle hexes, as Tactical Operations does not mention them while describing the rules for Jungle hexes.

Sleet01:
> > * Infernos not doing double damage to hexes when clearing - I could not find a rule reference for this.
> 
> Inferno SRMs should do 4 damage per missile to woods hexes when used to Clear them (Total Warfare, page 141). If this is happening, everything is fine; if not, that's a bug and should be fixed.
> 
> Interestingly, RAW they do not appear to have an equivalent effect on Jungle hexes, as Tactical Operations does not mention them while describing the rules for Jungle hexes.

Okay, I can confirm Inferno SRMs do 4 damage per hit to woods when clearing them, although I'm not sure what circuitous route the damage value takes to arrive at the `tryClearHex()` function with the correct number:

image

So I will go ahead and merge this.

PR #5762, opened 2024-07-21
Update artillery and bomb display with new graphics and revised/compacted logic

Sleet01:
Uses the new "Missed" icon for Artillery and Bombs, as well as:
- Drifted to / Drifted from hex info in Miss and Drift icons.
- Converts Drift icons into Hit icons (more fire) if they deal damage
- Cleans up some icon clearing and/or management
- Fixes a typo in the attack arrow color getter/setter methods, in prep for possibly adding drift arrows

New Artillery graphics:
image
image
image

New Bomb graphics:
image
image
image
image

Note: after further testing, I've determined that zooming in/out will trigger a redraw of the local client's board, so players can switch the Miss and/or Drift display options and immediately see the icons get displayed or hidden, just by zooming a little.  Outside of manually triggering a redraw after closing the Client option dialog, there's not a cleaner solution here, I don't think.

Testing:
- Various Bomb and Artillery attacks with, and on, a variety of units.


Sleet01:
Also note, the Drift->hit conversion currently only checks for damaging _units_; including a check for building damage could be done but would involve changing more existing functions, which I'm loathe to do.

megamek/src/megamek/common/weapons/ArtilleryWeaponIndirectFireHandler.java
```diff
@@ -579,8 +588,7 @@ protected int calcDamagePerHit() {
         }
         // BA Tube artillery is the only artillery that can be mounted by BA
         // so we do the multiplication here
-        if (ae instanceof BattleArmor) {
-            BattleArmor ba = (BattleArmor) ae;
+        if (ae instanceof BattleArmor ba) {
```
Sleet01, line 591: Idea cleaned these two lines up to a tidy check-and-cast.

HammerGS:
Curious in the last pic hex 1924, the red X with the fire exposions is that one hit and one bomb missed and drifted from their?

Sleet01:
> Curious in the last pic hex 1924, the red X with the fire exposions is that one hit and one bomb missed and drifted from their?

Yeah, 1924 was two Inferno bombs.
- one direct hit, using the Hit marker
- one miss, using the Miss marker
- - this drifted to1526 (the 2nd bomb picture) generating a Drift icon and an Inferno flamearker there.
- flames from the Inferno hit.

HammerGS:
> > Curious in the last pic hex 1924, the red X with the fire exposions is that one hit and one bomb missed and drifted from their?
> 
> Yeah, 1924 was two Inferno bombs.
> 
> * one direct hit, using the Hit marker
> * one miss, using the Miss marker
>   -- this drifted to 1526 (the 2nd bomb picture) generating a Drift icon and an Inferno flame marker there.
> * flames from the Inferno hit.

To me feels confusing, the tooltips are a big help. But that is a future thing. 

Tested and seems to work.

HammerGS (approved):
Tested and seems to work.

Sleet01:
As part of the clean-up, I also fixed #5763 without really thinking about it; I will merge this now so the fix goes in before the Nightlies run (I think).

Issue #5756, opened 2024-07-20
[0.50.0-SNAPSHOT] Unable to load savegames. 


Tzahr:
### Environment
02:51:53,126 INFO  [megamek.MegaMek] {main}
megamek.MegaMek.initializeLogging(MegaMek.java:142) - Starting MegaMek v0.50.0-SNAPSHOT
    Build Date: 2024-07-18T01:16:03.132246751
    Today: 2024-07-20
    Origin Project: MegaMek
    Java Vendor: Eclipse Adoptium
    Java Version: 21.0.3
    Platform: Windows 11 10.0 (amd64)
    System Locale: en_GB
    Total memory available to MegaMek: 2 GB
    MM Code Revision: Unknown
    MML Code Revision: Unknown
    MHQ Code Revision: Unknown

### Description
Cannot load any savegames in client, including those provided by others.

![image](https://github.com/user-attachments/assets/2d235a62-2841-41b9-b870-e395f77fa8a8) 

### Files
[megamek.log](https://github.com/user-attachments/files/16318114/megamek.log)

repligator:
If you save a game yourself, can you load that?

Tzahr:
Of course. 
First thing I did was check if I could load any of my own autosaves, but they all return the same error. 

[quicksave.sav.gz](https://github.com/user-attachments/files/16321738/quicksave.sav.gz)

repligator:
Well, good news is I can't load the save either.

HammerGS:
Any chance this was done between commits or nightlies? That could cause the issue.

Tzahr:
I had savegame loading issues in a previous Nightly - it'd load, but only to the lobby and lose all round progress, so the underway autosaves and quicksaves were useless. It was on my list to investigate further, and then I switched to a more recent build and MegaMek saves are not loading at all. 

I was given a test save to load for the Aerospace bombing bug when we were trying to replicate that, and that wasn't/isn't loading either.

HammerGS:
You have a Thunderbolt Mek (Any) by chance?

repligator:
> quicksave.sav.gz

This save _should_ fail. It makes perfect sense that it is failing. The very first entity never actually ends.
The fact that other saves, that are formatted correctly, are also failing for you is puzzling. They are even throwing the same error.
`XML document structures must start and end within the same entity.`
Despite the fact that I can open the save file in a text editor and clearly see that it does, in fact, start and end with the same entity.

[...]

PR #5764, opened 2024-07-21
Save File Fix

rjhancock:
Updates to Serialization Helper
Minor clean ups to AbstractGameManager, GameManagerSaveHelper, and Server.
Removed a nested try/catch block.

### DIFF
--- a/megamek/src/megamek/server/GameManagerSaveHelper.java
+++ b/megamek/src/megamek/server/GameManagerSaveHelper.java
@@ -18,19 +18,27 @@
  */
 package megamek.server;
 
-import megamek.MMConstants;
-import megamek.common.net.enums.PacketCommand;
-import megamek.common.net.packets.Packet;
-import megamek.common.util.SerializationHelper;
-import org.apache.logging.log4j.LogManager;
-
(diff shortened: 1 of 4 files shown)


rjhancock:
Should resolve #5756

megamek/src/megamek/server/GameManagerSaveHelper.java
```diff
@@ -95,8 +108,10 @@ public void sendSaveGame(int connId, String fileName, String localPath) {
                 finalFileName = fileName + MMConstants.SAVE_FILE_GZ_EXT;
             }
         }
-        localPath = localPath.replaceAll("\\|", " ");
+
+        localPath = localPath.replace("|", " ");
```
Sleet01, line 112: Was the escaping of the "|" actually unnecessary here?

rjhancock: `replaceAll` defaults to a regex pattern and create unneeded overhead if it's just a string.
`replace` defaults to string and only moves to regex when needed.

Sleet01 (approved):
LGTM

PR #5775, opened 2024-07-23
Name Validation Utility Fix.

rjhancock:
Removed lines that were duplicated left/right
Refactored Utility to DRY up the code.

### DIFF
--- a/megamek/src/megamek/utilities/NameChangesValidator.java
+++ b/megamek/src/megamek/utilities/NameChangesValidator.java
@@ -21,8 +21,12 @@ package megamek.utilities;
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/utilities/NameChangesValidator.java
```diff
@@ -62,33 +65,57 @@ public static void main(String... args) {
...
-                        errors++;
-                    }
-                }
+
+                lines.add(line);
             }
-        } catch (Exception ex) {
-            logger.error(ex, STRING_EXCEPTION);
+
+            br.close();
```
IllianiCBT, line 89: Do we need to close the reader in the catches too? As (unless I'm thinking about things wrong) wouldn't we get a memory leak if the streamed resources were left dangling?

rjhancock: Utility exists if there is an exception so it'll close automatically.

IllianiCBT (approved):

PR #5786, opened 2024-07-24
MMLogger Migrations

rjhancock:
Move from LogManager to MMLogger.
License Updates
Removal of embedded try block.
Added debug log level reporting.
Adjusted info, warn, debug to accept variable args for String.format.


SJuliez (approved):
well, I guess.

PR #5797, opened 2024-07-26
Fix #5795: Implement TW LOS and Forest rules for Low Altitude maps

### LINKED ISSUE #5795: Low-altitude Map Terrain not Obstructing Fire
### Environment
0.49.19 & 0.49.19.1

### Description
Terrain on the low-altitude map is not obstructing line-of-sight for attacks, as described on _TW_ p. 242

### Files
Image provided by RaozSpaz.
image-48

Sleet01:
It looks like this has been broken since before 0.48.0, maybe since 2003: we have been treating Low-Altitude map terrain as if it were Ground Map terrain, which means it never obscures Aerospace units from each other.

Buildings and landed Dropships (which are treated as buildings) should not obscure lines of site between any Aero units using Altitude on a Low-Altitude map.
Additionally, forests should only rise 1 level above the terrain they are on when Low Altitude maps are created from Ground Maps - thankfully the Hex class provides a mechanism to tweak this stat for forest / jungle / foliage hexes.

Total Warfare p. 81 says:

```
Each level on a ground mapsheet hex corresponds to an
aerospace altitude. For example, a hill marked Level 2 equals
Altitude 2. Units can enter that hex only if flying at Altitude 3
or higher (or 2 if landing; see p. 87). Wooded hexes rise one
altitude above the underlying terrain. For simplicity sake,
building hexes do not rise above the underlying terrain and
are ignored by aerospace units (unless the unit is attempting
to land, see p. 87).
```

Unfortunately it was much simpler to reverse the normal conversion and use the units' Altitudes as Elevation, rather than convert Elevation into Altitudes, due to the way that LosEffects is written.

Testing:
- Various ASFs and Conventional Fighters on a multitude of Low Altitude maps constructed from existing templates.

Close #5795 

image

### DIFF
--- a/megamek/src/megamek/common/Board.java
+++ b/megamek/src/megamek/common/Board.java
@@ -1020,6 +1020,10 @@ public class Board implements Serializable {
                         args[i++] = st.ttype == StreamTokenizer.TT_NUMBER ? (int) st.nval + "" : st.sval;
                     }
                     int elevation = Integer.parseInt(args[1]);
+                    if (mapType == T_ATMOSPHERE) {
+                        // All foliage on low-altitude maps have height 1
+                        args[2] = args[2].replaceAll("foliage_elev:\\d{1,}", "foliage_elev:1");
+                    }
                     // The coordinates in the .board file are ignored!
                     nd[index] = new Hex(elevation, args[2], args[3], new Coords(index % nw, index / nw));
                     index++;
--- a/megamek/src/megamek/common/Board.java
+++ b/megamek/src/megamek/common/Board.java
@@ -1020,10 +1020,6 @@ public class Board implements Serializable {
                         args[i++] = st.ttype == StreamTokenizer.TT_NUMBER ? (int) st.nval + "" : st.sval;
                     }
                     int elevation = Integer.parseInt(args[1]);
-                    if (mapType == T_ATMOSPHERE) {
-                        // All foliage on low-altitude maps have height 1
-                        args[2] = args[2].replaceAll("foliage_elev:\\d{1,}", "foliage_elev:1");
-                    }
                     // The coordinates in the .board file are ignored!
                     nd[index] = new Hex(elevation, args[2], args[3], new Coords(index % nw, index / nw));
                     index++;
--- a/megamek/src/megamek/common/LosEffects.java
+++ b/megamek/src/megamek/common/LosEffects.java
@@ -1224,7 +1224,7 @@ public class LosEffects {
             // Intervening Smoke and Woods
             int woodsLevel = hex.terrainLevel(Terrains.WOODS);
             int jungleLevel = hex.terrainLevel(Terrains.JUNGLE);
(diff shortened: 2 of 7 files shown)


megamek/src/megamek/common/Board.java
```diff
@@ -1020,6 +1020,10 @@ public void load(InputStream is, @Nullable List<String> errors, boolean continue
                         args[i++] = st.ttype == StreamTokenizer.TT_NUMBER ? (int) st.nval + "" : st.sval;
                     }
                     int elevation = Integer.parseInt(args[1]);
+                    if (mapType == T_ATMOSPHERE) {
+                        // All foliage on low-altitude maps have height 1
+                        args[2] = args[2].replaceAll("foliage_elev:\\d{1,}", "foliage_elev:1");
```
SJuliez, line 1025: Won't this affect graphics? Showing all woods as "foliage" (low bushes)
It is sad that we actually have Terrains.getTerrainElevation(...) that already retrieves the height depending on if the map is atmospheric. It doesnt seem to be used in LosEffects.
Also, I think this change would much more naturally fit into Board.initializeFoliageElev(). This does post load processing anyway.

megamek/src/megamek/common/LosEffects.java
```diff
@@ -1174,38 +1192,42 @@ private static LosEffects losForCoords(Game game, AttackInfo ai,
...
                 }
             }
-            
+
+
             // Intervening Smoke and Woods
-            int smokeLevel = hex.terrainLevel(Terrains.SMOKE);
             int woodsLevel = hex.terrainLevel(Terrains.WOODS);
             int jungleLevel = hex.terrainLevel(Terrains.JUNGLE);
             int foliageElev = hex.terrainLevel(Terrains.FOLIAGE_ELEV);
```
SJuliez, line 1227: maybe, Terrains.getTerrainElevation() can be used to find foliageElev instead of the terrain level with lowAltitude as parameter so the entire Board change is unnecessary?

SJuliez (commented):
I have some comments.

megamek/src/megamek/common/Board.java
Sleet01: It does reduce the graphics to foliage, although I personally found it a good reminder that the "level" would only be one higher.

I looked at getTerrainElevation() but wasn't sure that it would meet our needs.  I'll re-examine it, and look at initializeFoliageElev().

megamek/src/megamek/common/LosEffects.java
Sleet01: I'll check it out.

Sleet01: Incorporated `getTerrainElevation()` where possible.

megamek/src/megamek/common/Board.java
Sleet01: Removed this stuff; unnecessary.

SJuliez (approved):

PR #5801, opened 2024-07-26
BoardView sprite adaptations

SJuliez:
This is an addition for #5706 to either pull into that one or merge by itself. It allows Carryable sprites to draw under units.


SJuliez:
Now part of #5706

PR #5809, opened 2024-07-27
Use best hex for FoV highlighting for multi-hex units

### LINKED ISSUE #5785: Dropship Line of Sight Issue - Megamek 0.49.15.1
### Environment

This issue applies to the 0.49.19.1 release of MegaMek, and has been reliably replicated in both standalone and MekHQ's megamek apps. 

I am currently running the apps on Windows 10 AMD 64-bit.

Java Version: 21.0.2

### Description

It appears that grounded dropships are blocking their own line of sight. All grounded dropships appears to have no line of site to any space outside of their own hull. I suspect that this is because line of sight is being drawn from the hex within the center of the dropship, but is being blocked by the hexes composing the dropship's hull, like other unit's LOS are supposed to be. 

The attached save file is using a Union-class to demonstrate, but this problem has persisted among all spheroid and aerodyne dropships

 I've tested. The instant the unit's sprite becomes the seven-hex version used for ground maps, the unit goes blind. I've also tested launching dropships back to the air, and the instant they revert to their one-hex, in-flight sprite, their LOS returns. 

This issue persists through the movement and shooting phases. However I have to report that, counter to my previous reports, I've discovered that this bug may be mostly graphical. On further testing, I can still select and target any valid units within valid firing arcs on the ship and attack them - there's just no way to tell which enemy units are in LOS before clicking on them, cycling to a valid weapon group, and seeing if they're actually targetable. Which is a bit clunky. 

### Files

Here's a save game to help demonstrate, on wide-open terrain. The dropship is currently at 1 altitude, so it can easily be demonstrated that this is an issue with grounded ships, only. 
[Dropship-Error.sav.gz](https://github.com/user-attachments/files/16362341/Dropship-Error.sav.gz)

SJuliez:
FoV highlighting was not considering secondary positions. Fixes #5785 
This also adds a little test scenario with a grounded DS which is going to continue to work in the next version (unlike a savegame).

### DIFF
--- a/megamek/data/scenarios/Test Setups/DS_LOS.mms
+++ b/megamek/data/scenarios/Test Setups/DS_LOS.mms
@@ -1,0 +1,15 @@
+MMSVersion: 2
+name: Grounded DS Test
+planet: None
+description: A grounded DS on a small map for use in range/LOS/FoV testing
+
+map: AGoAC Maps/16x17 Grassland 2.board
+
+factions:
+- name: Test Player
+
+  units:
+  - fullname: Intruder (3056)
+    at: [ 10, 9 ]
+    altitude: 0
+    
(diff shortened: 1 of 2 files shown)


Sleet01 (approved):
LGTM!

PR #5814, opened 2024-07-28
Fix 5796: Safety check for entities with no hex location, e.g. just-loaded infantry

### LINKED ISSUE #5796: MegaMek battle received error "uncaught java.lang.NullPointerException detected."
### Environment
15:24:49,046 INFO  [megamek.MegaMek] {main}
megamek.MegaMek.initializeLogging(MegaMek.java:118) - Starting MegaMek v0.49.20
	Build Date: 2024-06-28T16:07:06.718979300
	Today: 2024-07-24
	Origin Project: MekHQ
	Java Vendor: Eclipse Adoptium
	Java Version: 17.0.12
	System Locale: en_US
	MM Code Revision: Unknown
	MML Code Revision: Unknown
	MHQ Code Revision: Unknown
### Description
During battle at end of phase, game experience a null pointer error

### Files
[issue report.zip](https://github.com/user-attachments/files/16368429/issue.report.zip)

Sleet01:
Prevents the NPE when hex is null due to the unit having just been loaded the previous Movement phase.

Testing:
- Ran saves from OP with and without fix, and confirmed via debugging that the just-loaded unit was at fault.

Close #5796

### DIFF
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -10042,6 +10042,10 @@ public abstract class Entity extends TurnOrdered implements Transporter, Targeta
         boolean canHit = false;
         boolean friendlyFire = game.getOptions().booleanOption(OptionsConstants.BASE_FRIENDLY_FIRE);
 
+        if (getPosition() == null) {
+            return false; // not on board?
+        }
+
         if ((this instanceof Infantry)
                 && hasWorkingMisc(MiscType.F_TOOLS,
                         MiscType.S_DEMOLITION_CHARGE)) {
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/Entity.java
```diff
@@ -10046,6 +10046,9 @@ public boolean isEligibleForPhysical() {
                 && hasWorkingMisc(MiscType.F_TOOLS,
                         MiscType.S_DEMOLITION_CHARGE)) {
             Hex hex = game.getBoard().getHex(getPosition());
+            if (hex == null) {
```
kuronekochomusuke, line 10049: there is a check not on board later in this function, should that be moved before this block.  to protect other calls beside just this one also?
```
       if (getPosition() == null) {
            return false; // not on board?
        }
```

kuronekochomusuke (changes requested):

megamek/src/megamek/common/Entity.java
Sleet01: Good find!  I've moved the getPosition() check to the top.  I will keep the hex == null check as well, as I have a sneaking suspicion it's possible for units with a  non-null position to still have a null hex value...

kuronekochomusuke (approved):

PR #5816, opened 2024-07-28
[After #5706] Carryable objects scenario v2 language

SJuliez:
This moves part of the carryable code up from Game to AbstractGame as I don't see why other games types wouldnt use this. Also, adds MM scenario language support and a test setup scenario (can be found in the scenario chooser in the "Test Setups" tab.


Sleet01:
I've noticed that loading pre-#5706 in-progress games into the latest code causes NPEs when mousing over the map, because the loaded Game's groundObjects Map isn't initialized and every tooltip is now trying to get info from it.

Do we care enough to safety the calls that refer to it, or are we willing to live with this due to the short time before 0.50.0 goes out?

NickAragua (approved):
This looks fine, but I'll need to adapt the heavy lifter PR once this is merged.

HammerGS:
This has picked up some conflicts.

SJuliez:
@NickAragua I updated this a bit. I added some commenting to ICarryable and changed the damage() method to return as a boolean, if the object was destroyed. My thought was that there could be object classes that dont detract damage from tonnage at all (or at a low rate) but are destroyed when one damage block is big enough (hit by AC/20). Other objects might be destroyed by any damage regardless of their tonnage. I worded the comments accordingly.

Issue #4902, opened 2023-11-19
49.16 - Beast infantry & Underwater Units - Princess cannot Move Orcas or Hipposaurs, or other Conventional underwater units.


Thom293:
Title.  Similar to the other.  She cant move the underwater units.  I added conventional underwater units to oppose her, and she didnt move them either for some reason.  If I add mechs to the Opfor they will travel the whole map and klll the Orcas.  Sometimes the Orcas and Hipposaurs will fire back, but usually not fire at all.  The but the conventional units on their team will never move from their starter hex.  Also, the HIpposaurs drown on turn 3 one game.  Not sure why.  

I am honestly unfamiliar with Underwater units and infantry, so I have no idea if this is a setting or intended behaviour. 

Sample scenario attached. 

Screen shows units I added in lobby and drowning.

![image](https://github.com/MegaMek/megamek/assets/23645569/bdb2d692-f9e4-40d4-89fe-86f1c0a5ba0e)

![image](https://github.com/MegaMek/megamek/assets/23645569/b6fee714-fe81-4f51-82be-63d00f453d83)

[waterbeasts.sav.gz](https://github.com/MegaMek/megamek/files/13402517/waterbeasts.sav.gz)

HammerGS:
Related #4901

Thom293:
Was looking for something else and came across this.  This was fixed by this PR:  https://github.com/MegaMek/megamek/pull/5822

Closing.

Issue #5361, opened 2024-04-13
RFE:  Princess needs "Hazard" terrain list updated


Thom293:
Im putting this up to remind myself, but its for anyone that wants to handle it.  I thought I had put it up before but I guess not.

The below would need similar fixes to the Ice fix I did before.  I think.

Right now, princess does not see at least the following hexes as a hazard, or doesnt see them as a hazard in all of the necessary situations:

Swamp
Mud
Quicksand

She will gleefully walk/run/jump into any of those hex and strand herself even when there are ample other alternatives.  There are probably more but I dont have my notes as Im not at home.

On the opposite end of the spectrum, she will not move underwater units, or at least some of them, at all.  My guess is that she sees high depth water hexes like a land unit would and so doesnt move UWU into them.  I think a UWU move exception needs to be coded for water, and that would solve the issue.  That is just a guess though.  Related:  https://github.com/MegaMek/megamek/issues/4902

repligator:
I need to look into this more but I don't think Princess understands black ice either.

Thom293:
Lave and related hotfoot things.  May have a few other things to add too.

Thom293:
It was suggested a few weeks back that I ping some old bugs before .50 ships, just in case people are looking. I forgot to do so, so am doing now.

This is an old persistent one that is related to MM basic map navigation and it would be nice if it were fixed, if someone has time to look at it. Thank you.

Related to #5614

Sleet01:
I can take a short pass at this today.
We will want to _brutally_ test any changes prior to cutting the next release, however.

Thom293:
If you are able to get these working I will commit to testing them.  Thank you.

Sleet01:
I will examine the following:

Swamp
Mud
Quicksand
Magma 
UWU underwater

Lava doesn't seem to be defined separately; it is just a synonym for "Magma" used in some of the map-tagging code.

Thom293:
I'm not super sure on lava. One is open firey stuff and one is firey stuff under a crust you can fall through. So akin to water under ice, I'd guess. I don't have access to my books just now. Sorry.

Sleet01:
Yeah, we model that as "Crust" and "Liquid" Magma, which is how TO:AR defines it.

Looks like "Lava" is a special form of flooding Magma and will require flow modeling, so that should get wrapped into future Flood work.

Thom293:
[...]

PR #5111, opened 2024-02-01
Princess non-combat PSR reduction.

Thom293:
Summary: This PR changes the maximum fall shame values and prevents Princess from taking most unnecessary non-combat PSR rolls, and does not hurt her combat performance at all.  Resulting in drastically fewer "she killed half of her force before I ever saw her" issues with the same (or better) combat performance (becase she has more living units).  

The "Fall Shame" scale maximum has been increased from 200 to 500 and 500 has been made default.  The old default of 100 has been moved down to spot 4, but I dont think there is really any reason to ever use it, honestly.  

I have tested it extensively.  Its impossible to test every possible combination, but I have tested extremes that I think will cover most cases.  I tested Mechs, Vehicles, VTOL, ASF and infantry.  Its been tested via terrain hazards, pavement, gravity (high and low) and the most extreme weather and light conditions, including combinations thereof.  There has been no performance reduction in any condition that I could find.  Detailed testing info posted below, for anyone that cares to look at it.

Caveat 1:  She will still take non-combat PSRs in Swamp, Quicksand, Ice (with no water under it), Mud, and a few other terrains because she does not currently "see" them as hazards.  I hope to fix those issues in a future PR.

Caveat 2:  She will still take in-combat PSRs and sometimes wreck herself, but is a much smaller issue because if there is a time to risk a PSR its in combat.

This fixes #4878.  I will add another issue later for the separate Swamp/Mud/Quicksand, and other hazards.

Detailed Testing, and its TLDR:
image

image

PR #5822, opened 2024-07-29
Implement RFE 5361: enhance Princess handling of hazardous terrain

Sleet01:
Adds additional hazard analysis for:
- Magma (Crust and Liquid, aka "Lava")
- Swamp
- Mud
- Water (for Submarine / UMU units)

in order to prevent Princess from exposing some unit types to excessive hazard while choosing paths.
Implements RFE #5361 

Testing:
- Ran some Bot v Bot scenarios with the relevant Terrains and movement types.
- Added new unit tests for specific interactions.

Note: UMU and Submarine Infantry units can now traverse Water hexes more easily, but pathing in/out of Water hexes seems to be broken.

### DIFF
--- a/megamek/src/megamek/client/bot/princess/BasicPathRanker.java
+++ b/megamek/src/megamek/client/bot/princess/BasicPathRanker.java
@@ -789,7 +789,9 @@ public class BasicPathRanker extends PathRanker implements IPathRanker {
                                               Terrains.WATER,
                                               Terrains.BUILDING,
                                               Terrains.BRIDGE,
-                                              Terrains.BLACK_ICE));
+                                              Terrains.BLACK_ICE,
+                                              Terrains.SWAMP,
+                                              Terrains.MUD));
 
         int[] terrainTypes = hex.getTerrainTypes();
         Set<Integer> hazards = new HashSet<>();
@@ -834,6 +836,12 @@ public class BasicPathRanker extends PathRanker implements IPathRanker {
                     break;
                 case Terrains.BLACK_ICE:
                     hazardValue += calcIceHazard(movingUnit, hex, step, movePath, jumpLanding, logMsg);
+                    break;
+                case Terrains.SWAMP:
+                    hazardValue += calcSwampHazard(movingUnit, hex, step, movePath, jumpLanding, logMsg);
+                    break;
+                case Terrains.MUD:
+                    hazardValue += calcMudHazard(movingUnit, hex, step, movePath, jumpLanding, logMsg);
                     break;
             }
         }
@@ -959,6 +967,12 @@ public class BasicPathRanker extends PathRanker implements IPathRanker {
             return 0;
         }
 
+        // Submarine units should be fine; Orca-riding Infantry goes here.
+        if (EntityMovementMode.SUBMARINE == movingUnit.getMovementMode()) {
+            logMsg.append("Submarine locomotion unit (0).");
+            return 0;
+        }
+
         // if we are crossing a bridge, then we'll be fine. Trust me.
         // 1. Determine bridge elevation
         // 2. If unit elevation is equal to bridge elevation, skip.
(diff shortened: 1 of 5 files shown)


Thom293:
Thank you Sleet.  Will test in a few hours. 

I believe this will also close #4902 ?

HammerGS:
Does this mean that Princess would be able to operate on water fully now?

Thom293:
RE: Pathing in and out of water.  I dont know for sure.  Maybe it was broken already.  But I did make some heavy handed changes before to Ice and PSR tolerance that may have some influence.  Maybe not, but for full disclosure, here they are if someone wants to see if I broke the water pathing:

https://github.com/MegaMek/megamek/pull/5041
https://github.com/MegaMek/megamek/pull/5111

Sleet01:
> Does this mean that Princess would be able to operate on water fully now?

I definitely saw UMU, Orca, and SCUBA units moving in and out of water, and at their full MP within water.
The water pathing may still need work, though: I only saw a couple of units exit to land (or enter the water _from_ land) in my testing.

Sleet01:
> Thank you Sleet. Will test in a few hours.
> 
> I believe this will also close #4902 ?

I believe so; I'll keep testing when I can.

repligator:
Everything is a definite improvement so far. Princess still makes some questionable decisions (for example choosing to walk over the hex where the magma crust has already cracked instead of one where it hasn't), but the important part (seeing the terrain as hazardous in the first place) is working.

Sleet01:
> Everything is a definite improvement so far. Princess still makes some questionable decisions (for example choosing to walk over the hex where the magma crust has already cracked instead of one where it hasn't), but the important part (seeing the terrain as hazardous in the first place) is working.

I think I can explain this as: there's a chance to get stuck if walking over a Crust hex, if the crust hex cracks.  That's a 1/6 chance, but then the expected damage from getting stuck is (Damage from Lava * (1.0 + chance of getting stuck + number of turns to escape)).  So it's very possible for the heat and damage from walking through Liquid Magma to be a _lower_ risk than possibly getting stuck after the Magma Crust breaks, especially for lower-PS pilots.

Sleet01:
Added increasing hazard from Lava as unit takes damage.

Sleet01:
It looks like no amount of damage, heat, or hazard will prevent determined Berserker Princess units from at least dipping their toes in some Liquid Magma, especially now that we've removed the (erroneously applied) +4 PSR to bog down when just crossing Liquid Magma hexes.

But the latest update should dissuade all but the most bloodthirsty.

megamek/src/megamek/client/bot/princess/BasicPathRanker.java
```diff
@@ -959,6 +1004,12 @@ private double calcWaterHazard(Entity movingUnit, Hex hex, MoveStep step, MovePa
             return 0;
         }
 
+        // Submarine units should be fine; Orca-riding Infantry goes here.
+        if (EntityMovementMode.SUBMARINE == movingUnit.getMovementMode()) {
```
SJuliez, line 1008: the movement mode has some useful bool methods, .isSubmarine()

megamek/src/megamek/common/Terrain.java
```diff
@@ -607,6 +607,7 @@ public int getBogDownModifier(EntityMovementMode moveMode, boolean largeVee) {
                     return 0;
                 }
             case MAGMA:
+                // Only applies when jumping into a hex.
```
SJuliez, line 610: seeing this and that the method already takes some additional parameters it's questionable if this should actually be in Terrain at all. Nothing's easy in BT. (Just commenting, no action required)

SJuliez (approved):
looks good to me. EntityMovementMode has some nice boolean methods (e.g. isHoverOrWiGE()) but those can be refactored in at another point.

PR #5835, opened 2024-08-01
Allow bot vs bot in ScenarioV2 and test setup

SJuliez:
A tiny change that allows leaving the human player in a scenario with no units and thus have two bots fight it out.

Best hide whitespace when reviewing. The diff looks big, but it only surrounds the unit reader with a check if there is a unit node, new line 218


Sleet01 (approved):
LGTM

PR #5842, opened 2024-08-02
SBF updates

SJuliez:
Some additions to SBF play. Minor impact on anything else.


HammerGS:
Is this a draft?

SJuliez:
Only in the sense that SBF is not complete :)

repligator:
`map: AGoAC Maps\16x17 Desert 2.board` won't load on Linux. Loads fine when changed the `\` is changed to `/` . I realize that SBF is very much WIP, and multi OS support for parsing MMS files may not even be started yet, but I reckoned it was worth pointing out regardless.

SJuliez:
@repligator Ah thanks, good point. I just hadnt thought about it when copying in the filename.

Sleet01 (approved):
Looking good!  This is really clean so far.

PR #5845, opened 2024-08-03
Force destruction of out-of-control DropShips and larger

### LINKED ISSUE #770: New Errata - Impacting crashes of Aero Units
http://bg.battletech.com/forums/index.php?topic=54315.0

Crashes (p. 81)
⑤ Replace the first paragraph with the following:

     If a unit’s altitude matches the level of the hex it occupies and it does not attempt to land (see Landing, p. 87), it crashes. If the unit was Out of Control when the crash happened and is a DropShip or larger craft, it is destroyed. Otherwise, the unit’s controller rolls 2D6, multiplying the result by 10 and then multiplying again by the current velocity of the unit. If the unit had no velocity when it crashed, use a value of 2 for velocity when calculating damage. The final result is the number of damage points the unit suffers.
     Apply damage to randomly determined locations in ten-point groups. If the unit is a fighter, aerodyne DropShip or aerodyne small craft, use the Nose Hit Location Table for the appropriate unit class. If the unit is a spheroid DropShip or small craft, use the Aft Hit Location Table instead.

Sleet01:
This is the final step to completing the Errata from #770: enforce destruction of DropShips (and larger) that crash while out of control.

Note: this may be gilding the lily, as the chances of most DropShips surviving crashes is slim already.  But this will ensure that, after all damage is dealt and terrain is re-arranged, the crashing unit will be utterly kaput.

Testing:
- Crashed some out-of-control DropShips

Close #770

### DIFF
--- a/megamek/src/megamek/common/MovePath.java
+++ b/megamek/src/megamek/common/MovePath.java
@@ -15,7 +15,6 @@
 package megamek.common;
 
 import megamek.client.bot.princess.Princess;
-import megamek.common.MovePath.MoveStepType;
 import megamek.common.annotations.Nullable;
 import megamek.common.options.OptionsConstants;
 import megamek.common.pathfinder.AbstractPathFinder;
@@ -207,9 +206,9 @@ public class MovePath implements Cloneable, Serializable {
     }
 
     public MovePath addStep(final MoveStepType type, final Map<Integer, Integer> additionalIntData) {
-    	return addStep(new MoveStep(this, type, additionalIntData));
+        return addStep(new MoveStep(this, type, additionalIntData));
     }
-    
+
     public MovePath addStep(final MoveStepType type, final boolean noCost, final boolean isManeuver, final int maneuverType) {
         return addStep(new MoveStep(this, type, noCost, isManeuver, maneuverType));
     }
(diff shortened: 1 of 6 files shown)


megamek/src/megamek/common/MovePath.java
```diff
@@ -207,9 +206,9 @@ public MovePath addStep(final MoveStepType type, final boolean noCost) {
     }
 
     public MovePath addStep(final MoveStepType type, final Map<Integer, Integer> additionalIntData) {
-    	return addStep(new MoveStep(this, type, additionalIntData));
+        return addStep(new MoveStep(this, type, additionalIntData));
```
Sleet01, line 209: IDEA just reallllly disliked the formatting in this file.

megamek/src/megamek/server/GameManager.java
```diff
@@ -4898,6 +4898,7 @@ else if (game.getBoard().onGround()) {
     private Vector<Report> processCrash(Entity entity, int vel, Coords c) {
         Vector<Report> vReport = new Vector<>();
         Report r;
+        boolean destroyDropShip = false; // save for later
```
Sleet01, line 4901: New code part 1.

megamek/src/megamek/server/GameManager.java
```diff
@@ -4913,6 +4914,11 @@ private Vector<Report> processCrash(Entity entity, int vel, Coords c) {
             vReport.add(r);
             entity.setDoomed(true);
         } else {
+            // If out of control, destruction (of DropShips and larger) is assured.
+            destroyDropShip = (entity instanceof Dropship || entity instanceof Jumpship)
+                              && ((IAero) entity).isOutControlTotal();
+
+            // Technically bring to a halt; actual landing is handled elsewhere.
```
Sleet01, line 4921: New code part 2.

megamek/src/megamek/server/GameManager.java
```diff
@@ -5161,6 +5168,14 @@ private Vector<Report> processCrash(Entity entity, int vel, Coords c) {
...
         }
 
+        if (destroyDropShip) {
+            // Out-of-control DropShip that crashes is automatically destroyed.
+            r = new Report(9708, Report.PUBLIC);
+            r.indent();
+            r.addDesc(entity);
+            vReport.addAll(destroyEntity(entity, "Crashed while out of control!"));
+        }
+
```
Sleet01, line 5178: New code part 3; actually finishes the job.

megamek/src/megamek/server/GameManager.java
```diff
@@ -18629,7 +18644,8 @@ else if ((entity.heat >= 14) && !entity.isShutDown()) {
                 r.add(boom);
                 if (entity.getCrew().hasActiveTechOfficer()) {
                     rollValue += 2;
-                    String rollCalc = rollValue + " [" + diceRoll.getIntValue() + " + 2]";;
+                    String rollCalc = rollValue + " [" + diceRoll.getIntValue() + " + 2]";
+                    ;
```
NickAragua, line 18648: may as well get rid of the extra semi-colon

megamek/src/megamek/server/GameManager.java
```diff
@@ -4930,6 +4936,7 @@ private Vector<Report> processCrash(Entity entity, int vel, Coords c) {
             crateredElevation = h.getLevel() - 2;
         }
         if (entity instanceof Dropship) {
+            Dropship ds = (Dropship) entity;
```
NickAragua, line 4939: Why are we declaring this variable and not using it?

NickAragua (changes requested):
While it looks fine (other than the minor cleanup), I'm likely to introduce an unofficial option to keep large craft alive upon crashing. Because, dammit, if the thing's hull integrity is intact, we should be able to keep it going.

megamek/src/megamek/server/GameManager.java
Sleet01: Oops, good eye.  I think I was looking at replacing some casts, but it may no longer be necessary.

megamek/src/megamek/server/GameManager.java
Sleet01: What in the what?

Sleet01:
> While it looks fine (other than the minor cleanup), I'm likely to introduce an unofficial option to keep large craft alive upon crashing. Because, dammit, if the thing's hull integrity is intact, we should be able to keep it going.

I can add an option to turn this rule off, but as a reminder, it's only for Large Craft crashing _while out of control_.  If they crash due to loss of thrust or some other reason, they'll be treated as before.

NickAragua (approved):

PR #5855, opened 2024-08-05
Check if Edge is enabled in Game Options before using it

### LINKED ISSUE #5587: [Nightly] Edge being used despite being disabled in both MHQ and MM
**Environment**

2024-06-17T01:16:51.850219101
Java 17.0.11
Linux

**Description**
Edge is being triggered during battles despite being disabled in the settings for both MM and MHQ. I initially thought this was more on the MHQ side of things, but after further investigation, it looks like a lot of this bug is with Megamech. If you start up MM independent of MHQ, turn edge on, give pilots edge, then turn it off again, they will use the edge you gave them, even though it is off.

Algebro7:
When determining whether to roll edge on crits, the game only checked whether crews had Edge points remaining and not whether Edge was actually enabled. This led to weird situations where pilots who had edge points assigned would still roll edge even though it was disabled (as described in https://github.com/MegaMek/megamek/issues/5587).

This PR adds an additional check to see if Edge is enabled before checking whether the crew has edge points. I wish there was a cleaner way to do this than at each call site but it would've required more substantial refactoring which I was afraid to do. I looked at implementing it in the `Crew` object but it doesn't extend `Entity` and thus does not have access to `game`.

I tested in a game with all Mechs and they stopped rolling Edge when it was disabled, and rolled Edge when it was enabled, as expected.

Closes #5587

### DIFF
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -15903,4 +15903,11 @@ public abstract class Entity extends TurnOrdered implements Transporter, Targeta
     public boolean countForStrengthSum() {
         return !isDestroyed() && !isTrapped() && !isPartOfFighterSquadron();
     }
+
+    /** @return True if the unit should use Edge based on the current options and assigned Edge points */
+    public boolean shouldUseEdge(String option) {
+          return (game.getOptions().booleanOption(OptionsConstants.EDGE)
+              && getCrew().hasEdgeRemaining()
+              && getCrew().getOptions().booleanOption(option));
+    }
 }
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -15907,6 +15907,7 @@ public abstract class Entity extends TurnOrdered implements Transporter, Targeta
     /** @return True if the unit should use Edge based on the current options and assigned Edge points */
     public boolean shouldUseEdge(String option) {
           return (game.getOptions().booleanOption(OptionsConstants.EDGE)
+              && getCrew() != null
               && getCrew().hasEdgeRemaining()
               && getCrew().getOptions().booleanOption(option));
     }
--- a/megamek/src/megamek/common/Mech.java
+++ b/megamek/src/megamek/common/Mech.java
@@ -1971,7 +1971,8 @@ public abstract class Mech extends Entity {
                 // normal front hits
                 switch (roll) {
                     case 2:
-                        if ((getCrew().hasEdgeRemaining()
+                        if (game.getOptions().booleanOption(OptionsConstants.EDGE)
+                            && (getCrew().hasEdgeRemaining()
                                 && getCrew().getOptions().booleanOption(OptionsConstants.EDGE_WHEN_TAC))
                                 && !game.getOptions().booleanOption(OptionsConstants.ADVCOMBAT_NO_TAC)) {
                             getCrew().decreaseEdge();
@@ -1997,7 +1998,8 @@ public abstract class Mech extends Entity {
                     case 11:
                         return new HitData(Mech.LOC_LARM);
                     case 12:
-                        if (getCrew().hasEdgeRemaining()
+                        if (game.getOptions().booleanOption(OptionsConstants.EDGE)
+                                && getCrew().hasEdgeRemaining()
                                 && getCrew().getOptions().booleanOption(
                                         OptionsConstants.EDGE_WHEN_HEADHIT)) {
                             getCrew().decreaseEdge();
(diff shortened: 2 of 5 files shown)


megamek/src/megamek/common/Mech.java
```diff
@@ -1971,7 +1971,8 @@ public HitData rollHitLocation(int table, int side, int aimedLocation, AimingMod
                 // normal front hits
                 switch (roll) {
                     case 2:
-                        if ((getCrew().hasEdgeRemaining()
+                        if (game.getOptions().booleanOption(OptionsConstants.EDGE)
+                            && (getCrew().hasEdgeRemaining()
```
Sleet01, line 1975: It looks like we just have a bunch of this conditional repeated throughout Mech.java (and its inheritors), just with different boolean options.
Would it be viable to add, e.g.

```
public boolean shouldUseEdge(String option) {
    return (
                      game.getOptions().booleanOption(OptionsConstants.EDGE)
                                && getCrew().hasEdgeRemaining()
                                && getCrew().getOptions().booleanOption(option)
                );
```
or something?  Then each check becomes something like this:

```
    case 2:
        if (shouldUseEdge(OptionsConstants.EDGE_WHEN_TAC)
        && !game.getOptions().booleanOption(OptionsConstants.ADVCOMBAT_NO_TAC))
```

Which is somewhat cleaner when spread across all these cases.

megamek/src/megamek/common/TripodMech.java
```diff
@@ -1282,7 +1290,8 @@ && getCrew().getOptions().booleanOption(
                         case 11:
                             return new HitData(Mech.LOC_LARM, true);
                         case 12:
-                            if (getCrew().hasEdgeRemaining()
+                            if (game.getOptions().booleanOption(OptionsConstants.EDGE)
+                                && getCrew().hasEdgeRemaining()
                                 && getCrew().getOptions().booleanOption(
                                     "edge_when_headhit")) {
```
Sleet01, line 1296: We should probably take this opportunity to replace all of these with ```EDGE_WHEN_HEADHIT```

Sleet01 (changes requested):
Functionally looks good, could use a little cleanup.

megamek/src/megamek/common/Mech.java
Algebro7: Would it make sense to implement that method in the `Entity` base class all the various Mech and Aero units extend? Looks like it has a crew member so I think it should have all the information it needs

megamek/src/megamek/common/TripodMech.java
Algebro7: Good call, done in 62ec6e8

megamek/src/megamek/common/Mech.java
Sleet01: I think that's acceptable; the downside is you would also have to check that `getCrew()` doesn't return `null` prior to calling `.hasEdgeRemaining()` or `.getOptions()`, because while an Entity subclass like Mech _must_ have crew, there's no guarantee that a bare Entity will.

megamek/src/megamek/common/Entity.java
```diff
@@ -15903,4 +15903,11 @@ public Base64Image getBase64Icon() {
     public boolean countForStrengthSum() {
         return !isDestroyed() && !isTrapped() && !isPartOfFighterSquadron();
     }
+
+    /** @return True if the unit should use Edge based on the current options and assigned Edge points */
+    public boolean shouldUseEdge(String option) {
+          return (game.getOptions().booleanOption(OptionsConstants.EDGE)
+              && getCrew().hasEdgeRemaining()
```
Sleet01, line 15910: For safety, will need `&& getCrew() != null` immediately above this line.

megamek/src/megamek/common/Mech.java
```diff
@@ -2049,7 +2045,8 @@ && getCrew().getOptions().booleanOption(
                     case 11:
                         return new HitData(Mech.LOC_RLEG);
                     case 12:
-                        if (getCrew().hasEdgeRemaining()
+                        if (shouldUseEdge(OptionsConstants.EDGE)
+                                && getCrew().hasEdgeRemaining()
```
Sleet01, line 2049: This looks incorrect; should replace all three lines with

```
if (shouldUseEdge(OptionsConstants.EDGE_WHEN_HEADHIT)) {
```

megamek/src/megamek/common/Entity.java
Algebro7: Good call, done in 16d2e9c

megamek/src/megamek/common/Mech.java
Algebro7: Yep you're right! Fixed in 0821323

megamek/src/megamek/server/GameManager.java
```diff
@@ -20451,7 +20451,8 @@ private Vector<Report> resolveCrewDamage(Entity e, int damage, int crewPos) {
                     // return true;
                 } // else
                 vDesc.add(r);
-            } while (e.shouldUseEdge(OptionsConstants.EDGE_WHEN_KO)
+            } while ((e.shouldUseEdge(OptionsConstants.EDGE_WHEN_KO)
+                    && e.getCrew().isKoThisRound(crewPos))
```
Algebro7, line 20455: This check was accidentally removed when refactoring, which led to entities rerolling edge until the pilot blacked out.

Sleet01: NGL, the effect was hilarious.  But good find!

Sleet01 (approved):

NickAragua (approved):

PR #5859, opened 2024-08-05
Use OptionsConstants instead of hard-coded strings

Algebro7:
I found a few more instances in the codebase where game options were being queried with hard-coded strings rather than using the `OptionsConstants` enum. This was using a find all for `game.getOptions()`--I checked all of them and I believe these are the last.


Sleet01 (approved):
Nice!  LGTM

PR #5860, opened 2024-08-05
Tweak automatic munition selection for low-ammo-count Autocannons

Sleet01:
QA testing revealed that some units with single bins of AC10 and AC5 ammo would get Precision Autocannon munitions loaded, despite this significantly _decreasing_ their mission endurance.

The issue was caused by a couple things:
1. Precision and Armor-Piercing munitions were not listed in the category "AMMO_REDUCING_MUNITIONS", so reducing the weight of the category would not reduce the weights of these munitions.
2. Standard munitions might not be selected as the primary ammo when two or more munitions tie for highest weight, so even if 1 is fixed, Standard munitions might not be selected.

Testing fixes for 1 and 2 revealed an opportunity to extend the Caseless ammo selection code for AC/20s to specific AC/10 and AC/5 situations as well.

Testing:
- Tested on MHQ campaign submitted by reporter of original bug (Thanks @Thom293 !)

### DIFF
--- a/megamek/src/megamek/client/generator/TeamLoadoutGenerator.java
+++ b/megamek/src/megamek/client/generator/TeamLoadoutGenerator.java
@@ -113,7 +113,7 @@ public class TeamLoadoutGenerator {
             "Acid", "Laser Inhibiting", "Follow The Leader", "Heat-Seeking", "Tandem-Charge",
             "Thunder-Active", "Thunder-Augmented", "Thunder-Vibrabomb", "Thunder-Inferno",
             "AAAMissile Ammo", "ASMissile Ammo", "ASWEMissile Ammo", "ArrowIVMissile Ammo",
-            "AlamoMissile Ammo"
+            "AlamoMissile Ammo", "Precision", "Armor-Piercing"
     ));
 
     public static final ArrayList<String> TYPE_LIST = new ArrayList<String>(List.of(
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/client/generator/TeamLoadoutGenerator.java
```diff
@@ -652,15 +652,62 @@ public static ReconfigurationParameters generateParameters(
...
+                .filter(w -> w.getName().toLowerCase().contains("ac") && w.getName().contains(size)).count();
+    }
+
+    private static int getACAmmoCount(Entity e, String size) {
+        return (int) e.getAmmo().stream()
+                .filter(w -> w.getName().toLowerCase().contains("ac") && w.getName().contains(size)).count();
+    }
+
+    // Set low-ammo-count AC5 carriers to use Caseless if ammo is <= 1/2 ton per tube
+    private static boolean setAC5Imperatives(Entity e, MunitionTree mt, ReconfigurationParameters rp) {
```
NickAragua, line 671: This function is almost identical to the AC/10 (and probably the AC/2 and AC/20 one). Let's re-work it into one function with an extra "AC Rating" parameter.

NickAragua (changes requested):
Let's reduce copy-pasta a little bit.

megamek/src/megamek/client/generator/TeamLoadoutGenerator.java
Sleet01: Normally I would agree, but I intentionally differentiated these, as an example of how I would like additional imperatives to be added.
I want it to be very simple to add a new imperative for a specific weapon type and size, even for novice programmers, and I want the flow in this section to be very clear.
And making a single function that takes the AC rating and branches on that would likely produce as much code as we have now, without being any clearer.

So I must respectfully decline.

NickAragua: Not sure how thes approach below produces more code than copy+pasting setACImperatives 4 times. Following function calls is a basic programming skill that it's better that novices learn. I would also argue that exposing novice programmers to "copy + paste identical method with slight modification" programming technique rather than "pull common logic out into separate function" is not what we want to be teaching novice coders.

Finally, if you decide to change the "setACImperatives" logic across the board, you have to remember to change it in 3-4 different methods, rather than changing it in just the one. 

setAC2Imperatives(entity, munitiontree, reconfigurationparameters) {
 setACImperatives(entity, munitiontree, reconfigurationparameters, 2)
}

setAC5Imperatives(entity, munitiontree, reconfigurationparameters) {
 setACImperatives(entity, munitiontree, reconfigurationparameters, 5)
}

setAC20Imperatives(entity, munitiontree, reconfigurationparameters) {
 setACImperatives(entity, munitiontree, reconfigurationparameters, 20)
}

setAC10Imperatives(entity, munitiontree, reconfigurationparameters) {
 setACImperatives(entity, munitiontree, reconfigurationparameters, 10)
}

setACImperatives(entity, munitiontree, reconfigurationparameters, barrelsize) {
getACWeaponCount(barrelsize);
aerocheck() or no acweaponcount;
if (acammocount <= acbarrelcount) use caseless
use any
}

Anyway, I'll let someone else tie-break this one.

Sleet01:
Roping in @IllianiCBT as they are some fresher eyes on the project, and have not worked on this code before.

Do you mind taking a look at my proposed code, and @NickAragua 's suggested fix?

megamek/src/megamek/client/generator/TeamLoadoutGenerator.java
```diff
@@ -652,15 +652,62 @@ public static ReconfigurationParameters generateParameters(
...
+            return true;
+        }
+        return false;
+    }
+
+    // Set low-ammo-count AC10 carriers to use Caseless if ammo is <= 1 ton per tube
+    private static boolean setAC10Imperatives(Entity e, MunitionTree mt, ReconfigurationParameters rp) {
+        int ac10Count = getACWeaponCount(e, "10");
+
+        // TODO: remove this block when implementing new anti-ground Aero errata
```
IllianiCBT, line 693: I know @Algebro7 is implementing some of the new Aero errata, but I don't know if they're doing the anti-ground errata. Tagging them here so that they're aware of this TODO (in case there is any cross-over).

megamek/src/megamek/client/generator/TeamLoadoutGenerator.java
```diff
@@ -652,15 +652,62 @@ public static ReconfigurationParameters generateParameters(
...
     // region Imperative mutators
     private static void setACImperatives(Entity e, MunitionTree mt, ReconfigurationParameters rp) {
+        setAC5Imperatives(e, mt, rp);
+        setAC10Imperatives(e, mt, rp);
         setAC20Imperatives(e, mt, rp);
     }
 
+    private static int getACWeaponCount(Entity e, String size) {
+        return (int) e.getWeaponList().stream()
+                .filter(w -> w.getName().toLowerCase().contains("ac") && w.getName().contains(size)).count();
```
IllianiCBT, line 662: Can we have a new-line at `.contains` and `.count`, it might not be strictly necessary from a Style Guide perspective, but it's my experience that it makes the stream much easier to parse at a glance

megamek/src/megamek/client/generator/TeamLoadoutGenerator.java
```diff
@@ -652,15 +652,62 @@ public static ReconfigurationParameters generateParameters(
...
+    }
+
+    private static int getACAmmoCount(Entity e, String size) {
+        return (int) e.getAmmo().stream()
+                .filter(w -> w.getName().toLowerCase().contains("ac") && w.getName().contains(size)).count();
```
IllianiCBT, line 667: Here too

megamek/src/megamek/client/generator/TeamLoadoutGenerator.java
Sleet01: Side note: anyone working on this will also need to touch the other MM, MML, and MHQ ammo validation code - it's not a small lift.

megamek/src/megamek/client/generator/TeamLoadoutGenerator.java
IllianiCBT: Speaking as one of the aforementioned novice programmers, I am tilted towards a preference for single methods over multiple similar methods. The biggest reason is from a maintainer perspective: it's a lot easier to adjust behavior reliably when you only have the one place to look. Having multiple methods increases the likelihood of one getting missed during an update pass.

Now, I will say that both of you are far more experienced at this than I, so I really don't feel like I'm qualified to make a decision, but you brought me into this PR so I figured I'd offer my thoughts.

IllianiCBT:
> Do you mind taking a look at my proposed code, and @NickAragua 's suggested fix?

So I took a look, but I really don't feel qualified to make a decision given the substantially more experience you both have over me.

Sleet01:
> > Do you mind taking a look at my proposed code, and @NickAragua 's suggested fix?
> 
> So I took a look, but I really don't feel qualified to make a decision given the substantially more experience you both have over me.

Alright, thanks for giving it a look-over.

PR #5862, opened 2024-08-05
Overloaded wordWrap Method with Default Max Length

IllianiCBT:
Introduced a new `wordWrap` method that defaults to a max length of 100 characters. Also updated parameter naming for clarity in the word wrapping logic.


Sleet01 (approved):
LGTM

PR #5863, opened 2024-08-05
Refactored Gender Enum to Include 'Other' Genders

IllianiCBT:
Cleaned up the class and added handlers for personnel with one of two 'Other' genders.

This PR does nothing with this functionality, it is added with a mind towards being used by another mhq-origin PR.


Sleet01 (approved):
LGTM

PR #5864, opened 2024-08-06
Tweak princess gravity PSR handling

Sleet01:
This increases Princess's caution when moving on maps with high or low gravity:
- On low-gravity ground maps, ground units will recognize the PSR danger of exceeding their normal Run / Flank / Jump MP
- On high-gravity ground maps, ground units will recognize the PSR danger of using jump MPs, but Run / Flank will not be penalized.

Also reverted an inadvertent change to the Low/High-G nag checks that I did during the bog-down work.
Additionally, reverts MoveStep.java to space-based indentation.

Testing:
- Ran 12 v 12 bot matches with VTOL, hover, ground tank, and fast Light 'mechs (with and without jumping) to compare low- and high-gravity performance across various G levels.

### DIFF
--- a/megamek/src/megamek/common/MovePath.java
+++ b/megamek/src/megamek/common/MovePath.java
@@ -114,6 +114,8 @@ public class MovePath implements Cloneable, Serializable {
 
     // is this move path being done using careful movement?
     private boolean careful = true;
+    private boolean gravityConcern = false;
+    private final float gravity;
 
     /**
      * Generates a new, empty, movement path object.
@@ -121,6 +123,9 @@ public class MovePath implements Cloneable, Serializable {
     public MovePath(final Game game, final Entity entity) {
         this.setEntity(entity);
         this.setGame(game);
+        // Do we care about gravity when adding steps?
+        gravity = game.getPlanetaryConditions().getGravity();
+        gravityConcern = (gravity != 1.0F && game.getBoard().onGround() && !entity.isAirborne());
     }
 
     public Entity getEntity() {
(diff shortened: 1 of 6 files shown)


megamek/src/megamek/common/MovePath.java
```diff
@@ -114,13 +114,18 @@ public int hashCode() {
...
 
     /**
      * Generates a new, empty, movement path object.
      */
     public MovePath(final Game game, final Entity entity) {
         this.setEntity(entity);
         this.setGame(game);
+        // Do we care about gravity when adding steps?
+        gravity = game.getPlanetaryConditions().getGravity();
+        gravityConcern = (gravity != 1.0F && game.getBoard().onGround() && !entity.isAirborne());
```
NickAragua, line 128: You could make this a little more as well - the only time gravity becomes a concern for low gravity is when the entity's MP actually changes as a result of gravity.

The reason I say this is because gravity and entity move MP don't change, but addStep() is called very frequently, so if we can move at least some of the MP checks out of the addStep method, that will help with performance.

NickAragua (changes requested):

megamek/src/megamek/common/MovePath.java
Sleet01: Don't MPs change as soon as gravity is over 1.0?  Let me double-check...

Sleet01: Actually, I guess we can just ask the entity if its MP are different from no-gravity MP...

Sleet01: Done.

PR #5879, opened 2024-08-09
Fix 5614 princess does no pathing during deployment

### LINKED ISSUE #5614: 49.20 Nightly - Princess Appears to do no pre-deployment pathfinding check.
This is a long-standing issue and not unique to the nightly.  I dont recall a bug report on it and didnt see one on a quick search.

This Locust deployed on this building on turn 1.  It was competely unnecessary as there were plenty of empty roads in our deployment zone.  He has been there the entire game shooting the building, trying to break free.  

She exhibits similar behavior by deploying on islands, or in a landlocked hex in a group of tall buildings, or on swamp/water, etc.  So, seems she may check the deployment zone and see any hex in it as valid, without checking the actual hex itself?

image

image

Sleet01:
Adds hazard assessment for initial deployment hexes, and path opportunities + hazard assessments in a radius around the highest-ranking (lowest hazard) hexes.
Falls back to original deployment hex selection algorithm if the new search algorithm fails.

Performance appears to be the same as, or faster than, the original naive deployment algorithm while avoiding various pitfalls:
- Stranding non-jump units on buildings
- Stranding non-jump units in pits or inner courtyards that they cannot escape
- Placing units in hazardous hexes
The new algorithm uses sampling (of all valid deployment hexes) and a small kernel radius for determining surrounding path options and hazards to rank the best of a subset of candidate hexes against each other.
Because of this, there will be some (very few) instances when deployment may slow down, and units may be placed sub-optimally, due to the sampled subset containing no valid hexes and the fall-back algorithm being subsequently used.
I believe the improvements are worth this possible slight inconvenience, however:

image

Notice that no units were deployed in cul-de-sacs or trapped on top of buildings, nor in water or on hazardous terrain hexes.

Note: this patch does not provide an option to disable the new search algorithm, nor controls to adjust the kernel size.

Testing:
- Ran a number of tests with various unit types in cities, hazardous environments / maps, jump and non-jump-equipped units.

Close #5614

### DIFF
--- a/megamek/src/megamek/client/bot/princess/Princess.java
+++ b/megamek/src/megamek/client/bot/princess/Princess.java
@@ -13,6 +13,7 @@
  */
 package megamek.client.bot.princess;
 
+import megamek.MegaMek;
 import megamek.client.bot.BotClient;
 import megamek.client.bot.ChatProcessor;
 import megamek.client.bot.PhysicalCalculator;
@@ -44,9 +45,8 @@ import megamek.common.util.StringUtil;
 import megamek.common.weapons.AmmoWeapon;
 import megamek.common.weapons.StopSwarmAttack;
 import megamek.common.weapons.Weapon;
+import megamek.logging.MMLogger;
 import org.apache.logging.log4j.Level;
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
 
 import java.io.File;
 import java.text.DecimalFormat;
@@ -56,6 +56,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 
 public class Princess extends BotClient {
+    private static final MMLogger logger = MMLogger.create(Princess.class);
     private static final char PLUS = '+';
     private static final char MINUS = '-';
 
@@ -171,7 +172,7 @@ public class Princess extends BotClient {
     public static Princess createPrincess(String name, String host, int port, BehaviorSettings behavior) {
         Princess result = new Princess(name, host, port);
         result.setBehaviorSettings(behavior);
-        LogManager.getLogger().debug(result.getBehaviorSettings().toLog());
+        logger.debug(result.getBehaviorSettings().toLog());
         return result;
     }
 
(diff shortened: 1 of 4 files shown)


Sleet01:
As a note: the latest commits switched Princess to the MMlogger!  This appears to be working fine, but in case it breaks anything: I did it here!

megamek/src/megamek/client/bot/princess/Princess.java
```diff
@@ -629,10 +631,152 @@ private int calculateTurretDeploymentValue(final Coords coords) {
...
+        Collections.shuffle(localCopy);
+
+        // Hacky, but really, "DEPLOY" should be a path step...
+        MovePath mp = new MovePath(game, deployedUnit);
+        mp.addStep(MoveStepType.NONE);
+        MoveStep deployStep = mp.getLastStep();
+        IPathRanker ranker = getPathRanker(deployedUnit);
+        HashMap<Double, ArrayList<Coords>> rankedCoords = new HashMap<>();
+
+        if (!deployedUnit.isAero()) {
```
NickAragua, line 707: Probably add VTOL and WIGE to this as well.

NickAragua (changes requested):
One minor optimization. 

Also, I could have sworn there was an "am I deploying to an inaccessible zone" check already. Check out the call to calculateEdgeAccessFitness function in BotClient.getStartingCoordsArray().

megamek/src/megamek/client/bot/princess/Princess.java
Sleet01: Do we deploy WiGEs in flight?  If not, they probably still want all the hazard checks...

Sleet01: It's optional to deploy WiGEs in flight, so WiGEs that _don't_ will want the extra checks.

Sleet01: We do weight cul-de-sac hexes lower, and completely inaccessible hexes lower still, but it's fairly unsubtle.

After reviewing the botClient code, though, I am thinking we can remove the shuffle before sampling - the list of valid coordinates should be in a semblance of fitness order, so just iterating through the first 20 (+/- some) should be fine.

I'll test this.

Sleet01: Looks like _not_ re-shuffling leads to slightly less clumping, so let's go with that.

PR #5880, opened 2024-08-09
prevent various NPEs and other exceptions

NickAragua:
Prevents an infinite popup loop and an actual exception that occurs when selecting a different unit during the deploy phase after placing a unit on the board but not finalizing deployment.


Sleet01 (approved):
LGTM

PR #5886, opened 2024-08-11
#5885: Fix AC option checking

### LINKED ISSUE #5885: 50.00 - Increased AC damage rule seems to be on by Default, even though it is turned off.
Simply, option shows it is turned off, but AC2 and AC5 show they are doing 1 more damage each.  Seems like it is somehow turned on in the background?

Just thinking out loud - but I do have rapid fire AC and kinder rapid fire AC rule on.  Maybe this other alternate rule became entangled with them somehow?  

Also begs the question - what other options might be accidentally on without us knowing?  

image

image

image

SJuliez:
fixes #5885


Sleet01 (approved):
Looks good to me!

PR #5887, opened 2024-08-11
Torso twist fix

### LINKED ISSUE #5827: [0.49.20] Errata reversion: Implement/reinstate turret 'twist' in the indirect artillery phase
Bundled Megamek with MekHQ 0.49.20

Vehicle turrets mounting artillery should be able to turn those turrets in the indirect targeting phase in order to change artillery firing arc immediately prior to firing. This is how it was implemented in earlier versions before other, now reverted, errata.

https://bg.battletech.com/forums/index.php?topic=84849.msg2023681#msg2023681

Looks like earlier versions (eg 47.9) allowed two turret twists per round, rather than just one as per the current clarification.

NickAragua:
Fix #5827  - enable torso/turret twist for artillery, but you can only do torso twist once per round.

Note that "canChangeSecondaryFacing" already includes a check for "already twisted".


Sleet01 (approved):
LGTM!

rjhancock (approved):
From a. Sanity check, looks good

PR #5889, opened 2024-08-12
Tools for enhancing games, upgraded scenarios

SJuliez:
This upgrades MM scenarios along with some systems to present something more than just a few units on a map. The direct impact on MHQ games and lobby games should be zero. However, MHQ can make use of the new abilities.

Some of the new stuff is similar to what StoryArcs does and I have tried to keep it similar enough in code that things could be fused together in the future, although I'm not sure if that would be useful right now. However, the system of linked storytelling is not used in the same way.

This PR
- adds a Trigger system. Triggers are classes that react by returning true when their conditions in the game's state are matched. This is somewhat similar to current victory conditions, which are however much more limited by their code. There are various triggers available right now, subject to expansion where useful, and triggers can be combined in many ways to form more complex triggers
- adds scripted events. These are gathered in a list in AbstractGame. Each has a Trigger that the GameManager can check at various times. Unlike MHQ, we cannot check them at every Campaign change event, but they are checked at the beginning and end of all phases, and could potentially be checked even more densely if there is a need for it. Events can be added in code by `game.addScriptedEvent(TriggeredEvent event)` which, I think, can be used in MHQ's GameThread classes during game setup in the same way ScenarioV2 does it.
- adds in-game messages (which are scripted events). These are displayed very much like StoryArc messages and use almost the same dialog code. See below for an example. Messages may use markdown formatting (like in StoryArcs) and can show an image like in the example (This adds the commonmark dependency to MM). Unlike MHQ, the message event must supply the details, the dialog just displays them. Messages use a Trigger to determine when they are shown.
- adds game end events (which are also scripted events and also based on Triggers). These, unsurprisingly, end the game, moving it to the victory phase regardless of what victory, if any, happens. This is a more variable system than the fixed round limit we have in code now; the game can be made to end, e.g., when a number of units have fled or with any other trigger(s). The use for this (as opposed to pure victory conditions) that sometimes it is easier to write up separate game end triggers and victory triggers in scenarios. 
- adds victory events (which are also scripted events and also based on Triggers). These are independent from the current VictoryResult system and much more flexible, albeit not available from the lobby UI. Victory events can be set to be either checked at any round end and end the game, or, be only checked once the game has ended via a game end event.
- adds initial board post-processing abilities. With these, scenarios can apply changes to the game board once it has been assembled from board files or created in some other way. This can be expanded easily. For the two scenarios I added, I added post processors for changing the theme, removing terrain and changing terrain type/level. I have not yet added support for MHQ/lobby games but I should do this.
- adds two MM scenarios in the Kell Hounds tab (Lowering the Boom and To Save A Prince), based on the old FASA Kell Hounds book that make use of scripted events to show messages.
- adds scenario parsing for all the above, as well as remaining armor/internal parsing, unit force and pilot nickname

Some Issues/enhancements for the future:
- add code suport for board post-processing so it can be used from any game setup (lobby, MHQ)
- the client/server packet system may end up showing multiple modal dialogs at the same time (save to MUL dialog and victory message). Not sure how to deal with this
- messages are currently not targeted to a player
- the victory systems should work together; victory events should be able to process score points

image

image


Thom293:
Holy crap yes finally!  Amazing.

IllianiCBT:
Now this is very cool!

sshagent:
Wow, this is excellent stuff.  My group are 5 years in of minimum weekly games against Princess.  This is going to add so much life to our games.  Thank you very much.  Woohoo!

BLR-IIC:
Question:  Can these enhancements be used to automatically administer grinder-type games?

SJuliez:
> Question: Can these enhancements be used to automatically administer grinder-type games?

Please explain what "administer grinder-type" means.

BLR-IIC:
> > Question: Can these enhancements be used to automatically administer grinder-type games?
> 
> Please explain what "administer grinder-type" means.

As I'm familiar with it, a grinder game is a game where multiple players play against each other, either in a team or free-for-all format, with a unit that's taken from a list, either by choice or by random selection.  When that unit is destroyed, it's automatically replaced by another unit, typically one that's a weight class higher than the previous unit.  These are used at conventions to introduce players to the game, and I think some of them can run perpetually.  Other people here might have more experience with it than I do.

I think that would be a neat addition, but reviewing the PR's description again my question seems beyond the scope of this PR.

SJuliez:
What you describe is a respawn mechanic. While I havent added that with this PR, it is doable using the trigger system. It requires an additional event that replaces a killed unit with a new one. Replacing with the same unit would be relatively easy. Replacing with another unit would require some way encode that (allowing the player to choose would be comparatively difficult), e.g. a table of units. But it's possible and I had thought about a respawn system recently as carryable objects enable a capture the flag style game that would require respawns.

HammerGS:
Other than conflicts, I'd tested this in development and it worked great. 

Once the conflicts are resolved. LGTM!

HammerGS (approved):
Tested and works for me.

Issue #5913, opened 2024-08-17
Canon filter shows 0 units


repligator:
### Environment
Linux
Java 17
Recent versions of both MM (https://github.com/MegaMek/megamek/commit/5d74b455d1bfe3c55797e4e3daac3bd0168ca828) and MHQ (https://github.com/MegaMek/mekhq/commit/b3e0e4457a766489e5545398116ef0d341862003)

### Description
When any of the filters that involve canon units are checked (for example Allow Canon Units Only in MM, or allow Canon Refits Only in MHQ) no units at all will appear.

repligator:
Turns out MM/MHQ don't automatically spit out canon unit lists when run from IDEA, it needs to be done via the gradle tab. Once the task is run, everything is fine. Closing.

PR #5919, opened 2024-08-18
A simple test that checks canonicity is working as intended

repligator:
Checks if a known canon mech is understood by MM to be canon. Fails if it isn't. Note that the test should actually be failing at the moment, see https://github.com/MegaMek/megamek/issues/5913 .


mmetke-qumulo (changes requested):
Please make a copy of the mtf file and place it somewhere in the `/testresources/` directory.

The test is not whether the current version of the .mtf file is canon; the test is whether the current _code_ understands a _known-good_ file as canon.

All unit tests that rely on a specific file should have a test version of that file; no unit test should rely on the live data.
If you have the test rely on the live data, erroneous failures may arise.

Sleet01 (changes requested):
Please make a copy of the mtf file and place it somewhere in the `/testresources/` directory.

The test is not whether the current version of the .mtf file is canon; the test is whether the current _code_ understands a _known-good_ file as canon.

All unit tests that rely on a specific file should have a test version of that file; no unit test should rely on the live data.
If you have the test rely on the live data, erroneous failures may arise.

Sleet01 (approved):
LGTM!

PR #5925, opened 2024-08-19
Fix make mech file parser canon check unit testable

Sleet01:
Fixes the `testCanon()` test that is currently failing when run from the IDE.

As-is, this code has pretty strong code smells:
1. mostly unchanged for about 20 years,
2. does critical state initialization in the middle of a static method called by a bunch of instance methods,
3. Meaningless nested try/catch clauses that do nothing.
4. Actively breaks unit testing by requiring a specific environment state to work

This fix adds some dependency injection capabilities (the ability to change the internal data for testing purposes):
1. we can specify arbitrary canon unit name list files and containing dir, or
2. we can construct the canonUnitNames Vector explicitly and set it directly.

This will allow unit tests that want to check a unit file's canonicity during parsing to do so, even within the IDE.


repligator (approved):
LGTM

PR #5926, opened 2024-08-19
Fix #5669: Stop NPE when UI checks if Airborne artillery has Homing ammo 

### LINKED ISSUE #5669: NPE - isHomingAmmoInHomingMode()" because "ammo" is null
### Environment
Developmental
Java 17
Master (latest PR https://github.com/MegaMek/megamek/pull/5784)

### Description
Load the attached save, (join as Lyrans). Assign a bot to Princess. Start the game. Deploy your 3 dropships. Begin the first turn, and get hit by 10+ NPE's per second.

Edit:
It doesn't trigger the NPE if the Dropships start on the ground...

### Files
[Bolan_2.sav.gz](https://github.com/user-attachments/files/16074887/Bolan_2.sav.gz)
[megamek.log](https://github.com/user-attachments/files/16074888/megamek.log)

Sleet01:
(Hint: it does not)

We may need to update these checks when implementing that Aerospace ammo errata, but currently no Aerospace unit can fire Homing Artillery rounds from onboard launchers / tubes.

Testing:
- Recreated OP's save and confirmed NPE does not occur.
- Ran all 3 projects' unit tests (currently there is a failing canonicity test when run from Idea)

Close #5669

### DIFF
--- a/megamek/src/megamek/client/ui/swing/OffBoardTargetOverlay.java
+++ b/megamek/src/megamek/client/ui/swing/OffBoardTargetOverlay.java
@@ -21,6 +21,7 @@ import megamek.common.actions.ArtilleryAttackAction;
 import megamek.common.actions.WeaponAttackAction;
 import megamek.common.util.ImageUtil;
 import megamek.common.util.fileUtils.MegaMekFile;
+import megamek.common.weapons.bayweapons.BayWeapon;
 
 import javax.swing.*;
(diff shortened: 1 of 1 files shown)


Sleet01:
Note: this check would only cause an NPE on Bay-mounted artillery; AIV launchers onboard standard ASF would not show the error.

SJuliez (approved):

PR #5936, opened 2024-08-20
Fix #5903 with a null check before dereferencing Coords cp

### LINKED ISSUE #5903: [0.50.0-SNAPSHOT] NPE during deployment phase if a unit is placed on the board and another unit is selected before confirming the first deployment
### Environment
megamek.MegaMek.initializeLogging(MegaMek.java:146) - Starting MegaMek v0.50.0-SNAPSHOT
    Build Date: 2024-08-14T01:19:21.935476675
    Today: 2024-08-14
    Origin Project: MegaMek
    Java Vendor: Eclipse Adoptium
    Java Version: 17.0.6
    System Locale: en_US
    MM Code Revision: 73b326a9d0279583593f95e4a786452171085e4e
    MML Code Revision: fc321e451b068ba6292e43ecc1f6ce80345d79cb
    MHQ Code Revision: 97bfe2272ed3dd97837249f987f8617facdf82ac

### Description
(Updated to include new information about issue being specific to the Isometric View client setting)

NPE occurs if I perform the following steps:
1. load autosave_2024-08-14_14-15-17.sav.gz
2. hit T to enable "isometric view"
3. select warhammer for deployment
4. Place warhammer into hex 2533 but DO NOT click "deploy" to confirm the selection
5. click on the unit list on the right and select the next unit (Stinger), NPE should occur at this step

This NPE pops up any time you interact with the UI of megamek in any way, including attempting to save or close the game, I have to use task manager to end the process to escape.

The NPE also pops up again if I alt tab to the mekHQ window and attempt to interact with the interface there as well.

Sentry is enabled and should have a record of this NPE (probably several hundred)

### Files
I included my mekHQ campaign that the scenario was launched from but should not be necessary to reproduce the issue, I was able to repro it just by loading the save directly in a fresh MegaMek client

[DeploymentNPE.zip](https://github.com/user-attachments/files/16617996/DeploymentNPE.zip)

Sleet01:
@phoenixheart512 found a nasty NPE storm while switching units when deploying with Isometric view active.
It looks like this is caused by trying to get the sprite information for a unit that has not yet been placed on the map when preparing to draw the new unit's sprite, leaving the sprite position as null.

This fix just safeties that redraw operation in such a case.

Testing:
- Ran with OP's save and repro steps

Close #5903

### DIFF
--- a/megamek/src/megamek/client/ui/swing/boardview/BoardView.java
+++ b/megamek/src/megamek/client/ui/swing/boardview/BoardView.java
@@ -1126,7 +1126,7 @@ public final class BoardView extends AbstractBoardView implements BoardListener,
             Coords cp = sprite.getPosition();
             // This can potentially be an expensive operation
             Rectangle spriteBounds = sprite.getBounds();
-            if (cp.equals(c) && view.intersects(spriteBounds) && !sprite.isHidden()) {
+            if (cp != null && cp.equals(c) && view.intersects(spriteBounds) && !sprite.isHidden()) {
                 if (!sprite.isReady()) {
                     sprite.prepare();
                 }
--- a/megamek/src/megamek/client/ui/swing/boardview/BoardView.java
+++ b/megamek/src/megamek/client/ui/swing/boardview/BoardView.java
@@ -1124,9 +1124,12 @@ public final class BoardView extends AbstractBoardView implements BoardListener,
 
         for (HexSprite sprite : spriteArrayList) {
             Coords cp = sprite.getPosition();
+            if (cp == null) {
+                continue;
+            }
             // This can potentially be an expensive operation
             Rectangle spriteBounds = sprite.getBounds();
-            if (cp != null && cp.equals(c) && view.intersects(spriteBounds) && !sprite.isHidden()) {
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/client/ui/swing/boardview/BoardView.java
```diff
@@ -1126,7 +1126,7 @@ private synchronized void drawHexSpritesForHex(Coords c, Graphics g,
             Coords cp = sprite.getPosition();
             // This can potentially be an expensive operation
             Rectangle spriteBounds = sprite.getBounds();
-            if (cp.equals(c) && view.intersects(spriteBounds) && !sprite.isHidden()) {
+            if (cp != null && cp.equals(c) && view.intersects(spriteBounds) && !sprite.isHidden()) {
```
rjhancock, line 1129: I'd suggest moving the null check higher up and doing a continue before `sprite.getBounds()`


```java
if (cp == null) { 
  continue;
}

...
```

Sleet01: That works; implemented.

rjhancock (approved):

Issue #1531, opened 2019-11-03
Armored Cowls not working as intended.


HammerGS:
Update. From a discussion with @HammerGS, changes are necessary in:

- [x] it should only not apply when the attack comes from the precise *hex row* that the mek is looking at
- [ ] it should have the properties of the present head armor (hardened etc.) 
- [x] the cowl equipment in MiscType can be removed

From this errata forum post:

https://bg.battletech.com/forums/index.php?topic=67422.0

Basically Megamek is giving a constant 5 point reduction to damage, when it should function as an extra 5 point (or whatever the cowl is) on the location.

### Environment
46.0 on.

binaryspica:
Is Weak Head Armor also have the same problem? I didn't tested it but since both of them alts the basic armor on the unit so perhaps both of them have the same problem.

NickAragua:
I'm not sure that this is correct - I ran a test, and a cowl essentially adds 3 points to head armor, as long as the attack does not come from the front. Any hits after the first one result in 0 points being absorbed by the cowl. 

As for weak head armor, it's simply not implemented.

SJuliez:
Update: The cowl seems to work correctly in that it is a one-time 3 point damage reduction, only applied when attacks come from other than the front. From a discussion with @HammerGS, changes are necessary in:
- it should only not apply when the attack comes from the precise *hex row* that the mek is looking at
- it should have the properties of the present head armor (hardened etc.) 
- the code might set the cowl armor to 0 when the quirk is not present to simplify handling
- the cowl equipment in MiscType can be removed

Sleet01:
I've added some unit tests as part of the special armor and damage refactoring, and Cowl does seem to be working correctly.
Closing.

PR #5937, opened 2024-08-21
Cowl quirk updates

SJuliez:
This updates the cowl quirk to only be bypassed when the attack is coming from the hex row the mek is looking down. Also removes the nonexistent cowl equipment and updates some code around it.
Making it so the cowl armor behaves like the armor type on the head (hardened etc.) is a bigger issue and I'll leave that for another time.

This is not an important bugfix and can go in after 0.50 just as well.

See #1531


SJuliez:
No good. I realize the unit must be able to handle switchign quirks on and off.

megamek/src/megamek/common/Coords.java
```diff
@@ -456,6 +456,34 @@ public static Coords nextHex(Coords current, Coords destination) {
...
+     * Returns true when the given other Coords are exactly on the hex row (line) from this Coords in the given
+     * direction. For example, if the direction is 0 (north), returns true only for Coords that are above
+     * this Coords at the same x. Returns false when the other Coords are null, the other Coords are equal
+     * to this or the direction is outside of 0 to 5.
+     *
+     * @param direction The direction, 0 = N, 2 = SE ...
+     * @param other The Coords to test
+     * @return True when the other Coords are on the hex row from this Coords in the given direction
+     */
+    public boolean isOnHexRow(int direction, @Nullable Coords other) {
```
NickAragua, line 469: I'm amazed this wasn't part of the functionality previously.

NickAragua (approved):

Issue #4694, opened 2023-08-07
Silver Bullet Gauss Rifle Has Wrong Damage Profile on Aircraft


xiangyuhu1994:
Silver Bullet Gauss Rifle Has Wrong Damage Profile on Aircraft

### Environment

0.49.11, Windows 10, Java 11.0.14 

### Description
Silver Bullet Gauss Rifle does a single cluster of 5 damage in fighter vs fighter combat. Range does not matter for this. It should be doing 9. This does not appear to be an issue for any other LBX weapons.

Sleet01:
> Silver Bullet Gauss Rifle does a single cluster of 5 damage in fighter vs fighter combat. Range does not matter for this. > It should be doing 9. This does not appear to be an issue for any other LBX weapons.

A cursory examination seems to show that the Aero Attack Value is getting calculated correctly for this weapon, but I'm having a difficult time actually setting up a Silver Bullet ASF-on-ASF attack to repro the issue.

Do you have a craft file and/or save game file for this?  I'm not finding any ASF with Silver Bullet Gauss, Large Craft bays seem to calculate damage differently, and it would be good to test with the files you used (if possible).

Sleet01:
Okay, I see the issue now.  The prior dev set the IS Silver Bullet Gauss Rifle's Attack Value to the _book_ value, which is 9.  But the handler for the SB Gauss is the LBXHandler, which assumes the AV will be the ground unit damage, and multiplies it by 0.6 to get the _actual_ AV at runtime.

Fix is to update the AV values for `ISSilverBulletGauss.java` to 15.

SJuliez:
Fixed.

PR #5939, opened 2024-08-22
Edited Display Name for Transported Personnel Edge Use Option

IllianiCBT:
This reduces the length of the transported personnel edge use option, so that it doesn't get truncated when viewed through the Customize Person dialog in mhq.

This is the mm half of #4694


Sleet01 (approved):
LGTM.
Do we need to add a check for edge option displayableName length somewhere?

IllianiCBT:
That might not be a bad idea, but I'm not sure where the check should nor what shape it should be, as it only affects mhq and then only if the edit person menu is set at a width too narrow for the string to display

PR #5971, opened 2024-09-04
TWRuleHandler tweaks

SJuliez:
This plays a little with the TWRuleHandler thing, making it an interface instead of an abstract base class. This way rule handlers can be made records which cannot extend classes but can implement interfaces. This removes some boilerplate code from such rule handlers. 

Note that the losCache null guard that was in the MovementHandler constructor is now in the GameManager. There are two calls to the GameManager with the losCache; one of them already had the null check, I added it to the other.

Also added the copyright notice to the new classes and made some very simple code updates.


megamek/src/megamek/server/totalwarfare/MovementHandler.java
```diff
@@ -12,30 +30,20 @@
...
+ *
+ * @param gameManager The server's GameManager
+ * @param entity   The Entity that is moving
+ * @param md       The MovePath that defines how the Entity moves
+ * @param losCache A cache that stores Los between various Entities and
+ *                 targets.  In double blind games, we may need to compute a
+ *                 lot of LosEffects, so caching them can really speed
+ *                 things up.
+ */
+record MovementHandler(TWGameManager gameManager, Entity entity, MovePath md, Map<UnitTargetPair, LosEffects> losCache)
```
neoancient, line 44: Making this a record prevents the use of member fields, which are necessary for breaking the code down into smaller chunks.

neoancient (changes requested):
See #5974

SJuliez:
not worth keeping.

PR #5974, opened 2024-09-05
Movement handler refactor

neoancient:
Extracts the main loop of processMovement to its own method, splitting it into two methods of ~1k lines and ~2k lines. Local variables have been moved to member fields so that smaller methods can be extracted and work on the same data.

I also changed the name from MovementHandler to MovePathHandler, which better reflects its scope.


SJuliez (approved):
Difficult to review, but I had Princess fight one and the units seemed to move alright.

PR #5979, opened 2024-09-07
Recent boards

SJuliez:
Adds a list of recent board files to the Board main menu to (obviously) open them in the editor without file searching. The list managed in its own class and kept in mmconf/recent_boards.yml

image


HammerGS:
Tested and works for me.

HammerGS (approved):
Tested and works for me.

Issue #5997, opened 2024-09-14
Fall damage from high grav map using old ed rules.


BTSS88:
### Prerequisites and Pre-Issue Checklist

- [X] I'm reporting the issue to the correct repository:
- [MegaMek](https://github.com/MegaMek/megamek/issues)
- [MegaMekLab](https://github.com/MegaMek/megameklab/issues)
- [MekHQ](https://github.com/MegaMek/mekhq/issues)

- [X] I've tested the issue against at least the latest MILESTONE version
- [X] I've asked on the MegaMek Discord about the error
- [X] I've reviewed the BattleTech rules and MegaMek documentation, and I've confirmed that something isn't working as intended.
- [X] I've searched the Github tracker and haven't found the issue listed

### Severity *

Medium (Gameplay Limitation): Non-core functionality is impaired, providing a suboptimal but playable experience.

### Brief Description *

Fall damage from failing a jump PSR on a high grav world is calculated using old ed TacOp rules on p55. 
The griffin should take 0 internal damage based on the newer rule since high grav dmg only applies to every SECOND MP lost due to adjustment. In the old rule, you take 1dmg for EVERY MP lost which is what happened to the griffin. 

![Screenshot 2024-09-14 140049](https://github.com/user-attachments/assets/2cbf147b-1900-4ab1-a334-04a6f799cfe8)
![Screenshot 2024-09-14 142343](https://github.com/user-attachments/assets/9e2a80d7-fb02-46e3-bff0-590d02975e6f)
![Screenshot 2024-09-14 235950](https://github.com/user-attachments/assets/10f7b510-f153-4499-a5de-531141e8fffe)

### Steps to Reproduce

Just jump a 5/8/5 mech (griffin, wolverine) on a high grav world until you fail a PSR. 

### Operating System *

Windows 11

### Java Version *

17

### MegaMek Suite Version *

v0.50.0

### Final Checklist

- [X] I've checked to make sure that this issue has not already been filed
- [X] I'm reporting only one issue in this ticket for clarity and focus

DM0000:
This should be closed ref https://github.com/MegaMek/megamek/pull/6000

PR #5998, opened 2024-09-14
Fall damage from high grav map using old ed rules. #5997

DM0000:
In high gravity check for jump distance more than walkMP+1, reduced DMG taken to (WalkMP-gravWalkMP)/2.


HammerGS:
Looks like you've got conflicts against our current master.

PR #6009, opened 2024-09-15
Allow SRCS on SC

SJuliez:
Allows mounting of Smart Robotic Control Systems on SmallCraft.
MegaMek/megameklab#1559

@HammerGS IO p215 and p134 are contradictory but at least it seems clear that SC should be allowed to mount it and improve it. WS are unclear and the improved version is even more unclear (SV allowed?).


HammerGS:
> Allows mounting of Smart Robotic Control Systems on SmallCraft. [MegaMek/megameklab#1559](https://github.com/MegaMek/megameklab/issues/1559)
> 
> @HammerGS IO p215 and p134 are contradictory but at least it seems clear that SC should be allowed to mount it and improve it. WS are unclear and the improved version is even more unclear (SV allowed?).


Yeah, SC would be allows. WS are limited to SDS (CASPAR) DRONE CONTROL SYSTEM (SDS-DCS)

SV's would be allowed as well.

neoancient (approved):

PR #6018, opened 2024-09-18
Added JavaDoc building in Actions.

rjhancock:
Added CheckStyle and configured for Google Format Guide Added .editorconfig and set defaults based upon Style Guide and Google Style Guide


SJuliez:
May I strongly suggest to increase the line columns to 120 (if not more), the reasons being
- google uses 2 spaces indentation, we use 4, this adds 4 or more spaces to practically every line
- we want descriptive method names, class names and variable names. Those are longer and breaking lines down because of this is counter-productive.
- we live in 1440p or 4K times, at least for me there seems ample room for lines despite using big fonts and having some other windows/toolbars on the screen

rjhancock:
This is the input I'm wanting. I based this off of the Google version and know we have a few things different.

rjhancock:
Adjusted to 140 characters, added additional filters for XML, MTF, and BLK for extra space, trailing line, line endings, etc.

SJuliez (approved):
I'm not entirely sure what this will all mean but it seems ok.

rjhancock:
The changes on Java files will bring them more in line with the MM Style Guide.

The changes on the others is just for prettifying them so they are easier to work with (trailing white space, 4 spaces between tags in XML, line endings are all LF, etc).

The encoding SHOULD ensure all files are the same vs swapping between Windows and Unicode.

Configuration is the same for all 3 repos.

PR #6025, opened 2024-09-21
Weapon Handler Comment cleanup

SJuliez:
This is 99% comment cleanup in the weapon handler classes, removing those "non-javadoc"s, updating the copyrights, removing empty javadocs (I didnt bother filling in the constructor javadocs, as they mostly just call super). And maybe 3 minor and obvious code updates.


HammerGS (approved):
Not sure about the one security warning.

rjhancock:
If there is a variable in the parent class with the same name, they don't need to be redeclared and can cause issues

megamek/src/megamek/common/weapons/ASEWMissileWeaponHandler.java
```diff
@@ -88,24 +83,22 @@ protected void handleEntityDamage(Entity entityTarget,
...
-        if (entityTarget instanceof Dropship) { 
-            Dropship d = (Dropship) entityTarget;
+        //Large craft suffer a to-hit penalty for the location struck.
+        if (entityTarget instanceof Dropship d) {
             int loc = hit.getLocation();
             d.setASEWAffected(loc, 2);
             //Report the arc affected by the attack and the duration of the effects
             r = new Report(3472);
             r.subject = subjectId;
             r.add(entityTarget.getLocationAbbr(hit));
```
rjhancock, line 95: I know it's a short use one but.... could we not use one letter variable names here? If it's something like a stream, I don't see an issue as it's a one liner, maybe a loop counter variable. 

I think we still need to define out where the line is for this. Open for discussion.

megamek/src/megamek/common/weapons/ATMHandler.java
```diff
@@ -129,7 +111,7 @@ protected int calcHits(Vector<Report> vPhaseReport) {
     @Override
     protected int calcAttackValue() {
         int av = 0;
-        int counterAV = 0;
+        int counterAV;
```
rjhancock, line 114: Is this being initialized before use elsewhere in the code?

megamek/src/megamek/common/weapons/ASEWMissileWeaponHandler.java
SJuliez: ah yes. This was an IDEA warning and one-click change. I simply didnt look at it. I'll change it but I think we should not overdo things when it's legacy issues. Sometimes you really want to update code but then it can obscure what the PR does and other times you don't want to update every single legacy issue.

megamek/src/megamek/common/weapons/ATMHandler.java
SJuliez: I remove these on IDEA suggestion. Uninitialized ints are 0 anyway. I dont think this could possibly fail unless the compiler complains which it doesnt.

rjhancock: If the compiler doesn't complain, I'm good. Hard to keep track of which languages default the value and which don't.

megamek/src/megamek/common/weapons/ASEWMissileWeaponHandler.java
rjhancock: Fair, understandable, and I agree.

rjhancock (approved):

rjhancock (approved):

PR #6045, opened 2024-09-28
Hex areas and use for Deployment, Flee Zones and Triggers

SJuliez:
This PR continues my work to implement some (pretty simple) Kell Hounds scenarios from the old Kell Hounds book as MM scenarios with a bit of control. Changes:
- adds hex areas; they come in various basic forms and can be combined by adding, subtracting and intersecting to form more complex areas; hex areas are always evaluated against a board to determine if they contain a hex or to get all their coords; currently implemented areas include geometric ones like hex lines, half planes, circles, rays; others are board-dependent such as the board edges or hexes with a given terrain or hex level; hex areas are lightweight and good for transfer between server and client as they don't store hexes, only rules to create them;
- hex areas can be used to set deployment areas via code; it uses the same system as the terrain-type deployment zones
- players and individual units can now be given a hex area as a flee zone from which they are allowed to leave the battlefield; this can be anywhere, even in the middle of the board; the default behavior is unchanged from before (units dont have their own zone and players use the board edges); there is now a menu item to show the flee zone of a unit; note that Princess currently makes no use of this
- hex areas can be used to trigger events when one or more units are within a hex area
- MM scenario parsing for the above
- limits fleeing in Lowering the Boom and Saving A Prince scenarios in accordance with scenario rule


HammerGS (approved):
Tested and works, just waiting for Tex to confirm the Javadoc bit.

rjhancock:
/home/runner/work/megamek/megamek/megamek/megamek/src/megamek/common/hexarea/RectangleHexArea.java:36: error: malformed HTML

Replace the < with &lt; (html entity for less than sign) to fix the Javadoc issue

Issue #4937, opened 2023-12-02
[49.15] Indirect Fire Weapons Panel does not include Spotter bonus


ramrodofavalon:
### Environment
49.15 run MegaMek
macOS Sonoma 14.2
openjdk 11.0.18 2023-01-17

### Description
When using a Forward Observer and artillery, the Fire Weapons Panel does not update for Fire Adjust. However, the TN in Offboard Attack Phase is adjusted 

![image](https://github.com/MegaMek/megamek/assets/16293857/f143bd76-6829-4765-b4e7-8428bba94559)
![image](https://github.com/MegaMek/megamek/assets/16293857/d5848649-f9f4-46f6-8e52-1d91fe110b81)
![image](https://github.com/MegaMek/megamek/assets/16293857/2cee64a7-36a8-420d-9bc1-5c1e88b0382b)

### Files

megamek.log
Thjese are the only two entires in megemek.log, they repeat but I would assume not relevant
08:08:30,069 ERROR [megamek.client.ui.swing.boardview.FovHighlightingAndDarkening] {AWT-EventQueue-0}
megamek.client.ui.swing.boardview.FovHighlightingAndDarkening.getLosEffects(FovHighlightingAndDarkening.java:320) - Cannot process line of sight effects with a null source hex.

08:09:33,952 ERROR [megamek.client.ui.swing.minimap.Minimap] {AWT-EventQueue-0}
megamek.client.ui.swing.minimap.Minimap.getMinimapImage(Minimap.java:214) - 
java.lang.NullPointerException
	at megamek.client.ui.swing.minimap.Minimap.paintAttack(Minimap.java:869)
	at megamek.client.ui.swing.minimap.Minimap.drawMap(Minimap.java:555)
	at megamek.client.ui.swing.minimap.Minimap.getMinimapImage(Minimap.java:211)
	at megamek.client.ui.swing.minimap.Minimap$4.gamePhaseChange(Minimap.java:1427)
	at megamek.common.event.GamePhaseChangeEvent.fireEvent(GamePhaseChangeEvent.java:68)
	at megamek.common.Game.processGameEvent(Game.java:3011)
	at megamek.common.Game.setPhase(Game.java:717)
	at megamek.client.Client.changePhase(Client.java:417)
	at megamek.client.Client.handlePacket(Client.java:1522)
	at megamek.client.Client$1.lambda$packetReceived$0(Client.java:176)
[...]

PR #6046, opened 2024-09-29
Deleted `StratConSkillGenerator.java`

IllianiCBT:
Removed `StratConSkillGenerator.java` as it is no longer needed. This file contained methods for generating random skills. I found I could achieve the same goal better using `ModifiedConstantSkillGenerator.java`.

Introduced skill level adjustments based on dice rolls and added a new method in `SkillLevel` to parse integer values to skill levels.

This is the MM half of [#4937](https://github.com/MegaMek/mekhq/pull/4937)


HammerGS (approved):
Tested and I think it works.

Issue #5991, opened 2024-09-12
Mix tech mech not showing origin of weapon equipped. 


BTSS88:
### Prerequisites and Pre-Issue Checklist

- [X] I'm reporting the issue to the correct repository:
- [MegaMek](https://github.com/MegaMek/megamek/issues)
- [MegaMekLab](https://github.com/MegaMek/megameklab/issues)
- [MekHQ](https://github.com/MegaMek/mekhq/issues)

- [X] I've tested the issue against at least the latest MILESTONE version
- [X] I've asked on the MegaMek Discord about the error
- [X] I've reviewed the BattleTech rules and MegaMek documentation, and I've confirmed that something isn't working as intended.
- [X] I've searched the Github tracker and haven't found the issue listed

### Severity *

Low (Minor/Nuisance): Minor glitches or cosmetic issues that don’t affect gameplay and occur rarely.

### Brief Description *

Unit weapon screen page (F5) of a mix tech mech (Warhammer-C) not showing the origin of the weapon (IS/Clan). 
Weapon fire during round report DOES show the weapon is clan though. 

### Steps to Reproduce

Just select the mech in MM. I'm not sure if MekHQ would show things differently though. 

### Operating System *

Windows 11

### Java Version *

17

### MegaMek Suite Version *

v0.50.0

### Attach Files

![Screenshot 2024-09-12 121709](https://github.com/user-attachments/assets/82943852-06a0-4631-803a-af967039b69d)

### Final Checklist

- [X] I've checked to make sure that this issue has not already been filed
- [X] I'm reporting only one issue in this ticket for clarity and focus

DM0000:
Can change clan weapons to all have "Clan" in front of them.  Should I remove the clan from weapon screen too?

DM0000:
think this should be closed ref https://github.com/MegaMek/megamek/pull/6051.    Forgot to put closes in comments:(

DM0000:
think this should be closed ref https://github.com/MegaMek/megamek/pull/6051.    Forgot to put closes in comments:(

PR #6051, opened 2024-10-01
#5991  Mix tech mech not showing origin of weapon equipped.

DM0000:
mixedtech check and add appropriate tag
  
image

image

only displays if unit is mixedtech.

image


HammerGS (approved):
Tested and seems to work.

HammerGS:
Thank you for the contribution.

Issue #1486, opened 2019-09-13
Request: Add option to skip firing turn for mechs/vees that sprint


Tamren:
Since they can't fire any weapons while sprinting there is no point in controlling them during the weapon phase, so the game could skip them like it does for mechs with no melee targets in the physical attack phase.

inqistor:
But can they spot, while sprinting? What are the rules, when there is also gunner in cockpit of sprinting unit?

TenkawaBC:
From BMM:

A ’Mech that sprints may not make any deliberate attacks that
turn. Accidental charges as a result of skidding are still possible, and a
’Mech may still fire weapons to generate heat, though such firing has
no chance of doing damage. Additionally, the ’Mech may not spot
for indirect fire or take any action that would require it to sacrifice
its attack.

So it cannot spot, but for mechs with TSM, you can still shoot to keep your heat up, just not actually hit anything.  So auto-skipping isn't really an option, unless you also add a flag for TSM mechs to not skip, and there could be other scenarios I'm not thinking of.

gsparks3:
It could be made a game option, default off? The TSM mech example seems like the biggest reason why you wouldn't want to skip the weapons phase on a sprinting mech, and if you're using a TSM mech you'd know to keep it off, but others could turn it on to save time. Not 100% perfect, but still useful.

HoneySkull:
Is it possible in MegaMek to fire a weapon at no target in the case of TSM?  Or do you at least need to target a hex or building?

TenkawaBC:
> Is it possible in MegaMek to fire a weapon at no target in the case of TSM? Or do you at least need to target a hex or building?

Currently you must target something to fire weapons. Regardless of whether it is in range or not. So often for Tsm that means targeting a hex, which while sprinting you don't want to allow targeting, but maybe if it only allows hex targeting, or a phantom target hex (e.g.: auto target 99:99, almost guaranteed to be out of range)

DM0000:
Can anything besides Meks sprint?

gsparks3:
Tactical Operations: Advanced Rules page 23 has the rules for Overdrive, which is the vehicle equivalent of Sprinting (to the point of explicitly referring the mech sprinting rules).

RaozSpaz:
Couldn't this just be included within the 'skip invalid units during firing phase' rule option?

DM0000:
[...]

PR #6062, opened 2024-10-04
Issue 1486 Add option to skip firing turn for mechs/vees that sprint

DM0000:
added check for sprint and TSM in isEligibleForFiring

Only meks with TSM won't be skipped if option is checked

Closes https://github.com/MegaMek/megamek/issues/1486

### DIFF
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -84,6 +84,8 @@ import megamek.common.weapons.infantry.InfantryWeapon;
 import megamek.logging.MMLogger;
 import megamek.utilities.xml.MMXMLUtility;
 
+import static megamek.common.EquipmentTypeLookup.TSM;
+
 /**
  * Entity is a master class for basically anything on the board except terrain.
  */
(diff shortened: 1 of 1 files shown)


NickAragua:
What's special about TSM?

Also, what about industrial TSM?

DM0000:
So TSM users can fire weapons to keep heat up.  They should have 0% accuracy with "You have sprinted" msg.  I can add in the industrial check.

NickAragua:
Per "Sprinting" on page 16 of TO: AR, I'm pretty sure sprinting completely disallows weapons firing, so I don't think allowing weapons fire is the right approach here.

gsparks3:
> Per "Sprinting" on page 16 of TO: AR, I'm pretty sure sprinting completely disallows weapons firing, so I don't think allowing weapons fire is the right approach here.

In the exact same passage:
>[...] may still fire weapons to generate heat, though such firing has no chance of doing damage.

NickAragua:
Wow, never realized that. All right, good to go then.

NickAragua (approved):

Issue #4989, opened 2023-12-30
Vehicular Jump Jets aren't properly priced


Saklad5:
### Environment
MegaMek Version: MekHQ v0.49.15 Development Snapshot
Operating System: Mac
Java Version: 20.0.1

### Description

According to page 223 of the sixth printing of _Tactical Operations: Advanced Units & Equipment_, vehicular jump jets cost 200*[Unit Tonnage]*[Jump MP]² C-bills. MegaMekLab is currently skipping them in cost calculations.

---

The cost is supposedly a major barrier to their adoption, though personally I don't see how this could be considered that impactful.

PR #6068, opened 2024-10-05
Replace Manual GUI scaling with FlatLaf Scaling

SJuliez:
This PR removes most of the code of the old manual GUI scaling and replaces it with Flatlaf's built in scaling. It leaves some of the HTML-related methods in UIUtil and just "de-arms" and renames them as the tooltips make heavy use of them and I think we should rewrite those with CSS anyway. This PR also adds a util class for adding font and style to swing components using flatlaf's clientproperty style. The advantage of this is that it is less hassle than using setFont() and scales effortlessly. 

If anyone reviews this, better turn off whitespace changes. It's still a lot.

Merge together with MHQ #4989


SJuliez:
CodeQL warnings are because I marked the old html methods as deprecated. No real problem

HammerGS (approved):
While I didn't open every menu or screen, the scalilng works well.

kuronekochomusuke (approved):

PR #6082, opened 2024-10-10
Data update - line wrapping and consolidating availability ratings

SuperStucco:
Fixes the line wrapping that propagated from the mass naming/line length/general formatting fix.  Great for code, less so for data.

Also consolidates a number of redundant availability values to remove redundant data - lower chance of editing errors later, easier to automatically transcribe contents if at some point this is moved to YAML or other format.

As an example, an availability list of

`... ,CC:6, ... ,IS:6,FS:6, ... ,LA:6,FWL:6, ...`

is simplified down to `... ,IS:6, ...` as the major IS factions are already covered under the IS parent faction with the same availability value.

Similarly, major Periphery factions have been consolidated down under the parent Periphery faction and most of the Clan factions have been consolidated under the CLAN faction.

### DIFF
--- a/megamek/data/forcegenerator/2520.xml
+++ b/megamek/data/forcegenerator/2520.xml
@@ -180,8 +180,7 @@
 	</factions>
 	<units>
 		<chassis name='AC/2 Carrier' unitType='Tank'>
-			<availability>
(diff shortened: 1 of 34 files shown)


IllianiCBT:
Bless you for doing this

PR #6087, opened 2024-10-11
Read Mekset as UTF8

SJuliez:
Fixes unit icon lookup for units with diacritics in the name such as Götterdämmerung and Gùn by reading the mekset file as UTF8. This error only appeared in releases, not in the IDE. Also adds the Locale.ROOT to the uppercase calls to make it (supposedly) less dependent on locale. 
Also fixes small errors in the mekset itself.


SJuliez:
CodeQL is about logging stuff that had been logged before. Not relevant for this fix.

HammerGS (approved):

IllianiCBT:
@SJuliez Am I right to think the CodeQL errors are false-positives?

IllianiCBT:
I realized I could have got my answer by just reading one message up >_>

PR #6098, opened 2024-10-13
add unittooltip font size scaling

kuronekochomusuke:
- add unittooltip font size scaling
- add player color strip in unit tooltips

image

image


HammerGS (approved):
Tested and seems to work.

PR #6121, opened 2024-10-19
#5972: Game Options

### LINKED ISSUE #5972: [0.49.x-0.50.0+] Game Options ignore textbox value changes when in Search
### Environment
What version of MegaMek does your issue apply to?  megamek 0.49.11, 0.50.0, at least several versions before
What java version are you using?  17

### Description
When in the Game Options menu, if you use the Search tab to find options, the textbox fields value updates are ignored when okaying out of the menu.  Checkbox value changes do get applied properly.    For example, if you search for "Lance", some of the options are "TacOps Vehicle Lance Movement" and "Number of vehicles per lance".  If you change the lance movement checkbox and hit ok, the change is applied and you see a line of text in the chat window stating such.   If you change the number in the box (from say, 4 to 2,) and hit ok, the change is ignored and there is no line of text in chat.

Example: 

Before: 
search lance

After updating the values:
search lance edit

After hitting Okay, then loading the game options again (note that the checkbox toggle was applied, but changing 4 to 2 was not):
search lance after ok

Current workaround: you can still set the textbox values from their normal tabs (you just have to go look for them on their original tabs)

### Files
no files should be needed, this is entirely in the game options menu

SJuliez:
Fixes #5972 
Options in the search tab will now also update properly then the game options are changed elsewhere

### DIFF
--- a/megamek/src/megamek/client/ui/swing/DialogOptionComponent.java
+++ b/megamek/src/megamek/client/ui/swing/DialogOptionComponent.java
@@ -29,7 +29,8 @@ import megamek.client.ui.swing.util.UIUtil.FixedYPanel;
 import megamek.common.options.*;
 
 /** @author Cord Awtry */
-public class DialogOptionComponent extends FixedYPanel implements ItemListener, ActionListener, Comparable<DialogOptionComponent> {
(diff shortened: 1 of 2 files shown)


HammerGS (approved):
Seems to work, not sure about the security warning.

PR #6150, opened 2024-10-28
TRO View problem

SJuliez:
This reverts a recent change that broke the troview
image


IllianiCBT:
I have no idea why I used wordWrap here as it was completely unnecessary

PR #6153, opened 2024-10-29
RISC HS Override Kit: Gameplay

pavelbraginskiy:
Implements gameplay for the RISC Heat Sink Override Kit.

Additionally, the Kit is now displayed in the mek preview, when previously it would only be visible in record sheets.

### DIFF
--- a/megamek/src/megamek/server/totalwarfare/TWGameManager.java
+++ b/megamek/src/megamek/server/totalwarfare/TWGameManager.java
@@ -15466,28 +15466,38 @@ public class TWGameManager extends AbstractGameManager {
                         entity.setShutDown(true);
                     } else {
                         int shutdown = (4 + (((entity.heat - 14) / 4) * 2)) - hotDogMod;
+                        TargetRoll target;
                         if (mtHeat) {
                             shutdown -= 5;
+                            target = new TargetRoll(shutdown, "Base tacops shutdown TN");
                             switch (entity.getCrew().getPiloting()) {
                                 case 0:
                                 case 1:
-                                    shutdown -= 2;
+                                    target.addModifier(-2, "Piloting skill");
                                     break;
                                 case 2:
                                 case 3:
-                                    shutdown -= 1;
+                                    target.addModifier(-1, "Piloting skill");
                                     break;
                                 case 6:
                                 case 7:
-                                    shutdown += 1;
+                                    target.addModifier(+1, "Piloting skill");
                             }
+                        } else {
+                            target = new TargetRoll(shutdown, "Base shutdown TN");
+                        }
+                        boolean riscKit = false;
+                        Mek mek = (Mek) entity;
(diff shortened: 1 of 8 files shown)


megamek/src/megamek/server/totalwarfare/TWGameManager.java
```diff
@@ -15512,6 +15523,44 @@ else if ((entity.heat >= 14) && !entity.isShutDown()) {
...
+                            r = new Report(6328);
+                            r.subject = entity.getId();
+                            r.add("%d+1=%d".formatted(hits, hits+1));
+                            addReport(r);
+
+                            hits++;
+
+                            for (int j = 0; (j < 12) && (hits > 0); j++) {
+                                var crit = mek.getCritical(Mek.LOC_CT, j);
+                                if ((crit.getType() == CriticalSlot.TYPE_SYSTEM)
```
SJuliez, line 15557: suggest adding a (crit != null) && ... guard as empty slots will return null. It won't usually make a difference but experience says guard it all the same.

megamek/src/megamek/server/totalwarfare/TWGameManager.java
```diff
@@ -15466,28 +15466,38 @@ else if ((entity.heat >= 14) && !entity.isShutDown()) {
...
-                                    shutdown += 1;
+                                    target.addModifier(+1, "Piloting skill");
                             }
+                        } else {
+                            target = new TargetRoll(shutdown, "Base shutdown TN");
+                        }
+                        boolean riscKit = false;
+                        Mek mek = (Mek) entity;
+                        if (mek.hasRiscHeatSinkOverrideKit()) {
+                            riscKit = true;
```
SJuliez, line 15492: image
this could be rewritten as:
Mek mek = (Mek) entity;
boolean riscKit = mek.hasRiscHeatSinkOverrideKit();
if (riscKit) {

Actually, it isnt necessary to cache hasRiscHeatSinkOverrideKit() as it is already cached in Mek. Which makes me wonder: can the kit be damaged by crit or otherwise be nonfunctional? This is not reflected in the method call.

pavelbraginskiy: The RISC Kit isn't a piece of equipment, it's an intrinsic property of the 'Mech. Rules-wise it behaves like a quirk, except that CGL wanted it to appear in the Weapons and Equipment Inventory of record sheets. 

It's a lot like the full-head ejection system in that sense.

megamek/src/megamek/server/totalwarfare/TWGameManager.java
pavelbraginskiy: I think you can get down to 2 engine slots with a superheavy mech and a compact engine, so it's a realistic concern.

Done.

SJuliez (approved):

PR #6156, opened 2024-10-30
MM Scenarios: Princess settings

SJuliez:
This adds parsing support for bot settings in the scenario (V2) language and updates one of the new scenarios (Kell Hounds/Lowering the Boom) so that both sides can be played with the bot either fleeing or pursuing.


HammerGS (approved):
Tested and seems to be working.

PR #6163, opened 2024-11-01
Load units embedded into MUL file

pavelbraginskiy:
This PR allows for unit files to be embedded in a MUL file:
image

The construction_data field contains a base64'd and gzipped version of this unit's mtf (or blk) file.

If such a MUL file ([test.zip](https://github.com/user-attachments/files/17595092/test.zip)) is loaded in MM or MML with this PR, the file will load successfully, even if you don't have the contained custom marauder available.

THIS PR DOES NOT PROVIDE A WAY TO CONSTRUCT SUCH A MUL FILE, IT ONLY ENABLES THEM TO BE LOADED.

This PR provides the _code_ necessary to create a MUL file with custom units embedded, but there is no way to invoke this code from the GUI. 

Support for creating these MUL files will come in future PRs for MM and MML.


SJuliez (approved):
cool

PR #6164, opened 2024-11-01
feat: create unit list at the end of every encounter, not just the ones you have survivor units

Scoppio:
at the end of encounters, if you are playing bot against bot it does not ask if you want to save the unit list, the same if you have no surviving units, this happens because it looks for the PLAYERS list of surviving units at the end of the fight, I removed this check because the trouble of clicking cancel is lower than the trouble of not having the file when you need.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/ClientGUI.java
+++ b/megamek/src/megamek/client/ui/swing/ClientGUI.java
@@ -1839,6 +1839,7 @@ public class ClientGUI extends AbstractClientGUI implements BoardViewListener,
     public boolean doYesNoDialog(String title, String question) {
         ConfirmDialog confirm = new ConfirmDialog(frame, title, question);
         confirm.setVisible(true);
+        confirm.setAlwaysOnTop(true);
         return confirm.getAnswer();
     }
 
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/client/ui/swing/ClientGUI.java
```diff
@@ -2282,8 +2283,7 @@ public void gameEnd(GameEndEvent e) {
 
             // Allow players to save their living units to a file.
             // Don't bother asking if none survived.
```
pavelbraginskiy, line 2285: Remove this comment if it's not going to be true anymore

pavelbraginskiy (changes requested):

megamek/src/megamek/client/ui/swing/ClientGUI.java
Scoppio: Done

IllianiCBT:
Thank you so much for putting in the work to see this done. This has been on my 'want' list for months.

pavelbraginskiy (approved):

SJuliez:
would it make sense to avoid the yes no dialog entirely? This would show the file browser if Im not mistaken which the player can easily ESC out of.

pavelbraginskiy:
> would it make sense to avoid the yes no dialog entirely? This would show the file browser if Im not mistaken which the player can easily ESC out of.

I feel like it's good to have some kind of prompt before popping up a file browser. 

If you're not paying close attention a sudden file browser appearing without context could be pretty confusing.

I assume a game ending isn't something that happens so often that we're concerned over saving a single click.

Scoppio:
> would it make sense to avoid the yes no dialog entirely? This would show the file browser if Im not mistaken which the player can easily ESC out of.

I dont think its viable, the user has to know the context for the save dialog before opening it, of course there is the title of it, but the confirmation dialog spells out what its about so the user can make the decision. Someone who never used Megamek would be get by surprise with a dialog they know nothing about, with time they may not know the name of it or what its purpose is. I believe its always a good pratice to present the option in this kind of case.

HammerGS (approved):

PR #6168, opened 2024-11-02
#5956: Ice-related skidding PSR checks (underwater movement etc)

### LINKED ISSUE #5956: [0.49.19.1] Movement with Water & Ice
Windows 10
Java 17
Megamek 0.49.19.1

Edit: Issue 1 is fixed

[Fixed] 1. It is not currently possible to deploy a unit on the surface of frozen water. (I.e. if a hex has depth 3 water with ice on top, units will deploy at the bottom of the water, rather than giving an option to deploy on the ice surface.)
[Open Issue] 2. If a mech in depth 2 water with ice on top moves one hex (remaining in frozen depth 2 water), it is treated as both a) breaking through ice from below in the initial hex, and then b) at risk of falling through ice in the second hex. It does not appear possible to remain underwater and move beneath the ice.

Attached  are autosaves and logs of little test I ran.

[savegames.zip](https://github.com/user-attachments/files/16803770/savegames.zip)
[logs.zip](https://github.com/user-attachments/files/16803791/logs.zip)

Attached is the little 8x8 board I made for testing these, in case it will help with fixing.

[8x8 IceTest.zip](https://github.com/user-attachments/files/16803716/8x8.IceTest.zip)

SJuliez:
Adds an elevation check so that units only skid on ice when at elevation 0 = on the ice, not under or above it.

Fixes #5956
Climb mode affects how a mek moves when in depth 2 water. In "thru" climb it'll stay at the bottom, otherwise (default) it'll try to climb and may fall back in

### DIFF
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -84,8 +84,6 @@ import megamek.common.weapons.infantry.InfantryWeapon;
 import megamek.logging.MMLogger;
 import megamek.utilities.xml.MMXMLUtility;
 
-import static megamek.common.EquipmentTypeLookup.TSM;
-
 /**
  * Entity is a master class for basically anything on the board except terrain.
  */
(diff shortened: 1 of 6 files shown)


pavelbraginskiy:
Does this work if the hex is on a hill, so the surface of the ice isn't at level 0?

SJuliez:
Elevation is relative to hex level, so ice can only be at elevation 0, meaning it *should* work which is not to say that it might not actually fail.

pavelbraginskiy:
Hmm, there might still be an edge-case for bridges.

I'm not very familiar with how bridges or ice work in megamek, so maybe this isn't a possible scenario after all, but I believe under the rules you could conceivably have black ice on a bridge surface, with ice covering the water underneath the bridge, in which case you would have ice at two different elevations in one hex.

SJuliez:
ok, updated the code to handle black ice on bridges. Should be good now

pavelbraginskiy (approved):

PR #6181, opened 2024-11-05
Rename TroopSpace to InfantryCompartment

SJuliez:
... as it is called in TM p239
merge together with counterpart PRs in MML and MHQ


kuronekochomusuke:
should the advance search -> transport tab and mek summary also be updated with this?

image

HammerGS (approved):

SJuliez:
done

Issue #5965, opened 2024-09-03
RFE: Princess on Princess Action Pause


johnboyjjb:
An option in the Victory Conditions tab that allows some form of acknowledgement on games where there are only bots left in play to pause at the end phase of turns ending in 0 or 5. This way if the main player gets eliminated, they have an opportunity to /victory or /defeat. With the speed of the bots, this can be a one letter per phase as it currently sits.

Thom293:
Yes please.

FYI - as a current workaround, you can open the edit bots dialog and kick one of them.  Then type victory.  Then add them back and it should finish.  That dialogue isnt reset by phases and will allow you to interact with it.

IllianiCBT:
This would be amazing

HammerGS:
This is implemented last year

PR #6186, opened 2024-11-07
Allow pausing a Princess-only game

SJuliez:
Allows pausing the game by using Ctrl-Shift-P and unpausing with Ctrl-Alt-P.
- Both are normal keybinds that can be changed. 
- It only works if only bot units remain. This limitation is artificial and can be removed but that would have little use I guess. 
- This influences server packet handling so it is potentially dangerous; but as long as it isn't used I don't expect any errors from it
- The effect is purely server-side, meaning that Princess will always finish whatever she is doing which may take a while; but if things move fast, the pause happens immediately
- chat messages and log messages show pause/unpause happening


kuronekochomusuke (approved):

HammerGS:
Fixes: RFE: Princess on Princess Action Pause #5965

IllianiCBT:
Thank you so much, Julie. This is a great addition

PR #6188, opened 2024-11-08
Ulysses - 0.50 maps import

UlyssesSockdrawer:
Adding my new maps from 0.50 to the repo for distribution.


UlyssesSockdrawer:
I've requested a review on this one because when pushing the files it highlighted some tags getting removed in maps that might've already been added to the distro by Hammer. Not sure if there's been map editor code changes that I need to go and resave the maps against.

Let me know if an issue. :)

HammerGS:
We always run the board tagger on the new maps so any missing with get added.

PR #6190, opened 2024-11-09
Allow printing unit list from MM

pavelbraginskiy:
Requires MegaMek/megameklab#1656.

Adds a button to print the units of the selected player, without needing to save the list and opening it in MML:
image

This works by saving the unit list to a temporary file and then opening it in MML automatically.

If MegaMek and MegaMekLab are located in same directory (i.e., from a typical MekHQ distribution), MegaMek will automatically detect the MML executable and launch it. Otherwise, MegaMek will have to be told where MML is in the client settings:
image

This can be the actual MML executable (`MegaMekLab.exe` or `MegaMekLab.sh` as appropriate) or the gradle wrapper of the MegaMekLab git repo (`gradlew.bat` for windows, `gradlew` otherwise). If it's a path to a MegaMekLab executable everything will function as one might expect, if it's a gradle wrapper it will try to compile and run MML from source. 

This allows this feature to be used even with a development version of MegaMekLab.


SJuliez:
I suggest moving over the printing code from MML. That is eventually going to happen anyway.

pavelbraginskiy:
> I suggest moving over the printing code from MML.

That's a _huge_ refactoring. Why do you say it's something that's eventually going to happen anyway?

pavelbraginskiy:
I'll let you work on that then, but I think it's going to be harder than just move and change imports.
You'll have to figure out how to set up record sheet settings to be configurable in both MM and MML, how to handle the sheet template data, probably other stuff I'm missing.

SJuliez:
No idea when I'll get to it. Or if.

SJuliez (approved):
looks good. Suggest calling the button "Print Record Sheets" or at least note this in the tooltip. It's probably not that self-evident in MM.

PR #6196, opened 2024-11-10
add an advanced search for boards to the lobby

### LINKED ISSUE #433: RFE: Ability to search a Map Board by Name/Size/Features
As the RFE says.

Would be nice to be able to search a map by things like:

Name
Size
What features it has.

kuronekochomusuke:
- add an advanced search for boards to the lobby
- allows searching across board sizes
- filter by map tags
- filter by map path
- filter by width, height, and name 
- sort by the name or size
- selected board replaces any previous boards with the new selection when completing the dialog
- disable when using space map
- disable when map page size is not 1x1

image
image
image
image

fixes #433


HammerGS:
This is so cool.

HammerGS:
Been playing around with this and it's awesome. But I noticed the size filter doesn't seem to sort correctly
image

Made be realize our Boards folder is in a bit of sorry state with the names all over the place and lots of duplicates. 

Do you think there would be a way to identify duplicate boards?

The width and height is cool, but took me a couple of tries to get the usage correct. Maybe a helper tip?

kuronekochomusuke:
the size sorts as a string that is why it sorts that way.  could try 0 padding the width and height to get it to sort correctly.

could try doing a check  sum in the BoardsTagger and logging it there.  not sure if it would get them, the file may not be always written is the same order.  so if the order is slightly different they would show as different, even though they would produce the same map.  maybe have it write out the preview images into a temp folder and then doing a compare on those would be better.

there are also some that look like the would only differ by the theme.  since you can not select the theme for the board multiple were made with different ones.

can try adding a tooltip for the width and height inputs.

HammerGS (approved):
Been testing and seems to work with no issues.

PR #6208, opened 2024-11-18
Autoresolve support

Scoppio:
Extracts interface for some "end of game" events, this small necessary change is necessary for the implementation of an autoresolve feature that integrates with the Resolve Scenario wizard.

https://github.com/MegaMek/mekhq/pull/5155


HammerGS:
Is this still in Development, also could we get a description so we know what it does.

HammerGS (changes requested):
See note regarding Description and status

Scoppio:
> See note regarding Description and status

I talked with Illiani, we will move forward with this one, as the other type of autoresolve I am planning is in the back burner and may take longer than expected.

PR #6209, opened 2024-11-19
Implement RISC laser pulse module

neoancient:
This required changes to method signatures of EquipmentType::getToHitModifier and EquipmentType::getModesCount. getToHitModifier needs to know the weapon's mode and getModesCount needs to know whether the mount is linked by a laser pulse module. I added a getToHitModifier with no parameters to AmmoType (which is the old signature) to reduce the number of changes to the code.

### DIFF
--- a/megamek/src/megamek/common/EquipmentType.java
+++ b/megamek/src/megamek/common/EquipmentType.java
@@ -556,6 +556,15 @@ public class EquipmentType implements ITechnology {
     }
 
     /**
+     * @param mounted The equipment mount. In some cases the moudes are affected by linked equipment.
+     * @return the number of modes that this type of equipment can be in or
+     *         <code>0</code> if it doesn't have modes.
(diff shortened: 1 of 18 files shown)


megamek/src/megamek/common/EquipmentType.java
```diff
@@ -502,7 +502,7 @@ public boolean isSpreadable() {
         return spreadable;
     }
 
-    public int getToHitModifier() {
+    public int getToHitModifier(@Nullable Mounted<?> mounted) {
```
neoancient, line 505: There are some cases (particularly in AlphaStrike and BattleForce conversion code) where the Mounted is not available. In those cases a `null` is passed, which skips the check for the laser pulse module.

HammerGS (approved):
Tested and seems to work. Kinda fun piece of kit.

Issue #6069, opened 2024-10-05
Dropships can load BA SRM/LRM Rounds


repligator:
### Prerequisites and Pre-Issue Checklist

- [X] I'm reporting the issue to the correct repository:
- [MegaMek](https://github.com/MegaMek/megamek/issues)
- [MegaMekLab](https://github.com/MegaMek/megameklab/issues)
- [MekHQ](https://github.com/MegaMek/mekhq/issues)

- [X] I've tested the issue against at least the latest MILESTONE version
- [X] I've asked on the MegaMek Discord about the error
- [X] I've reviewed the BattleTech rules and MegaMek documentation, and I've confirmed that something isn't working as intended.
- [X] I've searched the Github tracker and haven't found the issue listed

### Severity *

Low (Minor/Nuisance): Minor glitches or cosmetic issues that don’t affect gameplay and occur rarely.

### Brief Description *

In the MM lobby, dropships show 2-4 different options for SRM and/or LRM ammo.
![Screenshot_20241005_120832](https://github.com/user-attachments/assets/52ab22e5-bde5-41a8-8082-173524bce691)
The first two (right to left) function normally. The last will only allow a single round to be added. This single round will be the BA equivalent, and will make the unit invalid, as it has less than the required ten shots.

Machine guns sometime have up to 5 different options, however these all seem to load normal ammo.

Possibly related to https://github.com/MegaMek/megamek/issues/6050 , but I've opened a new issue in case it isn't, feel free to close mine if it turns out to be a duplicate.

### Steps to Reproduce

1. Add a dropship with LRM/SRM to the lobby.
2. Configure ammo.
3. Right click the drop to view.
4. Scroll down to the invalid notice at the bottom.

### Operating System *

Linux

### Java Version *

17

### MegaMek Suite Version *

Free Text (type manually)

### Custom MegaMek Version

https://github.com/MegaMek/megamek/commit/e1d2728d2a51e221c92f9f8d1855e899303a4386

### Attach Files

[megamek.log](https://github.com/user-attachments/files/17266803/megamek.log)
[ba_on_DS.sav.gz](https://github.com/user-attachments/files/17266805/ba_on_DS.sav.gz)

### Final Checklist

- [X] I've checked to make sure that this issue has not already been filed
- [X] I'm reporting only one issue in this ticket for clarity and focus

PR #6220, opened 2024-11-25
fix: stops bays from loading BA and protomek ammunition

Scoppio:
Adds an extra filter when loading munitions which discard BA/Protomek exclusive ammo types for the `BayMunitionsChoicePanel`.

All the other stuff changed is non-functional, just for better name or type expression and to remove some odities of "very-old-code (tm)"

Fixes: #6069


megamek/src/megamek/client/ui/swing/BayMunitionsChoicePanel.java
```diff
@@ -226,33 +225,42 @@ class AmmoRowPanel extends JPanel implements ChangeListener {
...
                         || entity.hasWorkingMisc(MiscType.F_ARTEMIS_PROTO);
             }
-            if (atype.getMunitionType().contains(AmmoType.Munitions.M_ARTEMIS_V_CAPABLE)) {
+            if (ammoType.getMunitionType().contains(AmmoType.Munitions.M_ARTEMIS_V_CAPABLE)) {
                 return entity.hasWorkingMisc(MiscType.F_ARTEMIS_V);
             }
+            // A Bay should not load BA nor Protomek exclusive ammo/weapons
+            if (ammoType.hasFlag(F_BATTLEARMOR) || ammoType.hasFlag(F_PROTOMEK)) {
+                return false;
+            }
```
Scoppio, line 263: This is the only code newly added.

PR #6221, opened 2024-11-25
Corrects weight and critical slots for BNIU

sldfgunslinger2766:
Adjusts the weight/critical slots for the Battlemech Neural Interface Unit from 0/0 to .100/2


sldfgunslinger2766 (commented):
🤘

Issue #5395, opened 2024-04-22
Snord's Irregulars, Fidelis missing as factions


SuperStucco:
The ...\data\forcegenerator\factions.xml file is missing an entry for Snord's Irregulars, which was present in earlier releases.  The faction key 'MERC.SI' is still present in a number of entries in the force generator XMLs, such as the Black Knight:

https://github.com/MegaMek/megamek/blob/a0ed7c0ad75589d86f887b5d309b67a53348c687/megamek/data/forcegenerator/3039.xml#L1296-L1298

This might revert back through the parent mercenary/MERC faction as a default, but it would be best if the data was consistent i.e. MERC.SI as a valid faction key, or removed from the force generator XMLs.

Similarly the Fidelis/SL3.CSJ faction has been removed while the force generator XMLs have references to it.

Issue #5393, opened 2024-04-22
Minor change: Blessed Order faction key


SuperStucco:
The faction data in the force generator files use abbreviations for almost everything e.g. CS for ComStar, FS for Federated Suns, and so on.  There is one standout, 'Blessed Order' is spelled out and includes a space, which is not consistent with the rest of the naming:

https://github.com/MegaMek/megamek/blob/a0ed7c0ad75589d86f887b5d309b67a53348c687/megamek/data/forcegenerator/factions.xml#L33

I request that the key attribute change to something consistent with the rest of the faction keys, such as 'BLORD'.  This would need to be propagated to a number of data files such as the force generator XML files.  Thankfully there are very few instances of this.

PR #6231, opened 2024-11-28
Force generator data role and availability updates

SuperStucco:
Over the years I've been updating my own copy of the force generator availability files - doing some cleaning, adding role designations to specialist units, and so on.  This update ports over the applicable changes plus incorporates some reviews/updates.

Most of this is based on assigning roles.  To be specific these are the force generator roles, detailed in ...\docs\RAT and Force Generator Stuff\rat-generator.txt which allow for creating more specialized forces.  For example, generating a force with the URBAN role is more likely to have units with short ranged weapons, wheeled vehicles, and similar characteristics.  A force generated with the FIRE_SUPPORT role should be mostly long ranged weapons, while one with the CAVALRY role should be a strong balance of armor, speed, and concentrated firepower.

As I was updating those, I was also doing a quick back-check for faction availability using MUL as a guide.  This part wasn't exhaustive, there are many units which were not reviewed as they were not otherwise updated.  That's going to need an ongoing/rolling operation on a targeted basis e.g. FWL sub-states (those are a little chaotic and could use some work), Escorpion Imperio/Scorpion Empire (they have a lot of units inherited from many different factions), early Star League, early Pentagon/KLONDIKE era, etc.

Part of the faction availability checking was reviewing for consistency and optimizing.  Many early units have limited chassis availability e.g. TH:8 (Terran Hegemony) while their single model has been copied back from a later era, such as CC:8,IS:8,Periphery.Deep:8,FS:8,Periphery:8.  That can normally be simplified down to IS:8 (which covers the TH faction), but can be further optimized to TH:8 as the unit is only available to that faction.  The Balius OmniMek is a good example of this optimization - the chassis is limited to CHH (Clan Hell's Horses), with numerous model entries set to General:8.  When this data is processed it first checks for CHH, and not finding that it starts working it's way up through the faction parentage CHH > CLAN.IS > CLAN > General.  This optimization should save a few processing cycles and have a modest performance improvement, as well as improve readability for future updates.

Particular attention was paid to battle armor availability, ensuring that different factions get their appropriate squad sizes.  Most IS and Periphery nations will deploy (Sqd4) squads, while the Marian Hegemony deploys (Sqd5) contubernium, and ComStar/WOB deploys (Sqd6) Level Is.

Covers a couple of faction code updates, closing #5393 and #5395.


SuperStucco:
For those looking for the gory details, see attached.
[Unit update commit log.txt](https://github.com/user-attachments/files/17951437/Unit.update.commit.log.txt)

IllianiCBT:
A happy thanksgiving to you, too! I have been waiting eagerly for this and I’m really happy to see you got it across the finish line!

Does this include the expansion of Civilian, Support, and Cargo roles?

SuperStucco:
No, I'm going to do the Civilian RATs as a separate PR.  That's going to be another whopper, and I'm aiming for future data updates to be smaller and easier to review.

HammerGS (approved):

PR #6239, opened 2024-12-02
ACS-Like autoresolve

Scoppio:
Alot of the changes for [ACS-Like Autoresolve](https://github.com/MegaMek/mekhq/pull/5271/files) depend heavily in Megamek having a "sane" game interface which allow us detach entities from the gameboard.

However, many of the tools require the current TW Game as a basis, be it to declare a winner, be it to do checks like unit damage, limb blownoff, pilot survival, etc.

This Pull Request has multiple changes that has effect on absolutelly nothing. It just extract interfaces so we can implement and use stubs for the game object, publish end game events, etc. It also has many "weird" changes that do cause big effects, mostly improving initialization times, reducing number of function calls and reducing memory usage. Those changes are not earthshatering, they are more like "the full process takes 300ms, of those, 270ms are spent parsing a file, of that time, 70ms are spent loading the options for the hundreds of equipments, this change reduced from 70ms to 7ms the options loading, reducing the final time to 207ms

It also includes a change to log4j2 so that a developer may simply set an env var and have the logs show up in the console without the need for anything else.


megamek/mmconf/log4j2.xml
```diff
@@ -17,28 +17,30 @@
...
         <Logger name="mekhq" level="info" additivity="false">
             <AppenderRef ref="MekHQLog" />
-            <AppenderRef ref="Console" />
+            <AppenderRef ref="${env:mm.profile:-null}" />
         </Logger>
 
         <Root level="info">
-            <AppenderRef ref="Console" />
+            <AppenderRef ref="MegaMekLog" />
+            <AppenderRef ref="${env:mm.profile:-null}" />
```
Scoppio, line 43: I need to document this somewhere

megamek/src/megamek/client/generator/RandomUnitGenerator.java
```diff
@@ -497,14 +488,25 @@ public static synchronized RandomUnitGenerator getInstance() {
...
             rug.loader.setPriority(Thread.NORM_PRIORITY - 1);
             rug.loader.start();
         }
+
+        try {
+            latch.await();
+        } catch (InterruptedException e) {
+            logger.error("Initialization interrupted", e);
+            Thread.currentThread().interrupt();
+        }
```
Scoppio, line 508: substitutes the while sleep loop with a latch await

megamek/src/megamek/client/ui/swing/lobby/MekForceTreeRenderer.java
```diff
@@ -134,7 +134,7 @@ private void setIcon(Image image, int height) {
             int width = height * image.getWidth(null) / image.getHeight(null);
             setIcon(new ImageIcon(ImageUtil.getScaledImage(image, width, height)));
         } else {
-            logger.error("Trying to resize a unit icon of height or width 0!");
+            logger.debug("Trying to resize a unit icon of height or width 0!");
```
Scoppio, line 137: an error should be something that we need to fix, this apparently is here just to tell "hey, the image is not imaging"

megamek/src/megamek/common/Entity.java
```diff
@@ -9901,7 +9874,7 @@ public boolean isEligibleForFiring() {
         }
 
         // check game options
-        if (!game.getOptions().booleanOption(OptionsConstants.BASE_SKIP_INELIGABLE_FIRING)) {
+        if (!game.getOptions().booleanOption(OptionsConstants.BASE_SKIP_INELIGIBLE_FIRING)) {
```
Scoppio, line 9877: typo

megamek/src/megamek/common/Game.java
```diff
@@ -322,10 +329,10 @@ public void setupTeams() {
         Vector<Team> initTeams = new Vector<>();
         boolean useTeamInit = getOptions().getOption(OptionsConstants.BASE_TEAM_INITIATIVE)
                 .booleanValue();
-
+        var players = getPlayersList();
```
Scoppio, line 332: return the lpayer list only once.

megamek/src/megamek/common/options/BasicGameOptions.java
```diff
@@ -19,44 +19,61 @@
...
+    private static final Option.OptionValue[] victoryOptions = {
+        Option.of(OptionsConstants.VICTORY_CHECK_VICTORY, true)
+    };
+
+
+    private void addOptions(IBasicOptionGroup group, Option.OptionValue[] options) {
+        for (var entry : options) {
+            addOption(group, entry.getName(), entry.getType(), entry.getValue());
+        }
+    }
```
Scoppio, line 68: Big improvement in performance for loading options

megamek/src/megamek/common/options/StaticGameOptions.java
```diff
@@ -0,0 +1,88 @@
...
+import megamek.server.victory.BVDestroyedVictoryCondition;
+import megamek.server.victory.BVRatioVictoryCondition;
+import megamek.server.victory.EnemyCmdrDestroyedVictory;
+import megamek.server.victory.KillCountVictory;
+
+import java.util.Enumeration;
+import java.util.HashMap;
+import java.util.Map;
+
+public class StaticGameOptions implements IGameOptions {
```
Scoppio, line 12: Improve performance for auto-resolve scenario simulations by not having to load the same set of options hundreds of times.

PR #6245, opened 2024-12-05
feat: add new chat commands, updates a bit the commands gui, give nukes to GMs

Scoppio:
Even more GM tools
---

## Change player team

This GM command allow the GM to change players from teams manually, no need for doing it just one player per turn. It is still done at the end of the round, but the teams can be reassigned freely until them.

## End game

This command from the GM just ends the game, no need for voting, it may also be forced to end immediatelly. It also declares one team as victorious, but things like MHQ will keep doing what it does to decide who won the game.

## Nuclear Strike & Nuclear Strike Custom

Just more modern implementations of the old command "Nuke", these can only be called by the GM.

## GameListener gamePlayerStrategicAction

Triggers an event that can be picked up by MHQ whenever a player uses a nuclear strike or something else.

## GUI commands

Implements some intelligence/automation for GUI Commands (the chat commands shown in a GUI). If So now there are the PlayerArgument, UnitArgument, TeamArgument, so they list the players, teams and units. This allows for a more rich experience when using those tools instead of just random numbers appearing out of nowhere.

## Draw nuke

Now nukes from Arty should be drawn on the game too.


HammerGS (approved):
Tested and seems to be working.

HammerGS:
@Scoppio Just want to confirm this is complete and ready to merge.

PR #6246, opened 2024-12-06
Issue 1619 RFE: dirt and gravel roads

### LINKED ISSUE #1619: RFE: Dirt and Gravel Roads (TacOps p51)
Would it be possible to implement Dirt and Gravel roads (as per TacOps p51)?

At present Road = 2 provides an alternative visual for the road (with boundary shrubs).
Road = 3 (or greater) provide a different alternative visual on the Saxarba tilesets (these appear to be tyre-marks) but do not appear to be supported on other tilesets.

To avoid breaking existing maps, would recommend that unreserved values be used to indicate dirt or gravel roads. Appreciate this would also require amendments to the tilesets with the addition of new sprites for the new road surface types.

### LINKED ISSUE #5595: 49.19.1: RFE: Road Type 3 does not utilize exits to modify the sprite
### Environment
Windows
Java 11

### Description
When swapping Road terrain to type 3, it does not utilize the exits to modify the road sprite like the previous 2 types do.

I went through and created new sprites for the dirt road for each exit value, that will allow them to be linked together based on their exits.  See attached dirt_road.zip

### Files

road_nolink
road_link
[dirt_road.zip](https://github.com/user-attachments/files/15888992/dirt_road.zip)

psikomonkie:
Resolves #1619 
Resolves #5595

Contains new graphics for gravel roads (from stonewall072) and dirt roads (Morkath)


psikomonkie:
> Do we need art for this?

No & Yes. 

No, we don't need art for this to function. Currently a dirt road and gravel road look the same, but they function as they're supposed to. 

Yes, I think we'll want art for gravel.

I do not think we should hold off on merging this PR until we have gravel road art as the gravel roads art is purely decorative and won't effect the functionality of this. Gravel & Dirt roads should not show up during map generation, they should only appear if a map maker uses it.

psikomonkie:
Now includes art for gravel and dirt roads. Dirt roads use the art made by Morkath for #5595 and gravel roads use art made by stonewall072 for #1619 

Thanks!

HammerGS (approved):

PR #6254, opened 2024-12-11
Fix 6240: era-illegal asf bombs [Note: breaks MHQ]

### LINKED ISSUE #6240: 47.14 MekHQ/MegaMek Load Aircraft with bombs that don't exist at that time.
### Environment
HQ MegaMek
Win 10
Java  8 271
(Delete the above lines)

### Description
See attached save game.  Campaign file included.

Aircraft is loaded with Thunder ammo which is extinct until 3052 per TacOps p 360
Year of Campaign is 3047.  I believe the same is occurring with rocket pods which are not reintroduced until 3052.

### Files
[autosave_2021-01-19_22-12-11.sav.gz](https://github.com/MegaMek/megamek/files/5840136/autosave_2021-01-19_22-12-11.sav.gz)
[Tuesday Group30470508_AFC.cpnx.gz](https://github.com/MegaMek/megamek/files/5840137/Tuesday.Group30470508_AFC.cpnx.gz)

Sleet01:
Fixes the long-standing issue where automatically-generated ASF munitions ignore era restrictions, leading to e.g. early CI enemy forces with Thunder bombs.
The implementation re-uses existing era legality checks from the non-ASF munitions generation code.
Auto-generated ASF munitions should now all be restricted to era-appropriate options in both MM and MHQ.

As a sop to Princess, however, this also implements Rocket Launcher (Prototype), or RL-P, bomb pods.
This will allow Clan ASFs to be equipped with RL-P pods through 2823, and IS ASFs from the Early Spaceflight period on.  

Testing:
- Ran all 3 suites' unit tests (This breaks MHQ, requiring a separate patch there)
- Added unit tests to confirm era-appropriate munition selection for IS and Clan units.

NOTE: https://github.com/MegaMek/mekhq/pull/5337 needs to be pulled in close conjunction with this PR.

Close #6240

### DIFF
--- a/megamek/src/megamek/client/generator/TeamLoadOutGenerator.java
+++ b/megamek/src/megamek/client/generator/TeamLoadOutGenerator.java
@@ -1565,7 +1565,7 @@ public class TeamLoadOutGenerator {
      *         enums as the lookup e.g. [BombUnit.HE] will get the number of HE
      *         bombs.
      */
-    private int[] generateExternalOrdnance(
+     public int[] generateExternalOrdnance(
         int bombUnits,
         boolean airOnly,
         boolean isPirate,
--- a/megamek/src/megamek/client/generator/TeamLoadOutGenerator.java
+++ b/megamek/src/megamek/client/generator/TeamLoadOutGenerator.java
@@ -1771,8 +1771,8 @@ public class TeamLoadOutGenerator {
         }
 
         // Oops, nothing left - rocket launchers are always popular
-        // XXX: Add prototype RLs as well.
         if (Arrays.stream(bombLoad).sum() == 0) {
+            // Rocket Launchers are a good option after CI era
             if (
                 checkLegality(
                     BombType.createBombByType(BombType.B_RL),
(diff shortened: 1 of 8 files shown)


IllianiCBT (approved):
I am always nervous reviewing MM content, but this looks fine to my eyes. I would prefer a second opinion before it gets merged.

megamek/src/megamek/client/generator/TeamLoadOutGenerator.java
```diff
@@ -1708,37 +1741,61 @@ private static int[] generateExternalOrdnance(int bombUnits,
...
+                }
             }
         }
 
         // Oops, nothing left - rocket launchers are always popular
         if (Arrays.stream(bombLoad).sum() == 0) {
-            bombLoad[BombType.B_RL] = bombUnits;
-            return bombLoad;
+            // Rocket Launchers are a good option after CI era
+            if (
```
SJuliez, line 1776: If I may say so, this feels slightly like newline overkill

SJuliez (approved):

PR #6266, opened 2024-12-14
feat: env var mm.profile=dev makes log print on console

Scoppio:
The console will log everything form the logger if you set an environment variable with `mm.profile=dev`

This way making debuging and developing a bit easier


SJuliez (approved):
works

PR #6276, opened 2024-12-15
Issue 4259: Fixing Dropship Heat-by-Arc for Players and Princess

### LINKED ISSUE #4259: [0.49.12] Grounded Spheroid Dropships - Princess Does Not Fire Weapons Mounted In Front Left/Right Bays
Fill out the fields below with as much pertinent information as possible to help the developers solve the issue.
Search for your issue inside open and closed issues first before reporting so as not to make duplicates.
(Delete the above lines)

### Environment
Starting MegaMek v0.49.12
	Build Date: 2023-03-04T15:27:34.898759400
	Today: 2023-03-20
	Origin Project: MegaMek
	Java Vendor: Eclipse Adoptium
	Java Version: 17.0.6
	System Locale: en_US

### Description
Grounded spheroid dropships are only using weapons in their Aft-Left and Aft-Right bays while on the ground. Princess is not firing weapons in the front left/right bays.

Saves included to demonstrate issue:
dropshipFiringArcStart.sav - Very beginning of the game
dropshipFiringArcMoved.sav - After units moved into range
dropshipFiringArcFiring.sav - After princess has declared weapons fire for the dropship

Player name: Tempest Marauders

Note that in the dropshipFiringArcFiring.sav, she is only firing the ER Large Laser mounted in the left rear bay, even though the Gauss, LRM's, and ERPPC that are mounted in the left front bay are also in range.

There's also an NPE that occurs when I view the enemy dropship and then click the weapons to view the firing arcs. I don't know if it's related to this or not, but it should show up in the log at the bottom, as that was the last thing I did before exiting the game to submit this.

### Files

[Grounded Dropship Firing Arc Bug.zip](https://github.com/MegaMek/megamek/files/11020415/Grounded.Dropship.Firing.Arc.Bug.zip)

psikomonkie:
Fixes #4259 

It works, Princess now fires as many weapons as she can when using the default Dropship heat rules, without going over the heat limit of dropships.

The original bug wasn't present (only using Aft weapons & the NPE), but Princess & grounded dropships were still doing funny stuff. 

Bugs resolved:
- Princess was improperly calculating the optimal firing plan when using the default rules for dropship heat. 
- Grounded dropships for the player were improperly calculating heat
- Artillery fired by grounded dropships was not accounted for in Princess's firing plan, letting her go over her heat limit
- UI bug related to showing arc heat


psikomonkie:
This is ready for review. I would appreciate a code review for this as it's more than just a trivial change. 

Thanks!

megamek/src/megamek/client/ui/swing/unitDisplay/WeaponPanel.java
```diff
@@ -1195,12 +1195,10 @@ public void displayMek(Entity en) {
                     && game.getPhase().isFiring()) {
                 hasFiredWeapons = true;
                 // add heat from weapons fire to heat tracker
-                if (entity.usesWeaponBays()) {
+                if (entity.isLargeCraft()) {
```
Sleet01, line 1198: Are we _absolutely certain_ that this only applies to Large Craft, not other Aerospace units that can use bays?
Apparently Small Craft, Fighter Squadrons, and individual ASFs have all had the option to use bays at one point or another, but I'm not well-enough versed to say if that applies here or not.

megamek/src/megamek/client/ui/swing/unitDisplay/WeaponPanel.java
```diff
@@ -1257,16 +1255,20 @@ public void displayMek(Entity en) {
...
-            wArcHeatR.setVisible(true);
             m_chBayWeapon.setVisible(true);
             wBayWeapon.setVisible(true);
         } else {
-            wArcHeatL.setVisible(false);
-            wArcHeatR.setVisible(false);
             m_chBayWeapon.setVisible(false);
             wBayWeapon.setVisible(false);
         }
+        if (entity.isLargeCraft() && (game == null || !(game.getOptions().booleanOption(OptionsConstants.ADVAERORULES_HEAT_BY_BAY)))) {
```
Sleet01, line 1264: Is this correct?  The ArcHeat label will display if the game is null?

Sleet01 (approved):
A couple of questions but overall looks good.  I appreciate the code clean-up and safety checks you've added as well.

megamek/src/megamek/client/ui/swing/unitDisplay/WeaponPanel.java
gsparks3: Fighter Squadrons in particular use bays (at least tabletop-rules-wise). However, a recent-ish ruling explicitly says that ASF and SC should _not_ use weapon bays otherwise: https://bg.battletech.com/forums/index.php?topic=72939

megamek/src/megamek/client/ui/swing/unitDisplay/WeaponPanel.java
psikomonkie: Yeah, I think so. I originally didn't have that, but the code scan reported that the game may be null - This is at the beginning of the method:

```

// Grab a copy of the game.
Game game = null;

if (unitDisplay.getClientGUI() != null) {
  game = unitDisplay.getClientGUI().getClient().getGame();
}

```

Elsewhere in this method, when it wants to check the gameOptions, it first makes sure the game isn't null. If it is null, then it uses the default value:


```
 // Check Game Options for max external heat
int max_ext_heat = game != null ? game.getOptions().intOption(OptionsConstants.ADVCOMBAT_MAX_EXTERNAL_HEAT)
                : 15;

```

In this case, if the entity is a large craft, if we're unable to get the gameOption for ADVAERORULES_HEAT_BY_BAY we should assume the default setting - which means displaying heat by arc labels for large craft. I can change the block of code to be like this so it's using game !=null, but functionally it should be the exact same:


```
if ((!entity.isLargeCraft()) 
  || ((game != null) && (game.getOptions().booleanOption(OptionsConstants.ADVAERORULES_HEAT_BY_BAY)))){
  wArcHeatL.setVisible(false);
  wArcHeatR.setVisible(false);
}
else {
  wArcHeatL.setVisible(true);
  wArcHeatR.setVisible(true);
}

```

Thoughts?

megamek/src/megamek/client/ui/swing/unitDisplay/WeaponPanel.java
psikomonkie: > Are we _absolutely certain_ that this only applies to Large Craft, not other Aerospace units that can use bays? Apparently Small Craft, Fighter Squadrons, and individual ASFs have all had the option to use bays at one point or another, but I'm not well-enough versed to say if that applies here or not.

Good catch! I had made a helper method already that'll check if a weapon is a bay, and then either add the heat for the weapon if it isn't a bay, or add the heat for all the bay weapons if it is a bay. 

This if statement's "else" has been modified to handle it - if some entity, for some reason, has weapon bays but isn't a large craft, it should be able to properly add its heat now.

I'm going to check my other changes to make sure this same logic applies there.

psikomonkie: I updated this location & another to use the getHeatByBay to ensure it still works. The other locations where I changed checking if a has a weapon bay to checking if it's a large craft are truly only for large craft.

megamek/src/megamek/client/ui/swing/unitDisplay/WeaponPanel.java
Sleet01: I see what you're saying, yeah.
I _personally_ prefer the latter implementation, `if ((!X) || ((Y != null) && (Y.z))`, from a readability standpoint but since it's functionally identical there's no reason to change it.

psikomonkie: To be fair, when I saw it & read your comment, I was questioning what I did too. I agree that it's more readable, and made the change. Thanks!

psikomonkie:
Thank you @Sleet01! I made some changes at your suggestion.

Sleet01 (approved):
LGTM!

PR #6282, opened 2024-12-17
fix: fixes error with wrong type cast on GM commands for nukes and kick

Scoppio:
Some time ago I refactored some things and added a few extra argument types, but forgot to change in the places where I has manually casting then due to bad generic usage. Unfortunatelly those areas required a kind of approach that made it not be picked up by the compiler nor any parser/intellisense.


Sleet01 (approved):
LGTM

PR #6289, opened 2024-12-19
feat: allows gunEmplacement to be converted to Alpha Strike

Scoppio:
I am enabling the conversion of GunEmplacement to Alpha Strike element as even it not being covered by rules as written, it is possible to use it as is to create a valid unit, and ACAR does need it to run scenarios in facilities.


megamek/src/megamek/common/alphaStrike/conversion/ASConverter.java
```diff
@@ -206,14 +206,16 @@ public ConversionData(Entity entity, AlphaStrikeElement element, CalculationRepo
     /**
      * Returns true if the given entity can be converted to AlphaStrike. This is
      * only
-     * false for entities of some special types such as TeleMissile or
-     * GunEmplacement.
+     * false for entities of some special types such as TeleMissile.
+     * GunEmplacement can is being allowed conversion as of 50.02 even though its not
```
Sleet01, line 210: "can" is superfluous here.

Sleet01 (approved):
Looks good.  Comment could use a small clean-up but doesn't impact code.

HammerGS (approved):

PR #6328, opened 2024-12-30
Building type enum

NickAragua:
To prepare for some work regarding allowing players to put down buildings per scenario rules (a lot of scenarios call for a player to be able to "place X buildings of height Y and CF Z"), I've moved BuildingType to an enum, as that will make the UI and other relevant work a lot easier down the line.

Also fix a long-standing bug where buildings turned to rubble wouldn't actually allow picking up clubs.

### DIFF
--- a/megamek/src/megamek/common/enums/BuildingType.java
+++ b/megamek/src/megamek/common/enums/BuildingType.java
@@ -25,21 +25,27 @@ import java.util.ResourceBundle;
 import megamek.MegaMek;
 
 public enum BuildingType {
-    UNKNOWN("BuildingType.UNKNOWN.text", -1),
-    LIGHT("BuildingType.LIGHT.text", 1),
-    MEDIUM("BuildingType.MEDIUM.text", 2),
-    HEAVY("BuildingType.HEAVY.text", 3),
-    HARDENED("BuildingType.HARDENED.text", 4),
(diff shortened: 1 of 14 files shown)


megamek/src/megamek/common/enums/BuildingType.java
```diff
@@ -0,0 +1,63 @@
...
+    HARDENED("BuildingType.HARDENED.text", 4, 120),
+    WALL("BuildingType.WALL.text", 5, 120);
+    
+    private final String name;
+    private final int type;
+    private final int defaultCF;
+    
+    private BuildingType(String name, int type, int defaultCF) {
+        final ResourceBundle resources = ResourceBundle.getBundle("megamek.common.messages",
+                MegaMek.getMMOptions().getLocale());
```
Sleet01, line 41: Can this be moved to a class member?
Or is there no danger of map draw slowdowns because we only instantiate this once per Building Type, early on?

Sleet01 (approved):
This all looks great; I had a question about the resource loading in BuildingType, but I don't think it's critical.

megamek/src/megamek/common/enums/BuildingType.java
Scoppio: because this is an enum it will load only once

PR #6341, opened 2025-01-02
update status display for sim fire, sim  physicial to show remaining players

kuronekochomusuke:
- update status display for sim fire, sim  physicial to show remaining players
- stop my turn beep from firing multiple times with sim fire, sim  physicial

image


Sleet01 (approved):
LGTM!

PR #6343, opened 2025-01-02
Fix: Tank armor and invalid builds

Scoppio:
Fixes issue [MegamekLab #1674](https://github.com/MegaMek/megameklab/issues/1674).

It makes sure the correct armor tech level is loaded when loading the tank or SV.
It also makes sure the correct tech level is persisted when saving the non-mek entity.

This also adds a list of reasons/justifications to why a canon unit build is invalid, and allows it to be loaded.
If the unit is changed in a way to make it lose its canon status (change the MUL to -1 or change the name to something that is not in the canon names list) it will stop saving the reasons list for the build fail.

Also adds a bulk file verification test for all units in MegaMek.

Fixes:

- SV and Tank patchwork (IS and Clan)
- SV and Tank high tech armor level and rating (IS and Clan)

Implements:
- Build specific tag for canon units with the reason for failing build (only for BLK files)
- The invalid build reason tags are lost when making custom units derived from canon units (only for BLK files)

Example of use in a BLK file:

```

<invalidSourceBuildReasons>
UNIT_OLDER_THAN_EQUIPMENT_INTRO_YEAR
</invalidSourceBuildReasons>

```

### DIFF
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -14,16 +14,7 @@
  */
 package megamek.common;
 
-import java.awt.Image;
-import java.math.BigInteger;
-import java.util.*;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.function.Predicate;
-import java.util.stream.Collectors;
-import java.util.stream.IntStream;
-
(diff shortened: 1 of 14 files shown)


Sleet01 (approved):
Looks great; this will be a godsend.

PR #6375, opened 2025-01-08
fix: fixes a class cast bug and an error in the argument parsing algorithm

### LINKED ISSUE #6374: [0.50.02/0.50.03-SNAPSHOT] /nuke command custom nukes aren't working
### Brief Description *

Currently, when you try to use the /nuke command it will not trigger unless a "Type" has been specified. Including the Type parameter will override any other parameters you include in the command, making it impossible to use the command to generate a custom nuclear attack. 

### 3. Steps to Reproduce

1. Load into game
2. Attempt to use /nuke command using the parameters for a custom nuclear weapon
3. Nuke will not trigger and message in chat window will state that you must specify a type
4. Using 0-4 as the type parameter will allow the command to trigger, but any other parameters following that will be ignored

### Attach Files

_No response_

### Severity *

### MegaMek Suite Version *

0.50.02

### Operating System *

Windows 11

### Java Version *

17.0.11

### Final Verification

Scoppio:
fixes #6374 
The changes I made to the nuke command originally made it impossible to use the custom nuke command, this change allows the argument parser to discover which is the correct set of arguments to use in case you are writting only positional arguments.

This also fixes an exception that was happening due to a bad cast.

### DIFF
--- a/megamek/src/megamek/server/commands/ClientServerCommand.java
+++ b/megamek/src/megamek/server/commands/ClientServerCommand.java
@@ -85,9 +85,9 @@ public abstract class ClientServerCommand extends ServerCommand {
         return server.getGameManager().getGame().getPlayer(connId).getGameMaster();
     }
 
-    private void safeParseArgumentsAndRun(int connId, String[] args) {
+    protected void safeParseArgumentsAndRun(int connId, String[] args) {
         try {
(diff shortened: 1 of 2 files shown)


Sleet01 (approved):
LGTM

Issue #5730, opened 2024-07-11
Autoconfig Ammunition provides ASF with bombs on low altitude maps


repligator:
### Environment
Developmental
Linux
Java 17

### Description
I just hit autoconfig in MM for a low altitude map and my first Thunderbird has 5 cluster bombs, 3 inferno bombs, and a thunder bomb. On average bombs seem to be equipped for about half the fighters, with the other half getting proper AA ammo.

### Files
[bombs.sav.gz](https://github.com/user-attachments/files/16182727/bombs.sav.gz)

Sleet01:
@repligator Quick q, were you going through the context menu or the Player Config dialog?
If the former, did you select all units or just the ASF?

repligator:
I went through Configure Player.

Sleet01:
It looks like this is actually a Random Map generator bug: it doesn't set its onGround state correctly unless actual maps are selected manually.

PR #6383, opened 2025-01-09
feat: damage applier now forces unit devastation and more closely pro…

Scoppio:
feat: damage applier now forces unit devastation when necessary and more closely protects meks against unintended devastation

fixes: [#5730](https://github.com/MegaMek/mekhq/issues/5730)

### DIFF
--- a/megamek/src/megamek/common/autoresolve/damage/InfantryDamageApplier.java
+++ b/megamek/src/megamek/common/autoresolve/damage/InfantryDamageApplier.java
@@ -18,10 +18,23 @@ import megamek.common.Compute;
 import megamek.common.Crew;
 import megamek.common.Infantry;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
 /**
  * @author Luana Coppio
  */
 public record InfantryDamageApplier(Infantry entity, EntityFinalState entityFinalState) implements DamageApplier<Infantry> {
+
+    @Override
+    public int getRandomHitLocation() {
+        var entity = entity();
+        if (entity instanceof BattleArmor ba) {
+            return BattleArmor.LOC_SQUAD;
+        }
+        return Infantry.LOC_INFANTRY;
+    }
 
     @Override
     public int devastateUnit() {
@@ -31,7 +44,6 @@ public record InfantryDamageApplier(Infantry entity, EntityFinalState entityFina
                 dmg += te.getArmor(BattleArmor.LOC_SQUAD);
                 dmg += te.getInternal(BattleArmor.LOC_SQUAD);
                 te.setInternal(0, BattleArmor.LOC_TROOPER_1 + i);
-
             }
         } else {
             dmg += entity().getInternal(Infantry.LOC_INFANTRY);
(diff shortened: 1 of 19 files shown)


megamek/src/megamek/common/autoresolve/damage/InfantryDamageApplier.java
```diff
@@ -14,64 +14,105 @@
...
  */
-public record InfantryDamageApplier(Infantry entity) implements DamageApplier<Infantry> {
+public record InfantryDamageApplier(Infantry entity, EntityFinalState entityFinalState) implements DamageApplier<Infantry> {
 
     @Override
-    public boolean crewMustSurvive() {
-        return false;
+    public int getRandomHitLocation() {
+        var entity = entity();
+        if (entity instanceof BattleArmor ba) {
```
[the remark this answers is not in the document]
Sleet01: I think this can just be `if (entity instanceof BattleArmor) {`, right?

megamek/src/megamek/common/autoresolve/damage/InfantryDamageApplier.java
```diff
@@ -14,64 +14,105 @@
...
+        return Infantry.LOC_INFANTRY;
     }
 
     @Override
-    public boolean entityMustSurvive() {
-        return false;
-    }
+    public int devastateUnit() {
+        int dmg = 0;
+        if (entity() instanceof BattleArmor te) {
```
Sleet01, line 42: Same here.  Actually, what happens if you ask Java if an object is an instance of another instance, does it do an identity check?

Sleet01: Oh, wait, no, I see.
This is the newer Pattern Matching construct.
IGNORE ME!

Sleet01 (changes requested):
Just the one change, removing the `BattleArmor ba` inline Pattern Matching where `ba` is not used.

Otherwise looks very good!

megamek/src/megamek/common/autoresolve/damage/InfantryDamageApplier.java
Scoppio: This is just a shorthand for class casting.

```java
public static void main() {
    Object obj = new BattleArmor();
    
    if (obj.getClass().isAssignableFrom(BattleArmor.class)) {
        BattleArmor a = (BattleArmor) obj;
        System.out.println("It is a battle armor!" + a.toString());
    }
    
    if (obj instanceof BattleArmor) {
        BattleArmor a = (BattleArmor) obj;
        System.out.println("It is a battle armor!" + a.toString());
    }
   
    if (obj instanceof BattleArmor a) {
        System.out.println("It is a battle armor!" + a.toString());
    }
   
}
```
you cant check with `instanceof` against another instantiated object, only against classes, and I believe it has to be hardcoded.

Only way I remember right now to check class of an instance against the class of another is with `obj.getClass().isAssignableFrom(obj2.getClass())`

```java

public static void main() {
    Object obj = new BattleArmor();
    
    Object obj2 = new Infantry();
    
    if (obj.getClass().isAssignableFrom(obj2.getClass())) {
        System.out.println("They kind of are the same thing!");
    }
   
}

Sleet01 (approved):
Looks good!

PR #6387, opened 2025-01-10
Data update - Support/Civilian RAT integration

SuperStucco:
This pulls the contents of the old CivilianUnits.txt RAT file(s) into the force generator system, updated to include most of the other non-combat support and civilian units available in MegaMek that were added since those files were created.

Unlike combat units the MUL is treated more loosely as there are much fewer units with these roles.  For example, the Büffel is strictly Lyran in the MUL, but is the only vehicle with either a bulldozer or minesweeper for most eras; for our purposes it is more frequent in Lyran forces with the others less frequently fielding licensed copies or Quikscell-style low-quality bootlegs (head-canon I have the operators call that the Beefalo) all using the same base unit.

Availability numbers are adjusted by era following the general trend of BT history.  For example, fusion powered support and civilian units are common during the Star League but are very uncommon during the LosTech period and do not become common until after the initial surge in military production in the mid-3050s.  IndustrialMeks are uncommon even among the best rated (A/B equipment rating) during the same periods.

Armed IndustrialMeks fall into two categories.  Most of them, including the MOD designs, are considered military and will generate as part of the worst equipped units (D/F equipment rating) for Periphery forces in general and slightly higher frequency for pirates and Bandit Caste.  They will also appear in IS and mercenary forces during the WoB Jihad and later periods.  The rest, generally those designated as PoliceMeks, have the CIVILIAN role and be considered general law and order/public security forces.

The smaller number of overall units in each unit type (Mek, tank, VTOL, conventional fixed wing) means the availability numbers are balanced against using all weight classes rather than a specific one.  This means they are best generated using with either no weight class argument or all of the weight classes.  There may be cases where an empty table is generated as no units/no suitable units are available, so random generation should include a fallback if generating a unit is mandatory.

For now I've excluded adding more WiGE units as Princess has little to no idea on how to move them.  This is limited to a very small number of large cargo carriers which aren't very common.

RAT tables are still being affected by mixed weight class and Clan/SL/Omni percentage adjustments.  This particularly affects Clan IndustrialMek generation due to their high OmniMek and Clan percentages, which can lead to unexpected results.  I'll be dealing with that issue shortly in a separate PR.


HammerGS (approved):

HammerGS:
Thanks for all the work on this!

HammerGS:
@IllianiCBT We'll need to figure out the code support needed.

PR #6397, opened 2025-01-11
Fix 6395: Add knowledge of smoke zero damage to princess

### LINKED ISSUE #6395: Princess doesn't know that Smoke munitions do zero damage.
We've discovered that the reason Princess currently flings Smoke shots around willy-nilly is that she thinks they do the regular amount of damage (for whatever weapon fires them).

She should be aware that Smoke munitions do zero damage so she does not include them in her attack planning.

Sleet01:
Have the `WeaponFireInfo.computeExpectedDamage()` method return 0 if the associated munition is Smoke, Flare, or FASCAM/Thunder type.
This should give Princess a better idea of the utility of some munitions when planning turns and attacks, so she chooses them less often.
Similarly, explicitly set the damage value of Smoke, Flare, or FASCAM/Thunder artillery munitions to 0 for Artillery fire control computations.

We can either replace this with a generalized Smoke or Utility Munition planner, or implement that at a higher level, but for now we don't want Princess to treat zero-damage rounds as if they do damage to targets.

This also fixes an NPE path in minefield handling, when artillery-deployed minefields are used in counter-battery attacks.
Unfortunately it looks like all counter-battery attacks currently compute an expected damage of 0.0, possibly due to the targets being off-board, so Princess still chooses munitions for these attacks randomly.

Additionally, add trailing spaces to the Glancing Blow and Direct Blow messages so that they produce lines like:

`- Glancing Blow - 6 missile(s) hit`

rather than

`- Glancing Blow -6 missile(s) hit`

which can be confusing.

Testing:
- Ran several Princess-on-Princess battles with Smoke (and other) munitions equipped.

Close #6395

### DIFF
--- a/megamek/i18n/megamek/common/report-messages.properties
+++ b/megamek/i18n/megamek/common/report-messages.properties
@@ -348,11 +348,11 @@
 3181=Succeeds.
 3182=fails, subjecting the unit to particle feedback!
 3185=<newline>    (continuing hit report):
-3186=\ - Glancing Blow -
+3186=\ - Glancing Blow -\ 
 9985=\ - Glancing Blow due to Narrow/Low Profile -
 3187=does nothing (didn't target a unit).
 3188=<span class='success'><B>hits</B></span>, target tagged.
-3189=\ - Direct Blow -
+3189=\ - Direct Blow -\ 
 3190=<span class='success'><B>hits</B></span> the intended hex <data>.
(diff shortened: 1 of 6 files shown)


Sleet01:
Converted to draft; I'm also adding Thunder, FASCAM, maybe a couple others as well

HammerGS (approved):
Checked with Co-Pilot and IDEA AI

Issue #5760, opened 2024-07-21
Yellow Jacket Gunship (PPC) BV Calculation Incorrect


mrcglaisek:
### Environment
HQ 0.49.20
Windows
Java 17.0.1

### Description
The Yellow Jacket Gunship (PPC) variant shows a BV of 758, but on Master Unit List it is listed as 582. Other variants seem to be correctly valued. This appears to be due to using unpublished errata of the sourcebook which increases the movement of this variant.

HammerGS:
Moving to the Megamek tracker. This should be referred to the MUL team as it's likely using out dated calculations.

SJuliez:
The higher speed is according to TRO 3085 p.190 and there doesn't appear to be a MM issue here. Therefore closing this.

PR #6405, opened 2025-01-13
Issue 6385: Incorrect TMM bonus for WiGe

### LINKED ISSUE #6385: [Issue] WiGE getting a TMM bonus it shouldn't
### Brief Description *

I noticed that when moving units that use WiGE rules, they got an additional bonus to their TMM they are not supposed to have (underlined in red below). 
WiGE TMM
They should not be getting this bonus, as they are WiGEs and not VTOLs.

### 3. Steps to Reproduce

1. Move your WiGE unit.
2. Look at the TMM modifier (for a more verbose breakdown, try firing at it either by using another player or friendly fire).

### Attach Files

_No response_

### Severity *

### MegaMek Suite Version *

0.50.02

### Operating System *

Windows 10

### Java Version *

Java Vendor: Oracle Corporation     Java Version: 17.0.12

### Final Verification

psikomonkie:
Fixes #6385 

The TMM bonus for jumping, being a VTOL, or being an airborne WiGe are all the same bonus - TW.  pg 117 Jumped/Airborne (non-aerospace units) get +1 to hit modifier, 

Instead of checking and applying this bonus in different locations, the bonus is now applied in 1 location. I checked every usage of public static ToHitData getTargetMovementModifier(int, boolean, boolean, Game) to ensure it's properly utilizing the method. 

I updated JavaDocs and comments to ensure this is conveyed in the code to future contributors.

I verified in game that this is applying the TMM bonus when we want it, and it's not applying it twice.

I verified the BV of WiGes, VTOLs, and glider protomeks did not change. I noticed during this that they don't match the MUL already. I calculated the BV of the Swallow & Fensalir by hand using the TechManual and the BV I got matches what MegaMek has, not what the MUL has - so either the MUL is wrong, or there is a rule for WiGe BV that's not conveyed in the TechManual. I have a hunch it's the same root cause as #5760 but I didn't dig too deep into this. I think this should be treated as a separate issue. 

Finally, I fixed an additional bug: On the map, a unit's TMM shows for its last step as you're selecting where to move. This was not accounting for VTOL/WiGe flight - now it does. I couldn't find a existing issue for this, but I didn't look too hard.

### DIFF
--- a/megamek/src/megamek/client/bot/princess/FireControl.java
+++ b/megamek/src/megamek/client/bot/princess/FireControl.java
@@ -256,16 +256,16 @@ public class FireControl {
      *
      * @param hexesMoved The number of hexes the target unit moved.
      * @param jumping    Set TRUE if the target jumped.
-     * @param vtol       Set TRUE if the target is a {@link VTOL}.
+     * @param airborneNonAerospace       Set TRUE if the target is a {@link VTOL} or other airborne, non-aerospace unit.
      * @param game       The current {@link Game}
      * @return The target movement modifier as a {@link ToHitData} object.
      */
     @StaticWrapper()
     protected ToHitData getTargetMovementModifier(final int hexesMoved,
             final boolean jumping,
-            final boolean vtol,
+            final boolean airborneNonAerospace,
             final Game game) {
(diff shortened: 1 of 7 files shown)


HammerGS (approved):
Checked with Co-Pilot and IDEA AI.

PR #6415, opened 2025-01-15
fix: fixes an error where force consolidation would not set the correct owner of the force

Scoppio:
fix: fixes an error where force consolidation would not set the correct owner of the force


megamek/src/megamek/common/autoresolve/converter/ForceConsolidation.java
```diff
@@ -32,8 +32,6 @@
  */
```
Sleet01, line 32: Is this comment still accurate?

Sleet01 (approved):
One question about the docs given the refactoring of `ForceConsolidation.java`, but this is all still a work in progress and the docs don't impact functionality.

Looks good!

PR #6420, opened 2025-01-18
Fluff Hex and Minor Map Adjustments

x-Lanceman-x:
Enables some fluff art for railroad crossings that were in the folder already but never assigned in a tile set, just to give folks some more options. 

Also updating a couple of maps in my folder to better use the improved unpaved road art and fixing a silly typo in a board name.


HammerGS:
Just want to confirm that we don't need to add this fluff to other tileset.

stonewall072:
We shouldn't. Where saxarba uses it, it's part of the "StandardIncludes" so it's a tileset referenced within a tileset. Which should also catch it for hq_saxarba and any other theme that references the existing rail art in their tilesets.

Screenshot 2025-01-18 112741

Sleet01 (approved):
LGTM.

PR #6425, opened 2025-01-19
Issue 6390: Units switching from multi-hex to single-hex and vice-versa properly update the entityPosLookup cache

### LINKED ISSUE #6390: [Issue] Hotdrop units can land in a landing dropship's hex, causing both to get stuck
### Brief Description *

There is a chance when hot dropping near a hex which a dropship is landing in for the mech to land in that hex and become fused with the dropship, restricting the movement of both the dropship and mech. When this does happen, the dropship and 'mech become unable to move and are stuck for the rest of the game.
image
image
My guess for this behavior is that when the calculation for landing is done, it doesn't see the dropship's hex as occupied and so it lands the 'mech there and gets it stuck.

### 3. Steps to Reproduce

1. Load the provided save file 2 (it is a union surrounded by hot-dropping 'mechs). If you want to see where the 'mechs came from, look at #6389.
2. Have the dropship do a vertical land and end movement.
3. There is a chance at the end of movement that a 'mech lands in a hex that a dropship occupies (look at save file 3).

### Attach Files

[2.sav.gz](https://github.com/user-attachments/files/18384645/2.sav.gz)
[3.sav.gz](https://github.com/user-attachments/files/18384657/3.sav.gz)

### Severity *

### MegaMek Suite Version *

0.50.02

### Operating System *

Windows 10

### Java Version *

Java Vendor: Oracle Corporation     Java Version: 17.0.12

### Final Verification

psikomonkie:
Fixes #6390 

What a strange issue! When a Dropship changes from landed to flying, which switches it from multi-to-single hex, and vice-versa, the entityPosLookup "cache" inside of Game was improperly updating. Normally, getting a unit's currently occupied hexes, settings its new position, then getting the occupied hexes again will be accurate - a new occupied hex means the cache needs changed. 

When a Spheroid Dropship lands vertically the landing takes place before set position. When SetPosition executes, it will get the current occupied hexes and improperly consider the Dropship's position as not having changed. This fixes that by getting the "oldPositions" directly from the entityPosLookup "cache".


megamek/src/megamek/common/Dropship.java
```diff
@@ -498,7 +498,14 @@ public int getWalkMP(MPCalculationSetting mpCalculationSetting) {
     @Override
 
     public void setPosition(Coords position) {
+        // When a Dropship changes from being 1 hex to 7 getOccupiedCoords will return
+        // its changed secondary hexes. Instead, let's grab the cached ones for this to
+        // make sure we properly set out new positions.
         HashSet<Coords> oldPositions = getOccupiedCoords();
```
HoneySkull, line 504: What is the intent of keeping this call to `getOccupiedCoords()`?  It looks like you are setting `oldPositions` immediately after this assignment.  Should this be removed?

psikomonkie: It's a fallback in case the game is null. The method used to work this way, so I figured this was the best way to handle that.

Sleet01 (approved):
LGTM.

PR #6431, opened 2025-01-21
Fix Artemis derivatives getting bonus on indirect fire

Sleet01:
Problem: Artemis controllers _other_ than Artemis IV are getting their bonuses when firing in Indirect mode; this is incorrect:

> per BMM pg 30 Artemis FCS do not confer bonuses when fired indirectly

and BMM page 110, *ARTEMIS V FIRE CONTROL SYSTEM*:
> Indirect fire attacks do not receive these bonuses.

Solution:  move the existing not-indirect check for Artemis IV out to cover _all_ advanced FCS bonus checks within the LRM attack handler (other missile types are not subject to this check):
- Artemis IV
- Artemis V
- Prototype Artemis
- ATM
- NARC-capable missiles

Testing:
- Ran all the projects' unit tests
- Tested indirect fire with Artemis V FCS mek

### DIFF
--- a/megamek/src/megamek/common/actions/WeaponAttackAction.java
+++ b/megamek/src/megamek/common/actions/WeaponAttackAction.java
@@ -3352,6 +3352,7 @@ public class WeaponAttackAction extends AbstractAttackAction {
         }
 
         // Missile Munitions
+        boolean isIndirect = weapon.hasModes() && (weapon.curMode().isIndirect());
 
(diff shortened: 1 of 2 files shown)


Sleet01:
Need to track down the separate Artemis V FCS -1 to-hit bonus...

HoneySkull (approved):

PR #6432, opened 2025-01-22
Fix/princess should favor moving

Scoppio:
Currently Princess is "stuck", if she does not have something drawing her directly into battle, she will instead huddle up with all her units into a blob of death, and they will wiggle their way accross the map very slowly (less than 1 hex per turn).

Turns out that when there are no enemies in sight, the utility cost of all her MovePaths become overhelmed by the herding utility cost, which will make all movements that makes the unit land as close as possible to the center of the formation rise up to the top of best movements.

To counter that, more utility costs need to share the burden when there are very few things.

This change does the following:

- adds movement modifier: moving a large number of hexes and traversing a large distance are both positive things, so they are now being counted positively, it also counts the possible TMM the unit will gain as a multiplier.
- changes the arraylist of the "ranked paths" for a treeset, the tree set automatically sorts every ranked path that is added, so to pick the best you just have to grab the first.
- add a try catch inside the for-each loop that ranks every valid path, this turns out was a good thing because allowed me to find a couple of bugs and fix
- some parts of the code were changed because they were using exceptions as flow control, which is bad in JVM because it is a very slow process. Whenever you have an exception the stack is copied, the code has to exit, etc... and because those errors were in the hot-path, this improves a little bit the performance.


megamek/src/megamek/common/event/GameEntityChangeEvent.java
```diff
@@ -79,26 +79,22 @@ public Vector<UnitLocation> getMovePath() {
...
-                return getEntity().toString() + " probably deployed.";
-            }
-        } catch (NullPointerException npe) {
+        var entity = getEntity();
+        var movePathIsEmpty = movePath.isEmpty();
+        if (entity != null && !movePathIsEmpty && movePath.lastElement().getCoords() != null) {
+            return entity + " moved to " + movePath.lastElement().getCoords().toFriendlyString();
+        } else if (entity != null) {
+            return entity + " probably deployed.";
+        } else {
```
Scoppio, line 95: way cheaper to just checl if things are null or empty before texploding.

megamek/src/megamek/common/Coords.java
```diff
@@ -79,9 +79,9 @@ public Coords translated(int dir, int distance) {
     public Coords translated(String dir) {
         int intDir = 0;
 
-        try {
+        if (Character.isDigit(dir.charAt(0))) {
             intDir = Integer.parseInt(dir);
-        } catch (NumberFormatException nfe) {
+        } else {
```
Scoppio, line 84: a case of "exception oriented programming

Scoppio (commented):
basic stuff that I found

megamek/src/megamek/client/bot/princess/PathRanker.java
```diff
@@ -197,7 +196,7 @@ private List<MovePath> validatePaths(List<MovePath> startingPathList, Game game,
                 // Skip this part if I'm an aero on the ground map, as it's kind of irrelevant
                 // also skip this part if I'm attempting to retreat, as engagement is not the
                 // point here
-                if (!isAirborneAeroOnGroundMap && !getOwner().wantsToFallBack(mover)) {
+                if (!isAirborneAeroOnGroundMap && !getOwner().wantsToFallBack(mover) && !hasNoEnemyAvailable) {
```
Scoppio, line 199: added here because its useless to look for enemies again if there are no enemies around

megamek/src/megamek/client/bot/princess/PathRanker.java
```diff
@@ -250,8 +249,8 @@ private List<MovePath> validatePaths(List<MovePath> startingPathList, Game game,
      * @return "Best" out of those paths
      */
     @Override
-    public @Nullable RankedPath getBestPath(List<RankedPath> ps) {
-        return ps.isEmpty() ? null : Collections.max(ps);
+    public @Nullable RankedPath getBestPath(TreeSet<RankedPath> ps) {
+        return ps.isEmpty() ? null : ps.first();
```
Scoppio, line 253: TreeSet is ordered, this is implemented with reversed ordering so the first element is always the best ranked movepath

megamek/src/megamek/client/bot/princess/PathRanker.java
```diff
@@ -536,11 +535,8 @@ private boolean willBuildingCollapse(MovePath path, Game game) {
     }
 
     public static @Nullable Coords calcAllyCenter(int myId, @Nullable List<Entity> friends, Game game) {
-        if ((friends == null) || friends.isEmpty()) {
+        if ((friends == null) || friends.size() <= 1) {
```
Scoppio, line 538: units alone shuold not receive back a "allyCenter", if they have, they will "fight" against the location they currently are, so they won't move around.

megamek/src/megamek/client/bot/princess/RankedPath.java
```diff
@@ -72,13 +72,7 @@ public int compareTo(RankedPath p) {
...
             return 1;
         }
-        if (path.getKey().hashCode() < p.path.getKey().hashCode()) {
-            return -1;
-        }
-        if (path.getKey().hashCode() > p.path.getKey().hashCode()) {
-            return 1;
-        }
-        return 0;
+        return Integer.compare(path.getHexesMoved(), p.path.getHexesMoved());
```
Scoppio, line 75: I added the `hexes moved` as a second level of compare to break ties.

Scoppio (commented):
.

megamek/src/megamek/common/event/GameEntityChangeEvent.java
```diff
@@ -79,26 +79,22 @@
...
-        try {
-            return getEntity().toString() + " moved to "
-                    + movePath.lastElement().getCoords().toFriendlyString();
-        } catch (NoSuchElementException nsee) {
-            if (getEntity() != null) {
-                return getEntity().toString() + " probably deployed.";
-            }
-        } catch (NullPointerException npe) {
+        var entity = getEntity();
+        var movePathIsEmpty = movePath == null || movePath.isEmpty();
```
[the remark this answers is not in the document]
Scoppio: solved

IllianiCBT:
I've flagged this as for new dev cycle (50.04). Any dev can feel free to ignore it and merge away, I'm just wary of big Princess changes so close to launch. Especially after the 'Princess forgot how to shoot' bug a few versions back.

megamek/src/megamek/client/bot/princess/PathRanker.java
NickAragua: Does "friends" include the currently moving unit?

megamek/src/megamek/client/bot/princess/RankedPath.java
NickAragua: Hexes moved seems like the better option in the general case, as that drives TMM.

megamek/src/megamek/common/Coords.java
NickAragua: I bet that improves performance a bit.

megamek/src/megamek/common/MovePath.java
```diff
@@ -1341,10 +1369,10 @@ public boolean isMoveLegal() {
         }
 
         // for aero units move must use up all their velocity
-        if (getEntity().isAero()) {
-            IAero a = (IAero) getEntity();
+        // but only if it is actually IAero, because anything could return isAero() == true but not implement IAero
+        if (getEntity().isAero() && getEntity() instanceof IAero aero) {
```
NickAragua, line 1373: Do we know what returns isAero() but doesn't implement it? Feels like that's an oversight.

NickAragua (commented):
Looks good, just a couple of questions.

megamek/src/megamek/common/Coords.java
Scoppio: it is, removes the need for dumping the whole stack and memory.

megamek/src/megamek/client/bot/princess/PathRanker.java
Scoppio: it does, for absolutelly no good reason other than "we didnt took our time to filter out"

megamek/src/megamek/common/MovePath.java
Scoppio: I dont know, but I also dont want to risk casting to an interface without checking if it is castable

NickAragua (approved):

IllianiCBT:
@HammerGS I know I flagged this as for new cycle, but is it one of the features you want in 50.03? I know you're keen on the AI adjustments and with Nick having approved it I'm a lot less nervous.

PR #6433, opened 2025-01-24
Princess - Chat Commands fix

Scoppio:
The chat commands syntax and how they were used in the chat were wrong and did not align with how they actually work.

This not only fix that, as it also fix an error that I commited where I mistakenly believed that you could set the bot behavior using discrete values, this fix also allows that.


Sleet01 (approved):
LGTM.

Sleet01 (changes requested):
Failing tests need fixed now.

Sleet01 (approved):
Looks good!

PR #6436, opened 2025-01-24
feat: minimap preview and selection option

Scoppio:
Adds minimap theme selector with preview and 18 themes

https://github.com/user-attachments/assets/59611eb6-82d4-4473-9333-02ea1efbef76


megamek/src/megamek/client/ui/baseComponents/FileNameComboBoxModel.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @Override
+    public int getSize() {
+        return files.size();
+    }
+
+    @Override
+    public String getElementAt(int index) {
+        return files.get(index).getName();
```
HoneySkull, line 73: Does this need bounds checking?  Or should this method throw an exception that will occur if `index` is out of bounds?  (given this is a public method)

megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
```diff
@@ -1609,14 +1617,48 @@ private JPanel getOverlaysPanel() {
         return createSettingsPanel(comps);
     }
 
+    private BufferedImage boardImage;
+    private JLabel boardImageLabel;
```
HoneySkull, line 1621: If these are only referenced from `getMiniMapPanel()`, could these variables be local scope to the method instead of class variables?  If not, do they need to be initialized?

megamek/src/megamek/common/preference/ClientPreferences.java
```diff
@@ -108,6 +116,7 @@ public ClientPreferences(IPreferenceStore store) {
         store.setDefault(START_SEARCHLIGHTS_ON, true);
         store.setDefault(USER_DIR, "");
         store.setDefault(MML_PATH, "");
+        store.setDefault(MML_PATH, "");
```
HoneySkull, line 119: Just checking if this is a type - as it's a duplicate statement... should this be a different path here?  Should it be MML_PATH?

HoneySkull (commented):
Added a few nit-picky questions here.

megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
Scoppio: No, they cant, they have to be local to the dialog, otherwise the image wont ever be updated, as the reference to the object in the UI will be lost once you redesignate the local variable (the local  variable would have to be declared final, which makes it impossible to do what I am doing)

megamek/src/megamek/client/ui/baseComponents/FileNameComboBoxModel.java
Scoppio: Minor concern, you really should be greeted by a nullpointer exception in this situation.

megamek/src/megamek/client/ui/swing/CommonSettingsDialog.java
HoneySkull: Got it.  I just didn't see it being referenced anywhere, but this makes sense.

HoneySkull (approved):

PR #6438, opened 2025-01-25
Issue 1072: Can bomb and strafe the ground even if no LOS

### LINKED ISSUE #1072: Unable to "fire" energy weapons during Strafe.
### Environment
Megamek 0.44.0
MacOS 10.13.5
Java 8 Update 171

### Description
After setting up a Strafing run with Slayer SL-15 from Altitude 3 with Velocity 2 (attached), I'm able to select the hexes to strafe, but am unable to use "fire" button for the energy weapons to actually initiate the strafe. I'm wondering if this has something to do with planetary conditions (pitch black)?

### Files
[cantstrafe.sav.gz](https://github.com/MegaMek/megamek/files/2249696/cantstrafe.sav.gz)

psikomonkie:
Fixes #1072

Includes being able to altitude & dive bomb in low-vis conditions. 

I started on some unit tests to cover this and it's not going great. If this gets merged before I finish the unit tests, they'll come as a standalone PR (or with the next related bug, lol).

### DIFF
--- a/megamek/src/megamek/common/Compute.java
+++ b/megamek/src/megamek/common/Compute.java
@@ -1037,8 +1037,8 @@ public class Compute {
                 // If the target isn't spotted, can't target
                 if (game.getOptions().booleanOption(OptionsConstants.ADVANCED_DOUBLE_BLIND)
                         && ((!Compute.inVisualRange(game, los, other, target)
-                            && !Compute.inSensorRange(game, los, other, target, null))
-                        || (!target.isHexBeingBombed()))) {
+                            && !Compute.inSensorRange(game, los, other, target, null)
+                            && !target.isHexBeingBombed()))) {
                     mods.addModifier(TargetRoll.IMPOSSIBLE,
                             "outside of visual and sensor range");
                 }
--- a/megamek/src/megamek/common/Compute.java
+++ b/megamek/src/megamek/common/Compute.java
@@ -1036,8 +1036,9 @@ public class Compute {
                 ToHitData mods = los.losModifiers(game);
                 // If the target isn't spotted, can't target
                 if (game.getOptions().booleanOption(OptionsConstants.ADVANCED_DOUBLE_BLIND)
-                        && !Compute.inVisualRange(game, los, other, target)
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/common/Compute.java
```diff
@@ -1036,8 +1036,9 @@ public static Entity findSpotter(Game game, Entity attacker,
                 ToHitData mods = los.losModifiers(game);
                 // If the target isn't spotted, can't target
                 if (game.getOptions().booleanOption(OptionsConstants.ADVANCED_DOUBLE_BLIND)
-                        && !Compute.inVisualRange(game, los, other, target)
-                        && !Compute.inSensorRange(game, los, other, target, null)) {
+                        && ((!Compute.inVisualRange(game, los, other, target)
+                            && !Compute.inSensorRange(game, los, other, target, null))
+                        || (!target.isHexBeingBombed()))) {
```
HoneySkull, line 1041: This 'OR' here seems suspicious to me - it's going to short circuit the logic check here to return impossible anytime the target hex isn't being bombed?!?  Was this teste with non ASF scenarios?

HoneySkull (commented):
Question about the "OR" in the logic here for side-effects.

megamek/src/megamek/common/Compute.java
psikomonkie: Yeah it was just wrong, thanks HoneySkull.

psikomonkie:
> Question about the "OR" in the logic here for side-effects.

I fixed the issue you found but I've set this to draft while I do some more testing.

HoneySkull:
Thanks, I pulled the code and tried doing some standard indirect fire... I setup a match with an Archer and Locust against an Urbie on rolling hills map... the Indirect fire works because the method `findSpotter()` still returns the locust as a spotter... but if you have multiple spotters... it will likely pick the last or first one treating them as if they have the same toHit.

psikomonkie:
Removed some unnecessary (incorrect) changes and did some more thorough testing, including more with ground units. Thanks @HoneySkull for catching that issue.

HoneySkull (approved):

PR #6442, opened 2025-01-26
Unit Tests for WeaponAttackAction to help test megamek#1072

psikomonkie:
I mentioned I was working on unit tests in #6438 - I really was, and I got them working. Hopefully this can be expanded as we encounter and fix similar issues to #1072.


Sleet01 (approved):
Ship it, ship it now!

PR #6443, opened 2025-01-26
fix: ranked path fix

Scoppio:
# The bug
Due to an oversight of my part, I let it pass over me that when adding an object to a treeset, it may replace the object that is already present in the tree if the result of the comparison is equal to 0.
The consequence of this mistake is that if there were two path with the same rank, and same hexes moved, it would then substitute one of the for the other, which would cull the paths during insertion in the rankedPath list, unfortunatelly this does not provide much better performance, so there is no upside to this bug and it has to be addressed.

# This fix
This change makes that no two movepaths will never return 0 in the comparsion. It also implements a hashCode and equals for RankedPath and MovePath.

### DIFF
--- a/megamek/src/megamek/client/bot/princess/RankedPath.java
+++ b/megamek/src/megamek/client/bot/princess/RankedPath.java
@@ -20,6 +20,8 @@
 package megamek.client.bot.princess;
 
 import megamek.common.MovePath;
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
 
 import java.text.DecimalFormat;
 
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/client/bot/princess/RankedPath.java
```diff
@@ -72,42 +74,43 @@ public int compareTo(RankedPath p) {
...
         }
-        if (!path.equals(that.path)) {
-            return false;
+        if (expectedDamage < p.expectedDamage) {
+            return -1;
         }
-        if (reason != null ? !reason.equals(that.reason) : that.reason != null) {
-            return false;
+        if (p.expectedDamage < expectedDamage) {
+            return 1;
```
Sleet01, line 87: Do we store the hazard values anywhere in the RankedPath?  
If so, we should probably compare those as well; if not, we might want to take a look at it for a later update.

Sleet01 (approved):
Looks good!

megamek/src/megamek/client/bot/princess/RankedPath.java
Scoppio: Thats a good idea, will take a look into it later

PR #6450, opened 2025-01-28
Allow the print-unit-list code to run with MegaMekLab.jar

pavelbraginskiy:
Previously, when selecting a path to MegaMekLab, it had to be MegaMekLab.sh or MegaMekLab.exe (i.e., an executable file).

This allows you to select MegaMekLab.jar as well, and this should be what MegaMek tries to autodetect on Mac.

### DIFF
--- a/megamek/src/megamek/client/ui/swing/ClientGUI.java
+++ b/megamek/src/megamek/client/ui/swing/ClientGUI.java
@@ -39,6 +39,7 @@ import java.io.OutputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.*;
+import java.util.jar.JarFile;
 
 import javax.imageio.ImageIO;
 import javax.swing.*;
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/client/ui/swing/ClientGUI.java
```diff
@@ -2119,6 +2121,79 @@ public void saveListFile(ArrayList<Entity> unitList, String filename) {
...
+                if (autodetected) {
+                    logger.error("Could not auto-detect MegaMekLab! Please configure the path to the MegaMekLab executable in the settings.", "Error printing unit list");
+                } else {
+                    logger.error("%s does not appear to exist! Please configure the path to the MegaMekLab executable in the settings.".formatted(mmlExecutable.getName()), "Error printing unit list");
+                }
+                return null;
+            }
+
+            // The executable is a jarfile, so let's execute it.
+            var javaExecutable = ProcessHandle.current().info().command().orElse("java");
```
Sleet01, line 2181: I'm not a huge fan of this approach; if anything replaces the MML jar file with malicious Java, this will execute it without any warning.

Although, to be fair, the same thing would happen if the MML jar file were replaced with malicious Java and the user just double-clicked it or one of the batch files.

Given that MegaMek is the core program, I'd feel much better about moving low-level printing functionality to MM and then importing it into MML, rather than running MML as a sub-process.

That said, this doesn't seem to expose us to any more security risks, and is much more compact than the full re-write would be.

Sleet01 (approved):
LGTM.

megamek/src/megamek/client/ui/swing/ClientGUI.java
pavelbraginskiy: If MegaMekLab.jar is replaced with a compromised version, you're completely screwed anyway. There's no reason to suspect that MegaMek.jar wouldn't also be compromised in such a scenario.

PR #6457, opened 2025-01-29
Fix 5870: princess engaging targets outside visual range in heavy fog

### LINKED ISSUE #5870: [0.50.0-SNAPSHOT] Princess engaging targets outside of visual range in heavy fog.
### Environment
04:32:57,538 INFO  [megamek.MegaMek] {main}
megamek.MegaMek.initializeLogging(MegaMek.java:143) - Starting MegaMek v0.50.0-SNAPSHOT
    Build Date: 2024-08-04T01:23:25.882857488
    Today: 2024-08-07
    Origin Project: MekHQ
    Java Vendor: Eclipse Adoptium
    Java Version: 21.0.3
    System Locale: en_GB
    MM Code Revision: 0f0d10d35f6bbc190c42be24563140865de228c5
    MML Code Revision: 35aa5ce66ca59272c0d1e05c3d174cd4c0335b3c
    MHQ Code Revision: 76e9cea125956ab9a9d8af389ee60c2bb2eca81d
    
### Description
image

In the attached imagine, I have a Lance deployed with double-blind and sensor rules in operation. I also specifically do _not_ have Princess on single blind. The heavy fog weather modifier is in effect. Princess is able to fire on my units beyond visual range, seemignly ingoring the hard visual limits imposed by the fog. In fact,in my screenshot, you can even see that the black Warhammer being shot is displayed as semi-translucent, suggesting it should not be visible to Princess at all. 

This is more noticeable as ranges decrease, since if the shots miss from beyind visual range, they do not appear in the round report. 
I forced a /victory directy after this screenshot, and attached the gamelog - you can see repeated volleys being fired at my formation when they are in the long range and medium range bands. 
I have attached a quicksave of this round, as well as the one custom unit included that will be required to load this savegame. 

This can easily be replicated by having the lance remain at the edge of visual range, and monitor incoming fire. 

### Files
[megamek.log](https://github.com/user-attachments/files/16518292/megamek.log)
[gamelog.zip](https://github.com/user-attachments/files/16518310/gamelog.zip)
[quicksave.sav.gz](https://github.com/user-attachments/files/16518316/quicksave.sav.gz)
[Lancelot LNC25-01-Gr &quot;Grailsworn&quot;.zip](https://github.com/user-attachments/files/16518317/Lancelot.LNC25-01-Gr.quot.Grailsworn.quot.zip)

Sleet01:
Please see #5870 for analysis of the root cause.

This fix lets Princess use knowledge of the current visual range when guessing at attack effectiveness for movement planning.

It also removes the ability for C3-equipped units to completely bypass LOS restrictions as long as one C3 network member has a target within their sensor ranges.
Based on forum posts and TacOps, it looks like at least one member of the network needs to have _visual_ LOS, including any Planetary Conditions that restrict visual range, for the network to provide targeting capabilities to its members.  Otherwise it's just sharing the already-detected sensor pings.

Testing:
- Tested scenario from the linked issue repeatedly, with and without fixes.
- Updated FireControlTest to provide the necessary planetary environment data that Princess now checks.

Close #5870

### DIFF
--- a/megamek/src/megamek/client/bot/princess/FireControl.java
+++ b/megamek/src/megamek/client/bot/princess/FireControl.java
@@ -163,6 +163,7 @@ public class FireControl {
     static final TargetRollModifier TH_SWARM_STOPPED = new TargetRollModifier(TargetRoll.AUTOMATIC_SUCCESS,
             "stops swarming");
     static final TargetRollModifier TH_OUT_OF_RANGE = new TargetRollModifier(TargetRoll.IMPOSSIBLE, "out of range");
+    static final TargetRollModifier TH_OUT_OF_VISUAL = new TargetRollModifier(TargetRoll.IMPOSSIBLE, "out of visual targeting range");
     static final TargetRollModifier TH_SHORT_RANGE = new TargetRollModifier(0, "Short Range");
     static final TargetRollModifier TH_MEDIUM_RANGE = new TargetRollModifier(2, "Medium Range");
     static final TargetRollModifier TH_LONG_RANGE = new TargetRollModifier(4, "Long Range");
@@ -303,6 +304,11 @@ public class FireControl {
         final int maxRange = owner.getMaxWeaponRange(shooter, target.isAirborne());
         if (distance > maxRange) {
             return new ToHitData(TH_RNG_TOO_FAR);
+        }
+
+        final int maxVisRange = game.getPlanetaryConditions().getVisualRange(shooter, shooter.isUsingSearchlight());
+        if (distance > maxVisRange) {
+            return new ToHitData(TH_OUT_OF_VISUAL);
         }
 
         final ToHitData toHitData = new ToHitData();
--- a/megamek/src/megamek/common/actions/WeaponAttackAction.java
+++ b/megamek/src/megamek/common/actions/WeaponAttackAction.java
@@ -1380,7 +1380,9 @@ public class WeaponAttackAction extends AbstractAttackAction {
                 // c3 units can fire if any other unit in their network is in
                 // visual or sensor range
                 for (Entity en : game.getEntitiesVector()) {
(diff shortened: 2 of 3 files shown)


megamek/src/megamek/client/bot/princess/FireControl.java
```diff
@@ -305,6 +306,11 @@ ToHitData guessToHitModifierHelperForAnyAttack(final Entity shooter,
             return new ToHitData(TH_RNG_TOO_FAR);
         }
 
+        final int maxVisRange = game.getPlanetaryConditions().getVisualRange(shooter, shooter.isUsingSearchlight());
```
psikomonkie, line 309: Will Princess still be able to consider Entities illuminated by a different source as valid - like a different Mek with a searchlight illuminating the enemy?

psikomonkie (commented):
Nice!! I think someone will at some point want to add low-vis conditions this to Princess's fire control unit tests (I like double blind so I might).

Re: C3 & WAA changes: I was thinking of expanding the WAA JUnit tests I added last week during 50.04 if it really is a “stability” release - covering C3 in the manner you described would be a good test to add.

rjhancock (approved):
Looks clean.

megamek/src/megamek/client/bot/princess/FireControl.java
Sleet01: She will, but for the fast approximation path she's only going to consider the current unit's searchlight state and any cached illumination state on target units.

This path is used when planning where to move a unit, not just for the moving unit but for predicting the likely actions of all known enemies and allies as well.
So I wanted to add the planetary environment check but keep it as minimal as possible to avoid introducing a big slowdown.

The attack planning code switches to using the same to-hit calc code as player sees when using the GUI; there should be fewer options at that point, so it's fast enough.  
And that contains the full illumination check code.

PR #6459, opened 2025-01-30
feat: quick and simple hazard check for hex

Scoppio:
previous code did what "hashset.retainAll" does as standard.
Instead of instantiating one new array every time a hex is checked for hazards, it uses two default sets for hazard terrain types and choses one or the other to retain against the terrain types of the hex.


HoneySkull (approved):

Sleet01 (changes requested):
Please address the failing tests so that this can be merged.

Sleet01 (approved):
LGTM!

PR #6462, opened 2025-01-30
From StringBuilder to Logger

Scoppio:
Using logging trace level for all this stuff instead of debug because it really is a level lower than debug what it is doing.
Kept the reasoning in the RankedPath as it is a great tool to understand what happened to select that object, but other than that, not much was changed in code.


megamek/src/megamek/client/bot/princess/BasicPathRanker.java
```diff
@@ -1110,89 +1099,81 @@ private double calcWaterHazard(Entity movingUnit, Hex hex, MoveStep step, MovePa
...
-                    ProtoMek.LOC_HEAD == loc ||
-                    ProtoMek.LOC_TORSO == loc ||
-                    (!(movingUnit instanceof Mek) &&
-                            !(movingUnit instanceof ProtoMek))) {
-                logMsg.append(" breached and critical (1000).");
+            if ((Mek.LOC_HEAD == loc) ||
+                (Mek.LOC_CT == loc) ||
+                (ProtoMek.LOC_HEAD == loc) ||
+                (ProtoMek.LOC_TORSO == loc) ||
+                !(movingUnit.isMek() && !(movingUnit.isProtoMek()))
```
Sleet01, line 1119: I believe the first `!` must go inside the parentheses; this new construction will evaluate to `true` if the unit is a ProtoMek, but previously it would only evaluate to `true` if the unit is not a Mek _and_ not a ProtoMek.

Sleet01 (changes requested):
One update needed, otherwise looks great!

megamek/src/megamek/client/bot/princess/BasicPathRanker.java
Scoppio, line 1119:
```suggestion
                (!movingUnit.isMek() && !movingUnit.isProtoMek())
```

Sleet01 (approved):
LGTM!

PR #6463, opened 2025-01-30
feat: special negative-conditional cache for two properties that are very important for princess calculations

Scoppio:
this property only caches `false`, any other value and it will have to recalculate itself every time until it turns to false. So, for the most part, it is going to be fast.

Also, it looks like it has no side-effects on other places since other places rely on the unit during gameplay, or loading it directly from a file.

There is also a reset builtin the property to make sure it is renewed every turn, so if it does not have the property it will calculate it at least once.

It still needs some more information on the API of the methods to explain why and how, and better testing from my part, comparing with and without the variable enabled.


Scoppio:
Unnecessary after the flag optimization PR

PR #6468, opened 2025-01-31
Fix/npe clicking on board no unit selected

Scoppio:
fix NPE when clicking on the board if no unit is selected, also makes the "next unit" button select a unit if none is selected.


Sleet01 (approved):
LGTM

PR #6474, opened 2025-01-31
fix: makes Princess move way more than she otherwise would on no-contact

Scoppio:
Changes how calculateMovementMod works by making it conditional to not having any enemies wround, It is attached to the self-preservation value which will positively score paths which provide better TMM whenever Princess has to scout "cold" (no enemy contact at all). Once enemies appear, she ignores this modifier and fallsback to her normal combat behavior


Sleet01 (approved):
Looks good!

PR #6478, opened 2025-02-01
fix: a couple of signs in the wrong place on the scoring equation

Scoppio:
During a larger refactoring a few signs were left out of place, and one offBoard calculation ended up mangled.


Sleet01 (approved):
LGTM

PR #6479, opened 2025-02-01
fix: removes an argument and early exit added by mistake to hyperagreression

Scoppio:
removes an argument and early exit added by mistake to hyperagression mod


Sleet01 (approved):
LGTM!

PR #6481, opened 2025-02-01
Fix 5590 part 2 stop choosing zero damage attacks completely

Sleet01:
This is an addendum to the prior 5590 work.  Nice-to-have in 0.50.03 but not critical.

This updates the new logging to use properties to allow i18n support, adds a few lower-damage weapons to those that Princess knows are blocked by trees (when the TacOps Woods Cover rule is on), and now causes Princess to completely drop 0-damage attacks with good to-hit numbers.  Previously these 0-damage attacks might be selected when nothing else presented itself.

Testing:
- Ran several more mock battles
- Fixed one unit test.


rjhancock (approved):
Looks sane.

HammerGS (approved):

PR #6487, opened 2025-02-02
Update MASH equipment to use size values

SJuliez:
The old style core and theater equipment cannot be used and the units used to load incorrectly.


Scoppio (approved):
I was sure I had fixed this almost a month ago.

HammerGS (approved):

PR #6490, opened 2025-02-03
Added Infantry `Mob` Unit Files

IllianiCBT:
- Added "Mob" unit files for Small, Medium, Large, and Huge variants. Created by @HammerGS.

These are used by the upcoming Prisoner rework.

### DIFF
--- a/megamek/data/mekfiles/infantry/Mobs/Mob (Huge).blk
+++ b/megamek/data/mekfiles/infantry/Mobs/Mob (Huge).blk
@@ -1,0 +1,59 @@
+#Saved from version 0.50.03-SNAPSHOT on 2025-01-29
+<UnitType>
+Infantry
+</UnitType>
+
+<Name>
+Mob
+</Name>
+
+<Model>
+(Huge)
+</Model>
+
+<year>
+2400
+</year>
+
+<originalBuildYear>
+2400
+</originalBuildYear>
+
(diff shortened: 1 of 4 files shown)


HammerGS:
Thinking we might want to give them a more specific name, Prisoners vs Mob. Since Mob would reflect more civilian in nature

HammerGS:
> Won't we need a new Infantry or Entity subclass, since (from what little I've researched) Mobs have special rules for minimum size, disbandment, and adjacency?
> 
> If that's the case, we have a few options:
> 
> 1. Subclass "Mob" into Prisoner, Insurgent, Civilians, Partisans, etc. (most work),
> 2. Give Mob a special field that contains what it's composed of in case they have different behaviours (middle ground),
> 3. Just use the name to indicate what kind of Mob a Mob is, but still subclass it from Entity or Infantry to contain the Mob rules (least work).

I'd love us to fully implement the new Mob rules from IKEO, but has a short term bridge I think just making the unit files is the simplest option. But open to ideas, as mobs opens up some interesting options for Riot Duty contracts, etc

IllianiCBT:
If someone were willing to fully implement mobs I would absolutely make use of them.

rjhancock (approved):
@HammerGS needs to approve as this is data but looks fine to me.

PR #6491, opened 2025-02-04
AX SRM Hardened and Ballistic Reinforced Armor Fix

x-Lanceman-x:
Realized that the change made to AX SRMs was inadvertently causing double damage to hardened and ballistic armors. Adding this new generalDamageType allows the fine tuning of the Acid effects for these specialty armors, resulting in continued extra damage against Ferro-Lam (and all other applicable armor types) while letting Hardened and Ballistic-Reinforced treat the attack as a standard SRM attack per the rules in TO.


rjhancock:
Please add comments as to where (Book, page, and what edition) they are in the rules.

Something we need to start doing for these things so that when things change or are in question, we can find the correct info faster.

stonewall072:
> Please add comments as to where (Book, page, and what edition) they are in the rules.
> 
> Something we need to start doing for these things so that when things change or are in question, we can find the correct info faster.

Done!

rjhancock (approved):
Confirmed.

PR #6494, opened 2025-02-04
Cap splashscreen button width

pavelbraginskiy:
On ultrawide monitors and some multi-monitor setups that respond poorly to our monitor-width detection, the startup splash could be extremely wide or spill onto other monitors. 

MM, MML, and MHQ all draw their startup splash in essentially the same way, but MML had a few lines that solved this problem that were missing from MM and MHQ, that I just copied over, see here:

https://github.com/MegaMek/megameklab/blob/76fec6f90878379bdf417ab14250961f5259e875/megameklab/src/megameklab/ui/StartupGUI.java#L156


pavelbraginskiy:
Screengrab reported by a user:

image

I haven't been able to test that this fix works since I don't have a setup where this is a problem, but there's no reason it shouldn't work since Lab renders just fine on their machine and I'm just copying over what lab does.

rjhancock (approved):

Issue #6128, opened 2024-10-21
Hidden unit not actually revealed if it chooses to not take a point blank shot?


gsparks3:
### Prerequisites and Pre-Issue Checklist

- [X] I'm reporting the issue to the correct repository:
- [MegaMek](https://github.com/MegaMek/megamek/issues)
- [MegaMekLab](https://github.com/MegaMek/megameklab/issues)
- [MekHQ](https://github.com/MegaMek/mekhq/issues)

- [X] I've tested the issue against at least the latest MILESTONE version
- [X] I've asked on the MegaMek Discord about the error
- [X] I've reviewed the BattleTech rules and MegaMek documentation, and I've confirmed that something isn't working as intended.
- [X] I've searched the Github tracker and haven't found the issue listed

### Severity *

Medium (Gameplay Limitation): Non-core functionality is impaired, providing a suboptimal but playable experience.

### Brief Description *

Issue discovered while testing #6127 - if you choose to _not_ take the point-blank shot, the hidden unit is not actually revealed, despite an enemy unit having ended its movement adjacent to the hidden unit. 

Interestingly, while the unit is not revealed, and cannot be targeted during the weapon attack phase, it _can_ be targeted during the physical attack phase if the hex containing the still-hidden unit is selected as a "target".

### Steps to Reproduce

1. Load the provided save as AmbushVictim. Connect an additional client as HiddenUnit.
2. Walk the Locust southwest five hexes.
3. Click "No" on the PBS prompt on the HiddenUnit client.
4. Observe that for the AmbushVictim client, the hidden unit is not revealed.
5. Advance to the Weapon Attack phase and observe that the still-hidden unit cannot be targeted (even after selecting the hex containing said unit).
6. Advance to the Physical Attack phase and observe that selecting the hex containing the hidden unit allows you to declare a physical attack.

### Operating System *

Windows 10

### Java Version *

17.0.11

### MegaMek Suite Version *

v0.49.19.1

### Attach Files

[hidden_unit_softlock_test.sav.gz](https://github.com/user-attachments/files/17452659/hidden_unit_softlock_test.sav.gz)

### Final Checklist

- [X] I've checked to make sure that this issue has not already been filed
- [X] I'm reporting only one issue in this ticket for clarity and focus

Sleet01:
I'll take a look at this; I have a repro for #6127 that will let me test.
Most likely we aren't unsetting "hidden" unless the unit fires, but doing an adjacency check at the start of the next phase to reveal them.

PR #6496, opened 2025-02-05
Fix 6127: softlock during invalid Pointblank Shot

### LINKED ISSUE #6127: [50.02] Potential soft-lock during point-blank shot from hidden unit

- [X] I'm reporting the issue to the correct repository:

- [X] I've tested the issue against at least the latest MILESTONE version
- [X] I've asked on the MegaMek Discord about the error
- [X] I've reviewed the BattleTech rules and MegaMek documentation, and I've confirmed that something isn't working as intended.
- [X] I've searched the Github tracker and haven't found the issue listed

### Severity *

### Brief Description *

If a point-blank shot from a hidden unit is triggered, and the player of the hidden unit chooses to take the point-blank shot when prompted, but the hidden unit in question is incapable of attacking the unit that triggered the PBS, the game soft-locks. The "Skip Firing" button is available, but does not appear to do anything. 

Severity is a little tricky on this one - the error is currently unrecoverable once it occurs (short of reloading from an earlier autosave, which is highly inconvenient, especially in games with a lot of units), but encountering a situation where it occurs is relatively rare. Not using hidden units will avoid it entirely, and choosing not to take the point-blank shot will also avoid the soft-lock in the case that the hidden unit would not be able to attack the triggering unit (which is itself uncommon).

This bug was initially encountered when an airborne aerospace fighter triggered a point-blank shot from a hidden infantry unit, but it is arguable whether airborne aerospace units should detect hidden units at all, so the provided save reproduces an equivalent issue with two mechs instead.

### Steps to Reproduce

1. Load the provided save as AmbushVictim. Connect an additional client as HiddenUnit.
2. Walk the Locust southwest five hexes.
3. Click "Yes" on the PBS prompt on the HiddenUnit client.
4. Observe that the target is fully out of arc, so no fire can be declared to allow Done Firing to be pressed.
5. Observe that pressing Skip Firing has no effect.

### Operating System *

Windows 10

### Java Version *

17.0.11

### MegaMek Suite Version *

v0.49.19.1

### Custom MegaMek Version

_No response_

### Attach Files

[hidden_unit_softlock_test.sav.gz](https://github.com/user-attachments/files/17452592/hidden_unit_softlock_test.sav.gz)

### Final Checklist

- [X] I've checked to make sure that this issue has not already been filed
- [X] I'm reporting only one issue in this ticket for clarity and focus

Sleet01:
There were actually two issues in 6127:
1. Invalid units allowed to make Pointblank Shots against Aerospace units;
2. Skip Attack button had no event handler, did nothing whatsover in the PBS display.

This patch fixes both issues, so that only units that can make PBS attacks get the prompt, and the user can use the Skip Attacks button if they change their mind.

Testing:
- Added unit tests for new Compute function

Close #6127


Sleet01:
Converting to draft to pull in the fix for #6128 and update Aerospace spotting to match the errata.

NickAragua (approved):

PR #6497, opened 2025-02-05
Action Logger

Scoppio:
# What it does

Logs all actions taken by all server-side. The data is persisted in TSV format

## But, why?

This is a first implementation of what will allow us to have base game replays capabilities in the future, but for visual representation only. It currently lacks things like depth, height, altitude, velocity, etc.

This also helps with AI debugging, allowing to compare actions taken by the bot and by players manually.


megamek/src/megamek/common/GameDatasetLogger.java
```diff
@@ -0,0 +1,372 @@
...
+                logDir.mkdir();
+            }
+            if (PreferenceManager.getClientPreferences().stampFilenames()) {
+                filename = StringUtil.addDateTimeStamp(filename);
+            }
+            logfile = new File(LOG_DIR + File.separator + filename + LocalDateTime.now() + ".tsv");
+            writer = new BufferedWriter(new FileWriter(logfile));
+            initialize();
+        } catch (Exception ex) {
+            logger.error("", ex);
```
rjhancock, line 60: Please log something meaningful.

rjhancock (commented):
Im hesitant on this as there are plans for such a feature in the works and to be placed into a format that would allow visual replay.

Scoppio:
I understand your position, but this is not meant as a final feature, but as stopgap to allow the tuning of our bot. This is not meant to stop anyone from building a replay feature as this was VERY cheap to produce an I am more than willing of throwing it away for a better and more complete feature in the future that is not just a logger but a full replay feature.

HammerGS:
I'm going to merge this with the recognition that it's a stop gap and down the road will be replaced or optimized.

Issue #6303, opened 2024-12-21
[RFE] Implement Remaining Terrains and Terrain Modifications


stonewall072:
### Prerequisites and Pre-Issue Checklist

- [X] I'm reporting the RFE to the correct repository:
- [MegaMek](https://github.com/MegaMek/megamek/issues)
- [MegaMekLab](https://github.com/MegaMek/megameklab/issues)
- [MekHQ](https://github.com/MegaMek/mekhq/issues)

- [X] Does this conform with our Content Policy linked above?
- [X] I've searched the GitHub tracker and haven't found a similar feature request listed

### RFE Details *

This is an improvement to an existing feature

### Brief Description *

The following terrains and terrain modifications still need implementation in MegaMek:

_TacOps AR:_

- Gravel Piles 
- Half Levels
- Bug Storms
- Water Flow
- Extreme Depths
- Hazardous Liquid Pools
- Rails
- Earthquake
- Meteor Showers
- Tainted and Toxic Atmospheres
- Additional building features such as doors and hexside walls, tents, fences, Castles Brian 

_BattleMech Manual:_

- Mud (Simplified) 

_Map Pack Alien Worlds:_

- Atmosphereic Domes (w/ Explosive Decompression)
- Ultra Sublevels
- Fungal Spores
- Crystal

_Battle of Tukayyid:_

- Improved Fieldworks
- Steep Cliffs (I *think* this is just Sheer Cliffs from TO:AR with a different name)
- Unstable 

_BattleTech Essentials:_

- Hardpacked Gravel

### MegaMekLab Suite Version *

0.50.02

### Final Checklist

- [X] I've checked to make sure that this RFE has not already been filed
- [X] I am requesting implementation of only one relevant set of features in this ticket for clarity and focus

SimonLandmine:
The _Improved Fieldworks_ from _Battle of Tukayyid_ are a slight, weirdly directional, variation on our existing improved infantry positions, I believe, although it's been a while since I've looked at them.

stonewall072:
> The _Improved Fieldworks_ from _Battle of Tukayyid_ are a slight, weirdly directional, variation on our existing improved infantry positions, I believe, although it's been a while since I've looked at them.

Yeah, it certainly looks that way, says it uses the same rules as fieldworks but only if attacks come through the marked hex side. 

Which also reminds me to add the actual weird pseudo-building "improved positions" to the list.

HammerGS:
[...]

PR #6502, opened 2025-02-06
Issue 6303: RFE - Ultra Sublevel

psikomonkie:
Another step towards #6303 

It's a really deep hole from Alien Map Packs. 

Map Makers, I'd recommend you'd lower the elevation some relative to its surrounding terrain - if a VTOL "lands' on the surface of an ultra sublevel, it'll be destroyed. Lower the surface below the "rim", and it'll let players hover a VTOL over the ultra sublevel, and lower then they can lower the unit's elevation to use it for cover :)


megamek/src/megamek/client/ui/swing/boardview/AbstractWreckSprite.java
```diff
@@ -75,6 +75,12 @@ public Entity getEntity() {
      */
     @Override
     public void prepare() {
+        if (entity.getPosition() != null && entity.getGame() != null && entity.getGame().getBoard() != null
+            && entity.getGame().getBoard().getHex(entity.getPosition()) != null &&
+            entity.getGame().getBoard().getHex(entity.getPosition()).containsTerrain(Terrains.ULTRA_SUBLEVEL)) {
+           return; //Don't show wrecks for "Ultra Sublevels" - The unit fell through the map!
```
Sleet01, line 81: Glorious!

Sleet01 (approved):
LGTM!

megamek/src/megamek/client/bot/princess/BasicPathRanker.java
```diff
@@ -1352,6 +1355,32 @@ private double calcHazardousLiquidHazard(Hex hex, boolean endHex, Entity movingU
...
+                logger.trace("Hovering/VTOL while traversing ultra sublevel (0).");
+                return 0;
+            } else if (movingUnit.getElevation() > 0) {  //elevation of 0 is on the ground (not airborne), which would be destroyed
+                // Estimate chance of being disabled or immobilized over ultra sublevel; this is
+                // fatal!
+                // Calc expected damage as ((current damage level [0 ~ 4]) / 4) *
+                // UNIT_DESTRUCTION_FACTOR
+                dmg = (unitDamageLevel / 4.0) * UNIT_DESTRUCTION_FACTOR;
+                logger.trace("Ending hover/VTOL movement over ultra sublevel ({}).", dmg);
+                return dmg;
```
Scoppio, line 1377: if current unit damage level is 0 then the dmg it counts is 0, I suggest juts setting `return UNIT_DESTRUCTION_FACTOR`

Scoppio (changes requested):

megamek/src/megamek/client/bot/princess/BasicPathRanker.java
Sleet01: > if current unit damage level is 0 then the dmg it counts is 0, I suggest juts setting return UNIT_DESTRUCTION_FACTOR

We want unit damage level 0 to indicate that we expect no damage from crossing or stopping on this hex:
```
0/4 * UNIT_DESTRUCTION_FACTOR = expected hazard damage == 0
1/4 * UNIT_DESTRUCTION_FACTOR = expected hazard damage == UNIT_DESTRUCTION_FACTOR / 4
```
etc.
The goal is to let Princess drive more aggressively with fresh units, but become more cautious as damage increases.

psikomonkie: > > if current unit damage level is 0 then the dmg it counts is 0, I suggest juts setting return UNIT_DESTRUCTION_FACTOR
> 
> We want unit damage level 0 to indicate that we expect no damage from crossing or stopping on this hex:
> 
> ```
> 0/4 * UNIT_DESTRUCTION_FACTOR = expected hazard damage == 0
> 1/4 * UNIT_DESTRUCTION_FACTOR = expected hazard damage == UNIT_DESTRUCTION_FACTOR / 4
> ```
> 
> etc. The goal is to let Princess drive more aggressively with fresh units, but become more cautious as damage increases.

That's the intent, yeah, this is ripped from the Magma (technically Hazardous Liquid, but that was ripped from Magma) Path Ranker logic. In this specific case it's intended just for VTOLs - a fresh VTOL should feel safe ending its turn over a death hex. But increasingly damaged ones should view it as increasingly dangerous as they get closer to being immobilized and thus risk falling victim to an instant death.

It's why they share the common bug I elude to in a comment in BasicPathRankerTest - VTOLs don't actually consider the ground below them because of some logic at the start of BasicPathRanker::checkPathForHazards - and I'd rather not touch that if I don't have to, that could have other unintended consequences.

Scoppio: Fair, I missed the `(EntityMovementMode.HOVER == movingUnit.getMovementMode() || EntityMovementMode.WIGE == movingUnit.getMovementMode()) {` at the top

Scoppio (approved):

pavelbraginskiy:
> Map Makers, I'd recommend you'd lower the elevation some relative to its surrounding terrain - if a VTOL "lands' on the surface of an ultra sublevel, it'll be destroyed. Lower the surface below the "rim", and it'll let players hover a VTOL over the ultra sublevel, and lower then they can lower the unit's elevation to use it for cover :)

It would be great to figure out a way to make ultra sublevels be forced to be at an extremely low level, ideally somewhere near `Integer.MIN_VALUE`. 

There's ultra depth hexes in the upcoming legendary battles maps, which should allow UMU units to submerge to any depth of their choosing, so solving ultra sublevels the "right way" now should make that easier to do later.

psikomonkie:
> > Map Makers, I'd recommend you'd lower the elevation some relative to its surrounding terrain - if a VTOL "lands' on the surface of an ultra sublevel, it'll be destroyed. Lower the surface below the "rim", and it'll let players hover a VTOL over the ultra sublevel, and lower then they can lower the unit's elevation to use it for cover :)
> 
> It would be great to figure out a way to make ultra sublevels be forced to be at an extremely low level, ideally somewhere near `Integer.MIN_VALUE`.
> 
> There's ultra depth hexes in the upcoming legendary battles maps, which should allow UMU units to submerge to any depth of their choosing, so solving ultra sublevels the "right way" now should make that easier to do later.

That seemed more invasive than I was going for, but it did cross my mind. If y’all think it’s a good idea I’m down to implement it, either as part of this or an enhancement.

gsparks3:
I think my only worry with that would be switching to isometric view causing ... issues. Not necessarily a blocker, but worth considering.

PR #6503, opened 2025-02-06
fix: stop creating empty logs for the bot path ranker when its disabled

Scoppio:
the timestamp was written twice in the filename if you have timestamp active for logs, and it was always creating a new log file, even if you dont have the correct log level set (debug or trace or all)


Sleet01 (approved):
LGTM!

PR #6504, opened 2025-02-06
Feat/minimap move path

Scoppio:
# What it does?

Move path shown in the minimap, number of turns it persists is configurable from 0 to 100. Default set to 2 turns of paths.

round_1_18_FIRING


megamek/src/megamek/client/ui/swing/minimap/Minimap.java
```diff
@@ -572,7 +585,7 @@ public void run() {
                     drawMap();
                 } else {
                     try {
-                        Thread.sleep(50);
+                        Thread.sleep(16);
```
Sleet01, line 588: Should this be a constant?

Sleet01 (approved):
LGTM

megamek/src/megamek/client/ui/swing/minimap/Minimap.java
```diff
@@ -707,6 +744,19 @@ private void drawMap(boolean forceDraw) {
         }
     }
 
+    private void paintMoveTracks(Graphics g) {
+        Graphics2D g2d = (Graphics2D) g.create();
+        Stroke dashed = new BasicStroke(3, BasicStroke.CAP_BUTT, BasicStroke.JOIN_BEVEL,
+            0, new float[]{6}, 3);
+        g2d.setStroke(dashed);
+        Color previousColor = g.getColor();
```
SJuliez, line 752: With a new g2d created, I think the previous color doesnt need to be restored (?)

megamek/src/megamek/client/ui/swing/minimap/Minimap.java
```diff
@@ -584,6 +597,26 @@ public void run() {
...
+    private final List<Line> movePathLines = new Vector<>();
+    public record Line(int x1, int y1, int x2, int y2, Color color, int round) {};
+    private final Color MOVE_PATH_COLOR = new Color(0, 0, 0, 128);
+
+    private void addMovePath(List<UnitLocation> unitLocations, Entity entity) {
+        // values equal or lower than 0 mean no persistence
+        if (GUIP.getMovePathPersistenceOnMiniMap() <= 0) {
+            return;
+        }
+        Coords previousCoords = entity.getPosition();
```
SJuliez, line 609: getPosition can easily return null, e.g. when a unit is offboard (arty), a flyer goes offboard to return later, or a unit isnt deployed yet. A frequent cause of bugs. I don't know if this method can be called in that case, but I suggest giving it a test at least with an arty unit deployed offboard.

SJuliez (commented):
Nice feature! It looks like the color is just dark grey/black now, maybe it could be tinted with the player's color to give the traces a bit of distinction when units arent nicely separated?

megamek/src/megamek/client/ui/swing/minimap/Minimap.java
SJuliez: A unit that fled might also lead to problems.

PR #6505, opened 2025-02-06
Remove XStream from Image Atlas Generation

rjhancock:
* Removes the libraries from the CreateImageAtlas utility that relate to XStream.
* Adds 2 utility classes for storing data about the image atlases.
* Adjusts classes to ensure fewest code changes needed (converting from new classes to Map<String, String> and back as needed).
* Removes some additional Gradle warnings.


megamek/src/megamek/client/ui/swing/util/ImageAtlasRecord.java
```diff
@@ -0,0 +1,85 @@
+/*
+ * Copyright (c) 2024 - The MegaMek Team. All Rights Reserved.
```
IllianiCBT, line 2: Should this be 2025 now?

rjhancock: Yes it should. It's also 1130 at night and I've been up since 6a.

IllianiCBT: No criticism intended. Just spotted it while I was failing to sleep :D

rjhancock: lol I took none. I get you on it several times, I expect the same in return when I miss something.

HammerGS (approved):

PR #6507, opened 2025-02-06
MegaMekLab Issue 1662: Fixed issue with Mek exporting to text

psikomonkie:
Fixes MegaMek/MegaMekLab#1662 

`<#if riscKit>` was throwing an error when a null value for riscKit was provided. This lets it work properly if riscKit is null. 

This is handled the same way in `mek.ftlh`, line 39


Sleet01 (approved):
LGTM!

PR #4985, opened 2023-12-30
add in game force display dialog

kuronekochomusuke:
- add in game force display dialog, similar to the lobby force display 
- double clicking a unit views it in unit display and centers unit on game board
- popup menu for view readout of unit
- add an indicator if a unit is done or not
- show when a unit is destroyed

image
image

image

PR #6512, opened 2025-02-07
fix: bounds check for unit editor dialog (MHQ #4985)

NickAragua:
Fixes MHQ 4985 by bounds checking the array being accessed if the number of crits is somehow higher than the location can support.


Sleet01 (approved):
LGTM!

Issue #1487, opened 2019-09-13
BUG: Units with the Melee Master ability can attack twice with a tree club


Tamren:
Tree clubs are supposed to break after one use, but when a pilot has Melee Master all melee attacks are doubled. It's current bugged where it will break twice, once for each hit.

I guess the intended behaviour would be for the tree club to break after the first hit and be replaced by a punch? Or perhaps two punches because it takes two working hands to use a club.

![Tree Club](https://user-images.githubusercontent.com/25917471/64900190-55f8ce00-d644-11e9-9501-fe7ee819f9cd.png)

Windchild292:
Waiting on rules check: https://bg.battletech.com/forums/supplementary-rules/melee-master-and-tree-interaction/?PHPSESSID=q54aqf0m56c80e5224v0i2k8iv

Tamren:
The rule check was never answered so I think we could just do what makes the most sense. Tree clubs only last for one strike, and you need two working hands to use a tree club. So if you use melee master to attack with a tree, the tree breaks after the first hit and the pilot uses their now empty hands to execute a double punch for the second half of the attack.

HammerGS:
I’m on the rules team. Answer is still in review.

HammerGS:
You'll get one attack with the tree and it will break. But melee master would allow a second attack (a punch).

PR #6519, opened 2025-02-08
MML #1487: unit selector redundant scrollpanes

SJuliez:
I think this fixes MegaMek/megameklab#1487
- removes a few redundant scrollpanes from the unit selector and unit preview. The left side of the unit selector no longer gets a scrollpane outside the unit table; it doesn't require much space as the unit table is flexible. The unit preview no longer gets dual scrollpanes
- the tech level list (called "Group") was misbehaving by becoming pixel-sized when the unit selector was made smaller than some unreasonable threshold so I removed the scrollpane on it and made it multi-column instead, see screenshot
- also, some setXXXSize methods are now only called when a specific size is given. This is used in MHQ in the personnel market so I kept it; but these methods should be avoided when possible, overriding getPreferredSize instead. 

image


Sleet01 (approved):
LGTM

PR #6522, opened 2025-02-08
Force generator update - individual equipment availability values

SuperStucco:
The method of providing availability values covers most situations but has problems with a couple of edge cases.  When a chassis or model requires high availability but should be limited to one or two equipment levels there is no way of handling this.  For example, a MilitiaMek that is widely available in low-rated garrison forces needs a high availability value but that will spill into other equipment ratings e.g. FS:8- is effectively F 8, D 7, C 6, B 5, and A 4.  Which means it will be showing up in A-rated forces, which is not desired.  Keeping it out of A-rated forces means using a lower than desired availability value.

This fleshes out and extends some original code in reading the availability values.  There is a new formatting option of using "!" to identify specific equipment levels.  For example, FS!F:8!D:5 will make the availability value 8 for F-rated generation, 5 for D-rated, and 0 (not available) for the ones that are not included.

Also updates the TXT documentation to cover the new additional formatting option.


Sleet01 (approved):
Looks good!

Would you be willing to put in some unit tests showing how this works for specific entities or equipment that were causing problems before?
I believe there are some in `TeamLoadOutGenerator` that could be adapted.

PR #6523, opened 2025-02-09
feat: uses median of the 5 closest enemies for facing mod

Scoppio:
# What it does?

The `Better Facing Mod` uses the median of the 5 closest enemies as the center position to face towards. The number 5 is completely arbitrary, but I think it is a good compromise betwen "sheer numbers" and emergency.

The new algorithm added is very direct:

- It sort enemies for the distance away from the unit.
- Units that have not moved are considered as if they were going to move directly towards the unit.
- The top 5 closest units are chosen.
- The median of the current position of the units is picked up as the target to face towards.
- The final facing is decided also basing upon which side of the entity has more armor left. It is using a very simple calculation which should be further improved in the future to consider things like exposed internals, XL engines, sides containing ammo bins. It is also missing other types of vehicles, as only Meks are considered in this bias for facing left or right side.

The result we can see in this image, im yellow we have the warhammer looking towards the Wraith, its closest enemy. In Green we have the median position of the 5 enemies closest to the Warhammer, with the new algorithm, the Warhammer would be facing towards that point instead.

Screenshot 2025-02-08 at 22 49 49

### DIFF
--- a/megamek/src/megamek/client/bot/princess/BasicPathRanker.java
+++ b/megamek/src/megamek/client/bot/princess/BasicPathRanker.java
@@ -26,6 +26,7 @@ import megamek.client.bot.princess.BotGeometry.HexLine;
 import megamek.client.bot.princess.UnitBehavior.BehaviorType;
 import megamek.codeUtilities.MathUtility;
 import megamek.common.*;
+import megamek.common.annotations.Nullable;
 import megamek.common.options.OptionsConstants;
 import megamek.common.planetaryconditions.PlanetaryConditions;
 import megamek.common.util.HazardousLiquidPoolUtil;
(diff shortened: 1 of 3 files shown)


Scoppio:
Results are promising, but once the ball of death happens, things start to go out of the window:

Screenshot 2025-02-08 at 23 37 19
Screenshot 2025-02-08 at 23 37 49
Screenshot 2025-02-08 at 23 38 02
Screenshot 2025-02-08 at 23 41 03
Screenshot 2025-02-08 at 23 42 58
Screenshot 2025-02-08 at 23 43 18
Screenshot 2025-02-08 at 23 43 43
Screenshot 2025-02-08 at 23 45 00

Sleet01 (approved):
LGTM, although I think some of this code was also in the AI Tools PR so there may be some merge errors.

We might be able to get this behavior if we added facing awareness to the code that assesses Crit and Kill chances (when Princess is figuring out how painful a given path might be to the moving unit).
Currently it always assumes the shots are coming from the front, but making back shots accurately reflect their higher crit and kill chances might make Princess face differently for less work...

PR #6525, opened 2025-02-09
Fix 5429 and 6493 Arrow IV issues

### LINKED ISSUE #5429: [49.19] [Bug] Incorrect damage from Arrow IV Homing Missiles mounted in ASF bomb slots.
This is a weird one.  The bug performs differently depending on whether or not I have ground units on my side.  Either way the damage is wrong though.

With no ground units present on my side:  Arrow IV Homing bombs fired by my ASF's always "miss" the TAG'd target regardless of roll, doing 5 points of artillery damage to the hex rather than the expected 20 damage to a single location.

With at least one 'Mech present on my side:  The Arrow IV Homing bombs now properly "hit" but do 10 damage instead of the expected 20.

Included in the .zip are logs and saves for both of the above situations.

[ArrowIVBombBugs.zip](https://github.com/MegaMek/megamek/files/15140435/ArrowIVBombBugs.zip)

--

I don't know if this last one should be a separate report or not but it popped up while testing the bug above.

Every time I moused-over an incoming artillery indicator for an Arrow IV round fired by an ASF the previous turn, standard or homing, I get an NPE.  It doesn't seem to do anything but it is annoying.

06NPE

### LINKED ISSUE #6493: Homing Direct-fire artillery has an erroneous TMM.
### Brief Description *

Direct fire Arrow IV homing missiles are adding an erroneous TMM, direct fire artillery isn't supposed to be affected by TMM to begin with (TO:AR page 153), and Homing Arrow IVs are supposed to only have the 4+ to-hit roll with no other modifiers (TO:AUE page 166).

### 3. Steps to Reproduce

Use a unit with an Arrow IV, and tag another unit, make sure neither unit moves, when you target the tagged unit with the homing missile, it'll have a +5 TMM that shouldn't exist, and shouldn't be considered to begin with even if the target did move.

### Attach Files

Image

Seen here, using an Arrow IV on a tagged target, I must stress that neither the target nor the firer moved this turn.

### Severity *

### MegaMek Suite Version *

0.49.20

### Operating System *

Windows 11

### Java Version *

Eclipse Adoptium

### Final Verification

Sleet01:
Fixes the following Arrow IV / Homing Artillery issues:
- #5429
    - Arrow IV Homing Bomb rounds dealing incorrect damage on hits.
    - Arrow IV Homing Bomb rounds sometimes missing en masse, supposedly due to no TAG hits, despite TAGged units in range.
- #6493 
   - Direct-fire Homing munitions reporting a TMM modifier for target movement.
- Un-reported issues:
   - AE rounds dealing octuple damage to Conventional Infantry units in the open (should be quadruple only).
   - Incorrect target roll reporting for some Homing shots' last-ditch miss roll.
   - Possible NPE when checking for spotters before Vector was initialized.
   - Aerospace units restricted to only firing a single TAG, although the rules allow for firing multiple TAGs at the same target.
   - `ArtilleryWeaponIndirectHomingHandler.java` could produce other incorrect reports when dealing AE damage.

Root causes were:
1. Arrow IV Homing Bomb rounds being treated as Copperhead rounds (-10 to base damage) instead of Arrow IV for damage calculations.
2. A single ToHitData instance was reused when checking for valid TAG hits, and would continue reporting "IMPOSSIBLE" value if _any_ TAGs missed or were out of range of the current Homing round's target.
3. I put a bot-related modifier, meant to provide utility hinting to Princess while building fire plans, into code that is actually called by the UI as well.
4. Several damage-doubling checks were being done in both the AreaEffectHelper and in DamageEntity.  The newer checks in DamageEntity should take precedence.
5. There was no specific report for the case where TAG hits, but the Homing Attack rolls under 4 and only hits the target Hex, not the target Entity.
6. Addition of new calls, probably.
7. Overly restrictive check.
8. `ArtilleryWeaponIndirectHomingHandler.java` was using wrong values to call `AreaEffectHelper.artilleryDamageEntity`, but they happened to line up type-wise so there were no IDE warnings.

I also took the liberty of shrinking the Arrow IV Bomb weapon equipment and bomb ammo names, for clarity.

Testing:
- Ran copious games with Homing artillery of various types, and TAG units both ground and Airborne.
- Ran standard Battle Royale with 80k vs 80k bot armies of mixed units containing various artillery and TAG units.
- Added new unit tests for relevant code.

Close #5429 
Close #6493

### DIFF
--- a/megamek/i18n/megamek/common/messages.properties
+++ b/megamek/i18n/megamek/common/messages.properties
@@ -610,8 +610,10 @@ BeastSize.monstrous=Monstrous
 ArtilleryMessage.drifted=Artillery drifted here from
 BombMessage.drifted=Bomb drifted here from
 
+### Weapon Handlers
+ArtilleryIndirectHomingHandler.HomingArtyMissChance=Homing shot locked on! (can miss on 3 or less)
 
-
+### ACAR
 acar.invalid_attack=Invalid attack
 acar.invalid_skill=Invalid skill
 acar.skill=Skill
--- a/megamek/i18n/megamek/common/report-messages.properties
+++ b/megamek/i18n/megamek/common/report-messages.properties
@@ -366,7 +366,7 @@
 3198=<<span class='miss'><B>misses</B></span> and is rendered ineffective by scatter.
 3199=<data> rounds <span class='success'><B>hit</B></span> the intended hex <data> with <data>.
 3200=<<span class='miss'><B>misses</B></span> and scatters off the board.
-3201=<data> homing rounds arrive at the intended hex.
+3201=<span class='miss'><B>misses</B></span> Terminal guidance failed!  Missed target but hit hex.
 3202=<data> round <<span class='miss'><B>misses</B></span> and scatters to hex <data>.
 3205=<data> (<data>) now on fire for <data> turns.
 3210=<newline>In hex <data>:
--- a/megamek/src/megamek/client/bot/princess/WeaponFireInfo.java
+++ b/megamek/src/megamek/client/bot/princess/WeaponFireInfo.java
@@ -23,11 +23,13 @@ import java.text.DecimalFormat;
 import java.text.NumberFormat;
 import java.util.*;
 
+import megamek.client.ui.Messages;
 import megamek.common.*;
 import megamek.common.actions.ArtilleryAttackAction;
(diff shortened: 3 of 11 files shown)


HammerGS (approved):
Checked with Co-Pilot and IDEA AI they saw no issues.

PR #6526, opened 2025-02-09
Allow Arty correctly on DS/SC

SJuliez:
Fixes MegaMek/megameklab#1692

Code to filter long toms from fighters and all arty from capital craft was already present in AeroUtil, but the aero flag was missing.


Sleet01 (approved):
LGTM!

PR #6528, opened 2025-02-09
Issue 6524: Prevent weapon-malfunction-crits from being unjammed via Unjam RAC (movement phase) & prevent NPE if unit unjams rapid-fire weapon and flees

### LINKED ISSUE #6524: [Issue] Uncaught java.lang.NullPointerException at beginning of firing phase
### Brief Description *

Cause unknown. Happens consistently at the start of the firing phase for this save. Following the error, game hangs at 'Waiting to begin Firing phase'. No customs, no unusual events I was able to observe. Have been unable to find a way to continue.

### 3. Steps to Reproduce

_No response_

### Attach Files

[autosave.sav.gz](https://github.com/user-attachments/files/18723513/autosave.sav.gz)

[megamek.log](https://github.com/user-attachments/files/18723515/megamek.log)

Image

### Severity *

### MegaMek Suite Version *

50.03

### Operating System *

Windows 11

### Java Version *

Java 17

### Final Verification

psikomonkie:
Two for one bonus special!

Fixes #6524 

There are two issues at play here:
- An entity is trying to unjam an RAC on the same turn that its fled the map. Once an entity flees, it is null, so the unjam step cannot be properly processed.
  - This is handled by adding a null entity check in resolveUnjam.
  - I suspect a similar bug will occur for other cases in `TWGameManager::resolveAllButWeaponAttacks`. I considered putting the null entity check there, but several of these actions still need to be processed even if the entity has fled, like clearing a minefield or activating an AP pod. I think we should handle these cases as they occur.
  - This only became an issue when we enabled fleeing at the end of movement. Prior to that a unit would never be null before processing its unjam step.
- Tanks were able to unjam "weapon malfunction" crits (TW p. 195) in the movement phases using the "Unjam RAC" which is meant to be used for the RAC or other rapid fire weapons if the setting is enabled.
  - Tank weapons that are jammed because of that are added to a special list that only tanks have. When checking if a weapon can be unjammed in the manner of rapid-fire weapons we first check if the weapon was jammed from a weapon malfunction crit - if it was we can't unjam it using this action.

### DIFF
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -2847,6 +2847,17 @@ public abstract class Entity extends TurnOrdered implements Transporter, Targeta
      */
     public boolean canUnjamRAC() {
         for (WeaponMounted mounted : getTotalWeaponList()) {
+
+            // Tanks can suffer a "weapon malfunction" critical. The weapon is marked
+            // as jammed but it should not be unjammed in the movement phase, it should
+            // be unjammed in the firing phase - that is, if this weapon is jammed from
+            // a weapon malfunction critical, we shouldn't unjam it using this method.
+            if (isVehicle() && this instanceof Tank tank) {
+                if (tank.getJammedWeapons().contains(mounted)) {
+                    continue;
+                }
+            }
+
             int ammotype = mounted.getType().getAmmoType();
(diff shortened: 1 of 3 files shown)


Sleet01 (approved):
LGTM!

PR #6529, opened 2025-02-10
fix: check if it is a narcpod before trying to access getPosition

Scoppio:
Aparently 19 years ago someone had the idea that NarcPod is a "targetable" but it has no "position" and you should not ask its position, and if you do it throws a runtime exception at you. This is... Unexpected and unwelcomed, as it implements a common API for things that can be targeted I imagined that anything that implements this API would... well.. be targetable, narcpods are not, and if you dont first check if it is not a narcpod you cannot safely check the coordinates of an attack action.

So, this PR handles that problem in the dataset logger.

### DIFF
--- a/megamek/src/megamek/common/GameDatasetLogger.java
+++ b/megamek/src/megamek/common/GameDatasetLogger.java
@@ -176,7 +176,7 @@ public class GameDatasetLogger {
         if (target != null ) {
             targetId = target.getId() + "";
             targetType = target.getClass().getSimpleName();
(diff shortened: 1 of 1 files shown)


Sleet01:
I was suspicious, since INarcPods _are_ targetable by physical attacks, the same way that Swarming Battle Armor is.  See Total Warfare, pp 139 and 220.
Honestly, it should probably just return the position of the unit it's attached to, since INarcPods only get instantiated once an INarc shot hits something, and are essentially created attached to the target.
But apparently they were implemented this way on purpose?  So weird.

Sleet01 (approved):
LGTM

PR #6531, opened 2025-02-10
MML #1721: legacy infantry blk files

SJuliez:
Fixes MegaMek/megameklab#1721
This allows loading legacy infantry files that have a "secondn" entry but no "secondary" (a number of secondary weapons but no weapon type for it). It appears that as recently as 49.16 such files could be produced with MML. In any case, this file check/file correction doesnt hurt.


Sleet01:
Agle brackets need backslash escaping to show up in GitHub, I believe.

SJuliez:
good point :)

Sleet01 (approved):
LGTM!

PR #6538, opened 2025-02-12
Issue #6537: Improved name search for diacritics

### LINKED ISSUE #6537: [RFE] Improved name search for Diacritics
### Enhancement Type *

New Feature

Currently we need to have the correct names of units for Record Sheets, but the name search fails when looking for units with Diacritics.  The challenge involves typing "Götterdämmerung" on a North American keyboard, which means figuring out how to use diacritics like umlauts and accents.

We almost need a "true unit name" field and a searchable field. 

### Use Cases or Rationale

See above

### Attach Files

N/A

### MegaMek Suite Version *

0.50.0 and On

### Final Verification

SJuliez:
Fixes #6537 
In the unit selector, allows base latin characters to find diacritics. Both "Gött" and "Gott" will find the "Götterdämmerung", "Gun" will find the "Gún" and "Arana" the "Aran~a" (cant even find how to type that)

### DIFF
--- a/megamek/src/megamek/client/ui/swing/dialog/AbstractUnitSelectorDialog.java
+++ b/megamek/src/megamek/client/ui/swing/dialog/AbstractUnitSelectorDialog.java
@@ -24,6 +24,7 @@ import java.awt.event.ActionListener;
 import java.awt.event.KeyEvent;
 import java.awt.event.KeyListener;
 import java.awt.event.WindowEvent;
+import java.text.Normalizer;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


HammerGS (approved):

pavelbraginskiy:
Accent normalization doesn't cover some letters like `ø`. I don't know if this is a problem now, but it's possible it could be in the future. A way to solve this would be to add a dependency on [ICU4J](https://mvnrepository.com/artifact/com.ibm.icu/icu4j), which has a bunch of utilities for working with Unicode, including a transliterator that does exactly what we want here.

Example:

```java
var transliterator = Transliterator.getInstance("Latin-ASCII");
System.out.println(transliterator.transliterate("Gøtterdämmerung"));
// Gotterdammerung
```

In general, Unicode is hard, and I think it makes sense to leave figuring it out to the experts.

RaozSpaz (changes requested):
Was suggested on Discord to add a request to this for this issue: #6550

PR #6562, opened 2025-02-16
Issues 6532: Make sure target entity's are on the board before checking for hex/terrain-based defender SPAs

### LINKED ISSUE #6532: [Issue] 50.03 Release MegaMek - NPE & Game Freeze during Artillery Phase after Attempted Counter Battery Fire
### Brief Description *

Mission vs Clan Wolf out of StratCon, 3050. Both sides have off-board artillery and I was attempting to counter-battery fire. Two of my batteries went fine, but the 3rd started giving me NPEs. I managed to skip firing that one I believe, but the next one caused an NPE and that ended up freezing MegaMek on what appears to be the enemy's turn. I'm able to save but can't skip firing or do any action.

AFAIK no artillery on either side has died or been targeted yet.

### 3. Steps to Reproduce

Quick Save and the Artillery NPE saves are right at the error and do not load before MegaMek says I'm disconnected, the autosave for Round 4 is right before this phase and if you attempt to counter-battery the enemy Artillery bot disconnects and the game freezes on the enemy turn.

If at error, attempt to counter-battery fire the Clan artillery. If at beginning of round 4, attempt to counter-battery fire all available targets. If prior to the error round, my artillery fired at designated targets in previous rounds for most part.

### Attach Files

[Artillery NPE.zip](https://github.com/user-attachments/files/18741896/Artillery.NPE.zip)

Includes what seems like all relevant logs, multiple saves, and my customs folder.

### Severity *

### MegaMek Suite Version *

50.03

### Operating System *

Windows 11 (64 bit)

### Java Version *

17.0.11

### Final Verification

psikomonkie:
Fixes #6532

### DIFF
--- a/megamek/src/megamek/common/actions/WeaponAttackAction.java
+++ b/megamek/src/megamek/common/actions/WeaponAttackAction.java
@@ -5638,17 +5638,17 @@ public class WeaponAttackAction extends AbstractAttackAction {
             // Urban Guerrilla - Target gets a +1 bonus in any sort of urban terrain
             if (te.hasAbility(OptionsConstants.INFANTRY_URBAN_GUERRILLA)
                 && (game.getBoard().getHex(te.getPosition()).containsTerrain(Terrains.PAVEMENT)
-                || game.getBoard().getHex(te.getPosition()).containsTerrain(Terrains.ROAD)
-                || game.getBoard().getHex(te.getPosition()).containsTerrain(Terrains.RUBBLE)
-                || game.getBoard().getHex(te.getPosition()).containsTerrain(Terrains.BUILDING)
-                || game.getBoard().getHex(te.getPosition()).containsTerrain(Terrains.ROUGH))) {
+                    || game.getBoard().getHex(te.getPosition()).containsTerrain(Terrains.ROAD)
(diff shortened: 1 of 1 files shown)


DM0000:
maybe in the getPosition() we add check for offboard and add an "offboard" terrain type of 0;
Would make these changes redundant and catch more of these issues in the future.

Sleet01:
> maybe in the getPosition() we add check for offboard and add an "offboard" terrain type of 0; Would make these changes redundant and catch more of these issues in the future.

I think 0 is already a terrain type, but adding an explicit offboard type would be a good solution and has been mentioned in the past.

Sleet01 (approved):
This looks fine for now.

I'll file an RFE for the Offboard Hex fix, which should cover a lot of these corner cases.

DM0000:
I get a null when closing the choose target for counter fire. Doesn't cause crash might be unrelated to this bug.

Sleet01:
For future reference, RFE for "Off-Board" Hex value is here: https://github.com/MegaMek/megamek/issues/6564

psikomonkie:
> I get a null when closing the choose target for counter fire. Doesn't cause crash might be unrelated to this bug.
> 
> edit-different issue not related. caused by not handling not choosing target

Good find! Fixed it here: #6565

PR #6579, opened 2025-02-17
Refactor startup to query if the user would like sentry enabled

pavelbraginskiy:
This PR does two things.

First, it moves the common startup code which all three suite programs perform (sentry, OIF, global exception handler) into it's own utility class to avoid duplicating code.

Second, on startup, asks the user if they would like Sentry to be enabled if the `sentry.properties` file doesn't already exist. If this change goes through, releases should be created without a `sentry.properties` file so that one can be created through this method.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -100,3 +100,4 @@ OfficialUnitList.txt
 equipment.txt
 .github/copilot-instructions.md
 *.iml
+sentry.properties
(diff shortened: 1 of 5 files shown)


rjhancock (commented):
I applaud the effort unfortunately a unified start location is not feasible at the moment as it will require more work on a per project basis to complete.

Each project has it's own url for Sentry AND server name.

I'm not opposed to a dialog being displayed about Sentry and the .gitignore change wont have an impact as the file is already committed. It would only ignore changes to it.

Issue #6461, opened 2025-01-30
[Bug] Valid locations to deploy aerial units are not correctly hilighted


BlueThing00:
### Brief Description *

VTOLS do not hilight areas with trees.  

ASF do not hilight areas with water or trees.
![Image](https://github.com/user-attachments/assets/f56a3700-4aa4-4444-b31d-e36ac038e5b1)

### 3. Steps to Reproduce

1) Field aerial units on a map with water and/or trees in the deployment zone
2) Note that many areas are not correctly hilighted

### Attach Files

This is a saved game which has VTOL and ASF to deploy with a deployment zone that contains both water and trees.
[BugRepro.sav.gz](https://github.com/user-attachments/files/18596509/BugRepro.sav.gz)

### Severity *

Medium (Gameplay Limitation): Non-core functionality is impaired, providing a suboptimal but playable experience.

### MegaMek Suite Version *

50.02

### Operating System *

Windows 11

### Java Version *

17.0.11

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have included all necessary information and files to help reproduce this issue
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

Sleet01:
Oh, they don't highlight valid hexes for _deployment_ purposes, got it.

RaozSpaz:
Note that it just fails to highlight. You can still very much deploy in said hexes.

psikomonkie:
Please, when fixing this, don’t just highlight all valid hexes. Currently it is highlighting valid *ground* hexes to deploy to. 

It will be very difficult to find valid ground hexes for Dropships (which require 7 hexes) if this is changed without considering the current functionality.

BlueThing00:
I don't have a problem with having two colors, one for ground available and one for only airborne.  

I'd just like all valid hexes to be hilighted.

Sleet01:
This is actually preventing Princess from deploying ASFs over water completely.
She uses the same code that determines if a hex is a valid _ground_ hex to try to deploy Aerospace units on ground maps.

Sleet01:
So, it looks like the current code is mostly (barring some outliers like VTOL Infantry) doing the correct thing, but somewhat opaquely:

1. A flying unit is selected.
2. Its elevation is set to 0 by default prior to placing it on the board.
[...]

PR #6582, opened 2025-02-19
Fix 6461: clean up ASF / VTOL / Hover / WiGE deployment hex display

Sleet01:
Per the attached issue, the existing code is 90% correct but gives no hint that a unit with VTOL, Hover, WiGE, or Aerospace attributes might have more deployment hex options at a different elevation / altitude.

This fix:
1. Places Aerospace units at max _elevation_ for the purpose of determining determining valid deployment hexes.  This is separate from Altitude.
2. Adds Cyan-colored hexes denoting currently-illegal hexes that would be legal deployment hexes at a different ~~altitude~~ elevation; the existing code will prompt the user for a different elevation within the legal range of elevations when clicked.
3. Corrects an issue where VTOL or Hover Infantry units were considered ineligible to deploy over water at any elevation.

N.B. WiGE require special handling because, while it is _possible_ that a WiGE unit could perform a maneuver that would let it clear terrain features of height 1 or more, it cannot deploy above 1 elevation.  So `isLocationProhibited(coords, 3)` returns `false` for a WiGE in a light woods hex because it could traverse that terrain _if_ it could reach that elevation, but we need to restrict them to elevation 1 for the purposes of finding deployable hexes.

Examples:
1. Deploying a VTOL at ground level in a crowded, multi-level city:
image

2. Same VTOL with its elevation set to "2":
image

3. Now at elevation "4":
image

Note that the minimap displays all hexes where this elevation is valid for deployment, while the main map highlights taller hexes with blue.

Testing:
- Deployed a number of types of flying units manually over various terrain.
- Confirmed that the Bot can also do so for units with elevations/altitudes set in the lobby (this was broken but not reported yet).

Close: #6461

### DIFF
--- a/megamek/src/megamek/client/bot/BotClient.java
+++ b/megamek/src/megamek/client/bot/BotClient.java
@@ -679,7 +679,7 @@ public abstract class BotClient extends Client {
                 Coords c = new Coords(x, y);
                 if (board.isLegalDeployment(c, deployed_ent)
                         && !deployed_ent.isLocationProhibited(c,
-                            ((deployed_ent.isAirborne() | deployed_ent.getMovementMode().isHoverVTOLOrWiGE())
+                            ((deployed_ent.isAirborne() || deployed_ent.getMovementMode().isHoverVTOLOrWiGE())
                                 ? deployed_ent.getElevation() : 0))
                         && !deployed_ent.isLocationDeadly(c)) {
                     validCoords.add(new RankedCoords(c, 0));
--- a/megamek/src/megamek/client/bot/BotClient.java
+++ b/megamek/src/megamek/client/bot/BotClient.java
@@ -678,7 +678,9 @@ public abstract class BotClient extends Client {
             for (int y = 0; y <= board.getHeight(); y++) {
                 Coords c = new Coords(x, y);
                 if (board.isLegalDeployment(c, deployed_ent)
-                        && !deployed_ent.isLocationProhibited(c, (deployed_ent.isAirborne() ? deployed_ent.getElevation() : 0))
+                        && !deployed_ent.isLocationProhibited(c,
+                            ((deployed_ent.isAirborne() | deployed_ent.getMovementMode().isHoverVTOLOrWiGE())
+                                ? deployed_ent.getElevation() : 0))
                         && !deployed_ent.isLocationDeadly(c)) {
                     validCoords.add(new RankedCoords(c, 0));
                 }
--- a/megamek/src/megamek/client/bot/BotClient.java
+++ b/megamek/src/megamek/client/bot/BotClient.java
@@ -678,7 +678,7 @@ public abstract class BotClient extends Client {
             for (int y = 0; y <= board.getHeight(); y++) {
                 Coords c = new Coords(x, y);
                 if (board.isLegalDeployment(c, deployed_ent)
-                        && !deployed_ent.isLocationProhibited(c)
+                        && !deployed_ent.isLocationProhibited(c, (deployed_ent.isAirborne() ? deployed_ent.getElevation() : 0))
                         && !deployed_ent.isLocationDeadly(c)) {
                     validCoords.add(new RankedCoords(c, 0));
                 }
--- a/megamek/src/megamek/client/ui/swing/boardview/BoardView.java
+++ b/megamek/src/megamek/client/ui/swing/boardview/BoardView.java
@@ -1285,11 +1285,26 @@ public final class BoardView extends AbstractBoardView implements BoardListener,
         int drawHeight = (view.height / (int) (HEX_H * scale)) + 3;
 
         Board board = game.getBoard();
+        boolean airDeployGround = en_Deployer.getMovementMode().isHoverVTOLOrWiGE();
         // loop through the hexes
         for (int i = 0; i < drawHeight; i++) {
             for (int j = 0; j < drawWidth; j++) {
                 Coords c = new Coords(j + drawX, i + drawY);
-                if (board.isLegalDeployment(c, en_Deployer) &&
+                if (en_Deployer.isAero()) {
+                    // Aeros are always above it all
+                    if (board.isLegalDeployment(c, en_Deployer) &&
+                        !en_Deployer.isLocationProhibited(c, board.getMaxElevation())) {
+                        drawHexBorder(g, getHexLocation(c), Color.yellow);
+                    }
+                } else if (airDeployGround) {
+                    // Draw hexes that are legal at a higher deployment elevation
+                    Hex hex = board.getHex(c);
+                    if (board.isLegalDeployment(c, en_Deployer) &&
+                        !en_Deployer.isLocationProhibited(c, hex.ceiling() + 1)) {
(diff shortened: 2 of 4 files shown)


megamek/src/megamek/client/bot/BotClient.java
```diff
@@ -678,7 +678,9 @@
             for (int y = 0; y <= board.getHeight(); y++) {
                 Coords c = new Coords(x, y);
                 if (board.isLegalDeployment(c, deployed_ent)
-                        && !deployed_ent.isLocationProhibited(c)
+                        && !deployed_ent.isLocationProhibited(c,
+                            ((deployed_ent.isAirborne() | deployed_ent.getMovementMode().isHoverVTOLOrWiGE())
```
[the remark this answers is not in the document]
Sleet01: Whoops.

megamek/src/megamek/client/ui/swing/boardview/BoardView.java
```diff
@@ -1285,11 +1285,30 @@ private void drawDeployment(Graphics g) {
...
 
         Board board = game.getBoard();
+        boolean isAirDeployGround = en_Deployer.getMovementMode().isHover() || en_Deployer.getMovementMode().isVTOL();
+        boolean isWiGE = en_Deployer.getMovementMode().isWiGE();
         // loop through the hexes
         for (int i = 0; i < drawHeight; i++) {
             for (int j = 0; j < drawWidth; j++) {
                 Coords c = new Coords(j + drawX, i + drawY);
+                if (en_Deployer.isAero()) {
+                    // Aeros are always above it all
```
psikomonkie, line 1295: Does this provide highlighting for aeros that deploy on the ground?

psikomonkie (changes requested):
I'm not sure if this is what we want for deploying ASF/Dropships. 

There doesn't appear to be any way to tell where I can deploy a dropship on the ground, even if I set it deploy deploy at altitude 0 in the lobby:

image

The same is true with ASF:

image

megamek/src/megamek/client/ui/swing/boardview/BoardView.java
Sleet01: You know, I thought it did, but I may have changed it around after the initial tests.

I'd _wanted_ it to work the same for grounded VTOLs and Aerospace, but now I don't recall if `isAero()` requires the unit to be flying, or just to have Aerospace movement...

Regarding Dropship deployment, I thought you could put it anywhere as long as the central hex was over a valid hex.  Do you have a ref for the rules governing that?

psikomonkie: Are the trees valid hexes? Everything’s highlighted. But I’ll see if I can find the rules on Dropships landing. Notably I also remember preventing deploying Dropships on top of units - yes you can land on units but it destroys them, and deployment shouldn’t be destroying units or changing terrain imo.

psikomonkie: TW: 87 has the rules for landing Dropships. While a dropship can land on trees, buildings, etc. it destroys them in the process. We already don’t want to let players destroy enemy units during the deployment phase by deploying a grounded dropship on top of the unit so I’d think we’d apply the same logic here.

gsparks3: Yeah, I was definitely expecting DS to end up similar to VTOLs, with yellow hexes for valid landed hexes and cyan for invalid ground hexes (but valid if airborne).

Sleet01: It looks like we currently only mark the center hex, even if placing the dropship adjacent to something would destroy it.  Fixing the landed vs airborne thing is simple, but fixing dropship placement might take more time.

Sleet01: @psikomonkie @gsparks3 Actually, not all that difficult.

Currently, we only disallow DSes deploying on hexes that themselves are illegal terrain for tanks, or that have less than half of their adjacent hexes at the same height as the center one.
I can make a hex disallowed at ground level if any of its neighbors contains a unit, in which case the display looks like this:
image

Sleet01: Any blue hex will prompt the user to choose an altitude of 1 or higher.

gsparks3: Very nice! 

How does it interact with woods/buildings? And does the new code do anything weird when deploying on the atmospheric or space maps?

Sleet01: Woods/buildings/rubble etc. are treated as prohibited per either tank or ASF rules, I'm not sure which.
Haven't looked at Low Altitude or Space maps yet, but I think those are different code paths.

gsparks3: > Woods/buildings/rubble etc. are treated as prohibited per either tank or ASF rules, I'm not sure which.

That tracks with aerodyne DS and fighters acting like wheeled vehicles while landed, so that checks out. Excellent work!

Sleet01:
Fixes are up for:
1. Aerospace incorrectly showing all hexes as valid when set to ground level (Altitude == 0)
2. DropShips showing hexes with neighboring units as valid (anti-stacking code prohibits deploying next to units that would be destroyed, we just weren't displaying them correctly).
3. DropShips showing hexes occupied by units as valid (same deal)

I am _not_ going to fix the weird situation where a hex with 5 same-height neighbors and a single lower-level neighbor shows as invalid, as the placement code seems to agree and I can't find a ruling on it.
I did, however, mark the code for future revision.

psikomonkie (approved):
I would still _prefer_ if we applied the blue & yellow hexes to all ASF/Dropships on ground maps including ones that are set to deploy airborne as it's possible to change that while in the deployment phase (and I forget to change it in the lobby). If you think the issue we chatted about re: the negative level hexes showing as unavailable is related to prohibited terrain or some other external method then I think this looks good.

psikomonkie (changes requested):
Marking as requesting changes until Sleet fixes the issue with negative level hexes (or determines it's out of scope?) so this isn't accidentally merged.

Sleet01:
I think I've covered all the edge cases now.

We should definitely look at refining this code and UX, as well as doing away with using the same variable for Elevation _and_ Altitude when deploying Aerospace units.  It's madness!

psikomonkie (approved):
Nice! Thanks Sleet!! Looks great!

psikomonkie:
>as well as doing away with using the same variable for Elevation _and_ Altitude when deploying Aerospace units.  It's madness!

Hahaha I noticed that during the review and shuddered. Like, I think I get why it was done, a unit shouldn't have an elevation and an altitude and for checking with collisions with ground objects that's good enough. But also yeah I agree that it's _wrong_ to be doing it that way and I'm sure it will screw over some maintainer at some point in the future if we don't change it to be more explicit.

PR #1527, opened 2019-10-25
Spheroid dropship movement

NickAragua:
Fixes several "bugs" with airborne spheroid dropships in low atmosphere and on ground maps, such as them being unable to move at all on the former and zigzagging like a crazed teenager on the ground maps. Also addresses a few issues with firing arcs and enables the bot to use flying spheroids properly.

Detailed list:
- Player can now move spheroid dropships on low-atmo maps using TW rules (1 hex)
- Spheroid dropship movement on ground map is restricted to 8 hexes per TW and must follow jump-jet-like pattern (not explicitly in the rules, but...)
- Spheroid dropships can properly fire at higher airborne targets with nose weapons, and lower airborne targets with aft weapons.
- Bot can use spheroid dropships according to above rules
- Incidentally, bot infantry can now rotate after movement
- Fixed several issues where the bot was using the wrong FireControl class to evaluate damage potential
- Player must now manually "hover" their spheroid dropship if not changing altitude explicitly.

Addresses issue #1303

PR #6583, opened 2025-02-19
MML #1527: Armor validation 

SJuliez:
This PR adds an item to TestEntity unit validation that tests if more armor points are allocated on a unit than its armor tonnage allows. It moves some required code from MML's UnitUtil to TestEntity. Fixes MegaMek/megameklab#1527


Sleet01 (approved):
LGTM!

PR #6588, opened 2025-02-20
Issue 6569: Trailers towed on a tow hitch are no longer returned as an unloadable unit - they should not be unloaded, they should be disconnected.

### LINKED ISSUE #6569: [Issue] [0.54.0-SNAPSHOT] Trailers can be unloaded like infantry
### Brief Description *

Currently, when towing a trailer you are shown two options: Unload and Disconnect. Disconnect drops the trailer off in its current hex and is the opposite of the "Tow" command. Unload is what's used to deploy infantry from carriers.

If the Unload option is utilized there are two effects:

1. It allows you to place the trailer in any hex around the towing vehicle, which violates the towing rules
2. The trailer doesn't actually disconnect, but while it remains connected to the tractor, the Disconnect option is disabled, leaving the trailer perpetually attached to the tractor 

### 3. Steps to Reproduce

1. Load a tractor and a trailer unit in the lobby
2. Connect the tractor to trailer either in the lobby or in game
3. Use the Unload button instead of Disconnect
4. Observe that the trailer does not detach from the tractor and that on following turns the Disconnect button is disabled

### Attach Files

Image

### Severity *

### MegaMek Suite Version *

0.54.0-SNAPSHOT

### Operating System *

Windows 11 10.0

### Java Version *

Adoptium 17.0.11

### Final Verification

psikomonkie:
Fixes #6569 

Trailers/tow hitches are a real special kinda' Transport. When something is towed on a tow-hitch it shouldn't be unloaded, it should be disconnected - so it shouldn't be returned as something that's unloadable. There's a separate method that returns the list of trailers that can be disconnected, and I've included that in the JavaDoc.

### DIFF
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -8728,14 +8728,16 @@ public abstract class Entity extends TurnOrdered implements Transporter, Targeta
     /**
      * @return All Entities that can at this point be unloaded from any transports
      *         of this Entity which are
-     *         no Bays. This does not include any units that were loaded this turn.
-     *         Note that the returned list may be unmodifiable.
+     *         not Bays. This does not include any units that were loaded this turn.
+     *         Note that the returned list may be unmodifiable.This shouldn't return
+     *         towed entities, they're tracked separately.
      *
+     * @see #getLoadedTrailers()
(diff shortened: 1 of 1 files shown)


Sleet01 (approved):
LGTM

PR #6590, opened 2025-02-21
fix: make ejected and other removed units show up transparent on minimap

Scoppio:
Ejected and other units that are removed from the game but the entity is not marked as "destroyed" now show up as if they were destroyed in the minimap


Sleet01 (approved):
LGTM

PR #6592, opened 2025-02-21
fix: small fixes that were missed from a previous code review

Scoppio:
Just some code that should be added, it deals with possible error states and useless attribution


Sleet01 (approved):
LGTM

PR #6594, opened 2025-02-21
Chain Drape (Construction Only)

pavelbraginskiy:
Adds the Chain Drape equipment from Arcade Operations: UrbanFest.
Gameplay is not implemented, only construction in Lab. 
Some implementation notes:
- UrbanFest says to construct the three configurations of the Chain Drape by mounting it forwards, backwards, or both forwards and backwards. As far as I know MegaMek doesn't support some but not all of a Mounted's slots being rear-facing, and the three configurations have different names, so I just created them as three separate equipment types. 
- While I said gameplay isn't implemented, mp _is_ reduced, since that's visible on record sheets.
- There is no published BV information, so a chain drape doesn't affect BV. Maybe the mp reduction cancels out the added protection?
- There is also no tech information, so I filled in the chain drape as X/X-X-X-X with an intro date of early spaceflight. If I had not filled in this information at all, it would never be selectable in lab.

### DIFF
--- a/megamek/src/megamek/common/Mek.java
+++ b/megamek/src/megamek/common/Mek.java
@@ -1113,6 +1113,16 @@ public abstract class Mek extends Entity {
         return Math.max(mp, 0);
     }
 
+    public boolean hasChainDrape() {
+        for (MiscMounted mount : getMisc()) {
+            if (!mount.isDestroyed() && mount.getType().hasFlag(MiscType.F_CHAIN_DRAPE)) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
     public int getPartialWingJumpWeightClassBonus() {
         return (getWeightClass() <= EntityWeightClass.WEIGHT_MEDIUM) ? 2 : 1;
     }
--- a/megamek/src/megamek/common/verifier/TestMek.java
+++ b/megamek/src/megamek/common/verifier/TestMek.java
@@ -440,6 +440,16 @@ public class TestMek extends TestEntity {
                 return false;
             }
         }
+        if (mt.hasFlag(MiscType.F_CHAIN_DRAPE)) {
+            if (countCriticalSlotsFromEquipInLocation(entity, mounted, Mek.LOC_LT) != 3) {
+                buff.append("incorrect number of chain draop crits in left torso\n");
+                return false;
+            }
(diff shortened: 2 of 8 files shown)


megamek/src/megamek/common/verifier/TestMek.java
```diff
@@ -440,6 +440,16 @@ public boolean checkMiscSpreadAllocation(Entity entity, Mounted<?> mounted,
                 return false;
             }
         }
+        if (mt.hasFlag(MiscType.F_CHAIN_DRAPE)) {
+            if (countCriticalSlotsFromEquipInLocation(entity, mounted, Mek.LOC_LT) != 3) {
+                buff.append("incorrect number of chain draop crits in left torso\n");
```
Sleet01, line 445: Typo, "draop" -> "drape"

megamek/src/megamek/common/verifier/TestMek.java
```diff
@@ -440,6 +440,16 @@ public boolean checkMiscSpreadAllocation(Entity entity, Mounted<?> mounted,
...
+                return false;
+            }
+            if (countCriticalSlotsFromEquipInLocation(entity, mounted, Mek.LOC_RT) != 3) {
+                buff.append("incorrect number of chain draop crits in right torso\n");
```
Sleet01, line 449: Ibid.

megamek/src/megamek/common/Mek.java
```diff
@@ -1113,6 +1113,16 @@ public int getJumpMP(MPCalculationSetting mpCalculationSetting) {
         return Math.max(mp, 0);
     }
 
+    public boolean hasChainDrape() {
+        for (MiscMounted mount : getMisc()) {
+            if (!mount.isDestroyed() && mount.getType().hasFlag(MiscType.F_CHAIN_DRAPE)) {
```
Sleet01, line 1118: Reversing these conditional checks would be marginally faster for units with multiple MiscMounted items, although I'm guessing the difference would be nearly immeasurable.  Just something to consider when checking multiple conditionals: the one that is `true` the least should ideally be first, to maximize short-circuit benefits.

Sleet01 (changes requested):
Just a couple of typos, otherwise looks good.

pavelbraginskiy:
> Just a couple of typos, otherwise looks good.

Done.

PR #6595, opened 2025-02-22
Fix: Improvements for Dataset Logger and its javadocs

Scoppio:
Adds better javadocs, dataset logger now logs if the action is being taken by a bot, the parser is updated, places where the dataset logger logs data is changed, etc


HammerGS (approved):

PR #6596, opened 2025-02-22
feat: disable ACAR for all clients if using real-blind-drop

### LINKED ISSUE #6517: [RFE] option to prevent running ACAR when Real Blind Drop option enabled
### Enhancement Type *

Improvement to Existing Feature

Disable ACAR when the "Real Blind Drop" option is enabled, to keep information hidden in lobby.  

Or add an option to disable ACAR in the lobby.

By "lobby", I mean the pre-game unit selection screen in megamek.

### Use Cases or Rationale

This request is meant more for mulitplayer pvp-ish games with secret information, not the against the bot / stratcon / whatever the pve game is called now.   (Though, it could apply there as well if an option to hide generated opfor info in mekhq is added at some point.)

Currently, you can run ACAR simulation in the setup lobby.  The `Real Blind Drop` option hides all units except your own when in the setup lobby.  However, if you run ACAR, it simulates with all loaded units, even those you can't see, and lists all of the units in the final report, thus revealing all of the units currently loaded.

Disabling use of ACAR simulation when doing "Real Blind Drop" would be in line with the spirit of the Real Blind Drop option, keeping the selected units secret in the lobby.  (Obviously, the players could just avoid using the ACAR simulator so they can avoid having secret info revealed in the lobby.)

### Attach Files

_No response_

### MegaMek Suite Version *

0.50.03

### Final Verification

Scoppio:
Disables ACAR for all players in the lobby if Real Blind Drop is enabled
- Closes #6517

### DIFF
--- a/megamek/src/megamek/client/ui/swing/lobby/ChatLounge.java
+++ b/megamek/src/megamek/client/ui/swing/lobby/ChatLounge.java
@@ -1285,10 +1285,13 @@ public class ChatLounge extends AbstractPhaseDisplay implements
         refreshTeams();
         refreshMapSettings();
         refreshDoneButton();
+        refreshAcar();
     }
 
     public void refreshAcar() {
-        panAutoResolveInfo.setVisible(CLIENT_PREFERENCES.getShowAutoResolvePanel());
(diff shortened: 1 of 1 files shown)


HammerGS (approved):

PR #6597, opened 2025-02-22
feat: refactored princess chat commands to use a modular approach

Scoppio:
# What does this do?

Refactoring princess chat commands.

This simplifies a little bit how we create and add new princess chat commands, also makes it easier to parse commands, and adds support for localization.

All previous tests passes, I removed the Verbose tests and all verbose references I found in the code, as I will remove it in a future version.

## Anything weid?

A "BIG" part of the change is just whitespace on the test, so its not exactly 2k lines changed here.


HammerGS (approved):

PR #6600, opened 2025-02-23
fix: fix a parsing error and turns it into a singleton using enum

Scoppio:
fix: fix a parsing error in random callsign generator and turns it into a singleton using enum


SJuliez (approved):
All good. I assume you remove the background loading cause it can't possibly save more than 10 ms on a modern PC.
Not as a suggestion for this PR but I had the thought of storing this as a yaml list (HAMMER: 2 // JULIAANA: 1) - reading this is very little code and it allows comments. Something for another day.

Issue #1741, opened 2020-03-05
Anti-Personnel Weapons not staying on AP Weapon Mounts


GamerFoxNeumann:
MegaMek Lab 0.47.4

Battle Armor Bug/Problem I've been getting is that weapons that I originally put on AP weapon mounts migrate to the manipulators when loading again later. This is consistent and constantly makes my BA designs invalid. This happens with pistols and rifles. It also sometimes shifts weapons that I can't put on AP Weapon mounts like LAWs from the hands to an AP Weapon mount.

Is this a logged problem or is it just me?

Windchild292:
I think @neoancient should comment on this.

GamerFoxNeumann:
> I think @neoancient should comment on this.

I found out that I posted this in the wrong forum as it were... but if neoancient can ascertain about this problem, it would be helpful. :)

neoancient:
Yes, it's a bug. It was reported as MegaMek/megameklab#441 and fixed by #1691 for the next release.

PR #6601, opened 2025-02-23
Pintle turret and flags (MML #1741)

SJuliez:
See MegaMek/megameklab#1741 for more info (merge together)
This PR has a few unit updates as well as the described flag check changes. The description for pintle mounts now contain the weight if it has a fixed weight.


megamek/src/megamek/common/AmmoType.java
```diff
@@ -24,6 +25,8 @@
 
 public class AmmoType extends EquipmentType {
 
+    private static final MMLogger LOGGER = MMLogger.create(AmmoType.class);
```
Scoppio, line 28: nitpick, the standard for logger in the project is logger lowecase.

Scoppio (approved):
Seems pretty good, I hope we can find a better solution in the EquipmentBitSet.

megamek/src/megamek/common/AmmoType.java
SJuliez: I noticed :) 
The very widely adopted convention for constants (static final) however is ALL_CAPS. When Tex reworked the logger thing, he made another million changes and I didnt want to pick on that minuscule issue. But I'd really like to stay with our style guide and common styles unless there's good reason. Is there one?

SJuliez: @rjhancock This kinda involves you :)

rjhancock: This is a case of when I did the mass changes I was less worried about style guide and more worried about ensuring legal compliance.

Style guide says upper case (which is a best practice I ignored in this case due to a 2 week marathon run of more important issues), then it should be upper case.

PR #6605, opened 2025-02-23
Data updates - IS faction Heavy Mek review 2900-3049

SuperStucco:
Next in the ongoing series of general availability reviews, IS faction heavy Meks from the 2900-3049 years.  The overall goals are:

- checking intro dates
- checking faction availability
- add any missing/overlooked chassis and models
- a first-pass on Periphery availability

Another goal specific to this time period is addressing the lack of variation between A-rated (best equipped) and F-rated (worst-equipped) forces. This is particularly difficult here due to the lack of advanced technology that would otherwise be used in other periods. The +/- and ! availability adjustments are used so forces generated at the highest level have a representative portion of faction-associated Meks such as the Warhammer -6D/-6K and Archer -2S/2K models.  The forces at the lowest level are much more generic, mostly Meks that have the Ubiquitous (IS) quirk such as the Archer -2R, Crusader -3R, and Warhammer -6R models.

Two important points of note with this update.

First, the Stalker STK-4P model which is heavy weight class (75 tons) has been removed (availability only, unit data file is still there).  Because it shares the chassis data with the much more common assault-weight models it is also forced to be extremely common even though it is not.  For now the only way to handle this is to remove it; thankfully it is uncommon enough that excluding it shouldn't be a problem.

Second, the linear interpolation of advanced technology between era data points leads to some issues in the years 3047 and 3048, due to most advanced designs being introduced in 3049 or later.  Most IS factions are expected to have around 20-25 percent Star League/advanced IS technology units by this point.  With only one or two advanced units available in those years there can be some weirdness such as "oops, all Axman lance" because that and the Marauder -5D are the only advanced-tech heavy Meks for the Fed Suns at that time.  There is a code fix for this BUT that also leads to much more severe problems e.g. front line Clan forces having significant numbers of Star League era Meks rather than all Clan OmniMeks.  So the code fix will have to wait until most of the availability data has been reviewed and updated.


HammerGS:
Going to merge but curious for the reason in dropping the -4P

PR #6606, opened 2025-02-23
Enhancement - Improve Deployment-Phase Towing Logic

psikomonkie:
The original implementation of "put warning signs where your tractor/trailer wouldn't be valid" did not properly consider an already deployed tractors/trailers that were more than one link away. This is specifically tested now with `TowLinkWarningTest::testDeployTrailerMultiDeployed()`. This also includes a change to Princess to use the new method.


megamek/src/megamek/client/ui/swing/TowLinkWarning.java
```diff
@@ -44,50 +47,178 @@ public class TowLinkWarning {
...
+        Entity closestDeployedTrailer = null;
+        boolean useAdjacentHexForTractor = (towingEnt instanceof LargeSupportTank);
+        if (deployingEntity.getTowedBy() != Entity.NONE && game.hasEntity(deployingEntity.getTowedBy())) {
+            towingEnt = game.getEntity(deployingEntity.getTowedBy());
+            if (towingEnt != null) {
+                useAdjacentHexForTractor = !useAdjacentHexForTractor || towingEnt instanceof LargeSupportTank;
+                if (towingEnt.isDeployed()) {
+                    closestDeployedTractor = towingEnt;
+                }
+                while (towingEnt != null && towingEnt.getTowedBy() != Entity.NONE) {
```
Scoppio, line 100: add parenthesis for each individual component of the condition

megamek/src/megamek/client/ui/swing/TowLinkWarning.java
```diff
@@ -44,50 +47,178 @@ public class TowLinkWarning {
...
+                    towingEnt = game.getEntity(towingEnt.getTowedBy());
+                    useAdjacentHexForTractor = !useAdjacentHexForTractor || towingEnt instanceof LargeSupportTank;
+                    if (towingEnt != null) {
+                        if (closestDeployedTractor == null) {
+                            if (towingEnt.isDeployed()){
+                                closestDeployedTractor = towingEnt;
+                            }
+                        }
+                    }
```
Scoppio, line 109: encloses into a single conditional

megamek/src/megamek/client/ui/swing/TowLinkWarning.java
```diff
@@ -44,50 +47,178 @@ public class TowLinkWarning {
...
+                }
+            }
+        }
+
+        // Did we find a tractor?
+        Set<Coords> validTractorCoords = null;
```
Scoppio, line 115: sets should not initialize null

megamek/src/megamek/client/ui/swing/TowLinkWarning.java
```diff
@@ -44,50 +47,178 @@ public class TowLinkWarning {
...
+        if (closestDeployedTractor != null) {
+            Set<Coords> possibleCoords = new HashSet<>();
+            possibleCoords.add(closestDeployedTractor.getPosition());
+            Entity testEntity = game.getEntity(closestDeployedTractor.getTowing());;
+            do{
+                possibleCoords = getValidDeploymentCoords(testEntity, board, useAdjacentHex, possibleCoords);
+                testEntity = game.getEntity(testEntity.getTowing());
+                useAdjacentHex = !useAdjacentHex || testEntity instanceof LargeSupportTank;
+            }
+            while(testEntity != null && !testEntity.equals(deployingEntity));
```
Scoppio, line 126: do not use do while with an open conditional and without a safety exit

megamek/src/megamek/client/ui/swing/TowLinkWarning.java
```diff
@@ -44,50 +47,178 @@ public class TowLinkWarning {
...
+                                if (towingEnt.isDeployed()){
+                                    closestDeployedTrailer = towingEnt;
+                                    break;
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        }
```
Scoppio, line 156: reduce the number of nested conditionals

megamek/src/megamek/client/ui/swing/TowLinkWarning.java
```diff
@@ -44,50 +47,178 @@ public class TowLinkWarning {
...
+
+        Set<Coords> validTrailerCoords = null;
```
Scoppio, line 158: try to start with a non null set.

Scoppio (changes requested):
this PR uses alot of lists and sets, which is great, but it also uses alot of nulls and null checks, please take a look if you can change some or most of the nulls you are using for empty collections instead, just to reduce the number of situations where we have to check for null, also there is deeply nested code in many places, see if its possible to avoid some of the nesting and aggregate multiple conditionals in a single test.

psikomonkie:
> this PR uses alot of lists and sets, which is great, but it also uses alot of nulls and null checks, please take a look if you can change some or most of the nulls you are using for empty collections instead, just to reduce the number of situations where we have to check for null, also there is deeply nested code in many places, see if its possible to avoid some of the nesting and aggregate multiple conditionals in a single test.

The usage & null checking was intentional, null is returned for those sets/lists as an alternative to getting and returning every valid coord. I'll see if I can get this refactored to not need to do that.

megamek/src/megamek/client/ui/swing/TowLinkWarning.java
psikomonkie: Fixed! Thanks!

megamek/src/megamek/client/ui/swing/TowLinkWarning.java
psikomonkie: Fixed! Thanks!

megamek/src/megamek/client/ui/swing/TowLinkWarning.java
psikomonkie: Added to all 4 of the while loops, and added a unit to verify it'll throw the stack overflow exception if it somehow gets a ouroboros provided.

megamek/src/megamek/client/ui/swing/TowLinkWarning.java
psikomonkie: Reduced, thanks!

megamek/src/megamek/client/ui/swing/TowLinkWarning.java
psikomonkie: Fixed, thanks!

megamek/src/megamek/client/ui/swing/TowLinkWarning.java
psikomonkie: Fixed, thanks!

psikomonkie:
> this PR uses alot of lists and sets, which is great, but it also uses alot of nulls and null checks, please take a look if you can change some or most of the nulls you are using for empty collections instead, just to reduce the number of situations where we have to check for null, also there is deeply nested code in many places, see if its possible to avoid some of the nesting and aggregate multiple conditionals in a single test.

Removed all of the null lists, reduced conditional nesting, and added a manual stack overflow exception for if a tractor->trailer->tractor situation occurs (which should & is prevented outside of this class). Thanks Luana!

megamek/unittests/megamek/client/ui/swing/TowLinkWarningTest.java
```diff
@@ -190,12 +218,283 @@
...
+        void testInfiniteLoops() {
+            // Arrange
+            when(mockTractor.getTowing()).thenReturn(TRAILER_ID);
+            when(mockTrailer.getTowing()).thenReturn(TRACTOR_ID);
+            when(mockTrailer.getTowedBy()).thenReturn(TRACTOR_ID);
+            when(mockTractor.getTowedBy()).thenReturn(TRAILER_ID);
+
+            // Act / Assert
+            // This should throw a stack overflow error.
+            Throwable exception = assertThrows(StackOverflowError.class, () -> TowLinkWarning.findTowLinkIssues(mockGame, mockTrailer, mockBoard));
```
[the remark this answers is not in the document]
psikomonkie: Is it really a problem, GitBot?

Scoppio (approved):

PR #6612, opened 2025-02-25
feat: adding force view to commander gui, mouse clicks and more

Scoppio:
# What does it do?

Adds force view to the Commanders GUI in place of the weird entities list and some mouse actions on the map, also renames the "map" to StrategicView.

A few panels are also changed to better serve both ClientGUI and Commanders GUI.

A few minor improvements to the StrategicView in how it paint stuff, and I am adding a way to allow the player to select one or many units.


HammerGS:
With the dev gone and having so many conflicts closing.

PR #6615, opened 2025-02-25
feat: More princess commands - ignore player/turret, order offboard artillery

### LINKED ISSUE #3794: Please add additional controls for friendly Princess artillery
Princess is reasonably good at using artillery, up to 49.7 at least.  But friendly artillery gets a little out of hand with the 'area denial' aspect as I can't get my own units in there to do things like pick up a downed pilot, finish off a downed Mech cleanly to ensure salvage, or preserve buildings to satisfy scenario conditions.  At best I can try to remember to keep /kick 'ing the bot player before the artillery targeting phase.

To this end I would like to see a few commands added to the Princess repertoire specifically for handling friendly artillery.

[Princess name]: artillery: [ auto | halt | barrage: [hex] | volley: [hex] | single: [hex]  ]

Offboard Princess: artillery: auto
This is the behavior as it is now - automatically calculate target hexes and fire at will.  Princess always defaults to this when starting.

Offboard Princess: artillery: halt
Will not fire artillery until prompted with one of the other commands.

Offboard Princess: artillery: barrage 0808
This overrides the standard (auto) method of target hex generation with the specified hex, and fires all artillery at it until prompted otherwise.

Offboard Princess: artillery: volley 0815
This causes Princess to fire one round with each artillery piece at the specified hex.  If it has multiple artillery, such a Naga with dual Arrow IV or a triple Thumper turret, it will fire each weapon at the same hex.  In the following rounds it will not fire until prompted (effectively switching to 'halt' status).  Ideally this would also have an optional integer to indicate a number of successive rounds to fire, but that may be a bit too complicated.

Offboard Princess: artillery: single 1020
This fires a single artillery weapon at the designated hex, chosen by best odds of hitting/least scatter distance followed by most standard damage.  Ideally this would have an optional indicator for type of preferred round, such as "homing" or "illumination", but that may be a bit too complicated.

### LINKED ISSUE #5986: [0.50.0-SNAPSHOT] RFE - Add a way to have Princess ignore units by player/force
### Description
I would like to have a way to tell Princess to ignore all units of a specific player or force. The example that comes to mind is I would like to be able to tell Princess to ignore all Convoy units and only engage the enemy escorts. As an example of the reverse, sometimes I would like Princess to focus her attention on destroying evacuating units while I deal with the escorts.

Scoppio:
# What it does?

This PR allows players to select the type of special ammo they want princess to use when shooting, right now it only allows you to select between mine, smoke and flare.

It also gives 5 orders:
- halt: cease fire immediately.
- auto: normal behavior
- single: only one shooter will fire at a target location (if it has multiple target locations in memory it will pick the one that it has better chance of hitting.
- volley: every arty will shoot once at any of the targets given.
- barrage: every arty will shoot continuously at the targets provided.

It also adds two new commands:

- Ignore player: adds every unit from a specific player in the ignored list
- Ignore turrets: adds every turret in the game in the ignored list

## Important
There is already the "Clear ignored targets" which can be used to reset the ignored target units list.

- closes #3794 
- closes #5986


Sleet01 (approved):
Looks great!

PR #6617, opened 2025-02-25
Handheld weapons Part the First: Loading and Construction

pavelbraginskiy:
Adds initial support for Handheld Weapons (TO:AUE p128). 
Currently, these are don't do much: They don't work in gameplay and you can't produce record sheets. What you _can_ do is load a HHW file into MegaMekLab, edit it, and save it again.

Because HHW support in this state is useless and probably buggy, this "support" was made invisible to the end user. You can't search for HHWs or create a new one, you can only edit one from a file you already have. This means all this really does is setup the internals for future work to add proper support for either sheets or construction. Once HHWs work for sheets or construction, we can make their support obvious and bundle the few canon HHWs from TRO:Prototypes.

Since you'll need at least one HHW file to try them out, here's the set of weapons from TRO:Prototypes you can use, plus an extra one with a melee weapon: [Handheld Weapons.zip](https://github.com/user-attachments/files/18956419/Handheld.Weapons.zip)

As an implementation note, handheld weapons are currently missing some properties that every other kind of Entity has. For example, they don't have any structure, and they don't have any crit slots, their `Mounted` equipment just exists in LOC_GUN with no associated slots. This shouldn't _matter_, but if it does, yell at me and I'll implement it.


megamek/src/megamek/common/HandheldWeapon.java
```diff
@@ -0,0 +1,255 @@
+package megamek.common;
```
psikomonkie, line 1: License?

megamek/src/megamek/common/HandheldWeapon.java
```diff
@@ -0,0 +1,255 @@
...
+import megamek.common.cost.HandheldWeaponCostCalculator;
+import megamek.common.enums.AimingMode;
+import megamek.logging.MMLogger;
+
+import java.io.Serial;
+import java.util.Vector;
+
+import static megamek.common.ITechnology.TECH_BASE_ALL;
+
+public class HandheldWeapon extends Entity {
```
psikomonkie, line 14: `Entity` extends `TurnOrdered`. Should `HandheldWeapon`s be `TurnOrdered`? This might not be a problem for you/this PR though, this might be a task for the implementation.

megamek/src/megamek/common/battlevalue/HandheldWeaponBVCalculator.java
```diff
@@ -0,0 +1,34 @@
+package megamek.common.battlevalue;
```
psikomonkie, line 1: License?

megamek/src/megamek/common/loaders/BLKHandheldWeaponFile.java
```diff
@@ -0,0 +1,40 @@
+package megamek.common.loaders;
```
psikomonkie, line 1: License?

megamek/src/megamek/common/verifier/TestHandheldWeapon.java
```diff
@@ -0,0 +1,153 @@
+package megamek.common.verifier;
```
psikomonkie, line 1: License?

megamek/unittests/megamek/common/loaders/BLKHandheldWeaponFileTest.java
```diff
@@ -0,0 +1,38 @@
+package megamek.common.loaders;
```
psikomonkie, line 1: License?

psikomonkie (changes requested):
Nice!! I'm really excited for this feature. I think this is good for enabling the implementation while avoiding impacting the users. 

1. License is missing from the new files, and that's the 'requested change'.
2. A more general Q about HHWs being an Entity. If it is an issue, I would think this could be changed by the implementer. If you think it's an issue & not something that could easily be changed during implementation maybe we should find an alternate. 

Thanks!

megamek/unittests/megamek/common/loaders/BLKHandheldWeaponFileTest.java
rjhancock: Files need the license header at the top of all files.

megamek/src/megamek/common/HandheldWeapon.java
pavelbraginskiy: I'm not sure what `TurnOrdered` does so I'm not sure why this is a problem, but guessing from its name I think the various `isEligibleForFiring`, `isEligibleForMovement`, etc. methods can be overridden to all return `false` for this to not be a concern?

psikomonkie: I think that’ll work; this just stuck out to me & I wanted to ask about it. Thanks!

pavelbraginskiy:
>     2. A more general Q about HHWs being an Entity. If it is an issue, I would think this could be changed by the implementer. If you think it's an issue & not something that could easily be changed during implementation maybe we should find an alternate.

I considered simply implementing `ICarryable` instead, but being an Entity offers _incredible_ implementation advantages. 

A HHW has far more in common with any other Entity than it has different when it comes to "meta" out-of-game features. Being an Entity means BLK loading works almost out of the box, BV, Cost, and Weight calculators work, I don't have to refactor all of MegaMekLab. Proper handling of `Mounted` equipment is free. I could go on and on.

psikomonkie (approved):
Awesome! Thanks for answering my Q & adding the licenses. I think this looks good. Thanks!

Edit: That follow up answer is just perfect, thank you so much for that. I really appreciate it.

Sleet01 (approved):
LGTM

PR #6621, opened 2025-02-26
wip: adding image hex slider, hex scorch, etc

Scoppio:
Exploratory PR with some ideas for automatic image splitting, a kind of decal reference for plasting things in the board, etc... it doesnt work properly yet and I dont know why...


Scoppio:
This has a couple of things that I want to push forward, like the automatic image split thing, but it requires a bit more work.

PR #6626, opened 2025-02-26
fix: ejected crew, ignored targets and broken enemies wont be target of physical attacks anymore

Scoppio:
# What it does?

This fix makes Princess treat ejected crew with the same respect it was giving to mekwarriors (overall), it also makes Princess consider if the unit is in a broken state or in the list of ignored targets before deciding to kick it.


psikomonkie (approved):
Nice!

Sleet01 (approved):
Echoing @psikomonkie, looks good.

psikomonkie (approved):
Nice! I appreciate Princess's adherence to the Ares Conventions on surrendering forces.

PR #6627, opened 2025-02-27
Implement RFE 6147: move infantry loading to interface and allow cargo bays

### LINKED ISSUE #6147: [RFE] Allow infantry to be carried in both infantry and cargo bays on vehicles

- [X] I'm reporting the RFE to the correct repository:

- [X] Does this conform with our Content Policy linked above?
- [X] I've searched the GitHub tracker and haven't found a similar feature request listed

### RFE Details *

This is an improvement to an existing feature

### Brief Description *

The Karnov and Anhur VTOLs are built with an infantry bay, rather than the standard cargo bay indicated in the TROs.  This allows them to load infantry but are then not considered for cargo space for other purposes, such as in MekHQ or future improvements such as battlefield salvage operations.

Edit - similarly, the Anhur (P-Stealth) model is built with a cargo bay, meaning it cannot carry infantry which would be quite natural for the design.

Changes may need to extend into MML and MHQ if this requires a more abstract solution, such as implementing an 'infantry-capable cargo bay' in addition to a normal cargo bay and infantry bay for vehicles.

### MegaMekLab Suite Version *

v.050.0

### Attach Files

_No response_

### Final Checklist

- [X] I've checked to make sure that this RFE has not already been filed
- [X] I am requesting implementation of only one relevant set of features in this ticket for clarity and focus

Sleet01:
1. Moves some data from InfantryBay.java into InfantryTransporter.java, a new Interface that should provide the minimum functionality for an Infantry-loading bay.
2. Implement missing Infantry drop prerequisites to align with TW and SO rules.
3. Remove restriction on number of infantry units dropped per turn, per SO rules.
4. Add the ability to load Infantry units (but currently no others) to CargoBay.java
5. Return ability to use Cargo Container bays to Fighter units (this had worked but was broken by the Mounted changes)
6. Removed unused entity builders from AreaEffectHelperTest.java while adding them for CargoBayTest.java.

See https://discord.com/channels/458705327911731231/1344135419180351560 for rationale for some design decisions taken here.

Testing:
- Many manual drops to confirm UI, space usage, and unit compatilibity.
- Added unit tests for CargoBay class.

Merging this should unblock MegaMek/mekhq#5929

Close #6147

### DIFF
--- a/megamek/src/megamek/client/ui/swing/MovementDisplay.java
+++ b/megamek/src/megamek/client/ui/swing/MovementDisplay.java
@@ -2585,6 +2585,7 @@ public class MovementDisplay extends ActionPhaseDisplay {
         // *(excluding Infantry, see StratOps pg. 20)
         int doorsEligibleForDrop = bay.getCurrentDoors();
         List <Entity> droppableUnits = bay.getDroppableUnits();
+        boolean infantryTransporter = (bay instanceof InfantryTransporter);
 
         // No units == no drops
         if (droppableUnits.isEmpty()) {
@@ -2601,7 +2602,7 @@ public class MovementDisplay extends ActionPhaseDisplay {
 
         boolean hasDroppableUnit = false;
         for (Entity droppableUnit : droppableUnits) {
-            if (doorsEligibleForDrop > 0) {
+            if (infantryTransporter || doorsEligibleForDrop > 0) {
                 if (droppedUnits.contains(droppableUnit.getId())) {
                     // Infantry don't count against door usage
                     if (!droppableUnit.isInfantry()) {
@@ -2615,7 +2616,7 @@ public class MovementDisplay extends ActionPhaseDisplay {
                 }
             }
         }
-        if (doorsEligibleForDrop > 0 && hasDroppableUnit) {
+        if (hasDroppableUnit && (infantryTransporter || (doorsEligibleForDrop > 0))) {
             return true;
         }
         return false;
@@ -4052,8 +4053,11 @@ public class MovementDisplay extends ActionPhaseDisplay {
                 isInfantryT = true;
 
                 // Can't drop infantry from above 8 Altitude
-                if (isInfantryT && cmd.getLastStep().getAltitude() > 8) {
-                    continue;
+                if (isInfantryT) {
+                    int currAlt = (cmd.getLastStep() != null) ? cmd.getLastStep().getAltitude() : ce.getAltitude();
+                    if (currAlt > 8) {
+                        continue;
+                    }
                 }
 
                 for (Entity entity : iTransporter.getDroppableUnits()) {
--- a/megamek/src/megamek/client/ui/swing/MovementDisplay.java
+++ b/megamek/src/megamek/client/ui/swing/MovementDisplay.java
@@ -2600,6 +2600,8 @@ public class MovementDisplay extends ActionPhaseDisplay {
             return true;
         }
 
+        // If current entity has any viable bays with droppable units that aren't already dropping,
+        // activate the button.
         boolean hasDroppableUnit = false;
         for (Entity droppableUnit : droppableUnits) {
             if (infantryTransporter || doorsEligibleForDrop > 0) {
@@ -2616,7 +2618,7 @@ public class MovementDisplay extends ActionPhaseDisplay {
                 }
             }
         }
-        if (hasDroppableUnit && (infantryTransporter || (doorsEligibleForDrop > 0))) {
+        if (hasDroppableUnit) {
             return true;
         }
         return false;
@@ -2629,23 +2631,8 @@ public class MovementDisplay extends ActionPhaseDisplay {
      * @return true if there are available droppable units in this compartment
      */
     private boolean checkCompartmentDropEnable(InfantryCompartment compartment, Set<Integer> droppedUnits) {
-        // A compartment can drop any number of infantry and is "assumed to have one door".
         List <Entity> droppableUnits = compartment.getDroppableUnits();
-        boolean hasDroppableUnits = !droppableUnits.isEmpty();
-
-        // No units == no drops
-        if (!hasDroppableUnits) {
-            return false;
-        }
-
-        for (Entity droppableUnit : droppableUnits) {
-            if (!droppedUnits.contains(droppableUnit.getId())) {
-                // If we have even one boarded, droppable unit that didn't already drop, activate the button.
-                return true;
-            }
-        }
-        // No droppers means no enable
-        return false;
+        return droppableUnits.stream().map(Entity::getId).anyMatch(u -> !droppedUnits.contains(u));
     }
 
     private void updateDropButton() {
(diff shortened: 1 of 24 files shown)


megamek/unittests/megamek/common/CargoBayTest.java
psikomonkie, line 1: License?

psikomonkie (changes requested):
Nice! I'm really excited for this. Code-wise everything looks great. Missing license file on `CargoBayTest.java` will need added, and I think the javadoc for `InfantryTransporter.java` was copied from `Transporter.java` - I was confused by it for a moment while reviewing, so maybe that should be looked at. 

Thanks!

psikomonkie:
There are a few more `Transporter`s that are for transporting infantry, and I think the changes to `MovementDisplay.java` would prevent units from dropping out of these Transporters:

- `InfantryCompartment.java`
- `BattleArmorBay.java`

I don't think this should be implemented for the BA Handles even though those are for Transporting infantry, since this is specifically for infantry that are using their weight for loading.

Sleet01:
Moving to draft until I can get the BattleArmorBay done.

Sleet01:
@psikomonkie InfantryCompartment and BABay should both work correctly.
Also implemented the NOE infantry drop straight to ground from TW pg 225, with appropriate jump / VTOL mod.

megamek/src/megamek/client/ui/swing/MovementDisplay.java
```diff
@@ -2573,6 +2573,81 @@ private void updateLaunchButton() {
...
+        }
+
+        for (Entity droppableUnit : droppableUnits) {
+            if (!droppedUnits.contains(droppableUnit.getId())) {
+                // If we have even one boarded, droppable unit that didn't already drop, activate the button.
+                return true;
+            }
+        }
+        // No droppers means no enable
+        return false;
```
SJuliez, line 2648: I believe the whole marked area can be replaced by

```
// If we have one boarded, droppable unit that didn't already drop, activate the button.
return droppableUnits.stream().map(Entity::getId).anyMatch(u -> !droppedUnits.contains(u));
```

megamek/src/megamek/client/ui/swing/MovementDisplay.java
```diff
@@ -3983,31 +4038,63 @@ private TreeMap<Integer, Vector<Integer>> getDroppedUnits() {
...
         // cycle through the bays
         int bayNum = 1;
-        Vector<Bay> Bays = ce.getTransportBays();
-        for (int i = 0; i < Bays.size(); i++) {
-            Bay currentBay = Bays.elementAt(i);
+        Vector<Transporter> transporters = ce.getTransports();
+        List<Entity> currentUnits;
+
+        for (int i = 0; i < transporters.size(); i++) {
+            Transporter currentT = transporters.elementAt(i);
```
SJuliez, line 4045: Please avoid single letter abbreviations. currentT and iTransporter -> currentTransport and infantryTransporter. If this is a length issue then at least currentTrsp and infTransporter or something.

megamek/src/megamek/client/ui/swing/MovementDisplay.java
```diff
@@ -3983,31 +4038,63 @@ private TreeMap<Integer, Vector<Integer>> getDroppedUnits() {
...
-                } else {
-                    currentUnits.add(entity);
+            int doorsEligibleForDrop = 0;
+
+            // Handle infantry first
+            if (currentT instanceof InfantryTransporter iTransporter) {
+                isInfantryT = true;
+
+                // Can't drop infantry from above 8 Altitude
+                if (isInfantryT) {
```
SJuliez, line 4056: I fail to understand this if clause

megamek/src/megamek/client/ui/swing/MovementDisplay.java
```diff
@@ -3983,31 +4038,63 @@ private TreeMap<Integer, Vector<Integer>> getDroppedUnits() {
...
+                    int currAlt = (cmd.getLastStep() != null) ? cmd.getLastStep().getAltitude() : ce.getAltitude();
```
SJuliez, line 4057: I believe you can use the getFinalAltitude() method instead

megamek/src/megamek/common/Entity.java
```diff
@@ -8747,10 +8747,17 @@ public Vector<Entity> getDroppableUnits() {
 
         // I should only add entities in bays that are functional
         for (Transporter next : transports) {
-            if ((next instanceof Bay nextbay) && (nextbay.getCurrentDoors() > 0)) {
-                for (Entity e : nextbay.getDroppableUnits()) {
+            if (next instanceof InfantryCompartment nextCompartment) {
+                for (Entity e : nextCompartment.getDroppableUnits()) {
                     result.addElement(e);
                 }
```
SJuliez, line 8753: result.addAll(nextCompartment.getDroppableUnits());
?

megamek/src/megamek/common/Entity.java
```diff
@@ -8747,10 +8747,17 @@ public Vector<Entity> getDroppableUnits() {
...
+            } else if (next instanceof Bay nextBay) {
+                // Infantry (Conventional and BA) do not need doors to deploy (TM 209)
+                if (nextBay instanceof InfantryTransporter || nextBay.getCurrentDoors() > 0) {
+                    for (Entity e : nextBay.getDroppableUnits()) {
+                        result.addElement(e);
+                    }
```
SJuliez, line 8759: ditto

megamek/src/megamek/common/InfantryTransporter.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.Vector;
+
+/**
+ * Classes that implement this interface have the ability to load, carry, drop, and
+ * unload Infantry units in the game.
+ * This interface includes a subset of the Bay.java class methods and Transporter interface.
+ * @see megamek.common.Bay
+ * @see megamek.common.Transporter
+ */
+public interface InfantryTransporter extends Serializable {
```
SJuliez, line 27: Have you thought about this one extending Transporter? I have no idea if it would be better or work at all. Just a thought.

SJuliez (changes requested):
I have a few suggestions and questions. Leaving this as change requests so it doesnt get merged away immediately.

megamek/src/megamek/common/InfantryTransporter.java
psikomonkie: Transporter has a lot of unnecessary stuff for Infantry transport. I thought about asking this in my original review but I think the intent is to break up the functionality some. 

I’m thinking of doing something similar to support loading non-entities as cargo, having a CargoTransporter interface that can be implemented by any Transporter that can transport non-entity cargo. 

I might be totally off base though, I’m curious about Sleet’s thoughts.

megamek/src/megamek/client/ui/swing/MovementDisplay.java
Sleet01: Skips the current Transporter if it's an InfantryTransporter and the current / last altitude in the MP is > Altitude 8.

megamek/src/megamek/common/InfantryTransporter.java
Sleet01: Pretty much what @psikomonkie said: I don't believe all of the functionality of a Bay is required for an InfantryTransport (although currently most of the InfantryTransport functionality is already defined in Bay).
But not every Bay is an InfantryTransport, and vice-versa.
So it made more sense to extract the minimum set of functionality needed for infantry dropping and loading into its own interface, to my mind.

If somebody wants to completely redo Bays, Transporters, and InfantryTransporters, I'd like to see:
1. The base Bay class reduced to a minimum set of shared functionality.
2. Various other droppable / launchable type interfaces extracted out of Bay into their own Transporter interfaces.
3. <Type>Bay functionality created with composition, so that we can simply and quickly create bays that support multiple unit types (SmallCraftBay, for instance, should support launching and landing both Small Craft and AeroSpaceFighter units).
4. As much game logic as possible removed from the UI; this will require passing stuff like MoveStep, elevation, altitude, etc. in to various functions that report whether operations are allowed.

But that would all be for a future project.

megamek/src/megamek/client/ui/swing/MovementDisplay.java
SJuliez: isInfantryT is always true in l.4056 because l.4053 sets it to true.

megamek/src/megamek/common/InfantryTransporter.java
SJuliez: >But not every Bay is an InfantryTransport, and vice-versa.
That's an important point, so the interfaces must be kept apart as they are now

megamek/src/megamek/client/ui/swing/MovementDisplay.java
Sleet01: Aha, not anymore!
But also, there was a lot of refactoring going on >_>

SJuliez (approved):

SJuliez:
@psikomonkie There are still your change requests open, can you take a look if that's been resolved, thanks

PR #6629, opened 2025-02-27
Small Troopspace refactor

SJuliez:
This is a small code update around troopspace a.k.a. Infantry Compartment. It moves a few methods up to Transporter as defaults, removes redundant comments and reduces some extreme forms of over-verbose code.


psikomonkie:
What is this for? This looks like it'll conflict with #6627.

PR #6641, opened 2025-03-01
Prevent mixing special ammo kinds on HHWs

pavelbraginskiy:
Handheld weapons aren't allowed to mix different special munitions, so this adds a validation check to limit you to one ammo bin per weapon kind.

It's not yet clear if the restriction of one ammo type per weapon applies to MMLs and ATMs. This allows multiple ammo bins to be mounted if they differ in weapon type or rack size, so MMLs can have both SRM and LRM ammo, while ATMs are limited to a single type. 

The validation can easily be updated if we get a ruling for MMLs and ATMs.

The HHW default sprite is also updated.


megamek/src/megamek/common/verifier/TestHandheldWeapon.java
```diff
@@ -145,6 +164,24 @@ public StringBuffer printEntity() {
...
+            if (ammoKinds.contains(kind)) {
+                illegal = true;
+                buff.append("Handheld weapon can only mount a single ammo bin for a given kind of ammo.\n");
+                break;
+            }
+            ammoKinds.add(kind);
+        }
+
+        return illegal;
+    }
```
psikomonkie, line 183: How do you configure a HHW to use a single 3 ton ammo bin like the Heavy Thunderbolt Weapon instead of three 1-ton bins, which would return illegal for this method?  

image

image

psikomonkie (changes requested):
I'm kinda gonna miss seeing "GUN"!
I don't think `TestHandheldWeapon::correctEntity(StringBuffer, int)` is returning the values expected - can you take a look at that Pavel and let me know what you think? Thanks!

megamek/src/megamek/common/verifier/TestHandheldWeapon.java
pavelbraginskiy: HHW Ammo bins are resizable equipment, you can set the shot count the same as you would set the size of a Cargo. 
The value in the Size/Shots column can be edited, you can write an 18 in there instead of a 6.

pavelbraginskiy: I suppose I can actually throw a hint in the validation message about that.

psikomonkie: > The value in the Size/Shots column can be edited, you can write an 18 in there instead of a 6.
_Oooh_, thank you for explaining that. I was trying to edit the Tons column. Thank you for clarifying! A hint from the validation message would be great.

psikomonkie (approved):
Nice, thank you for explaining how to adjust the shots & adding that message to the popup.

PR #6643, opened 2025-03-01
MekHQ Issue 3655: Fix vehicle crew requirements for trailers 

psikomonkie:
Certain trailers do not require crew. This is the initial pass at better supporting that in MegaMek. On its own, this shouldn't change anything, except perhaps Princess opting to not abandon "immobilized" trailers (they're always immobilized, that's normal!).

Trailers in MegaMek will still be generated with crew. 

Trailers sent from MekHQ with no crew will receive a commander "No Commander" that isn't reported back to the `ScenarioResolutionWizard` in MHQ when the scenario is over.

Related to MegaMek/mekhq#3655

Associated MekHQ PR: MegaMek/mekhq#6116


HoneySkull (approved):

HoneySkull:
This change has a corresponding PR/change in MekHQ repo.

PR #6645, opened 2025-03-02
Support multiselect in the Add Unit dialog

pavelbraginskiy:
Allows selecting multiple units in the Select Unit dialog, instead of having to hit Select for each unit one at a time:

image

This might not be _incredibly_ useful for MegaMek, but this sets up for this functionality to be made available in Lab, where it should help Hammer make fluff edits to every unit of a given chassis in bulk.

### DIFF
--- a/megamek/src/megamek/client/Client.java
+++ b/megamek/src/megamek/client/Client.java
@@ -354,12 +354,13 @@ public class Client extends AbstractClient {
      * Sends an "add entity" packet that contains a collection of Entity
      * objects.
      *
-     * @param entities
-     *                 The collection of Entity objects to add.
+     * @see megamek.server.totalwarfare.TWGameManager#receiveEntityAdd(Packet, int)
+     *
+     * @param entities The collection of Entity objects to add.
+     *                 This should ideally be an {@link ArrayList<Entity>}, but other kinds of {@link List} will be converted to an {@link ArrayList}.
      */
     public void sendAddEntity(List<Entity> entities) {
         // Trying to pass a non-ArrayList jams the receiving client and prevents it from ever receiving more packets.
-        // I have no idea why this happens.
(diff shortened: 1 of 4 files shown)


megamek/src/megamek/client/Client.java
```diff
@@ -388,6 +358,11 @@ public void sendAddEntity(Entity entity) {
      *                 The collection of Entity objects to add.
      */
     public void sendAddEntity(List<Entity> entities) {
+        // Trying to pass a non-ArrayList jams the receiving client and prevents it from ever receiving more packets.
+        // I have no idea why this happens.
```
psikomonkie, line 362: `TWGameManager:receiveEntityAdd(Packet, int)` has the following:
```
        // Need to use a new ArrayLiut to prevent a concurrent modification exception
        // when removing
        // illegal entities
```

psikomonkie (commented):
Should there be a limit to how many units can be added a time? If someone clicked the first unit, scrolled to the end, and shift clicked the other unit (selecting every unit in the list), what would happen to the server when it receives that packet?

pavelbraginskiy:
> Should there be a limit to how many units can be added a time? If someone clicked the first unit, scrolled to the end, and shift clicked the last unit (selecting every unit in the list), what would happen to the server when it receives that packet?

Surprisingly, everything copes just fine.

Processing all the units and sending them to the client takes a couple seconds, and the client receives them just fine. The GUI does block for a couple minutes if you add _all 10k units_ but it does finish before the heat death of the universe and then works normally.

I think it's fine, if a user gets upset that they locked up their client for a little bit because they tried adding thousands of units, that's a skill issue on their part.

Sleet01 (approved):
LGTM.
I myself was wondering about the MegaMek utility, but yeah, having this in MML will be great.
And I'll likely use it from time to time in setting up debugging games.

PR #6657, opened 2025-03-05
MekHQ Issue 6165: Canonicity should be determined by MUL ID.

psikomonkie:
Fixes Megamek/MekHQ#6165

Canonicity should be determined by MUL ID, not the Official Unit List which was deprecated.

### DIFF
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -10971,12 +10971,12 @@ public abstract class Entity extends TurnOrdered implements Transporter, Targeta
     /**
      * Is this unit canon in Battletech? For Meks and
      * other normal entities we should check the MUL ID.
-     * MUL ID is -1, not canon - Any entity type that
+     * MUL ID less than 1, not canon - Any entity type that
      * doesn't follow this rule should override this
      * @return true if this unit is canon
      */
     public boolean isCanon() {
-        return hasMulId() && getMulId() > -1;
+        return hasMulId() && getMulId() > 0;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


IllianiCBT (approved):

rjhancock:
Change from -1 to be greater than 0. More definitive as MUL ID will always be positive when present.

And thank you for using the depreciation tags.

Scoppio:
this is fine I guess? If a player adds a new unit and sets a MUL ID in it, this would them mark the unit as canon... I'm thinking if we could do something different for the long term.

HammerGS:
I agree with Luana around the MUL ID. When the unofficial unit folders got removed it means the anything in the remaining folders is Canon or our attempt to get them close to Canon.

Scoppio:
We have units without MUL ID that are canon... how this would handle those?

megamek/src/megamek/common/Entity.java
```diff
@@ -10968,12 +10968,15 @@ public boolean hasTAG() {
...
+     * doesn't follow this rule should override this
+     * @return true if this unit is canon
+     */
     public boolean isCanon() {
-        return canon;
-    }
-
-    public void setCanon(boolean canon) {
-        this.canon = canon;
+        return hasMulId() && getMulId() > 0;
```
Scoppio, line 10979: Unfortunatelly we do have many units with no MUL ID that are canon. This change/feature has this big blindspot and will set a bunch of canon units as "custom"

Scoppio (changes requested):
Unfortunatelly this change will cause problems

megamek/src/megamek/common/Entity.java
psikomonkie: This is how the MegaMek lobby and MegaMek unit search determines canonicity, this is just making it consistent. I agree it's not ideal but this isn't breaking anything more than it already is.

PR #6659, opened 2025-03-06
Feat/signed canon units

Scoppio:
# What it does?

Functional and full feature proof of concept for digitally signing canon units to allow players to have always a list of canon units without worry that a custom unit may overwrite their units, this also does not depend on the MUL ID being present and differently from the OUL it don't suffer with name collisions.

It also has functionalities like generating signing keys,  signing the files using github actions, etc.

## How it works?

The current way to say a unit is canon is looking up it's name in the Oficial Units List, it works fine but has a few edge cases that it can't deal, like a canonical unit being customized, which should remove its canonicity but it does not, chassis and model collision-If the custom unit has the same model and chassis it may be read as canon. The OUL can also be generated at any moment and it will add to it any custom units you have as canonical.

The signing of the files can make sure that only the files that were signed with the private key used during this dist build are recognized as canon. If the unit is customized in any way its signature becomes invalid since it will change the checksum of the unit file. This feature deals with the problems where a non-canonical unit would by mistake be considered canonical, it does not solve the problem of people with malicious intent as it.

## Ok, but HOW it works?!

The unit file is hashed, the hash is signed by a private key, the resulting binary is converted to base64 and saved as the `<signature>...</signature>` at the bottom of the file.

When loading a file, the signature is striped of it and hold in a temporary variable, then the file is hashed and this hash is loaded into the public key, the public key then compares it to the signature and says if it matches or not. If there is any change to the file then the check will fail. If the file has no signature or an invalid signature it will also fail.

There is a gradel task added in this PR that generates the private public key pair, you have to run it in your own computer, it will generate one binary public key, you add it into the resources folder as `public.key`, and it will generate two private keys, one `private.key` and one `private.key.b64`, you will copy the contents of the `private.key.b64` and add it into the repository settings > security > secrets and variables > actions and inside of it you click the button `new repository secret`, you name it `FILE_SIGNING_PRIVATE_KEY` and then paste the contents of the private.key.b64 in there.

If you want to run it locally, sure, just add the private key in a place you wont publish it by mistake in the repo.

# Help! I published the private key by mistake!!!

Ouch, ok... generate a new one, substitute it in the github repo secrets, add the public key to the repo substituting the old one, generate a new release to sign all files again, publish the new release.

## Future improvements

Allows players in a match during the handshake/connection to exchange/verify the public keys in each others client, and flag any player that has a different public key as using a non-compliant client. This is to my point of view imporant since people do play tournaments and disputes over prize money and imaginary internet points, so a modicum of security is welcome.

We could even do a kind of "forced update" where we can use the public key of each new release to be different, this way two different version programs can't play in the same lobby because their keys are different.


megamek/src/megamek/common/Entity.java
```diff
@@ -10968,11 +10969,28 @@ public boolean hasTAG() {
         return false;
     }
 
+    /**
+     * Get the source file for this unit.
+     * @return The source file for this unit, null if it does not exists.
+     */
+    public @Nullable File getSourceFile() {
+        return MekSummary.getSourceFile(getChassis() + " " + getModel());
+    }
```
Scoppio, line 10978: turns out this piece of code is unnecessary

megamek/src/megamek/common/MekFileParser.java
```diff
@@ -151,7 +153,18 @@ public Entity getEntity() {
...
+        tempFile.deleteOnExit();
+        try (BufferedReader reader = new BufferedReader(new InputStreamReader(ns));
+             BufferedWriter writer = new BufferedWriter(new FileWriter(tempFile))) {
+            String line;
+            while ((line = reader.readLine()) != null) {
+                writer.write(line);
+                writer.newLine();
+            }
+        }
+        InputStream is = new FileInputStream(tempFile);
```
Scoppio, line 167: I hate it as much as everybody else... best way would be to move this to its own small method and nobody talks about it ever again

megamek/src/megamek/common/MekFileParser.java
```diff
@@ -785,16 +801,17 @@ else if (ent instanceof Infantry) {
...
-        }
+//        ent.setCanon(false);// Guilty until proven innocent
+//        if (canonUnitNames == null) {
+//            initCanonUnitNames();
+//        }
+
+//        int index = Collections.binarySearch(canonUnitNames, ent.getShortNameRaw());
+//        if (index >= 0) {
+//            ent.setCanon(true);
+//        }
```
Scoppio, line 814: Just a proof of concept, this would be removed when integrating it in the project.

megamek/src/megamek/common/MekSummary.java
```diff
@@ -1382,6 +1382,18 @@ public String formatSUA(BattleForceSUA sua, String delimiter, ASSpecialAbilityCo
         return null;
     }
 
+    public static @Nullable File getSourceFile(String fullName) {
```
Scoppio, line 1385: Missing javadoc

megamek/src/megamek/utilities/FileSignerTool.java
```diff
@@ -0,0 +1,293 @@
...
+     */
+    public static void main(String[] args) throws Exception {
+        if (args.length < 3) {
+            System.out.println("Usage: FileSignerTool <private_key_path> <directory1> [<directory2> ...]");
+            System.exit(1);
+        }
+
+        File privateKeyFile = new File(args[0]);
+        if (!privateKeyFile.exists()) {
+            System.out.println("Private key file not found: " + privateKeyFile.getAbsolutePath());
```
Scoppio, line 48: Have to go with system out here, otherwise it wont print in the github actions

megamek/src/megamek/utilities/KeyGeneratorTool.java
```diff
@@ -0,0 +1,64 @@
...
+            System.exit(1);
+        }
+        generateKeyPair(new File(args[0]), new File(args[1]));
+        System.exit(0);
+    }
+
+    /**
+     * Generate a new key pair and save to files.
+     */
+    public static void generateKeyPair(File privateKeyFile, File publicKeyFile) throws Exception {
```
Scoppio, line 42: I tried generating the key pairs using only linux commands, but I could not, after 5+ hours of multiple tries, make it generate a pair that would work.

megamek/data/mekfiles/meks/Shrapnel/Vol 18/Thunder Hawk TDK-7ZEM.mtf
```diff
@@ -158,7 +158,7 @@ Foot Actuator
...
-capabilities:Dominating the Thunder Hawk's arsenal are three Norse Gauss rifles, accounting for over half its weight. These long-range weapons can incapacitate most foes with a single barrage. Ammunition is distributed throughout the torso to mitigate critical damage risks. However, its Achilles' heel lies in the extra-light Vlar fusion engine. While heavily armored, any breach in the torso significantly compromises the BattleMech�s structural integrity, rendering it inoperable.
+capabilities:Dominating the Thunder Hawk's arsenal are three Norse Gauss rifles, accounting for over half its weight. These long-range weapons can incapacitate most foes with a single barrage. Ammunition is distributed throughout the torso to mitigate critical damage risks. However, its Achilles' heel lies in the extra-light Vlar fusion engine. While heavily armored, any breach in the torso significantly compromises the BattleMech�s structural integrity, rendering it inoperable.
```
Scoppio, line 161:
```suggestion
capabilities:Dominating the Thunder Hawk's arsenal are three Norse Gauss rifles, accounting for over half its weight. These long-range weapons can incapacitate most foes with a single barrage. Ammunition is distributed throughout the torso to mitigate critical damage risks. However, its Achilles' heel lies in the extra-light Vlar fusion engine. While heavily armored, any breach in the torso significantly compromises the BattleMech's structural integrity, rendering it inoperable.
```

.github/workflows/nightly-ci.yml
```diff
@@ -39,6 +39,21 @@ jobs:
...
+      - name: Setup private key for file signing
+        run: |
+          mkdir -p megamek/keystore
+          echo "${{ secrets.FILE_SIGNING_PRIVATE_KEY }}" > megamek/keystore/private.key
+          chmod 600 megamek/keystore/private.key
+
+      - name: Setup private key for file signing
+        run: |
+          mkdir -p megamek/megamek/keystore
+          echo "${{ secrets.FILE_SIGNING_PRIVATE_KEY }}" | base64 --decode > megamek/megamek/keystore/private.key
```
Sleet01, line 51: I assume these should not have the same name.

.gitignore
```diff
@@ -14,6 +14,11 @@ Thumbs.db
 /megamek/classes/
 /megamek/bin/
 
+# Sigining keys
```
Sleet01, line 17: s/Sigining/Signing/

megamek/src/megamek/utilities/KeyGeneratorTool.java
Sleet01: Probably better to not introduce a platform dependency anyhow.
This seems fine.

Sleet01 (changes requested):
A couple spots need updating, and the POC section that's commented out would need to go.

HammerGS:
Given the dev is no longer with the project, and conflicts closing.

PR #6661, opened 2025-03-06
Weapon panel null check

HoneySkull:
This PR adds some additional null checking to the `WeaponPanel::displayMek()` method.  We don't expect an entity on the board with a position that is off the map.  This particular code is calculating net heat effects for a unit to update the mek display particularly for effects of terrain hexes such as lava and fire.  

If the entity's position is not on the board, it cannot be affected by heat from terrain hexes anyway.

We log a warning here because something has gone wrong with the entity if it's not offboard and has a position not on the board.

Additionally, cleaned up unused imports.


psikomonkie (approved):
Nice!

PR #6670, opened 2025-03-09
fix: makes gif writer finish the gif at victory phase

Scoppio:
# What it does?

Makes Gif Writer finish the image when the victory phase is arrived at.

Also moves the dave dialog inside the gif writer thread, I know it is not the "bestest or correctest" place to leave it, but this deals with the problem where it is running in a separate thread inside another component that we don't have direct access to it, so it may be weird to do it this way, but it is reasonable given that this thread was purpose built for this function. So it is actually working as intended here.


psikomonkie (approved):
Nice! I verified this successfully saves the gif on Windows. Excited to see this feature working!

PR #6672, opened 2025-03-09
[Issue 6671] Fix incorrect unit valuations due to negative part costs

### LINKED ISSUE #6671: [Issue] illogical contract payment
### Brief Description *

Terminated a contract as per Individual command rights in order to accept a new contract, the contract accepted massively overpays the company to the tune of 33 quadrillion per month.

### 3. Steps to Reproduce

No repro step available, occurred as a result of extended play.  Was not immediately noticed until halfway through the travel period.  Will be attaching several save files.  The first save file, the contract is available in the contract market.  During play, the various settings were negotiated to 100%, support over battle loss, infantry pay not a factor as per settings.

The second is after the contract was accepted and XP was spent on the characters across the company.  This occurred before I noticed the logic failure.

Other information available on request.

### Attach Files

[The Gypsy Consortium30360218.cpnx.gz](https://github.com/user-attachments/files/19052621/The.Gypsy.Consortium30360218.cpnx.gz)

[The Gypsy Consortium30360220.cpnx.gz](https://github.com/user-attachments/files/19052640/The.Gypsy.Consortium30360220.cpnx.gz)

### Severity *

### MekHQ Suite Version *

0.50.03

### Operating System *

Windows 11 10.0 amd 64

### Java Version *

Eclipse Adoptium 17.0.9

### Final Verification

Dark-Hobbit:
Fixes #6671 

Currently we allow part costs to be negative, and in some unit cost calculation methods for aero units there was a supposed typo that multiplied the final unit cost by the part's cost if it was negative - wildly inflating the final valuations. In this specific case, it was driven by negative SI.

I have brought the aero cost calculation logic in line with all other methods, and put a safeguard treating negative SI as zero. However, we should probably decide what to do with negative unit part prices in principle.

### DIFF
--- a/megamek/src/megamek/common/cost/BattleArmorCostCalculator.java
+++ b/megamek/src/megamek/common/cost/BattleArmorCostCalculator.java
@@ -76,6 +76,9 @@ public class BattleArmorCostCalculator {
 
         costs[idx++] = (baseArmorCost * battleArmor.getOArmor(BattleArmor.LOC_TROOPER_1));
 
+        // For all additive costs - replace negatives with 0 to separate from multipliers
+        CostCalculator.removeNegativeAdditiveCosts(costs);
+
         // training cost and clan mod
         if (includeTrainingAndClan) {
             if (battleArmor.isClan()) {
@@ -89,14 +92,10 @@ public class BattleArmorCostCalculator {
 
         // TODO : we do not track the modular weapons mount for 1000 C-bills in the unit
         // files
-        costs[idx++] = CostCalculator.getWeaponsAndEquipmentCost(battleArmor, ignoreAmmo);
-        costs[idx++] = -battleArmor.getSquadSize();
+        costs[idx++] = Math.max(0, CostCalculator.getWeaponsAndEquipmentCost(battleArmor, ignoreAmmo));
+        costs[idx] = -battleArmor.getSquadSize();
 
-        double cost = 0; // calculate the total
-        for (int x = 0; x < idx; x++) {
-            cost += costs[x];
-        }
-
+        double cost = CostCalculator.calculateCost(costs);
         String[] systemNames = { "Chassis", "Jumping/VTOL/UMU", "Ground Movement", "Manipulators", "Armor",
                 "Clan Structure Multiplier", "Training", "Equipment", "Troopers" };
         CostCalculator.fillInReport(costReport, battleArmor, ignoreAmmo, systemNames, 7, cost, costs);
--- a/megamek/src/megamek/common/cost/BattleArmorCostCalculator.java
+++ b/megamek/src/megamek/common/cost/BattleArmorCostCalculator.java
@@ -94,11 +94,7 @@ public class BattleArmorCostCalculator {
 
         double cost = 0; // calculate the total
         for (int x = 0; x < idx; x++) {
-            if (costs[x] < 0) {
-                cost *= -costs[x];
-            } else {
-                cost += costs[x];
-            }
+            cost += costs[x];
         }
 
         String[] systemNames = { "Chassis", "Jumping/VTOL/UMU", "Ground Movement", "Manipulators", "Armor",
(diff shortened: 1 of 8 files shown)


Scoppio:
@SJuliez was the one to implement that equation 3 years ago, I have no way to understand why it was done the way it was done without studying more of the build-process, so I would like to wait for Juliez review here before merging it.

HoneySkull (changes requested):
I may have approved this prematurely - is it possible that the costs in the cost array are being set as negative to differentiate between incremental cost and cost multipliers?

SJuliez:
@HoneySkull Yes precisely.

megamek/src/megamek/common/cost/CostCalculator.java
```diff
@@ -42,11 +42,7 @@ public class CostCalculator {
     static double calculateCost(double[] costs) {
         double cost = 0;
         for (double v : costs) {
-            if (v < 0) {
-                cost *= -v;
-            } else {
-                cost += v;
-            }
+            cost += v;
```
SJuliez, line 45: The javadoc for this method a few lines above clearly says what negative values are being used for. This change cannot be made in this form.

megamek/src/megamek/common/cost/BattleArmorCostCalculator.java
```diff
@@ -94,11 +94,7 @@ public static double calculateCost(BattleArmor battleArmor, CalculationReport co
 
         double cost = 0; // calculate the total
         for (int x = 0; x < idx; x++) {
-            if (costs[x] < 0) {
-                cost *= -costs[x];
-            } else {
-                cost += costs[x];
-            }
+            cost += costs[x];
```
SJuliez, line 97: See the other instance of this change.

SJuliez (changes requested):
The marked changes go against what the methods are supposed to do. From what I can see this will break aero cost calculation.

Dark-Hobbit:
Thanks for the clarification. I see that a significant portion of unit cost calculators (warships, combat vehicles, fixed wing, jumpships, Meks, space stations) use a different approach, where all part costs are summed and what is supposed to be a multiplicator coefficient is explicitely used to multiply the final result. Which approach is the preferred one?

So far I can see three different approaches to solving this issue:
1. Wrap cost estimation for every individual part that is not a multiplicator in Math.max(x, 0) - prevents them from being negative and solves this specific issue, but still leaves us with two different approaches used throughout the code.
2. Switch all unit cost calculators to one approach - can reduce the future confusion, regardless of the choice (I personally think the one explicetely multiplying by the coefficient where necessary is a safer bet).
3. Change the base CostCalculator to accept two arrays (incremental costs and cost multipliers), or a map of coefficient and its type (incremental/multiplier) - actually removes two risks: negative part costs and accidentally placing an incremental cost after a cost multiplier in the array, which would break the current logic.

What would be your opinion on this? @SJuliez

SJuliez:
Unfortunately I dont remember why some unit types use the standard cost calc and some dont. One possibility would be that there are some actual negative costs (deductions) on unit types. But I don't really remember anything like that.
It is also a matter of the effort one wants to spend here. Clamping additive values to minimum 0 (option 1) is ok. This solves issues with negative SI and, possibly, also armor values.
If the calculations could be truly unified using a map (or a List of record <Double, Boolean> values) then that would be ideal, but it's going to be some effort. The value in unified cost calculation is also that a unified cost report can be generated in MML. IIrc this is done in MekCostCalculator by calling the standard cost calc, but only to create the report, not the actual cost.
So unifying it in the sense of calling a central method that does the calculation and report would definitely be good, but is going to be quite some work. Maybe not the scope of this PR.

Dark-Hobbit:
I made a very simple safeguard in a form of a simple method for ensuring non-negative values in the array. It gets applied before any multipliers are added to the array and only in the classes using the standard CostCalculator for calculating the final costs. That means that any class manually processing cost values will be unaffected in case there are some true negative part prices (which I couldn't find but still).

Does that look like a reasonable enough solution to you? I might still look at unifying the calculation approaches but, as you said, it might be out of scope for the original issue.  @SJuliez

SJuliez (approved):
Looks good to me.

HoneySkull (approved):

PR #5924, opened 2024-08-19
custom deployment zones

NickAragua:
Implements definition and assignment of custom deployment zones. Zones can be defined in the map editor by selecting the "deployment zone" terrain:
image

They can then be placed:
image

Finally, they can be assigned in the lobby to a player or unit:
image

Still a few tasks left:
- clear selected custom deployment zones when map selection changes in lobby
- easier configuration of multiple DZs in one hex

PR #6675, opened 2025-03-10
fix: use provided game year

Scoppio:
Game Year was being left unchanged when generating reconfiguration parameters, meaning it would load 99999 as the allowedYear for the reconfiguration parameters

- closes [#5924](https://github.com/MegaMek/mekhq/issues/5924)


Sleet01:
This shouldn't be necessary; the TeamLoadoutGenerator instance is supposed to get the year from its GameOptions reference, and uses that value to filter the bin types that it generates from the ReconfigurationParameters (which is more concerned with specifics about the fight rather than munition availability).

Sleet01:
As an aside, please assign bugs that you are working on to yourself.

HoneySkull (approved):

PR #6677, opened 2025-03-11
Better indicate effect of RISC ECS

pavelbraginskiy:
Adds the +6 heat sinking from RISC ECS to heat readouts. Compare the heat profiles of these two sheets:

image
image

Official sheets don't include the heat profile so I don't think this needs CGL approval.


Sleet01 (approved):
Looks good.  I think familiarity with RISC ECS rules will be required either way, but this is definitely clearer.

PR #6679, opened 2025-03-11
fix: cube-coords error with conversion to offset

Scoppio:
# What it does?

Fix cube coordinates so it converts correctly from CubeCoords to Coords


Sleet01 (approved):
LGTM!

PR #6692, opened 2025-03-13
refactor: cleaned up the mp candidate evaluation on longest path finder

Scoppio:
# What does it do?

cleaned up the mp candidate evaluation and transformed the error log into a debug log as there was no benefit at logging something that cant be acted upon

It was basically a click of the mouse and using the `refactor` action from the IDE, I also changed the `while(!v.empty())` for a `if (v.empty())` conditional check and early exit as it never did loop, it was just a flow control.


Sleet01 (approved):
LGTM

Issue #6259, opened 2024-12-12
Logistics Artifact 12/06 - "Returning no deployment position; THIS IS BAD!" in log, game was unable to process beginning of next round, hung game


OrbMonky:
### Prerequisites and Pre-Issue Checklist

- [X] I'm reporting the issue to the correct repository:
- [MegaMek](https://github.com/MegaMek/megamek/issues)
- [MegaMekLab](https://github.com/MegaMek/megameklab/issues)
- [MekHQ](https://github.com/MegaMek/mekhq/issues)

- [X] I've tested the issue against at least the latest MILESTONE version
- [X] I've asked on the MegaMek Discord about the error
- [X] I've reviewed the BattleTech rules and MegaMek documentation, and I've confirmed that something isn't working as intended.
- [X] I've searched the Github tracker and haven't found the issue listed

### Severity *

Critical (Game-breaking/Crash): The game crashes or a core feature (like saving, loading, or network connection) is completely unusable.

### Brief Description *

It appears an infantry platoon did not have a deployment coordinate available/present? Not sure what happened, it just gave an error and I couldn't proceed past the beginning report phase of the next round. I could still interact and open/close the report but there was no ability to move to the next step.

[logs.zip](https://github.com/user-attachments/files/18106321/logs.zip)

### Steps to Reproduce

Unable to determine

### Operating System *

Windows 10

### Java Version *

17

### MegaMek Suite Version *

None

### Custom MegaMek Version

December 6th Logistics Artifact

### Final Checklist

- [X] I've checked to make sure that this issue has not already been filed
- [X] I'm reporting only one issue in this ticket for clarity and focus

OrbMonky:
Upon reloading it was able to proceed, but for the first time I've ever seen, Princess seems to be in I think 'air mech' mode?
![image](https://github.com/user-attachments/assets/76a72fc4-347b-4958-9883-b356b6ebd5d6)

HammerGS:
A huge number of these

```
23:59:50,737 ERROR [megamek.common.Game] {Lyran Commonwealth OpFor Turn 3 Calc Thread}
megamek.common.Game.getEntitiesVector(Game.java:1479) - Foot Platoon (Laser) (Lyran Commonwealth OpFor) is not in Coords (0, 13); 0114!
```

and

```
23:59:50,734 ERROR [megamek.client.bot.BotClient] {Lyran Commonwealth OpFor Turn 3 Calc Thread}
[...]

PR #6694, opened 2025-03-13
fix: gracefully close gifWritter, and gracefully initialize it on a second run

Scoppio:
# What it does?

This change gracefully closes gifWritter if the client is suddenly disconnected or killed, and gracefully initialize it on a second run, as it was only checking if it was not null when initializing, now it checks if not null or if not alive, if not alive it will create a new gifWriterThread. This prevents MekHQ from freezing up due to the rogue Dialog to save the GIF that opens with reference to a dead client.

- Closes [#6259](https://github.com/MegaMek/mekhq/issues/6259)


Sleet01 (approved):
LGTM

PR #6697, opened 2025-03-14
fix: Correct the license for AI modules to have the correct license header

Scoppio:
Updating the license header on those files in the AI module


HoneySkull:
Several of the license changes refer to MekHQ.  Being this is the megamek repository - does this matter?  (I'm asking because I don't know and if it has any unintended consequences as far as licensing goes.

Scoppio:
> Several of the license changes refer to MekHQ. Being this is the megamek repository - does this matter? (I'm asking because I don't know and if it has any unintended consequences as far as licensing goes.

You are right, I just corrected it.

Sleet01 (approved):
LGTM

PR #6703, opened 2025-03-15
Updated SkillLevel to Use Arbitrary Values Instead of Ordinal

IllianiCBT:
- Added experience levels integer to SkillLevel enums

### Dev Notes
Based on feedback in https://github.com/MegaMek/mekhq/pull/6273 and a following discussion with @Scoppio it was noted that relying on a specific ordinal value in an enum is a weak link. Instead, assigning arbitrary values is preferred as this means that should the enum order ever change the 'value' of the enum remains consistent. While I can't imagine anyone is planning to reorder `SkillLevel` any time soon, I figured we might as well start as we mean to go on.


Sleet01 (approved):
Nice, looks good!

PR #6704, opened 2025-03-15
Add tactics command bonus to deploy turn init

Sleet01:
Loosen restrictions on applying init bonuses so that units deploying this turn benefit their side.
In a future PR I will move all command bonus calculations into Entity, Crew, Equipment, etc. so that only one call is required to determine the current Initiative bonus that a unit provides.

Testing:
- Ran all three projects' unit tests
- Tested with varying initiative bonuses including Command Consoles and Tactics-based Command Initiative bonuses, deploying on different turns.

This PR also replaces a number of errant tabs with spaces.


IllianiCBT (approved):
Thanks again :)

PR #6705, opened 2025-03-15
Fix #6301: phases are not skipped by /victory even with phase skipping enabled

### LINKED ISSUE #6301: 12/16 Logistics Artifact / 0.50.02 SNAPSHOT - /victory not skipping remaining phases even with phase skipping enabled

- [X] I'm reporting the issue to the correct repository:

- [X] I've tested the issue against at least the latest MILESTONE version
- [X] I've asked on the MegaMek Discord about the error
- [X] I've reviewed the BattleTech rules and MegaMek documentation, and I've confirmed that something isn't working as intended.
- [X] I've searched the Github tracker and haven't found the issue listed

### Severity *

### Brief Description *

These are the settings in MM
image

In the attached save, /victory was ran at the start of the phase just after the report was dismissed. The victory dialogue appeared to process correctly but it went to the fire phase and fire was prosecuted.

Here is a save in which this happened initially and after reload;
[logs.zip](https://github.com/user-attachments/files/18215563/logs.zip)
Round 11 save should do it, but I included 10 as well in case for testing

### Steps to Reproduce

_No response_

### Operating System *

Windows 10

### Java Version *

17

### MegaMek Suite Version *

None

### Custom MegaMek Version

12/16 Logistics Artifact / 0.50.02 SNAPSHOT

### Attach Files

_No response_

### Final Checklist

- [X] I've checked to make sure that this issue has not already been filed
- [X] I'm reporting only one issue in this ticket for clarity and focus

Sleet01:
Adds immediate check to skip all turns and remaining phases _if_ user has "skip all phases when forced victory occurs" enabled, types `/victory`, and gets all other players to accept defeat.

Note:
1. There may be a race condition between registering the forced victory state and bot clients returning their /defeat packets; I'm not sure if there is a good solution for this, but the following note makes it less of an issue.
2. Because of the way the `/victory` command works, we _should_ get all the triggers for a forced victory set prior to actually checking it and clearing the remaining turns because...
3. Actually checking for forced victory and clearing all remaining phases and turns requires a trigger of some kind, which in our case will be ending one unit's turn.

Testing:
- Tested `/victory` command in multiple phases and with both settings of the "skip all phases" option.

Close #6301

### DIFF
--- a/megamek/src/megamek/server/commands/VictoryCommand.java
+++ b/megamek/src/megamek/server/commands/VictoryCommand.java
@@ -14,6 +14,7 @@
 package megamek.server.commands;
 
 import megamek.common.Player;
+import megamek.common.options.OptionsConstants;
 import megamek.server.Server;
 import megamek.server.totalwarfare.TWGameManager;
 
--- a/megamek/src/megamek/server/totalwarfare/TWGameManager.java
+++ b/megamek/src/megamek/server/totalwarfare/TWGameManager.java
@@ -1360,6 +1360,8 @@ public class TWGameManager extends AbstractGameManager {
      * round.
      */
     public void forceVictory(Player victor, boolean endImmediately, boolean ignorePlayerVotes) {
+        // endImmediately if passed true, or if game option set
+        endImmediately |= game.getOptions().booleanOption(OptionsConstants.VICTORY_SKIP_FORCED_VICTORY);
         game.setEndImmediately(endImmediately);
         game.setIgnorePlayerDefeatVotes(ignorePlayerVotes);
         game.setForceVictory(true);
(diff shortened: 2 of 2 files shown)


Sleet01:
> The changes here look good to me. Do these changes behave as expected for multiplayer human? (i.e. if two or more human players are in a phase and one signals /victory and another doesn't agree, will it still force the victory?)

The other player(s) all have to use the `/defeat` command before the current unit's turn in the current phase ends, or nothing happens and the game moves on to the next unit's turn.
The possible race condition I saw was from bots on the same machine replying with `/defeat` prior to the game completing the handling of the initial `/victory` packet, but that may be due to the debug environment and wouldn't affect human players.

I haven't run a test with two human players yet, I'll see if I can set that up sometime today.

Sleet01:
@HoneySkull Confirmed that the turn just continues if the opponent(s) doesn't submit a `/defeat` prior to the current unit's turn ending.

HoneySkull (approved):

PR #6710, opened 2025-03-16
Fix/unit state serde

Scoppio:
# What it does?

Removes useless optimization code focused for fine tuning Princess: Because Princess model is too simple, the optimization was barely functional, mostly useless. It was removed as evolving the current Princess isn't a priority anymore.

A bug where the chassis name was written twice was fixed, it is annoying but this does not make the logs useless, I just need to make a pre-processing to fix it.


Sleet01 (approved):
Looks good to me.

Issue #6711, opened 2025-03-16
[Issue] Infantry appears to always survive transport destruction


BlueThing00:
### Brief Description *

I just fought a battle where I killed 4 transports containing infantry.  In no case did the Round Report indicate that the infantry rolled to avoid death.  In all four cases, the infantry survived.

For example, this is the vehicle destruction from Round 1:
![Image](https://github.com/user-attachments/assets/fcb7fed1-c685-4701-852d-25f725e83d5e)
![Image](https://github.com/user-attachments/assets/65880f61-d9f9-4ab9-90cd-041deaf71874)

This is RAW, per TW 223:
![Image](https://github.com/user-attachments/assets/c3bd65ea-6cd5-4de1-b4be-43330a7aba3a)

### 3. Steps to Reproduce

1. Destroy a transport
2. Infantry survives

### Attach Files

[Round-10-autosave_2025-03-16_07-02-13.sav.gz](https://github.com/user-attachments/files/19272395/Round-10-autosave_2025-03-16_07-02-13.sav.gz)

### Severity *

Medium (Gameplay Limitation): Non-core functionality is impaired, providing a suboptimal but playable experience.

### MegaMek Suite Version *

50.03

### Operating System *

Windows 11

### Java Version *

Adoptium 17.0.11

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have included all necessary information and files to help reproduce this issue
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

Sleet01:
I believe we should now be destroying all the crunchies appropriately, but I will double-check, and look at adding a succinct log to show which form of destruction is being visited on them (if not each individual roll).

PR #6714, opened 2025-03-17
Fix 6711: Add VTOL and WiGE airborne destruction as unsurviveable

Sleet01:
Also explicitly check for infantry survival when carrier destroyed otherwise.

Previously we assumed that the destruction checks for a carried unit would only apply to Infantry, as they were the most likely to be carried by a destroyed unit.
We also assumed that any Tank-derived unit was being destroyed on the ground; in fact we need to check if the unit was destroyed in the air first.

This PR adds an explicit check for Infantry vs other carried unit types, which should only be destroyed in specific circumstances handled elsewhere (or by the existing Airborne) check.
It also adds the missing destruction checks for infantry units per TW pg. 223.

Testing:
- Tested with loaded units in a variety of situations.

Close: #6711


HoneySkull (approved):
Looks good to me.

Issue #1769, opened 2020-03-15
0.47.5-SNAPSHOT - Board Editor RFE: Add Close Button to New Board Dialog


Windchild292:
Currently, if you open the new board dialog it will always regenerate the map. That could be undesirable if you misclick it.

PR #6728, opened 2025-03-19
Merge with MML #1769: utility method in Tank to determine if location is side location

SJuliez:
Adds a utility method to Tank to determine if a location is a side location (LS/RS, FRLS/RS, RRLS/RS)
This PR is required for Megamek/megameklab#1769


Sleet01 (approved):
LGTM!

PR #6729, opened 2025-03-20
Removed deprecated logger method.

rjhancock:
While updating MekHQ deprecation methods, ran into issues where logger.error was opening the dialog instead of logging the error.


Sleet01 (approved):
LGTM

PR #6735, opened 2025-03-20
Unify MHQ Faction and FactionRecord and use yaml faction data files

SJuliez:
This PR together with MML and MHQ PRs aims to unify the faction data of MHQ and MM on the MM side to make it available everywhere and use individual yaml files instead of a single huge xml file. 

- For now, the old classes are preserved and still used; they only get their data from the new Faction2 class (the best name I could come up with) instead of loading it from factions.xml (=2 separate and different files in MHQ and MM)
- FactionRecord also loads complicated era-based RAT data from other files, this is unchanged; it doesnt seem useful to place this inside the yamls as this would make the yamls unwieldy
- Factions2 reads its data from the data/universe/factions and data/universe/commands folders. These need to be replicated in MML and MHQ so long as we dont have a unified data repo.
- "commands" are the subfactions (those with a dot in the name), e.g. the Hesperus Guards; in other words, combat groups rather than political entities; sorting their yaml data in their own folder is purely for cleanliness, it doesnt matter where which yaml file is. In the future, it's possible that additional commands data will make separate classes (in addition to separate yamls) better, but for now it's still good
- data is also read from the user directory, if available, so [userdir]/data/universe/factions or /commands; factions there will be added; if they have the same key as a canon faction, they replace the original
- "startingplanet" has been renamed "capital" as it was used for that purpose only

This is pretty hefty for review, but when filtering out the .yml files, it's only a handful of classes. Ideally, the whole thing should have no visible effect in MM, MML or MHQ except fix a bug with MML faction selection which omitted a bunch of factions.


IllianiCBT:
Thank you so much for doing this, Juli. As someone who works with factions more than a little, I just wanted to say how appreciated this work is.

AaronGullickson:
Are any of the faction codes in MHQ actually different? Because all of the planetary data depends on these....

SJuliez:
The faction codes are generally equal aside from some MHQ factions that are missing in the RATs and vice versa and a very small number of, well, data bugs, e.g. we have three versions of the Tamar Pact with different codes.

rjhancock:
Taking a quick glance, the era mods might be an issue. Which mod works with which era? Needs labels.

SuperStucco:
Don't suppose you could update the Bandit Caste (BAN code) to include the full range of equipment ratings?  While under-equipped - top-rated Bandit Caste is certainly not the same as top-rated Clan forces - there should still be a range of availability from the Society-derived forces fielded by The Jaguar and Tanite/ex-Burrock forces to the dregs that are barely getting by.
As it stands they are coded for 'Solahma' only, which also kind of conflicts with the availability XMLs which provide two levels.

AaronGullickson:
> The faction codes are generally equal aside from some MHQ factions that are missing in the RATs and vice versa and a very small number of, well, data bugs, e.g. we have three versions of the Tamar Pact with different codes.

Whoops meant to quote reply and I edited above. If you are editing the factions, we should ensure these are not in use by planetary systems.

HammerGS:
Hi @SJuliez 

You seem to be on a break but their is interest in merging this if its ready to go, can you advise?

Cheers
Hammer

AaronGullickson:
Well, there are conflicts so no merge at the moment. Also, I need to make some changes to factions for the big SMACK project (which I have not been able to find time for). I don't think that will likely affect this much as it is mostly about removing unused factions but its probably worth checking.

SJuliez:
Replaced by #7246

PR #6743, opened 2025-03-21
Formatting, remove armored component weight from aero weight calc (MML#1173)

SJuliez:
comes together with Megamek/megameklab#1773 but they're not dependent. This is a bit of formatting and it removes the armored component weight from aero weight calculation as only BM/IM can have armored components and only TestMek will ever return anything other than 0.

### DIFF
--- a/megamek/src/megamek/common/verifier/TestEntity.java
+++ b/megamek/src/megamek/common/verifier/TestEntity.java
@@ -1933,8 +1933,12 @@ public abstract class TestEntity implements TestEntityOption {
                 + getEntity().getYear() + ")\n";
     }
 
+    /**
+     * @return The total additional weight due to critical slot armoring. Does not include the weight of the equipment
+     * itself.
+     */
     public double getArmoredComponentWeight() {
-        return 0.0;
+        return 0;
     }
 
     public static boolean usesKgStandard(Entity entity) {
--- a/megamek/src/megamek/common/verifier/TestMek.java
+++ b/megamek/src/megamek/common/verifier/TestMek.java
@@ -786,21 +786,16 @@ public class TestMek extends TestEntity {
         return "Mek: " + mek.getDisplayName();
     }
 
-    /**
-     * calculates the total weight of all armored components.
-     */
     @Override
     public double getArmoredComponentWeight() {
         double weight = 0.0;
         double cockpitWeight = 0.0;
-        for (int location = Mek.LOC_HEAD; location < mek.locations(); location++) {
+        for (int location = 0; location < mek.locations(); location++) {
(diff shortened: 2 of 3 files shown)


Sleet01 (approved):
LGTM

megamek/src/megamek/common/verifier/TestMek.java
```diff
@@ -786,21 +786,16 @@ public String getName() {
...
 
-    /**
-     * calculates the total weight of all armored components.
-     */
     @Override
     public double getArmoredComponentWeight() {
         double weight = 0.0;
         double cockpitWeight = 0.0;
-        for (int location = Mek.LOC_HEAD; location < mek.locations(); location++) {
+        for (int location = 0; location < mek.locations(); location++) {
```
rjhancock, line 793: Why the change from an ENUM to an Integer?

megamek/src/megamek/common/verifier/TestEntity.java
```diff
@@ -1933,8 +1933,12 @@ public String printTechLevel() {
                 + getEntity().getYear() + ")\n";
     }
 
+    /**
+     * @return The total additional weight due to critical slot armoring. Does not include the weight of the equipment
+     * itself.
+     */
     public double getArmoredComponentWeight() {
-        return 0.0;
+        return 0;
```
rjhancock, line 1941: Internally I know it'll convert, but why removal of the `.0`?

megamek/src/megamek/common/verifier/TestMek.java
SJuliez: It was not an enum, just a number. In this spot, the LOC_HEAD is meaningless, it actually wants to cycle from 0 to the number of locations. Having LOC_HEAD there is not in any way more informational, it's nonsense.

megamek/src/megamek/common/verifier/TestEntity.java
SJuliez: what is 0.0 good for? Zero is zero, doesnt get better as 0.0. Why have letters that are meaningless and not particularly informative?

megamek/src/megamek/common/verifier/TestMek.java
SJuliez: Put differently: I have to hover the LOC_HEAD to see that it's 0 and this actually cycles all locations. LOC_HEAD could be 2, then this for() would be special. But it isnt special, and there is no reason not to see that immediately.

rjhancock: Ahh. So it was a static value. Was just curious. These questions aren't meant to stop the merge. Just curiosity.

PR #6757, opened 2025-03-24
Restrict user dir loading to the correct subfolders

SJuliez:
This was requested by Hammer a while ago: loading content from the userdir now requires it to be placed in the same folder structure that is used internally even for those files where it wouldnt technically be necessary like .mtf or .ttf. Units must now be somewhere in userdir/data/mekfiles etc.

### DIFF
--- a/megamek/docs/UserDirHelp.html
+++ b/megamek/docs/UserDirHelp.html
@@ -6,9 +6,9 @@
 
 <P>How to place files within the user data directory:</P>
 <UL>
-    <LI><B>Fonts</B> can be placed anywhere in the user data directory (i.e., in the directory itself or in any subfolder)
+    <LI><B>Fonts</B> must be placed in &lt;user data directory&gt;/data/fonts/. In this subfolder, further subfolders can be used (optional).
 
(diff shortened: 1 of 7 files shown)


pavelbraginskiy:
Wait this sucks. Why would we want that.

rjhancock (approved):

Sleet01:
This is going to cause massive user grief, and that's just describing how I _personally_ will feel.
If we don't provide users with a way to set this up, we're going to get tons of bug reports when their existing user folders stop working for them.

What is the desired end result from this change?

HammerGS:
Simple, we have players that have dropped stuff assuming it works where ever it works then it doesn't. They come to Discord and we have to spend a lot of time troubleshooting. Where if we had a uniform consistent stucture the question becomes. Did you put it in this folder.

It also maintains a consistency with the existing folders which lets us easier explain how to set up the folders.

HammerGS:
Going to convert to a draft, till we discuss.

gsparks3:
At a minimum, if a matching folder structure is to be required, I would want some way of automatically _generating_ the correct folder structure within a selected user folder.

PR #6793, opened 2025-03-30
Update the Packet Class with type specific getters.

rjhancock:
This PR updated the Packet Class with type specific getters in both single (nullable) form and collection form. The old `getObject` method has been deprecated for purposes of facilitating this with the intention of making it private only.

`getObject` was adjusted to include out of bounds checking on the internal data array.

All other new methods include `instanceof` checking before either returning the value or creating an internal collection and returning it.


rjhancock:
Too many things are breaking by being strict. Only way to solve this would be a network re-write entirely.

PR #6807, opened 2025-04-01
Deprecate Aero#initializeSI

pavelbraginskiy:
- Aero#initializeSI is identical to Aero#setOSI, so deprecate the former in favor of the latter.
- Rename `Aero#set0SI` to `Aero#setOSI`. Why was there a zero there.
- Call autoSetSI when weight or engine rating changes.

### DIFF
--- a/megamek/src/megamek/common/Aero.java
+++ b/megamek/src/megamek/common/Aero.java
@@ -724,7 +724,7 @@ public abstract class Aero extends Entity implements IAero, IBomber {
     /**
      * @deprecated use {@link Aero#setOSI(int)} instead
      */
-    @Deprecated
+    @Deprecated(since = "0.50.05")
     public void initializeSI(int val) {
         setOSI(val);
     }
--- a/megamek/src/megamek/common/Aero.java
+++ b/megamek/src/megamek/common/Aero.java
@@ -686,7 +686,12 @@ public abstract class Aero extends Entity implements IAero, IBomber {
         currentDamage = i;
     }
 
-    public void set0SI(int si) {
+    /**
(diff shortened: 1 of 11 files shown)


megamek/src/megamek/common/Aero.java
```diff
@@ -716,9 +721,12 @@ public void autoSetFatalThresh() {
...
     }
 
+    /**
+     * @deprecated use {@link Aero#setOSI(int)} instead
+     */
+    @Deprecated(since = "0.50.05")
     public void initializeSI(int val) {
-        orig_structIntegrity = val;
-        setSI(val);
+        setOSI(val);
```
IllianiCBT, line 729: Something you could do here, if you care, is removed the deprecated code and just have the method call `setOSI` instead. That way, in the unlikely event it's being used elsewhere that your IDE didn't catch, you don't end up in a weird situation where things are acting odd and you can't figure out why.

megamek/src/megamek/common/loaders/BLKSpaceStationFile.java
```diff
@@ -94,7 +94,7 @@ public Entity getEntity() throws EntityLoadingException {
         if (!dataFile.exists("structural_integrity")) {
             throw new EntityLoadingException("Could not find structual integrity block.");
         }
-        a.set0SI(dataFile.getDataAsInt("structural_integrity")[0]);
```
IllianiCBT, line 97: We will never know why that was an `0` instead of `O`, but man was it a weird choice.

IllianiCBT (approved):
LGTM

PR #6816, opened 2025-04-02
parse mek/blk from string

exeea:
adds the possibility to read a saved state from a string using the original parse function.
To keep a single parse function it detects and fakes the filename, then streams the string.


rjhancock:
What's the use case for this? Is this something that will be in use in this or another part of the project or just a "nice to have" feature that doesn't currently have a use?

Just asking.

exeea:
@rjhancock is needed for implementation of the undo/redo/restore in MML.

PR #6820, opened 2025-04-03
Feat/caspar ai

Scoppio:
Adds basic stuff needed for Caspar, and updates the games action log.


megamek/src/megamek/codeUtilities/MathUtility.java
```diff
@@ -33,7 +33,7 @@ private MathUtility() {
      *         maximum. Otherwise, this will return the rounded integer between the
      *         two points
      */
-    public static int lerp(final int min, final int max, final double f) {
+    public static int lerp(int min, int max, double f) {
```
Scoppio, line 36: `final` is useless in arguments, so I just removed it.

rjhancock: It's actually not useless. It prevents the programmer from re-assigning a value to them.

megamek/src/megamek/codeUtilities/MathUtility.java
```diff
@@ -133,8 +133,28 @@ public static float clamp(final float value, final float min, final float max) {
...
+    }
+
+    /**
+     * Clamp a value between 0 and 1.0 (inclusive).
+     * @param value the float value to clamp
+     * @return The value clamped
+     */
+    public static double clamp01(double value) {
+        return Math.min(Math.max(value, 0f), 1f);
+    }
```
rjhancock, line 158: Although meant to be default clamp ranges, why not just be helper methods for the regular `clamp` methods and pass in the default 0 and 1 values?

rjhancock (commented):
How often will the binary files change?

Other than a few comments and a few classes could use (not required) some top level JavaDocs, it looks good for the initial PR.

megamek/unittests/megamek/ai/neuralnetwork/NeuralNetworkTest.java
```diff
@@ -0,0 +1,80 @@
+package megamek.ai.neuralnetwork;
+
+import megamek.client.bot.caspar.ClassificationScore;
+import megamek.client.bot.caspar.MovementClassification;
+import megamek.common.Coords;
+import org.junit.jupiter.api.AfterEach;
```
rjhancock, line 6: Copyright....

megamek/src/megamek/codeUtilities/MathUtility.java
Scoppio: its minimalistic enough that it makes no difference.

Scoppio:
> How often will the binary files change?
> 
> Other than a few comments and a few classes could use (not required) some top level JavaDocs, it looks good for the initial PR.

It depends... Ideally I want to create at least 5 AI models (Beginer, easy, normal, hard, hardcore), and maybe two or three "flavors" for each of those too would be interesting (if viable).

I may try to make them read directly from zip files, or at least tar all the model files into one thing instead? Need to check Tensorflow's API. But the models sohuld not change alot once done, unless someone is doing anything with them, like adding new stuff or tuning, in any way, its usually just a file replace.

rjhancock:
Alright. The speed of change was my concern as the way Git handles binary is it just re-adds the entire file. Not changing often, not really an issue.

rjhancock (approved):

HammerGS:
@Scoppio Failing checks

`The job fails due to multiple compilation errors in /src/megamek/ai/dataset/UnitPath.java. The issue is that the UnitAction class is missing the methods being called, such as currentCubePosition(), finalCubePosition(), currentPosition(), finalPosition() and others.`

HammerGS:
Given the dev has left, and the huge number of conflicts closing.

PR #6821, opened 2025-04-03
No Clone Method - Scenario V1

rjhancock:
Adds the `clone()` method as is required by it's parent classes.

### DIFF
--- a/megamek/src/megamek/common/AmmoType.java
+++ b/megamek/src/megamek/common/AmmoType.java
@@ -74,10 +74,12 @@ public class AmmoType extends EquipmentType {
     public static final int T_EXLRM = 37;
     public static final int T_APGAUSS = 38;
     public static final int T_MAGSHOT = 39;
+
     // Removed
     // public static final int T_PXLRM = 40;
     // public static final int T_HSRM = 41;
     // public static final int T_MRM_STREAK = 42;
+
     public static final int T_MPOD = 43;
     public static final int T_HAG = 44;
     public static final int T_MML = 45;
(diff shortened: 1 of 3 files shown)


megamek/src/megamek/codeUtilities/MathUtility.java
```diff
@@ -18,75 +18,81 @@
...
-     * @return The value if it is inside the range given by the limits (inclusive);
-     *         the min value if value is below that range and the max value if value
-     *         is above that range.
-     *         clamp(2, 6, 8) returns 6,
-     *         clamp(7, 6, 8) returns 7,
-     *         clamp(12, 3, 5) returns 5.
+     *
+     * @return The value if it is inside the range given by the limits (inclusive); the min value if value is below that
+     *       range and the max value if value is above that range. clamp(2, 6, 8) returns 6, clamp(7, 6, 8) returns 7,
+     *       clamp(12, 3, 5) returns 5.
```
Sleet01, line 95: I'd argue that the list format here was A) deliberate, and B) clearer.

megamek/src/megamek/codeUtilities/MathUtility.java
```diff
@@ -96,12 +102,10 @@ public static int clamp(final int value, final int min, final int max) {
...
```
Sleet01, line 108: Same comment for subsequent entries.

megamek/src/megamek/common/AmmoType.java
```diff
@@ -537,6 +539,21 @@ public EnumSet<Munitions> getMunitionType() {
         return munitionType;
     }
 
+    public boolean notAllowedByClanRules() {
```
Sleet01, line 542: This will probably need to be incorporated into the ammo auto-config validity check.  I'll take a look.

Sleet01 (approved):
Looks good, especially the panels and utility stuff.
The changes to the list-based comments are a matter of taste, so shouldn't hold up the PR.

megamek/src/megamek/codeUtilities/MathUtility.java
rjhancock: Easy enough to revert back and put into a list form.

megamek/src/megamek/codeUtilities/MathUtility.java
rjhancock: Turned it into a list.

megamek/src/megamek/common/AmmoType.java
rjhancock: It was in two locations so just made a helper method for it. Seemed better to put it here vs elsewhere. It should be able to tell us if it's valid and gives us a single place to work with.

PR #6824, opened 2025-04-03
Don't fail loading unit with MASS in wrong location

pavelbraginskiy:
Fixes MegaMek/megameklab#1809.

If the MASS is in the wrong location, the unit will be marked as Invalid, so there's no reason to also throw an exception on load, so this just removes that exception. 

Also, the check to throw the exception (but not the check for validation) was wrong anyway.


Sleet01 (approved):
Looks good to me.  I see that almost the exact same check happens in `TestMek.java`, which should be what's responsible for the Invalid result.

rjhancock (approved):

PR #6840, opened 2025-04-07
fix: changed Internationalization to i18n

Scoppio:
Finally


rjhancock (approved):
Added bonus, this gives me a class to work from instead of replacing it later. Can use it as the stepping stone.

rjhancock (approved):

PR #6841, opened 2025-04-07
tip of the day

exeea:
add common Tip Of The Day implementation with i18n support.


megamek/src/megamek/common/util/TipOfTheDay.java
```diff
@@ -0,0 +1,219 @@
...
+ *  This program is free software; you can redistribute it and/or modify it
+ *  under the terms of the GNU General Public License as published by the Free
+ *  Software Foundation; either version 2 of the License, or (at your option)
+ *  any later version.
+ *
+ *  This program is distributed in the hope that it will be useful, but
+ *  WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ *  or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ *  for more details.
+ */
```
rjhancock: Use the correct copyright that is listed within the Wiki at https://github.com/MegaMek/megamek/wiki/MegaMek-Coding-Style-Guide and update the required fields accordingly.

megamek/src/megamek/common/util/TipOfTheDay.java
```diff
@@ -0,0 +1,219 @@
...
+        if (availableWidth <= 0)
+            return; // Not enough space to draw
+
+        Graphics2D g = (Graphics2D) g2d.create();
+
+        try {
+            g.setRenderingHint(RenderingHints.KEY_ANTIALIASING, RenderingHints.VALUE_ANTIALIAS_ON);
+            g.setRenderingHint(RenderingHints.KEY_TEXT_ANTIALIASING, RenderingHints.VALUE_TEXT_ANTIALIAS_ON);
+            g.setRenderingHint(RenderingHints.KEY_RENDERING, RenderingHints.VALUE_RENDER_QUALITY);
+            g.setRenderingHint(RenderingHints.KEY_STROKE_CONTROL, RenderingHints.VALUE_STROKE_PURE);
```
rjhancock, line 140: Please use more descriptive variable names (g2d -> graphics2D).

Unless I'm mistaken, the Rendering hints should be automatic.

rjhancock (changes requested):

megamek/src/megamek/common/util/TipOfTheDay.java
exeea: already changed, I copy pasted it randomly from another file and took the wrong one by mistake.

megamek/src/megamek/common/util/TipOfTheDay.java
exeea: g2d to graphics2D done.
the Rendering hints are needed or the text and the stroke appears jagged. (tested)

pavelbraginskiy (changes requested):
On lower-resolution screens the tip is too large relative to the splash.

rjhancock:
> On lower-resolution screens the tip is too large relative to the splash.

That would be due to the hints. What they do internally is essentially double the size of the render and scale it down. On HiDPI/Retina screens they show correctly. On "normal" screens they are twice the size.

Or that is my theory on it anyways.

exeea:
Originally I did that with a layeredpane but I was getting very bad aliasing and ugly stroke. This was the most visually pleasing approach.

megamek/src/megamek/common/util/TipOfTheDay.java
```diff
@@ -0,0 +1,246 @@
...
+ * (at your option) any later version.
+ *
+ * MegaMek is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with MegaMek. If not, see <http://www.gnu.org/licenses/>.
+ */
```
rjhancock, line 18: Please use the updated license as is linked to on the wiki.

rjhancock (changes requested):
The license header needs to be updated to the correct version and you'll need to update latest master into this and fix the calls to the Internationalization class to I18N.

Almost there.

megamek/src/megamek/common/util/TipOfTheDay.java
```diff
@@ -0,0 +1,254 @@
...
+                if (layout != null) {
+                    tipLayouts.add(layout);
+                    totalTipHeight += layout.getAscent() + layout.getDescent() + layout.getLeading();
+                } else {
+                    break; // Should not happen with LineBreakMeasurer unless width is tiny
+                }
+                if (measurer.getPosition() == layout.getCharacterCount() + measurer.getPosition()
+                        && measurer.getPosition() < tipAS.getIterator().getEndIndex()) {
+                    break;
+                }
```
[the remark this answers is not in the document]
exeea: fixed

megamek/src/megamek/client/ui/swing/util/UIUtil.java
```diff
@@ -521,11 +524,88 @@ public static int getScaledScreenHeight(DisplayMode currentMonitor) {
...
+            try {
+                gc = GraphicsEnvironment.getLocalGraphicsEnvironment()
+                    .getDefaultScreenDevice()
+                    .getDefaultConfiguration();
+            } catch (HeadlessException e) {
+                return new Dimension(800, 600); 
+            }
+        }
+        Rectangle bounds = gc.getBounds();
+        return new Dimension(bounds.width, bounds.height);
```
rjhancock, line 608: Although this is short enough to not need it, please have more descriptive variable names in the future.

Also, is there a reason we aren't logging the exception? I know it's being dealt with, don't need to pass the exception in just a comment about it would be a good thing in the logs.

megamek/src/megamek/common/util/TipOfTheDay.java
```diff
@@ -0,0 +1,256 @@
...
+                count++;
+                String tip = I18n.getTextAt(bundleName, TIP_BUNDLE_KEY + count);
+                if (tip.startsWith("!") && tip.endsWith("!")) {
+                    return count - 1;
+                }
+            }
+        } catch (Exception e) {
+            // When we get an exception, we've found all tips
+            return count - 1;
+        }
```
rjhancock, line 127: Has to be a better way to do this to get the number of tips.

rjhancock (approved):

megamek/src/megamek/common/util/TipOfTheDay.java
exeea: I don't see many other options:
a) or we store, in the file, also the number of tips in a dedicated key=value (making the i18n file store also numeric constants)
b) or we ditch the key=value file format, and we use a dedicated file for specifically only the tips, one per line, and we use the lines count to know how many tips we have

psikomonkie: If you get the `ResourceBundle` you're using, you would be able to get the list of its keys. From there you could filter it to keys that contain `TipOfTheDay.tip` (maybe change it to be longer/more specific so there's no false positives?).  This would also give the list of tips, and wouldn't require them to be numbered sequentially. Something like: 
```
ResourceBundle bundle = I18n.getInstance().getResourceBundle(bundleName);
Enumeration<String> keys = bundle.getKeys();
```
?

rjhancock: I was thinking making a resource file JUST for the tips for easier translation and just get the keys that way.

The ResourceBundle will handle I18N for us anyways and I plan to update the I18N class itself to be more dynamic in nature with ability to have defaults for the class as well as add in additional bundles as needed. A way to keep the resource files smaller and easier to manage for translations.

psikomonkie (approved):
I love it!

In the future, I'd love to see an enhancement that lets values within the tip be dynamically replaced - like `tip.7=You can press "{HOTKEY:VIEW_TOGGLE_ISOMETRIC}" to switch between top view and isometric view.` and then `TipOfTheDayJava` would replace that with the actual hotkey before returning it to the player. That's for another day though, 

I think this is a wonderful foundation for this feature and I'm excited to see this eventually get added to MegaMekLab and MekHQ as well.

exeea:
> I think this is a wonderful foundation for this feature and I'm excited to see this eventually get added to MegaMekLab and MekHQ as well.
The PR for MML is already up!

exeea:
> I think this is a wonderful foundation for this feature and I'm excited to see this eventually get added to MegaMekLab and MekHQ as well.

The PR for MML is already up!

exeea:
The PR for MML is already up!

HammerGS:
@pavelbraginskiy Have you're requests been met.

Scoppio (approved):

PR #6848, opened 2025-04-08
Added `escapeWithSlashes` Method for Better String Escaping

IllianiBird:
- Introduced `escapeWithSlashes` method to escape XML strings and replace forward/backslashes with underscores.

### Dev Notes
This PR is required by https://github.com/MegaMek/mekhq/issues/6341 - we had an issue where we were unable to save units that contained `/` or `\` in their name, as these were not caught by the stock escaping we had already implemented.

I opted to add a new method, rather than updating the original, as I didn't want to run the risk of breaking the cases where the original escaping was in-use.

### DIFF
--- a/megamek/src/megamek/utilities/xml/MMXMLUtility.java
+++ b/megamek/src/megamek/utilities/xml/MMXMLUtility.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018-2022 - The MegaMek Team. All Rights Reserved.
+ * Copyright (c) 2018-2025 - The MegaMek Team. All Rights Reserved.
  *
  * This file is part of MegaMek.
  *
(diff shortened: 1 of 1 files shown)


rjhancock:
Look up CDATA in XML. That would be a much better thing to implement and one that could be added to all XML output with no (or should not be) changes to the reading side.

IllianiCBT:
Something like this... ?

image

rjhancock:
No. CDATA is part of the spec itself for XML and allows you to put most items within an attribute without there being any issues.

https://en.wikipedia.org/wiki/CDATA

IllianiBird:
AHHHHH, I'm a donkey. Ignore me and this PR. I can just handle this in mhq.

Though the CDATA thing is good to know. So this wasn't a wasted exercise.

PR #6856, opened 2025-04-09
Updated Logging, Clone Handling, and Added `toString` Methods

IllianiBird:
- Introduced logging with `MMLogger` across `CancelAction`, `CloseAction`, `OkayAction`, and `Parameters` classes to log errors during cloning.
- Enhanced `clone` method by marking it with `@Nullable` and adding detailed error logs for failed cloning operations.
- Implemented `toString` methods in the aforementioned classes for better state representation during debugging.

### Dev Notes
During the investigation of an unrelated mhq issue I spotted that these methods could silently throw up a `null` return. In mhq this was causing problems, so rather than just fixing the logging in mhq I decided to go through the other instances in MegaMek to ensure any cloning failure is logged so we can spot it in the logs.

I would have gone through and added null protection for any instances where `clone` was going to be called from these classes, but IDEA couldn't find any and neither did a good ol' fashioned project search. So I'm assuming these are just utility methods, or required when extending the parent class.

Either way, we're logging now so in the event these do mysteriously start breaking after 5 years we'll know about it.

### DIFF
--- a/megamek/src/megamek/client/ratgenerator/Parameters.java
+++ b/megamek/src/megamek/client/ratgenerator/Parameters.java
@@ -32,12 +32,16 @@ import java.util.Collection;
 import java.util.EnumSet;
 
 import megamek.common.EntityMovementMode;
+import megamek.common.annotations.Nullable;
+import megamek.logging.MMLogger;
 
 /*
  * A class that holds all the parameters used to generate a table and is used as the key for the cache.
  *
  */
 public final class Parameters implements Cloneable {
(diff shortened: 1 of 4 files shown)


rjhancock:
The `clone()` method was added as it was throwing errors about objects being `Cloneable` with custom properties but NOT overriding the `clone()` method from it's parent and accounting for it.

IllianiBird:
Ah ha, of course, because they're implementing Cloneable! Even if indirectly by extending AbstractAction (which implements Cloneable). That makes complete sense.

PR #6869, opened 2025-04-12
Added @Nullable to `getEngine`

IllianiBird:
- Added nullability annotation `@Nullable` to the `getEngine` method in `Entity` to clarify potential null return.

### DIFF
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -10990,7 +10990,7 @@ public abstract class Entity extends TurnOrdered
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


psikomonkie (approved):
Trailers love you

IllianiBird:
And infantry, it turns out. I've opened up a sister PR over in MegaMek so that `getEngine` is tagged as `Nullable` moving forward. It makes perfect sense why it'd return `null`, _after the fact_, but I imagine I won't be the only beneficiary of that warning.

IllianiBird:
Oh, this _is_ the sister PR :D

PR #6875, opened 2025-04-13
Issue MekHQ#6619: Move `RestoreUnitAction::copyC3Networks` from MekHQ to MegaMek's `C3Utils`

psikomonkie:
Move `RestoreUnitAction::copyC3Networks` from MekHQ to MegaMek's `C3Utils`.

First step to resolving MegaMek/mekhq#6619.

PR #6878, opened 2025-04-14
generalized pilot dialog and C3 linking class

exeea:
- Pilot dialog (and his views) have been generalized to support instancing without the need of a clientgui or lobby
- C3 code have been moved fully to the C3Util class (that already existed but was incomplete) instead of being split between multiple classes

this PR is needed to allow to use the MM Pilot dialog + ammo selection + C3 linking into MML/MHQ without the need to instance a full lobby.

This PR is required for https://github.com/MegaMek/megameklab/pull/1835


IllianiBird:
Just a heads up if you're not already aware of it, we also have this C3 PR in the queue https://github.com/MegaMek/megamek/pull/6875 which might (I don't know if it will) conflict.

exeea:
> Just a heads up if you're not already aware of it, we also have this C3 PR in the queue #6875 which might (I don't know if it will) conflict.

We touch the same class but I see no conflicts. Thanks for the heads up!
We should move all possible C3 related code to this class.

megamek/src/megamek/client/ui/swing/unitDisplay/UnitDisplay.java
```diff
@@ -142,8 +150,8 @@
         displayP = new JPanel(new CardLayout());
         mPan = new SummaryPanel(this);
         pPan = new PilotPanel(this);
-        aPan = new ArmorPanel(clientgui != null ? clientgui.getClient().getGame() : null, this);
-        wPan = new WeaponPanel(this, clientgui != null ? clientgui.getClient() : null);
+        aPan = new ArmorPanel(client.getGame(), this);
```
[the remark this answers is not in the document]
exeea: solved

megamek/src/megamek/client/ui/swing/lobby/LobbyUtility.java
```diff
@@ -285,14 +285,6 @@ public static HashSet<Force> getForces(Game game, String idList) {
...
 
-    /**
-     * Returns true if a and b share at least one non-hierarchic C3 system (C3i, Naval C3, Nova CEWS). Symmetrical (the
-     * order of a and b does not matter).
-     */
-    public static boolean sameNhC3System(Entity a, Entity b) {
-        return (a.hasC3i() && b.hasC3i()) || (a.hasNavalC3() && b.hasNavalC3()) || (a.hasNovaCEWS() && b.hasNovaCEWS());
-    }
-
     /** Returns the string with some content shortened like Battle Armor -> BA */
```
rjhancock, line 288: Don't just remove methods. Mark for Deprecation, which version, and (if not in use) add the `forRemoval=true` tag.

megamek/src/megamek/client/ui/swing/unitDisplay/UnitDisplay.java
```diff
@@ -119,14 +122,19 @@ public class UnitDisplay extends JPanel {
...
+    public UnitDisplay(ClientGUI clientgui, @Nullable MegaMekController controller) {
+        this(clientgui.getClient(), controller);
+        this.clientgui = clientgui;
+    }
+
+    public UnitDisplay(Client client,
             @Nullable MegaMekController controller) {
         super(new GridBagLayout());
-        this.clientgui = clientgui;
+        this.client = client;
```
rjhancock, line 137: As indicated in the comment, this was left `Nullable` to support other clients such as MekWars.

Please fix this by re-adding the tag and adjusting the code accordingly. I'll bring MekWars support up with the rest of the Dev team to determine if we still wish to support it to some degree or try to merge that general idea into MegaMek.

rjhancock (changes requested):
Overall looks good but some issues need to be addressed first.

PR #6883, opened 2025-04-15
fix: fixes the file entity conversion for ACAR by adding move by copy

Scoppio:
# What it does

Fixes the ammo type being messed up when running ACAR on MekHQ, makes the AFS unit load the structure and armor correctly so it uses the original/starter values instead of the current for the "max" stats. It also deals with some random problems like units being discarded during resolution for "never being deployed" when they were in fact deployed, equipment not being destroyed when they should, etc.

# Fixes

- fixes [#6633](https://github.com/MegaMek/mekhq/issues/6633)


psikomonkie (approved):
Nice!

Issue #4696, opened 2023-08-09
Request for SPAs for the Battletech Missions series


Jamespryde:
### Environment
What version of MegaMek does your issue apply to?  
MekHQ 0.49.14.
What operating system are you using?  
Windows
What java version are you using? 
	Origin Project: MekHQ
	Java Vendor: Eclipse Foundation
	Java Version: 11.0.12
	Platform: Windows 10 10.0 (amd64)

### Description
SPAs used in the Missions set are different then are available. Understand some are not possible due to limitations but any that could be implemented would be appreciated. Breakdowns are below see attached.
[Player Rules 3.1.pdf](https://github.com/MegaMek/megamek/files/12307198/Player.Rules.3.1.pdf)

Capellan Confederation (Liao) - Fanatic
For the Chancellor! (15 XP, Max PS 5): You may take a pilot hit to add 1 to an attack roll after it
is rolled.
After you, I insist! (30 XP, Max PS 4): At any time before you move for a turn, you may take a
pilot hit to move at the end of initiative, after all other units have moved. In effect, this moves you
out of the initiative order (requiring recalculation of which friendly units must move in each
phase).
Fanaticism (45 XP, Max PS 3): Once per scenario, during the end phase of a turn, you may
declare that you are using Fanaticism. During the following turn, all attacks made by you are
made with a -2 to hit bonus. However, all attacks made against you are also made with a -2 to
hit bonus.
Free Worlds League (Marik) - Teammaker
Plays well with others (15 XP, Max PS 5): Once per scenario, for one turn only, you may
declare one other pilot to be your partner and an enemy unit that you can see within 10 hexes
as your target. All your and your partner’s attacks against the target have a -1 to hit bonus.
Take the Hit (30 XP, Max PS 4): When one of your teammates is hit by a weapon attack, if you
are at least as close to the attacker as you are to your teammate and in the attacker’s same
firing arc, you can take the damage instead. Usable once per round.
Share the wealth (45 XP, Max PS 3): If both players agree, any player on your team can use a
PSA or faction ability possessed by another. This ability must be limited in some way. If it is a
“Usable <n> times per scenario” ability, then this use counts for the owning player. If the ability
requires a pilot hit to use, you take the pilot hit, and must roll consciousness accordingly.
Draconis Combine (Kurita) - Samurai
Bushido (15 XP, Max PS 5): Once per scenario you may declare a target for a duel. Attacks
against that target are made with a -1 to hit bonus. However, attacks against any other target
[...]

PR #6884, opened 2025-04-15
Initial dev work.  Additions to messages.properties additions and Opt…

TheodoreBrinkman:
…ionsConstants, along with a couple of the low-lying-fruit implementations, and the beginnings of Tactical Genius (implementing a way to determine if the player/team wins initiative ties).

I Downloaded Sarna (Not sure if LegAttackHandler is needed for this?) Cluster Hitter Tactical Genius (implemented a way to determin if the player/team wins initiative ties).

Implementation for https://github.com/MegaMek/megamek/issues/4696


TheodoreBrinkman:
Ok, I think I got the pull request figured out.  Here's hoping.  (Moving from TFS experience to Git is a real change of process.)

Scoppio (changes requested):
Hi @TheodoreBrinkman! Thanks for showing interest into Megamek!

While we appreciate any help with the project, we won't be able to accept those changes, as we have a strong policy about only implementing rules and other things that change and effect the gameplay that come from up to date and errata'd official sources and only from the source books that are officialy supported by the project.

You are free to implement this work in your own fork of the project, but we won't be able to accept this pull request.

I will be closing this pull request later.

PR #6885, opened 2025-04-15
fix: wrong access for weapon arc

Scoppio:
A tipical error caused by primitive obsession.

Function should be called with equipment ID/number but instead I was passing the equipment location.


mmetke-qumulo (approved):
LGTM!

psikomonkie (approved):
Interesting! Good fix!

Sleet01 (approved):
LGTM!

rjhancock (approved):

PR #6886, opened 2025-04-15
Change getWeaponArc signature and add JavaDocs

psikomonkie:
After seeing #6885 and seeing I'd made the same mistake in the past too, I've changed the signature of the `getWeaponArc` method so it's `weaponNumber` instead of `wn`, and added JavaDocs specifying it should be the equipment number.

I also fixed this bug in `MultiTargetFireControl`.


Scoppio (approved):
Good fix, doesnt solve the problem but helps ALOT! And good catch with the bug

rjhancock (approved):

PR #6890, opened 2025-04-17
corrected destroyed heatsinks count

exeea:
all heatsinks counted as 1, now properly distinguished between normal and double.


rjhancock (approved):
Ok, this was smaller than I thought. Is it possible to get tests for this?

PR #6891, opened 2025-04-17
Fix version comparisons now that isExtra is never null

psikomonkie:


psikomonkie:
I'm not sure if this will fully work - when we compare two versions we compare their `extra` too. How will versions tracked by git-hash or branch be able to compare properly? Should extra be completely removed from the comparison?

rjhancock:
If this fixes the issue in MHQ, it's fine. When it comes to the `extra`, the only time that matters would be for nightlies which would be sorted correctly. PRs don't matter in that context as we can match the hash. 

The full version should be newer than an `extra` version.

rjhancock (approved):

psikomonkie:
> If this fixes the issue in MHQ, it's fine. When it comes to the `extra`, the only time that matters would be for nightlies which would be sorted correctly. PRs don't matter in that context as we can match the hash.
> 
> The full version should be newer than an `extra` version.

How should those interact with each-other? They're both in the `extra`, so is it correct to consider a MM/MHQ version newer/older when comparing an artifact branch hash to a nightly date? I'm not sure a text comparison of this field for comparing if a version is newer or older is appropriate. 

This may make testing artifacts difficult, as MekHQ saves will have version comparison issues that block loading when switching between the nightly and artifacts. 

This is also something that could be handled in MekHQ - instead of using `Version::compareTo` directly, we just look at the major, minor, and patch. I don't think that's a good option though.

rjhancock:
Testing artifacts shouldn't be tested against each other, they should be tested against the previous released version.

With testing of artifacts, the extra data is meant to tell us which commit they are from as well as whether it's a nightly or a PR. That is all. It's just more information for us.

megamek/src/megamek/Version.java
```diff
@@ -299,12 +299,12 @@ public int compareTo(final Version other) {
...
-        } else if (getExtra() == null) {
-            return -1;
-        } else if (other.getExtra() == null) {
+        } else if (getExtra().isEmpty()) {
             return 1;
+        } else if (other.getExtra().isEmpty()) {
+            return -1;
         }
 
         return getExtra().compareTo(other.getExtra());
```
rjhancock, line 310: Just remove the comparison on `extra` and return `0` at this point. Let's just not worry about the extra.

psikomonkie: Done

PR #6893, opened 2025-04-17
Properly calculate armor crit slot count for superheavy meks

pavelbraginskiy:
Sets armor crits to CRITICALS_VARIABLE

### DIFF
--- a/megamek/src/megamek/common/MiscType.java
+++ b/megamek/src/megamek/common/MiscType.java
@@ -1383,6 +1383,7 @@ public class MiscType extends EquipmentType {
             return (int) Math.ceil(size);
         }
         // right, well I'll just guess then
+        LOGGER.error("Failed to calculate crit slots for equipment %s".formatted(this.getName()));
         return 1;
     }
 
--- a/megamek/src/megamek/common/MiscType.java
+++ b/megamek/src/megamek/common/MiscType.java
@@ -1383,7 +1383,7 @@ public class MiscType extends EquipmentType {
             return (int) Math.ceil(size);
         }
         // right, well I'll just guess then
-        LOGGER.error("Failed to calculate crit slots for equipment %s".formatted(this.getName()));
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/common/MiscType.java
```diff
@@ -1383,6 +1383,7 @@ public int getCriticals(Entity entity, double size) {
             return (int) Math.ceil(size);
         }
         // right, well I'll just guess then
+        LOGGER.error("Failed to calculate crit slots for equipment %s".formatted(this.getName()));
```
rjhancock, line 1386: Replace with 

`LOGGER.error("Failed to calculate crit slots for equipment {}", this.getName());`

megamek/src/megamek/common/equipment/ArmorType.java
```diff
@@ -391,7 +391,7 @@ private static ArmorType createClanFerroFibrous() {
         armor.addLookupName("Clan Ferro-Fibrous Armor");
         armor.addLookupName("Clan Ferro Fibre");
         armor.cost = 20000.0;
-        armor.criticals = 7;
+        armor.criticals = CRITICALS_VARIABLE;
```
rjhancock, line 394: How does these changes impact non-superheavy units?

megamek/src/megamek/common/MiscType.java
pavelbraginskiy: Done

megamek/src/megamek/common/equipment/ArmorType.java
pavelbraginskiy: They don't. Nothing accesses `MiscType.criticals` directly, it goes through the `getCriticals` method. `getCriticals` already had the correct CRITICALS_VARIABLE handler for all armor types to report the number of crits to be the standard value in all cases except for superheavy meks, but that handler wasn't being invoked because armor wasn't actually CRITICALS_VARIABLE.

rjhancock (approved):

PR #6896, opened 2025-04-18
clone an entity

exeea:
this adds two methods:
- clone() // allows to clone an Entity (@Scoppio implementation from MHQ)
- repair() // repairs an unit, all the criticals/equips and, optionally, heals the crew

### DIFF
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -15677,13 +15677,13 @@ public abstract class Entity extends TurnOrdered
             }
         }
         if (healCrew) {
+            Crew crew = this.getCrew();
+            crew.setDoomed(false);
+            crew.setEjected(false);
             for (int i = 0; i < this.getCrew().getSlotCount(); i++) {
-                Crew crew = this.getCrew();
                 crew.setHits(0, i);
                 crew.setDead(false, i);
                 crew.setCrewFatigue(0, i);
-                crew.setDoomed(false);
-                crew.setEjected(false);
             }
         }
     }
--- a/megamek/src/megamek/common/Entity.java
+++ b/megamek/src/megamek/common/Entity.java
@@ -15,6 +15,10 @@
 package megamek.common;
 
 import java.awt.Image;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
 import java.io.Serial;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
@@ -25,6 +29,7 @@ import java.util.stream.IntStream;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import megamek.MMConstants;
 import megamek.client.bot.princess.FireControl;
+import megamek.client.generator.RandomNameGenerator;
 import megamek.client.ui.Base64Image;
 import megamek.client.ui.swing.GUIPreferences;
 import megamek.client.ui.swing.calculationReport.CalculationReport;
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/Entity.java
```diff
@@ -15609,4 +15614,78 @@ public boolean isJumpingWithMechanicalBoosters() {
...
+        }
+        for (int loc = 0; loc < this.locations(); loc++) {
+            for (int slot = 0; slot < this.getNumberOfCriticals(loc); slot++) {
+                CriticalSlot cs = this.getCritical(loc, slot);
+                if (cs != null) {
+                    cs.setDestroyed(false);
+                    cs.setMissing(false);
+                    cs.setBreached(false);
+                    cs.setHit(false);
+                }
```
Scoppio, line 15671: It looks more logical to me if CriticalSlot implemented a `reset` function

megamek/src/megamek/common/Entity.java
```diff
@@ -15609,4 +15614,78 @@ public boolean isJumpingWithMechanicalBoosters() {
...
+        } catch (Exception e) {
+            throw new AssertionError("Clone not supported", e);
+        }
+    }
+    
+    /**
+     * Repairs an entity.
+     * @param healCrew Whether to heal the crew or not
+     */
+    public void repair(boolean healCrew) {
```
Scoppio, line 15650: it `repairs` or `resets`?
Repair is a game action/rule, involves rolls and etc, and may also have consequences like things being partically repaired. I suggest calling it `reset` as it is setting the unit back to its "initial state".

megamek/src/megamek/common/Entity.java
```diff
@@ -15609,4 +15614,78 @@ public boolean isJumpingWithMechanicalBoosters() {
...
+                    try(ByteArrayInputStream byteArrayInputStream = new ByteArrayInputStream(serializedData)) {
+                        try (ObjectInputStream objectInputStream = new ObjectInputStream(byteArrayInputStream)) {
+                            return (Entity) objectInputStream.readObject();
+                        }
+                    }
+                }
+            }
+        } catch (Exception e) {
+            throw new AssertionError("Clone not supported", e);
+        }
```
Scoppio, line 15643: I implemented something very close to this, but I see it has at least one mistake.
First is that it is catching an exception and transforming it into an AssertionError.
This is not an assertion, It's a runtime exception, I dont know which exception this is supposed to throw, so I would use the exception the ByteArrayOutputStream tells you it throws (IOException maybe???)
Then if you want to raise another exception, raise a new exception specific to EntityCloning, like... EntityCloningException, may extend it from runtime exception so the API does not complain about adding an exception to the signature of the function.

Scoppio (changes requested):

megamek/src/megamek/common/Entity.java
```diff
@@ -15609,4 +15614,78 @@ public boolean isJumpingWithMechanicalBoosters() {
...
+                    cs.setMissing(false);
+                    cs.setBreached(false);
+                    cs.setHit(false);
+                }
+            }
+            this.setInternal(this.getOInternal(loc), loc);
+            this.setArmor(this.getOArmor(loc), loc);
+            if (this.hasRearArmor(loc)) {
+                this.setArmor(this.getOArmor(loc, true), loc, true);
+            }
```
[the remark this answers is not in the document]
exeea: that would need investigation. Thanks for the tip.

megamek/src/megamek/common/Entity.java
```diff
@@ -15609,4 +15614,78 @@ public boolean isJumpingWithMechanicalBoosters() {
...
+            Crew crew = this.getCrew();
+            crew.setDoomed(false);
+            crew.setEjected(false);
+            for (int i = 0; i < this.getCrew().getSlotCount(); i++) {
+                crew.setHits(0, i);
+                crew.setDead(false, i);
+                crew.setCrewFatigue(0, i);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
exeea: I couldn't find one. There is resetPilotAndEntity, that internally is relying on Campaign.clearGameData(Entity entity) that does many other things, including setting the Game and Owner.
But for example it doesn't restore damaged equip, critical slots, armor, etc...

megamek/src/megamek/common/Entity.java
exeea: reset() is a better name.

PR #6905, opened 2025-04-19
Issue #6904: Flak attacks must target entities, that must be considered when princess is targeting

### LINKED ISSUE #6904: [Issue] 0.50.05-SNAPSHOT - Princess not firing artillery (Arrow IV specifically) with onboard units
### Brief Description *

I have noticed Princess does not seem to be firing onboard artillery at all, neither as indirect artillery or flak/direct attacks. I have so far only seen this with Arrow IV, but I have ONLY fought against arrow IV units so far so it probably affects standard artillery too.

There may be some other explanation as to why she's not firing and if so great, but I am attaching an example savegame so that it can be looked into.

Unit in question from the save that has Arrow IV and isn't firing: Hadur (hovertank)

### 3. Steps to Reproduce

Observe issue in game

### Attach Files

[ArtilleryNotFiring.zip](https://github.com/user-attachments/files/19817123/ArtilleryNotFiring.zip)

### Severity *

### MegaMek Suite Version *

0.50.05-SNAPSHOT     Build Date: 2025-04-15T03:37:36.449977996

### Operating System *

Windows 10

### Java Version *

Java Vendor: Eclipse Adoptium     Java Version: 17.0.6

### Final Verification

psikomonkie:
Fixes #6904 

Conventional unguided Arrow IV rounds are tagged as flak, so when checking if an attack is flak we can't just look at the ammo, we also need to consider if we're targeting an entity or a hex. If it's a hex, it's not flak.

### DIFF
--- a/megamek/src/megamek/client/bot/princess/ArtilleryTargetingControl.java
+++ b/megamek/src/megamek/client/bot/princess/ArtilleryTargetingControl.java
@@ -456,7 +456,11 @@ public class ArtilleryTargetingControl {
 
 
                     for (Targetable target : targetSet) {
-                        boolean targetingEntity = target.getTargetType() == Targetable.TYPE_ENTITY;
+                        boolean attackOnAirborneEntity = (target.getTargetType() == Targetable.TYPE_ENTITY) &&
+                                                              (target instanceof Entity targetedEntity) &&
+                                                              ((targetedEntity.isAirborne()) ||
+                                                                     (targetedEntity.isAirborneVTOLorWIGE()) ||
+                                                                     (targetedEntity.isAirborneAeroOnGroundMap()));
                         double damageValue;
                         if (isZeroDamageMunition) {
                             // Skip zero-damage utility munitions for now.
@@ -473,7 +477,7 @@ public class ArtilleryTargetingControl {
                             }
                         } else {
                             // Flak Artillery need to be made during direct fire, not as Indirect
-                            if (targetingEntity) {
+                            if (attackOnAirborneEntity) {
                                 damageValue = damage;
                             } else {
                                 if (!isADA) {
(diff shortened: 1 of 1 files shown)


Sleet01 (approved):
LGTM.  Was this causing Princess to skip all indirect shots, then not have any direct fire options during the attack phase?

psikomonkie:
> LGTM. Was this causing Princess to skip all indirect shots, then not have any direct fire options during the attack phase?

She seemed to direct fire on me fine, it was just indirect that was having issues. Standard HE rounds for all arty are flagged as “F”lak.

Sleet01:
> > LGTM. Was this causing Princess to skip all indirect shots, then not have any direct fire options during the attack phase?
> 
> She seemed to direct fire on me fine, it was just indirect that was having issues. Standard HE rounds for all arty are flagged as “F”lak.

Actually, do we need to explicitly restrict entity targets to airborne here?  I'm on my phone so I can't easily check if we're doing that, and I didn't recall of the top of my head.

psikomonkie:
> > > LGTM. Was this causing Princess to skip all indirect shots, then not have any direct fire options during the attack phase?
> > 
> > 
> > She seemed to direct fire on me fine, it was just indirect that was having issues. Standard HE rounds for all arty are flagged as “F”lak.
> 
> Actually, do we need to explicitly restrict entity targets to airborne here? I'm on my phone so I can't easily check if we're doing that, and I didn't recall of the top of my head.

Should we? Probably. When stepping through the code I think I even saw that check when building the list of targets, so I could just copy it from there. 

Do we need to? I don’t think so, it’d cause a problem if there was an ammo type that targets an entity and is flak. That’d get weird. But looking at the rules there are no ammo like that. 

I’ll switch it up on my local and see how it is.

Sleet01 (approved):
LGTM!

PhoenixHeart512:
What about the Cluster rounds she wasn't firing either, are those also tagged as Flak?

Sleet01:
@PhoenixHeart512 Yep, Cluster Artillery is Flak (iirc)

PR #6910, opened 2025-04-20
Fix cbf class cast exception

Sleet01:
Also prevents Princess firing Homing munitions at airplanes (seen during testing of the above fix).

The issue is that we narrowed our check for flak target unit types (correctly) but this exposed that we've been handling
Counter-Battery Fire targets against off-board units _incorrectly_.
After some discussion we decided not to change how CBF targeting works at this late date, but we do wish to prevent the Class Cast exception.
Since we only damage the target off-board unit when making CBF attacks, we should treat the damage value the same as for Flak: only the target itself will be damaged.

I also added some clauses to try to prevent Princess from firing Homing rounds at Airborne Aerospace units.
While the other prerequisites might be in place (TAG unit near target, Homing round does more damage, Firing phase) it's not possible to guide Homing munitions to these units so the effective possible damage _is_ zero.

Testing:
- Ran stress game with multiple ASFs, on-board and off-board artillery, and varied munitions.

### DIFF
--- a/megamek/src/megamek/client/bot/princess/ArtilleryTargetingControl.java
+++ b/megamek/src/megamek/client/bot/princess/ArtilleryTargetingControl.java
@@ -335,6 +335,7 @@ public class ArtilleryTargetingControl {
             }
         }
 
+        // TODO: Counter-battery fire must target a hex (TO:AR p 154); needs better off-board unit deploy logic
         for (Entity enemy : game.getAllOffboardEnemyEntities(shooter.getOwner())) {
             if (enemy.isOffBoardObserved(shooter.getOwner().getTeam())) {
                 targetSet.add(enemy);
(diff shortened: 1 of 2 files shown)


psikomonkie (approved):
Nice!!

psikomonkie (commented):
Actually I do have one question - currently I thought we do apply splash damage if we land "near" the target for counter battery. Can we still use the other method that calculates that damage as opposed to using the full damage here when judging counter battery shots? (`calculateDamageValue`)

Sleet01:
> Actually I do have one question - currently I thought we do apply splash damage if we land "near" the target for counter battery. Can we still use the other method that calculates that damage as opposed to using the full damage here when judging counter battery shots? (`calculateDamageValue`)

That function is basically doing the same thing except it adds the damage for any nearby units as well.  Since we currently deploy all off-board artillery into the same hex, this would give us damageValue = `N * rackSize()` where N is the number of units in that hex.
So this would give us a very high damage value.
But we're treating CBF as if it only targets / can hit one unit, so we don't want Princess to know that high value.

When we have implemented distributed deployment of off-board artillery, then we definitely _should_ switch the non-Airborne-but-still-Entity-target path to use that method.

PR #6916, opened 2025-04-22
Fix #6915: Fixed Status of Uncrewed Vehicles

### LINKED ISSUE #6915: Fully repaired vehicles listed as "Crippled" in the hanger, when all they need is a crew.
Not really a bug per-se. Just a minor inconsistency in the interface.

But this appears to have been present in most previous patches, but checking its there in 49.19

So situation. If "Tank" and "Naval" classed vehicles have 0 crew, their Status is listed as "Crippled" in the Hangar tab which is a tad confusing to look at especially if the machine looks fully repaired/isn't showing in the repair bay, etc. 

If the unit has at least 1 crew member assigned it's status changes to "Undamaged"

Clicking through and checking most other unit types, other machines be they Mechs, ASF, CF, Dropships, Warships, VTOLs and many more if they lack crew but are in otherwise working order are listed "Undamaged".

Just for consistency sake, I feel it might be nice to change it so that the status of  the Tank and Naval units be "Undamaged" if they aren't damaged and just lack crew. That or if possible, perhaps add a new status of "Needs Crew" which can be applied to the other unit types as well.

Thank you for your consideration, have a good day.

IllianiBird:
- Vehicles are immobilized when their crew is missing. This caused `isPermanentlyImmobilized` to return `true` when crew was missing, even if 'checkCrew' was `false`. Vehicles already have an override that checks for actual movement damage, so this return was incorrect

Fix #6915

### Dev Notes
This is the cause of vehicles displaying as 'crippled' in MekHQ when they are without crew and why no other unit type was affected in this way.


IllianiBird:
@psikomonkie I've added you as a reviewer to this, as it required some reworking of one of your methods and I wanted to verify that I haven't unintentionally broken something.

megamek/src/megamek/common/Entity.java
```diff
@@ -1913,14 +1913,16 @@ public boolean isPermanentlyImmobilized(boolean checkCrew) {
...
             return false;
-        } else {
+        } else if (!(this instanceof Tank)) { // this is already handled in the tank specific override
             return ((getOriginalWalkMP() > 0) || (getOriginalRunMP() > 0) || (getOriginalJumpMP() > 0))
                          // Need to make sure here that we're ignoring heat because that's not actually "permanent":
                          &&
                          ((getWalkMP(MPCalculationSetting.PERM_IMMOBILIZED) == 0) &&
                                 (getRunMP(MPCalculationSetting.PERM_IMMOBILIZED) == 0) &&
                                 (getJumpMP(MPCalculationSetting.PERM_IMMOBILIZED) == 0));
         }
```
psikomonkie, line 1923: `// this is already handled in the tank specific override` is an interesting comment, because it's not an `@Override`, but it could be, and I think it would make the code a bit neater. This block (without the ! instance of Tank check) could `isMovementHit()`. Throw the `@Override` annotation on the `Tank::isMovementHit` and remove the now-unnecessary override `Tank::isPermanentlyImmobilized` (which just calls `isMovementHit`, but now that method is already called inside of it.

psikomonkie (approved):
Nice! One comment, not necessary just a slight refactor your comment inspired. No functional difference, just something to make you think 😉

megamek/src/megamek/common/Entity.java
IllianiBird: While this _is_ a good idea, the override is used in 8 different methods, including a bunch of tests, and I really don't have the heart to futz with it this late in the dev cycle.

IllianiBird:
I'm holding this back till 50.06. I can't quantify what it is, but the thought of merging this is making the hairs on my neck standout. I just _know_ this is going to break something, I just can't say what. And given we want 50.05 to be a milestone I'm not willing to ignore that little sixth sense voice.

megamek/src/megamek/common/Entity.java
rjhancock: Good news! New dev cycle!

If there is a better way, take some time to do it right and clean up as you go.

PR #6928, opened 2025-04-24
Issue #6918: If `target` is not null, but it's an entity, and the entity is no longer in the game, then it's fled/dead

### LINKED ISSUE #6918: Counter Battery Fire: Splash damage hitting other offboard units
### Brief Description *

from discussion in the discord: "Splash damage hitting other offboard units"

Rarely, Counter Battery fire is dealing splash damage to all units in the same (simulated?) offboard location.  For example, all units offboard North, 17 hexes away count as the same hex.   This doesn't always happen, unsure what causes it to start happening.

Example game log of a few shots doing this (all of the units here are offboard:
[example log - Copy.txt](https://github.com/user-attachments/files/19858919/example.log.-.Copy.txt)
see shots by `Marksman Artillery Vehicle #5`, `Marksman Artillery Vehicle #6`.

I am unsure if this is correct according to the rules or not, as theoretically all of the units being damaged are in the same hex.  But they probably shouldn't be.

possible workaround:  stagger the offboard distance of artillery units so they count as being in different hexes

### 3. Steps to Reproduce

1. use Counter Battery fire to target an offboard enemy unit that is at the same direction+distance as other offboard enemy units
2. wait for one to hit.  Most of the time, it will only damage the targeted unit, as it should.  However, in some cases, ALL units at the offboard location take damage.  There is likely some secondary factor, because I only triggered this a few times in one game, late into the game.

Side note: in round 11 Offboard Attack Phase, some of the shots targeting the offboard Naga will also hit every target in the offboard hex (even if the Naga has Disengaged).  

Can provide more save files from various rounds/phases in this game if needed,

### Attach Files

I was unable to reproduce in my smaller test game, so here are saves from the game where it happened

fire2 is during the TAG phase directly before the offboard splash damage happens.  Perform or skip `Mannheim`'s TAGs to reach the interesting artillery fire results.   One of the automatically hitting counter battery shots (see https://github.com/MegaMek/megamek/issues/6917) also triggers this splash damage when it hits the offboard target.
fire3 is after the artillery fire has resolved, check the game logs then.

[fire2 autosave_2025-04-21_23-42-27.sav.gz](https://github.com/user-attachments/files/19858867/fire2.autosave_2025-04-21_23-42-27.sav.gz)
[fire3 autosave_2025-04-21_23-43-42.sav.gz](https://github.com/user-attachments/files/19858866/fire3.autosave_2025-04-21_23-43-42.sav.gz)

Interestingly, when I was testing the save files, the same unit triggered the splash damage by rolling a 9 when needing a 10.

### Severity *

### MegaMek Suite Version *

0.50.04

### Operating System *

Windows 10

### Java Version *

    Java Vendor: Eclipse Adoptium     Java Version: 17.0.12

### Final Verification

psikomonkie:
Fixes #6918


Sleet01 (approved):
LGTM!
We discussed a little bit of the finer details on the Discord as well.

PR #6937, opened 2025-04-27
Calculate heat for vibroblades

pavelbraginskiy:
I believe this has no effect in gameplay, but it does mean that Vibroblades on record sheets will show the correct heat value instead of a dash.


Sleet01 (approved):
LGTM!

PR #6944, opened 2025-04-28
feat: adds CONVOY behavior setting preset and flag to ignore damage output on pathing

### LINKED ISSUE #3250: RFE: Add Convoy Bot
This calls for the ability to define a bot that handles a convoy.

This would preferrably be a subset of Princess, but may need to be a new Bot.

### LINKED ISSUE #4308: Teach Princess to handle certain units as non-combat regardless of weapons
Convoy escort scenarios from MekHQ get a little annoying at times, since Princess likes to do things like rush a MASH truck across the field and straight into enemy fire just because it has a machine gun or two it wants to get into long range (3 or 4 hexes away).  That's just following the standard algorithm which assumes everything is reasonably combat oriented.  There isn't an easy way to automatically determine which units are genuinely 'combat oriented' and which are not so needs to be done manually by the user.

In the Princess configuration dialog, add a multi-select list of all units that bot player is responsible for.  Every entry would be deselected by default for a new bot (e.g. after a /kick and Replace Player... operation), and the user selects the units which Princess should consider non-combat.  During the movement phase these units, regardless if they are armed or not, would be considered to have a potential damage of 0 (including physical attacks for Mech-type units).  This should put them in the position of only defensive positioning - not getting shot - since they are not considered capable of shooting back.  The weapons and physical attack phases are treated as normal, with those units taking shots and using searchlights as warranted.

### LINKED ISSUE #6260: [RFE] Have a princess setting for "Convoys" where they only walk or try to stay away from the enemy instead of sprinting their doom.

- [x] I'm reporting the RFE to the correct repository:

- [X] Does this conform with our Content Policy linked above?
- [X] I've searched the GitHub tracker and haven't found a similar feature request listed

### RFE Details *

This is an improvement to an existing feature

### Brief Description *

On "Escort Convoy" missions both stratcon and normal generated variants, the convoy ALWAYS charges to their doom into the enemy lines, and if they have hover vehicles half the convoy is dead by turn 3 before your mechs can even catch up to defend them. My suggestion, make a flag or setting that all convoy units "walk" or stay a certain distance away from all enemy units. Because as of now the only way to prevent this is to manually take control of the units

### MegaMekLab Suite Version *

v0.50.01

### Attach Files

_No response_

### Final Checklist

- [X] I've checked to make sure that this RFE has not already been filed
- [X] I am requesting implementation of only one relevant set of features in this ticket for clarity and focus

Scoppio:
# What is it about?

Adds **CONVOY behavior setting** preset, a behavior controlling flag to *ignore damage output* (damage that would be caused by this unit onto enemies) when calculating pathing utility score was added, also adds special case where unit will give priority to face towards its destination when fleeing but only if it has the flag `ignore damage output` enabled.

## What it does?

This small feature adds a simple behavior control toggle

## Anything else?

Alot of Javadoc was added on BasicPathRanker to better explain what is happening.

## Issues:
This PR potentially resolves the following RFE's

- Closes #4308
- Closes #3250
- Closes #6260


SuperStucco:
The clamp01() method indicates "The value if it is inside the range given by the limits (inclusive)", but it only has one input - the value to be clamped.  It should include either values to clamp between as arguments, or the JavaDoc updated to show that it is clamping between 0.0 and 1.0 (or whatever values are hard coded).

rjhancock (approved):
Im glad to see more documentation get added in.

That being said, the nitpick CoPilot picked out is valid. :)

Scoppio:
> The clamp01() method indicates "The value if it is inside the range given by the limits (inclusive)", but it only has one input - the value to be clamped. It should include either values to clamp between as arguments, or the JavaDoc updated to show that it is clamping between 0.0 and 1.0 (or whatever values are hard coded).

Well... it is the `CLAMP 0 1` method....

Issue #6950, opened 2025-04-28
[Issue] Customizing Units in The Lobby (and potentially in-game) Results in NPE & Loss of Deployment Information


IllianiBird:
### Brief Description *

Customizing a unit in the lobby results in an NPE due to null portrait (even for units with crew that have portraits). It also wipes the units' deployment information. This results in the unit being unable to be deployed.

### 3. Steps to Reproduce

1. Load the provided save and launch the scenario
2. Attempt to edit any character you'll be hit with an NPE due to null portrait and the unit's deployment information will be removed.
3. Launch the game, the unit will be selected for deployment but any deployment with that unit will be impossible as they have no assigned deployment zone.

### Attach Files

[customizeUnitErrors.cpnx.gz](https://github.com/user-attachments/files/19948510/customizeUnitErrors.cpnx.gz)

[mekhq.log](https://github.com/user-attachments/files/19948539/mekhq.log)

### Severity *

High (Major Disruption): A major feature is broken or incorrect, but a workaround exists.

### MegaMek Suite Version *

0.50.06 (current master)

### Operating System *

Mac OS X 15.4.1 (aarch64)

### Java Version *

17.0.14

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have included all necessary information and files to help reproduce this issue
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

IllianiBird:
Confirmed that this does trigger in-scenario, which leads me to believe it's behind the NPEs Jester was getting during their last stream.

PR #6954, opened 2025-04-29
fix for null clientgui and memory leaks

exeea:
fixes NPE and the issue preventing to edit deployment and portrait of unit.
This is solved by splitting the two versions of the constructor instead of relying on the one without clientGUI (that would cause the UI initialization without it)
Fixed also some memory leaks caused by this dialog, the damage dialog and the camo dialog.

Fixes https://github.com/MegaMek/megamek/issues/6950


rjhancock:
For the time being, with the new settings on this repo, please add a line to the history.txt file as well.

rjhancock (approved):
Other than the history.txt entry, sanity check looks good.

rjhancock (approved):

PR #6965, opened 2025-05-01
Distinguish Mek weapons from BA weapons in advanced search

### LINKED ISSUE #6924: [RFE] Disambiguate equipment in filter list
A lot of equipment seemingly shows up twice when filtering units by weapon:

Image

I believe this is caused by both the BA and Mek-scale version of the weapon being listed. 
It would be good to add a note to clarify which one is which.

### MegaMek Suite Version

0.50.04

pavelbraginskiy:
Adds a new `uniqueName` attribute to `EquipmentType`s which can be set when the equipment shares it's name with something else. Currently only distinguishes BA weapons from Mek weapons, but this should make it easier to resolve other collisions which I didn't know about at time of making this PR.

Closes #6924.

### DIFF
--- a/megamek/docs/history.txt
+++ b/megamek/docs/history.txt
@@ -15,6 +15,7 @@ MEGAMEK VERSION HISTORY:
 + Fix #6955: fuel tanks take zero damage from AE
 + Fix #6956: Correct fluff art for HHWs
 + Fix #6957: BA Artillery doing no damage
++ PR #6965: Disambiguate BA vs Mek weapons with the same name
 
 0.50.05 (2025-04-25 1800 UTC)
 + Fix #6652: ProtoMek BV fixed for melee weapons and LRM-1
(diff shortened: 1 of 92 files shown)


pavelbraginskiy:
For reviewers: This PR features 89 files of nothing but `uniqueName = name + " [BA]";`, the "actual" code changes are just in [WeaponsTableModel.java](https://github.com/MegaMek/megamek/pull/6965/files#diff-178a53b997388ae341b7f3241c0e12b8ae28cb12cd2c695c5637e1dcba485d2f) and [EquipmentType.java](https://github.com/MegaMek/megamek/pull/6965/files#diff-552baa42f708e256fa98bd82b60d540ab560efe178e4f3a4d4cde3c8fc3a7cd4).

Sleet01:
We already have members that will let us disambiguate BA equipment from other equipment.  Why not just use those rather than adding another member - another member that _does not guarantee uniqueness_ - strictly for this purpose?

Issue #6947, opened 2025-04-28
[Issue] When Tiling Maps (2x2) Right Side Maps Hexes Appear Scrambled


Shazrael:
### Brief Description *

[Princess is stuck deploying](https://i.imgur.com/RwRr9WQ.png)

### 3. Steps to Reproduce

1. Start a new game. Create the map as seen. Import my militia .mul list. Generate the army for Princess via "Create Random Army"
_For ease, I have already included a pre-game save file. See "Pregame.sav.gz"_
2. Load in and start deploying. I guess at some point I forgot to turn on ghost targets for my single ECM equipped mech and turned that on (as well as ghost targets for the communications equipment) on my side. You can see my most recent deployment in "Stuck.sav.gz"

Basically all I did was get in and start deploying as normal and the moment it got to Princess turn, she was stuck. The only fix I had was to kick her from game, /skip her turn, and then replace her back into the bot.

### Attach Files

[Pregame.sav.gz](https://github.com/user-attachments/files/19940194/Pregame.sav.gz)
_The Pregame setup_

[Stuck.sav.gz](https://github.com/user-attachments/files/19940197/Stuck.sav.gz)
_The current turn where she is stuck_

[megamek.log](https://github.com/user-attachments/files/19940314/megamek.log)
_megamek log_

[Militia List.zip](https://github.com/user-attachments/files/19940399/Militia.List.zip)
_My custom militia list .mul I imported from .04. Not included is the camo I made_

![Image](https://github.com/user-attachments/assets/18843040-86bb-438a-8dd8-f985560decf4)
_The Camo I made, if that is relevant(?)_

CUSTOM UNITS
[Custom Units.zip](https://github.com/user-attachments/files/19940350/Custom.Units.zip)
_so you can load the save game_

### Severity *

High (Major Disruption): A major feature is broken or incorrect, but a workaround exists.

### MegaMek Suite Version *

0.50.05

### Operating System *

Windows 11

### Java Version *

Java Version: 18

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have included all necessary information and files to help reproduce this issue
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

psikomonkie:
[...]

PR #6969, opened 2025-05-03
fix: add an info nag that informs the user about the danger of tiling odd width size maps

Scoppio:
# What it does?

Opens an info nag for the user which tells them about the risk of using odd width boards when trying to tile them together.
Also adds a new type of dialog box with a "dont bother me" checkbox.

## What else?

A couple of `dispose` were added in strategic places to make sure the dialogs were properly disposed of when its use has finished.

Also adds an entry to recover the nag later in the client settings alongside the other nags.

## Closes

- Partially closes the issue #6947 (it's "enough" for now, but to properly close this problem we need a connector between the boards).


Scoppio:
The PR #6971 is a way better solution than this one. I will close it, and open a new PR with the small memory leak fixes this had in it.

Scoppio:
Adding this, then remove it after the 06 release

Sleet01 (approved):
LGTM!

PR #6970, opened 2025-05-03
feat: add stale action to deal with ever growing number of issues that may not even be valid anymore

Scoppio:
# What it does?

Adds the action `staler` which closes stale issues and PRs based on a set of very simple rules.

Can be configured to consider issues and PRs exempt if they were assigned to specific people or have specific labels or milestones.

This PR is more for you Tex, as you will be the one with the permission level to finish its setup.

### DIFF
--- a/.github/workflows/stale-issues.yml
+++ b/.github/workflows/stale-issues.yml
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/stale@v9
         with:
           stale-issue-message: 'This issue is stale because it has been open 45 days with no activity. Remove stale label or comment or this will be closed in 15 days.'
-          stale-pr-message: 'This PR is stale because it has been open 45 days with no activity. Remove stale label or comment or this will be closed in 10 days.'
+          stale-pr-message: 'This PR is stale because it has been open 45 days with no activity. Remove stale label or comment or this will be closed in 15 days.'
           close-issue-message: 'This issue was closed because it has been stalled for 15 days with no activity.'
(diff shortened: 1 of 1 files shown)


BLR-IIC:
Funny this came up - I actually went through my issues (bug reports) a few weeks ago to see if I should close anyway.  I determined that I could close one of them anyway.  I'm curious to see how many issues end up being stale.  There's 951 of them at the time of this comment.

HammerGS:
> After reviewing alot of old issues (bugs and RFEs) I see that many are still valid and valuable.
> 
> It should only close RFE's that are 12 months old or more, and issues 24 months older or more. And we need a few tags to mark the issue as exempt:
> 
> * CGL related ones
> * Needs investigation
> * Blocker / Severity: Critical
> * Accessibility
> * legal compliance
> * Help Wanted
> * Good First Issue
> 
> We should also have some tags that are to be deleted "soon" after a reasonable period (15 days)
> 
> * Can't reproduce
> * Needs more info
> * Won't fix
> * Invalid
> * Unofficial
> 
> @HammerGS what do you think of those rules?

I think we should take this to the Dev chat on Discord.

Sleet01:
Does this need further review?

HammerGS (approved):

IllianiBird:
Flipping to draft so there is no risk of it being accidentally merged until a consensus has been reached.

HammerGS:
With the dev leaving the projects, closing this.

PR #6973, opened 2025-05-03
fix: i18n support for sorters, reduced a bit the code duplication, prevents NPE if player is null

Scoppio:
# What it does?

Adds i18n support for sorter display names, reduces code duplication, simplifies compare operations, and prevents NPE if player is null in any of those cases.

## Exception
- [NPE](https://sentry.tapenvy.us/organizations/tapenvyus/issues/4659/?alert_rule_id=11&alert_type=issue&notification_uuid=a15a7860-7fa2-46a6-b47a-032275d9efe0&project=8&referrer=discord)


megamek/src/megamek/common/AbstractGame.java
```diff
@@ -110,6 +110,7 @@ public synchronized Forces getForces() {
     }
 
     @Override
+    @Nullable
```
Scoppio, line 113: the original function this overrides is `@Nullable`, so it's important to keep the annotation in the override function, otherwise the IDE may not figure out that a call can return null (which is exactly what was happening).

megamek/src/megamek/common/UnitRole.java
```diff
@@ -411,38 +417,25 @@ public boolean qualifiesForRole(AlphaStrikeElement unit, double tolerance) {
...
-
-        boolean fits(BTObject unit) {
-            return fits.test(unit);
+    private String capitalizeName() {
+        StringBuilder sb = new StringBuilder();
+        for (String word : name().split("_")) {
+            sb.append(Character.toUpperCase(word.charAt(0)))
+                  .append(word.substring(1).toLowerCase()).append(" ");
         }
+        return sb.toString();
```
Scoppio, line 439:
```suggestion
        return sb.toString().trim();
```

megamek/src/megamek/client/ui/swing/lobby/sorters/MekTableSorter.java
```diff
@@ -49,16 +115,27 @@ default boolean isAllowed(IGameOptions opts) {
 
     /** Returns the sorting direction. */
     default Sorting getSortingDirection() {
+        // This probably should be Sorting.ASCENDING, having a default is better than relying on a null behavior
+        // that is implementation dependent
         return null;
     }
```
rjhancock, line 121: Go ahead and assign a default value. I have no objection to that.

rjhancock (changes requested):
I'd want one other to give it a once over as I'm a bit tired while looking at it.

Otherwise it looks good to me and just one change requested.

HammerGS (approved):

PR #6982, opened 2025-05-05
NPE fix on Aero BV Calculation

exeea:
NPE fix on Aero BV Calculation


Scoppio (approved):
Approving, but a minor nitpick, our current style ask to add all individual binary operations on chained tests to be separated by parenthesis. Example.:

```java

if (a != null && b != null) { // non-conforming

...

if ((a != null) && (b != null)) { // conforming
```

I wont enforce, but keep it in mind so next time if you find and work on a non conforming code we expect you to also rewrite that whole expression to be conforming :)

Scoppio (approved):

PR #6984, opened 2025-05-05
Fix: makes sure Mast Mount and its equipments aren't destroyed if the rotor takes 0 damage

Scoppio:
# What it do?

Mast Mount on VTOLs should be destroyed if any damage is caused to the rotor, this fix makes sure that this only happens if any damage (more than zero) is taken at all by the rotor.


HammerGS:
Any Impacts to MekHQ and repairs?

Scoppio:
> Any Impacts to MekHQ and repairs?

Not that I am aware, the change is on TWGameManager, let me take a look on MHQ repair system.

rjhancock (approved):

HammerGS (approved):

HammerGS (approved):

rjhancock (approved):

Issue #1783, opened 2020-03-21
Minefield Graphical Issues


Windchild292:
The first one is that the Conv. 15 is not readable, and would be better located beneath the numbers for the hex..
The second is more arguable, but I believe the minefield should be centered.

![image](https://user-images.githubusercontent.com/39067288/77237245-cad13580-6b9c-11ea-9f7e-5a94463a7582.png)

PR #6985, opened 2025-05-05
fix: moves minefield marker to its appropriate hex (issue #1783) and …

Scoppio:
# What it do?

Moves minefield marker to its appropriate hex (issue #1783) and corrects some messages
Screenshot 2025-05-05 at 17 11 30

The position of the minefield marker was put in that specific height on the hex because it leaves the level label visible 

# Closes

- Fix longstanding issue #1783


Sleet01 (approved):
LGTM!

Issue #6410, opened 2025-01-15
[Issue] Concurrent Modification Error when ragequitting at the end of a bad firing round


OrbMonky:
### Brief Description *

![image](https://github.com/user-attachments/assets/ca4539bd-a1f1-4202-9a7a-ac8690990ba3)

Let's just say my Archer pilot needs to take lessons on how to stand up in a fully functional battlemech, but aside from that, this error presented on the resolution of the firing phase when the report was displayed. I quit the session without dismissing the window and immediately got the above error.

[logs.zip](https://github.com/user-attachments/files/18418280/logs.zip)

### Severity *

Medium (Gameplay Limitation): Non-core functionality is impaired, providing a suboptimal but playable experience.

### MegaMek Suite Version *

0.50.03 Snapshot from latest CO IIC Snapshot

### Operating System *

Windows 10

### Java Version *

17

### Final Verification

- [X] I confirm this is a single, unique issue that hasn't been reported before
- [X] I have included all necessary information and files to help reproduce this issue
- [X] I have asked on MegaMek Discord about this issue
- [X] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

Sleet01:
This has likely been fixed by #6770, which specifically addresses the code in the error.  Closing.

PR #6989, opened 2025-05-06
fix: makes BV Calculator be less prone for concurrent modification exceptions

Scoppio:
# What is the problem?

Concurrent Modification Exception is an error that happens during iteration on lists if that list is altered. There are a couple of things that can be done to deal with that, for example making the access to such lists an immutable copy of the original list... another option is to remove the statefulness of it.

So, this change removes the hashmaps and list from inside the BV Calculator and transform into a message passing.

# Closes

- #6410


Sleet01 (approved):
Looks good to me, although I should note that there are several ways to cause a ConcurrentModificationException that can still happen with a copy of a list;
Q.V. https://docs.oracle.com/javase/8/docs/api/java/util/ConcurrentModificationException.html

I don't think we necessarily need to go to the extents mentioned above, but be aware that if we need to make this class truly CME-incapable, we'll probably need to implement more strict usage of iterators.

PR #6991, opened 2025-05-06
fix: fixed error with armor calculation which would not lower-cap the value at zero.

Scoppio:
# What it does

Fixed error with BV armor calculation which would not lower-cap the value at zero, interpreting instead the values -1, -2 and -3 (which are special values) as negative armor value instead.

A couple of other places had a quick makeup touch-over.

## Closes

- MegaMek/mekhq#6890

### DIFF
--- a/megamek/src/megamek/common/battlevalue/BattleArmorBVCalculator.java
+++ b/megamek/src/megamek/common/battlevalue/BattleArmorBVCalculator.java
@@ -55,10 +55,6 @@ public class BattleArmorBVCalculator extends BVCalculator {
         super(entity);
     }
 
-    private BattleArmor getBattleArmor() {
-        return (BattleArmor) this.entity;
-    }
-
     /**
      * Calculates the Battle Value of a single trooper of this BattleArmor. This
      * value is not influenced by the pilot skill or any force bonuses.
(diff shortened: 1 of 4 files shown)


megamek/src/megamek/common/battlevalue/BattleArmorBVCalculator.java
```diff
@@ -204,30 +220,31 @@ protected void processSummarize() {
...
-                    TechConstants.isClan(battleArmor.getArmorTechLevel(BattleArmor.LOC_TROOPER_1)));
+            final String armorName = EquipmentType.getArmorTypeName(getBattleArmor().getArmorType(BattleArmor.LOC_TROOPER_1),
+                  TechConstants.isClan(getBattleArmor().getArmorTechLevel(BattleArmor.LOC_TROOPER_1)));
             final EquipmentType armor = EquipmentType.get(armorName);
             modifier = " (" + armor.getName().replaceAll("^BA\\s+", "") + ")";
         }
-        defensiveValue += battleArmor.getArmor(currentTrooper) * armorBV + 1;
-        String calculation = "1 + " + battleArmor.getArmor(currentTrooper) + " x " + formatForReport(armorBV)
-                + modifier;
+        int currentArmor = Math.max(0, getBattleArmor().getArmor(currentTrooper));
```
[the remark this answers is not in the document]
Scoppio: ambivalent, not relevant here since it is just class casting Entity

megamek/src/megamek/common/battlevalue/BattleArmorBVCalculator.java
```diff
@@ -35,12 +49,14 @@
...
 
-    private final BattleArmor battleArmor;
     private int currentTrooper;
 
     BattleArmorBVCalculator(Entity entity) {
         super(entity);
-        battleArmor = (BattleArmor) entity;
+    }
+
+    private BattleArmor getBattleArmor() {
```
Sleet01, line 58: NGL, this is pretty gross.  I'd argue that _ideally_ it should either be getEntity() and the caller is responsible for casting, or getBattleArmor() but the class is instantiated with a BattleArmor instance, not an Entity instance.
I guess that would require rewriting the whole BVCalculator hierarchy, though...

Sleet01:
What's the rationale behind doing all the getBattleArmor() refactoring?
I didn't see any connection to the armor value clamp.

Scoppio:
> What's the rationale behind doing all the getBattleArmor() refactoring? I didn't see any connection to the armor value clamp.

To remove a duplicated instance of the entity in itself

megamek/src/megamek/common/battlevalue/BattleArmorBVCalculator.java
Scoppio: PS.: I have a possible solution for all those stateful calculators that are reutilized while not reutilizing anything and only making the code prone to ConcurrentModificationExceptions just because someone changed screens too fast.

Sleet01 (approved):
Looks good!

PR #6993, opened 2025-05-06
Allow formatting run MP without the current game state

pavelbraginskiy:
Required for MegaMek/megameklab#1873.


Scoppio (approved):
Looks ok, not really great to be honest, but if its needed...

PR #6994, opened 2025-05-06
Fix Artillery units not losing autohit mod after spending mp

### LINKED ISSUE #6929: Artillery auto-hit hexes not lost on MP expenditure
### Brief Description *

Tactical Operations: Advanced Rules page 150 indicates that if an artillery unit expends MP after successfully hitting a targeted hex, that hex can no longer be automatically hit. Megamek does not appear to implement this rule.

### 3. Steps to Reproduce

1. Load the provided save with an auto-hit hex active.
2. Expend MP in any way (turn in place, move forward and back into the same hex, etc.).
3. During the Targeting Phase of the next round, observe that the hex can still be automatically hit.

### Attach Files

Save with auto-hit active:
[autohit-active.sav.gz](https://github.com/user-attachments/files/19900543/autohit-active.sav.gz)

A save prior to the targeted hex being struck (with a significant fire correction bonus) is also provided in case anything that occurs when the hex is hit would be useful for debugging this issue.
[pre-autohit.sav.gz](https://github.com/user-attachments/files/19900538/pre-autohit.sav.gz)

### Severity *

### MegaMek Suite Version *

0.50.04

### Operating System *

Windows 10

### Java Version *

Java 17.0.11

### Final Verification

Sleet01:
Fix for Artillery units moving but keeping their auto-hit mods on hexes they've already hit.
This patch adds a new method to the ArtilleryTracker to remove _only_ those autohit mods when a unit expends MP.

Note that other mods (such as base artillery mod, spotter adjustment mod, etc.) should not be modified by this method, as only the auto-hit mod is lost after movement (per TO:AR rules).

Testing:
- Tested with games vs Princess including on- and off-board artillery on various unit types
- Added unit tests

Close #6929

### DIFF
--- a/megamek/src/megamek/server/totalwarfare/MovePathHandler.java
+++ b/megamek/src/megamek/server/totalwarfare/MovePathHandler.java
@@ -125,6 +125,12 @@ class MovePathHandler extends AbstractTWRuleHandler {
     }
 
     void processMovement() {
+        if (md.getMpUsed() > 0) {
+            // All auto-hit hexes for this unit (not including preset targets) are cleared
+            // if any MP are expended.
+            entity.aTracker.clearHitHexMods();
+        }
+
         if (md.contains(MovePath.MoveStepType.EJECT)) {
             if (entity.isLargeCraft() && !entity.isCarcass()) {
                 r = new Report(2026);
--- a/megamek/src/megamek/common/ArtilleryTracker.java
+++ b/megamek/src/megamek/common/ArtilleryTracker.java
@@ -93,6 +93,7 @@ public class ArtilleryTracker implements Serializable {
     /**
      * Remove all autohit mods from hexes that were hit previously; used when artillery unit moves.
      * This _should_ be thread-safe.
+     * Only autohit mods are lost when an artillery unit spends MPs (TO:AR pg. 150)
      */
     public void clearHitHexMods() {
         for (Vector<ArtilleryModifier> modVector : weapons.values()) {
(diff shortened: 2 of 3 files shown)


megamek/src/megamek/server/totalwarfare/MovePathHandler.java
```diff
@@ -125,6 +125,12 @@ class MovePathHandler extends AbstractTWRuleHandler {
     }
 
     void processMovement() {
+        if (md.getMpUsed() > 0) {
+            // All auto-hit hexes for this unit (not including preset targets) are cleared
+            // if any MP are expended.
+            entity.aTracker.clearHitHexMods();
```
Scoppio, line 131: I think a report would be reasonable for this. But not vital.

Scoppio (approved):

megamek/src/megamek/server/totalwarfare/MovePathHandler.java
Sleet01: Sorry, I had stepped away for dinner.
The only issue with reporting is that this is happening during the movement phase, as each move is executed, so reports inserted here would be interleaved with the normal movement reporting.
It feels a little wrong to insert reports about lost auto-hit artillery mods into the move report, but we could store the removed Coords _in_ the ArtilleryTracker and report them during the next Indirect Phase report...
Actually, it might be good to report any auto-hit mods on a per-unit basis as well since we don't do that either.
I'll file an RFE, but I think it's outside the sope of this fix.

PR #6999, opened 2025-05-07
chat retain focus, blinking cursor

### LINKED ISSUE #1058: Chat Box Constantly Loses Focus
### Environment
Megamek 0.44.0 standalone
Java 1.8
Windows 8.1

### Description
If playing a multiplayer game and the players are typing in the chatbox, every time a new phase or round report begins the chatbox loses focus and the player's key presses then begin to trigger the hard coded keybinds of megamek and are no longer typing in the chatbox.  I dont know how difficult this would be to fix / make better with how old that code may be but, I'm sure this contributes to the opinion that MegaMek is an old clunky program.  My screen space is limited so, I wouldnt vote for a pop-out chat window like youtube does but, maybe if the user has selected the chatbox and is typing megamek should allow them to continue as the program is cycling through the phases and doing it's various things without interrupting the player.

### Files
NA

exeea:
This PR makes all the dialogs during a game NOT steal focus.
There is also a tracking for the focus of the main window and, if it was in focus when a new dialog appears, it will try to retake it.
Additionally, the cursor is now blinking and the input text doesn't disappear if the focus is lost (on purpose or not)

Fixes #1058

### DIFF
--- a/megamek/src/megamek/client/ui/swing/ClientGUI.java
+++ b/megamek/src/megamek/client/ui/swing/ClientGUI.java
@@ -34,6 +34,8 @@ import java.awt.event.ActionListener;
 import java.awt.event.ComponentAdapter;
 import java.awt.event.ComponentEvent;
 import java.awt.event.ComponentListener;
+import java.awt.event.FocusEvent;
+import java.awt.event.FocusListener;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


Sleet01 (approved):
LGTM!

Scoppio:
I am merging and will run multiple tests, if this ends up causing a bunch of unrelated bugs we revert and fix those bugs in the second merge of this, ok @exeea @Sleet01 ?

HammerGS:
> I am merging and will run multiple tests, if this ends up causing a bunch of unrelated bugs we revert and fix those bugs in the second merge of this, ok @exeea @Sleet01 ?

Go for it

PR #7043, opened 2025-05-13
removed shortName from AmmoType

exeea:
This PR removes the shortName from AmmoType that was hiding the shortName from EquipmentType.


pavelbraginskiy (approved):

Sleet01 (approved):
LGTM

PR #7044, opened 2025-05-13
Quad vee fixes

### LINKED ISSUE #5752: Quadvees in Vehicle Mode PSR Issues
Quadvees have lots of little things

### Environment
Tested on 49.19.1

### Description
1. Quadvee's are getting a -1 'dedicated pilot' bonus to PSRs.  In interstellar operations, they get no such 'dedicated pilot' bonus.  In vehicle mode, they dont take PSRs from damage though, which is working correctly.  Tripods get a -1 'dedicated pilot' bonus 'when operating a tripod' which is probably where the quadvee bonus came from, but this -1 bonus only applies to tripods, not quadvees.

2. Quadvee's wheel bonus is not supposed to be effected by things like superchargers/tsm, per forum ruling on the bg battletech boards.  However the Notos prime with wheels can supercharge for 12 in vehicle mode.

3. Wheel hits should reduce the wheel'd MP by 1/4th per crit, like the track equipment in TW.  In testing, a wheel crit did not reduce the notos's speed down from 6/9.  This is confirmed on the forums as well.

4. Physicals kicks against a quadvee *in vehicle mode* should be on the *punch table 'from above', for being a mech 1 level less then another mech.  In testing, they are on the full body table like if the quadvee was prone.  "If subject to physical attacks from other ’Mechs, the QuadVee is treated as a ’Mech that stands 1 level lower than its present level (thus a QuadVee being attacked by a BattleMech on the same level is treated as if it is actually standing one level below 
its attacker)."

6. The BV for speed increase from wheels isnt being applied to quadvees, so the 6/9 Notos with wheels is only paying an offensive speed factor of 5/8 in the megamek calculations.  But this may require a ruling from the Errata people on the forums.

MOD NOTE: Edit for readability

DM0000:
fixes #5752

Removed superCharger bonus from quadVee in vehicle mode and set vehicle mode height to 1.


megamek/src/megamek/common/QuadVee.java
```diff
@@ -423,7 +419,7 @@ public int conversionCost() {
     @Override
     public int height() {
         if (getConversionMode() == CONV_MODE_VEHICLE) {
-            return 0;
+            return 1;
```
Scoppio, line 422: I'm not confortable with this, the rulebook says that for LOS it is height 1 (which would me 0 in the code), for physical combat it is heigh 1 (which would mean 0 in the code), changing it from 0 to 1 will put it at the same height as a normal biped mek, showing up behind 1 height buidings or hills, being able to have it's head above the water in depth 1.

Scoppio (changes requested):

megamek/src/megamek/common/QuadVee.java
DM0000: Could add this to melee attacks-----

int quadVeeHeightAdjustment = 0; // Need to adjust height of quads +1 if in vehicle mode
     if(target.isQuadMek()){
      quadVeeHeightAdjustment = game.getEntityOrThrow(target.getId()).getConversionMode() ==CONV_MODE_VEHICLE ? 1 :0;
        }
        
        final int targetElevation = target.getElevation()+ targHex.getLevel()+quadVeeHeightAdjustment ;

DM0000: or a method like that in  PhysicalAttackAction to get target height then implement the method in the other attacks

Scoppio: Dont the entity itself provide the correct "hit table" when being hit on different heights and at different positions?
Maybe the `public abstract HitData rollHitLocation(int table, int side)`is the correct place to make changes?

Scoppio: I believe you are correct, it will need to be handled in the Physical attack action to get target height and elevation difference and refactor the physical attacks to use that instead. In the PAA you can then verify if there are any special cases (quadvee specifically) and be done with it.

Scoppio: Use the static constants instead of the raw int value for comparing the conversion mode, and you can chain the conditions tests, but other than that, I think this looks good.

HammerGS:
Do we have an update on this.

DM0000:
Just want to clarify that if your attacking from below you would use normal table not kick table due to the -1 height?

DM0000:
I'm gonna redo this pr and submit it again.

PR #7045, opened 2025-05-13
Fix: Fixed Naming of MekTech Portrait Folders

IllianiBird:
The 'Mek Tech' folder should be named 'MekTech'. A separate PR has handled this change in MekHQ


psikomonkie (approved):
Nice!

PR #7050, opened 2025-05-14
fix: stops jumps from being made into buildings through roofs

### LINKED ISSUE #2935: 48.0 Infantry can jump through buildings contrary to TW rules
### Description
Per Total Warfare p. 168, if an infantry unit is moving from a building hex to an adjacent building hex (including another hex of a multi-hex building), it must use Ground MP. Similarly, per TW p. 169, when changing levels inside a building, infantry must use Ground MP. 

They may jump _into_ a building hex (at a level below that of the roof) from outside the building, but their movement ends immediately as soon as they enter the first building hex, and they must have line of sight to said hex and elevation from their initial position. Jumping onto the roof is the same as jumping onto a clear hex at higher elevation and is not subject to such restrictions.

Currently, Megamek allows infantry with Jump MP to jump to any elevation of any building hex within their jump radius, regardless of how many floors or walls this involves jumping through. The desired hex is selected by clicking on it as normal, while the desired elevation is selected by using the "Go Up" and "Go Down" buttons. This should be adjusted to match the TW rules.

Reproduction steps:
1. Open a fresh Megamek lobby and select a map with buildings (16x17 Forward Base works well).
2. Add jump-capable infantry. I use Jump Platoon (Rifle) and Cavalier BA, but the type is not particularly important.
3. Experiment with jumping through said buildings in ways that violate the TW rules.

Examples: 

Jumping from outside the building to the ground floor of an interior hex
image
Jumping from the ground floor of one corner to the roof of the far corner
image
Jumping from the roof to the first floor of an adjacent hex
image

_Correct_ behavior would be: jumping from outside the building to any elevation of any exterior building hex that was within LoS of the initial position, jumping from inside the building to any point where the jump path does not pass through any other building hexes, or jumping onto / off of the roof normally. 

TLDR - you can jump into or out of a window but you shouldn't be able to jump through the floors, walls, or ceilings.

Scoppio:
# What it does?

It stops infantry and BA from jumping through roofs into different floor levels on buildings they can try to move through windows but thats all.

# Related Issues

- Closes #2935
- The same issue has a comment which points to another bug, where the PSR isnt requested in every situation, I will address it a next pull request.


megamek/src/megamek/common/Board.java
```diff
@@ -1023,6 +1024,14 @@ public void load(InputStream is) {
         load(is, null, false);
     }
 
+    public void load(String boardString) {
+        try (InputStream is = new ByteArrayInputStream(boardString.getBytes(StandardCharsets.UTF_8))) {
+            load(is, null, false);
+        } catch (IOException ex) {
+            logger.error(ex, "Error loading string to build board - {}", ex.getMessage());
+        }
+    }
```
Scoppio, line 1033: Allows me to load boards from hardcoded strings, great to make mock boards very quickly for tests

megamek/src/megamek/common/Entity.java
```diff
@@ -7182,7 +7182,9 @@ else if (!getCrew().isActive()) {
         }
 
         // okay, let's figure out the stuff then
-        roll = new PilotingRollData(entityId, getCrew().getPiloting(moveType), "Base piloting skill");
+        roll = new PilotingRollData(entityId, getCrew().getPiloting(moveType), (this instanceof Infantry) ?
+                                                                                     "Anti-Mek skill":
+                                                                                     "Base piloting skill");
```
Scoppio, line 7180: I believe this is correct, but if anyone wants to correct me, please do.

PR #7057, opened 2025-05-14
Fix: fixed a couple more problems with jumping in and out of buildings with BA/Infantry and Nags for jumping into buildings

Scoppio:
# What it does?

When you jump into a building through the window with a BA or infantry you have to roll an AntiMek check, this fix makes sure it happens if jumping into a building lower down or into a floor higher than you currently are.

Also makes it illegal to jump inside a floor to move, you should not be able to jump inside a bulding, only to leave the building or to enter the building.

Also stops you from moving through floors from inside to the roof.

A bunch of unit tests that evaluate jumping paths, this also serves as a first example on how to test movepaths without having to mock everything, and adds a base class to make it easier to implement this kind of test for movePaths


Sleet01 (approved):
LGTM!

PR #7074, opened 2025-05-15
PR: Added keyboard shortcuts for forward and backward movement.

GamesByFelix:
Added keyboard shortcuts to move one hex forward or backward, so that fewer mouse interactions are required during the movement phase. With Shift+WASD one can now freely move around without the mouse.


GamesByFelix:
Please re-open. I lost my fork, etc :-(

Issue #7002, opened 2025-05-08
[Issue] [0.50.06-nightly-2025-05-07] Ammunition resetting to top bin when weapon is selected in list


PhoenixHeart512:
### Brief Description *

When a weapon has multiple ammunition bins/types available, it is resetting to the top bin in the list if the weapon is selected again after ammo is changed.

### 3. Steps to Reproduce

For example, if you have an LBX10 with slug in the top bin and cluster in the bottom bin, if you:
- Set the weapon to use cluster
- Click to a different weapon
- Then click back to the LBX10
- The ammo will reset back to using the top bin (slug)

### Attach Files

[AmmoBug.sav.gz](https://github.com/user-attachments/files/20097251/AmmoBug.sav.gz)

### Severity *

High (Major Disruption): A major feature is broken or incorrect, but a workaround exists.

### MegaMek Suite Version *

0.50.06-nightly-2025-05-07

### Operating System *

Windows 10

### Java Version *

Java Vendor: Eclipse Adoptium     Java Version: 17.0.6

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have included all necessary information and files to help reproduce this issue
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

PhoenixHeart512:
@Scoppio Illiani asked me to tag you. It looks like this was merged 8 hours ago, but II tested the 5/17 nightly just now and it is still occurring in that nightly. Not sure if it just didn't make it in in time, or if it did then the issue is still occurring.

Scoppio:
It was merged after the nightly was buikt

Scoppio:
It must have been added after the nightly was built, because I tested it a few times, both in the same round and between rounds.

PR #7078, opened 2025-05-16
fix: AmmoBin selection switching back to the first AmmoBin in the weapons display

Scoppio:
# What it does?

For some reason, it was trying to select an ammo bin from an uninitialized array which should contain all ammobins for that weapon. This would make the selector always reset to the first one. This fix makes sure that the ammo array is initialized before it is accessed.

# QOL

This also adds a more comprehensive warnin log for wrong equipment flag checks, and fixes one instance of a flag being checked in the wrong equipment type.

# Related issues

- closes issue #7002


Sleet01 (approved):
LGTM

PR #7080, opened 2025-05-16
fix: minimap and unit display and mini report dialog priority against other windows

Scoppio:
# What it does?

Keeps minimap dialog and unit display dialog always ontop of the board view, makes the report dialog not show always ontop of the rest this way it does not cover the save dialogs.


Sleet01 (approved):
LGTM

PR #7092, opened 2025-05-18
fix: unit tooltip fixed for aero, also fixed one mistake with html tags for pilot tooltip

### LINKED ISSUE #6870: [Issue] Dropship weapon ranges tooltip incorrect
### Brief Description *

While the ranges are correct on the Unit Display and seem to be correct in what's fired, the hovering tooltip insists that every weapon has a range profile of 6/12/20. 

Image

### 3. Steps to Reproduce

_No response_

### Attach Files

[Dropship-Ranges.sav.gz](https://github.com/user-attachments/files/19716906/Dropship-Ranges.sav.gz)

### Severity *

### MegaMek Suite Version *

50.04

### Operating System *

Windows 11

### Java Version *

Java 17

### Final Verification

Scoppio:
# What it does?

Unit tooltip for landed aero units was showing aerospace range brackets instead of TW range brackets. There as also one minor mistake with Pilot tooltip html which was adding a broken tag `<<div>` which would then close with `</<div>`, apparently nothing broke because of that error, but it was corrected and the final result still the same.

# Related issues
- Closes #6870


Sleet01 (approved):
LGTM

megamek/src/megamek/common/weapons/WeaponHandler.java
```diff
@@ -171,9 +171,9 @@ protected int getLargeCraftHeat(Entity entity) {
                     int loc = prevWeapon.getLocation();
 
                     // create an array of booleans of locations
-                    boolean[] usedFrontArc = new boolean[ae.locations()];
-                    boolean[] usedRearArc = new boolean[ae.locations()];
-                    for (int i = 0; i < ae.locations(); i++) {
+                    boolean[] usedFrontArc = new boolean[entity.locations()];
+                    boolean[] usedRearArc = new boolean[entity.locations()];
+                    for (int i = 0; i < entity.locations(); i++) {
```
Scoppio, line 176:
```suggestion
                    boolean[] usedFrontArc = new boolean[ae.locations()];
                    boolean[] usedRearArc = new boolean[ae.locations()];
                    for (int i = 0; i < ae.locations(); i++) {
```

this should not be in this PR, I am reverting it

Sleet01: I was looking at that recently so it seemed perfectly natural -_-;

Scoppio: I believe it is correct and should be entity, but anyway... I was not working with this file, so it should not be pushed

PR #7096, opened 2025-05-19
Fix: Keyboard shortcuts didn't work during the Deployment Phase.

GamesByFelix:
The "Next Unit" keyboard shortcut never seemed to work during the Deployment Phase. I fixed this and also added a keyboard shortcut to the "Turn" Unit button to make deployment quicker.


GamesByFelix:
Please re-open. I lost my fork, etc :-(

Issue #3396, opened 2022-01-23
Specialty Armor for ASF and damage Thresholds


Logion567:
Fill out the fields below with as much pertinent information as possible to help the developers solve the issue.
Search for your issue inside open and closed issues first before reporting so as not to make duplicates.
(Delete the above lines)

### Environment
MM version .49.5
### Description
Specialty Armors for ASF still trigger Damage Thresholds critical checks for their respective damage type. 
![image](https://user-images.githubusercontent.com/98195429/150662672-0ef1de14-caf5-4adb-a4f5-d368fc66cc4e.png)
this forums post says that only damage counted in the record sheet/against the armor after damage reduction counts for these thresholds https://bg.battletech.com/forums/total-warfare/(answered)-prevented-or-reduced-damage-and-psr/

### Files
[DAMAGE.sav.gz](https://github.com/MegaMek/megamek/files/7919632/DAMAGE.sav.gz)
the overall setup is there for ease of testing, should give plenty of room for figuring out how the logic is breaking

My Name is Logion567 (as seen in the screenshot)

HammerGS:
Related to https://github.com/MegaMek/megamek/issues/2032

Issue #2209, opened 2020-09-11
Battle armor reactive armor not working against artillery weapons


AlphaSled:
Version 0.47.7

New ticket created at the behest of Windchild so that this doesn't get forgotten.

When mounted on battle armor, reactive armor is not reducing damage against standard artillery (both high explosive and homing) or artillery cannons.

Relevant rules may be found on page 282 of Tactical Operations.
Attached is a game file where most of these were tested.

[explosive.test.sav.gz](https://github.com/MegaMek/megamek/files/5210637/explosive.test.sav.gz)

Issue #2032, opened 2020-06-27
Specialty armors handling PSRs incorrectly


AlphaSled:
v.47.7

This relates to the issue that has previously been posted here: https://github.com/MegaMek/megamek/issues/1377
In addition to hardened armor, every other specialty armor is currently handling PSRs based off of damaged received before the damage reduction happens (eg: 2 PPCs vs reflective armor), not the actual damage the unit receives after the damage reduction occurs (eg: 10 damage, 5 from each PPC).

Armors that were tested and that are affected are Ballistic-Reinforced, Reflective, Reactive, Impact-Resistant, Ferro-Lamellor, and of course Hardened. Interestingly, Blue Shield and Spikes are OK and handle things correctly.

Relevant rules on in the BattleMech Manual, page 53.

I can attach some game files if they would be any help.

Logajam:
Curses.  Someone reported this already.  Now I have nothing to complain about.  Darn.

HammerGS:
Adding a clarification I found Xotl posted.

_The main change is a clarification that PSRs caused by damage are calculated based on the damage actually dealt to the mech, after any modifications to that total.  However, hardened armour only counts each full armour circle it loses as "one point of damage" for this purpose.  So you'd have to do 40 damage to a unit with hardened armour to reach that 20 pt PSR threshold._

HammerGS:
See also https://github.com/MegaMek/megamek/issues/3396 Specialty Armor for ASF and damage Thresholds

HammerGS:
Related errata post - https://bg.battletech.com/forums/errata/battlemech-manual-16-august-2020-v3-21/msg1879926/#msg1879926

BMM Page 106, AP Ammo, last Game Rules para:

This ammunition loses its AP ability against hardened or ferro-lamellor armor. Its damage is not reduced by the latter, however.
Change to:
This ammunition loses its AP ability against hardened or ferro-lamellor armor. However, the ammunition treats both armor types in all other ways as standard armor (no damage reduction, 1 point of damage eliminates one armor circle).

PhoenixHeart512:
Still present in 49.12, and I am still interested in seeing it fixed as someone who enjoys using hardened armor on things that walk

HammerGS:
> Still present in 49.12, and I am still interested in seeing it fixed as someone who enjoys using hardened armor on things that walk

If the report is open it's likely unresolved. We've looked into this and it's a very complex fix given how the armor code is currently written.

[...]

PR #7108, opened 2025-05-21
Fix advanced armors handling PSRs, Piloting Rolls, damage reduction incorrectly

Sleet01:
This PR contains fixes for a number of advanced armors not preventing PSRs or piloting rolls as they should.
It also contains updates for some Aerospace and BA armor handling edge cases.
Finally, it implements a TWDamageManager class and moves the entirety of direct damage assignment function
`damageEntity` out of TWGameManager into a separate class that TWGameManager will configure through composition.

The rationale behind splitting damageEntity() out of TWGameManager is three-fold:
1. It allows for switching between the current and new code with a single option, so that breaking bugs needn't hold up other testing or development;
2. It allows refactoring of the original function into specialized functions which will simplify coding and unit testing;
3. It moves code out of TWGameManager, which is bloated, monolithic, and handling far too much.

A new game option has been added to allow disabling the new damage manager code if desired / needed while further fixes are implemented.

### Going Forward ###
There are a few more edge-case bugs that may need to be added to this PR but I would like to get the initial work in for testing purposes.

There is also a good argument for pulling more damage-adjacent handling code out of TWGameManager into TWDamageManager, but that is outside my remit - the work so far was in service of simplifying `damageEntity` so that specific issues around advanced armor could be fix, and no more.
If it makes sense to move functions out of TWGameManager then another PR may be in order.

Testing:
- Added new unit tests specific to the reported issues;
- Ran all 3 projects' existing unit tests;
- Ran large bot-controlled battles for several hours each.

Fix: #2032 
Fix: #2209 
Fix: #3396


megamek/src/megamek/server/IDamageManager.java
```diff
@@ -0,0 +1,10 @@
+package megamek.server;
+
+import megamek.common.DamageInfo;
+import megamek.common.Report;
+
+import java.util.Vector;
+
+public interface IDamageManager {
+    public Vector<Report> damageEntity(DamageInfo damageInfo);
```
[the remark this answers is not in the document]
Sleet01: Not at this time.

megamek/i18n/megamek/common/options/messages.properties
```diff
@@ -85,6 +85,9 @@ GameOptionsInfo.option.hide_unofficial.displayableName=Don't show unofficial gam
 GameOptionsInfo.option.hide_unofficial.description=If checked, unofficial game options will be deactivated and hidden
 GameOptionsInfo.option.hide_legacy.displayableName=Don't show legacy game options
 GameOptionsInfo.option.hide_legacy.description=If checked, legacy game options, i.e. official rules that are superseded by more modern rules, will be deactivated and hidden
+GameOptionsInfo.option.new_damage_manager.displayableName=Use new damage handling system
+GameOptionsInfo.option.new_damage_manager.description=Use new code to handle damage application, may be buggy
```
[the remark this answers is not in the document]
Sleet01: This will stay until 0.50.07 is fully tested.

HammerGS:
Asked AI about the Java Doc failure. Says this. 

1. Fix Javadoc Errors in Game.java
Errors and warnings in Game.java include missing @param, @return, and @throws tags for certain methods.

Code Suggestions:
Method: getNextHiddenEntityNum Add the missing @param description for start.

```
/**
 * Retrieves the next hidden entity number.
 * 
 * @param turn the current game turn
 * @param start the starting point for the search
 * @return the next hidden entity number
 */
public int getNextHiddenEntityNum(GameTurn turn, int start) {
    // Method implementation
}
```
Method: removeFirstTurnFor Add the missing @param, @return, and @throws tags.

```
/**
 * Removes the first turn for the given entity.
 * 
 * @param entity the entity whose turn is to be removed
 * @return the removed game turn, or null if none exists
 * @throws Exception if an error occurs during removal
 */
public @Nullable GameTurn removeFirstTurnFor(final Entity entity) throws Exception {
    // Method implementation
}
```

Method: setAllReports Add the missing @param tag.

```
/**
 * Sets all reports for the game.
 * 
 * @param v a list of lists of reports to set
 */
public void setAllReports(List<List<Report>> v) {
    // Method implementation
}
```

2. Fix XML Tag Issues in TWGameManager.java
The error indicates improperly closed <code> tags in the Javadoc comments.

Code Suggestions:
Replace the incorrect <code> tags with properly closed ones.

```
/**
 * <code>Report</code>s to. It may be empty, but not null.
 */

```

Sleet01:
@HammerGS The Game.java stuff is outside my scope, but I can try fixing the tag error.

megamek/unittests/megamek/server/totalwarfare/TWDamageManagerTest.java
```diff
@@ -0,0 +1,544 @@
+package megamek.server.totalwarfare;
+
+import megamek.common.*;
+import megamek.common.options.GameOptions;
+import megamek.common.options.IOption;
+import megamek.common.options.Option;
+import megamek.common.options.OptionsConstants;
+import megamek.common.weapons.DamageType;
+import org.junit.jupiter.api.BeforeAll;
```
rjhancock, line 9: Copyright Header....

rjhancock (changes requested):
My only sticking point is the copyright headers.

The rest are just requests along with... please do something about the variable names.

megamek/unittests/megamek/server/totalwarfare/TWDamageManagerTest.java
Sleet01: Renaming the variables will take a while.

Sleet01: Actually, can I get some clarification on which variables need renaming?
I don't plan to do any additional work on the code in `TWDamageManager::damageEntity` because it is meant to be as close to the original damage application code as possible.

rjhancock: It was items such as `r -> report` kind of stuff. Make the variables more human readable. Things that don't impact the code except make it more readable.

It's just a request, not a requirement.

Sleet01: I imagine that future me will appreciate it if we do this now.  :)

Sleet01:
One of the recent changes has introduced an OOM at the end of the firing phase on very large matches; I've confirmed it's not in the old damage code so I'm taking this to Draft to find the issue.

Sleet01:
Turns out early return escapes need to be replaced with _something_ after they are removed.
Now with the deprecation updates as well.

megamek/src/megamek/common/DamageInfo.java
```diff
@@ -0,0 +1,172 @@
...
+ * Microsoft's "Game Content Usage Rules"
+ * <https://www.xbox.com/en-US/developers/rules> and it is not endorsed by or
+ * affiliated with Microsoft.
+ */
+
+package megamek.common;
+
+import megamek.common.weapons.DamageType;
+
+public record DamageInfo(
```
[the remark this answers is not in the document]
Scoppio: Ok, can we not have entity be `te`, or hit data be `hit`?
I also have no idea what means nukeS2S or `areaSatArty`, those booleans also make it hard to parse what is happening here. I'd prefer to see a bunch of enums to make it more explicit but... Let's not get ourselves stoped because of this, those things can go on a second step, but would be nice if that second step happened in the near future.

megamek/src/megamek/common/DamageInfo.java
```diff
@@ -0,0 +1,172 @@
...
+        this(
+              te,
+              hit,
+              damage,
+              ammoExplosion,
+              DamageType.NONE,
+              false,
+              false
+        );
+    }
```
Scoppio, line 82: I know this completely personal preference, but I'd like to expose that my personal preference is to use static functions that build the class instead of having multiple constructors, this way there is only "one right way" to construct the class and all the others are effectively auxiliary functions to facilitate on that process.

megamek/testresources/megamek/common/units/Flashman FLS-10E.mtf
```diff
@@ -0,0 +1,178 @@
...
+
```
Scoppio, line 170: Unless those units were selected for very special reasons, I suggest to instead of adding them as test resources for you to hardcode on a constant the chassis + model name on the test and use the MekSummary to load the entities, they are accessible in tests (I did that a couple of times :) )

megamek/testresources/megamek/common/units/Seydlitz C.blk
```diff
@@ -0,0 +1,113 @@
+#building block data file
+<BlockVersion>
+1
+</BlockVersion>
+
```
Scoppio, line 5: I think @rjhancock will agree with me that adding (duplicating) canon data in the tests may not be the best option given how we want to manage datafiles on the project, so I suggest that you either do the thing where it loads the file it's own data, or instead you add unnoficial units built exclusively for tests

Scoppio (dismissed):
If TWDamageManager needs both GameManager and Game... can't it use the game reference from the GameManager itself? Also, since it is so vital for it, would be best if the TWDamageManager had to be be initialized on its constructor with a TWGameManager, solving all possible error states because of bad initialization.

Scoppio:
Also, can you rename the "new" damage manager to something else?
Like "TWDamageManagerModular", the other one could be named "TWDamageManagerLegacy", this way we dont endup with any `NTWDamageManagerNewNewThisOneFinal`

rjhancock (approved):

megamek/testresources/megamek/common/units/Seydlitz C.blk
rjhancock: Actually, data for tests should be isolated from everything else and placed IN the test resources location SPECIFICALLY to deal with test isolation and ensure we can test good and bad files without impacting live data.

megamek/src/megamek/common/DamageInfo.java
Sleet01: This is a lightweight dataclass to simplify the interface to TWDamageManager and reduce some of the multiple chained function calls in TWGameManager; adding any further functionality would just increase bug surface area and require further refactoring of TWGameManager.

Sleet01:
> If TWDamageManager needs both GameManager and Game... can't it use the game reference from the GameManager itself? Also, since it is so vital for it, would be best if the TWDamageManager had to be be initialized on its constructor with a TWGameManager, solving all possible error states because of bad initialization.

Are you familiar with the "Chicken and Egg" problem? :smile_cat: 
I've created a mutual dependency between TWGameManager, which now cannot run a game without TWDamageManager, and TWDamageManager, which needs functions on TWGameManager to do its work.
This means that only one of them can instantiate the other within its constructor, not both.
Some languages would be able to resolve this with a double-pass approach under the hood, but it doesn't look like Java is one of them.
So one of the two classes _cannot_ require the other for its constructor, and I decided that would be TWDamageManager as this seemed the simplest to deal with.

Further, while we _currently_ automatically create a Game instance when instantiating TWGameManager, we may not always - and decoupling the Game instance from the TWGameManager instance makes for easier testing.

In the near future I plan to move more functionality out of TWGameManager into TWDamageManager, which should remove this mutual dependency, but for now I wanted to get the new damage handling code out and tested by players.

Sleet01:
> Also, can you rename the "new" damage manager to something else? Like "TWDamageManagerModular", the other one could be named "TWDamageManagerLegacy", this way we dont endup with any `NTWDamageManagerNewNewThisOneFinal`

I can do this, but the "Legacy" TWDamageManager is going away once we're comfortable with the new code, and the new code will replace its innards.  So we'll only have TWDamageManagerModular for 1~2 releases.

PR #7122, opened 2025-05-23
check actuator before add damage to shield + fixed shields rules ref

exeea:
This PR fixes fixes the fact that mechs with shields mounted on arms without specific actuators, started with the shield "damaged" (less pips).
Additionally, fixed rules reference.

### DIFF
--- a/megamek/src/megamek/common/MiscType.java
+++ b/megamek/src/megamek/common/MiscType.java
@@ -3972,7 +3972,7 @@ public class MiscType extends EquipmentType {
         misc.setModes(modes);
         misc.baseDamageAbsorptionRate = 3;
         misc.baseDamageCapacity = 11;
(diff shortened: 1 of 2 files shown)


Scoppio (approved):

Schnueferl:
Hi!
I've just started to use shields in my games, so I'm very grateful to all you the people who spend your time and effort to fix it!
Just wanted to point that there might be some misunderstanding (or misunderstanding from my side, English is not my native language). The construction rules on p.100 of TO:AUE state that "Shields have no special actuator requirements, but each missing arm or hand actuator in the arm containing the shield reduces its DA and DC by 1 (see p. 103)." So it looks like the mechs with shields mounted on arms without specific actuators must start with the shield having less hit points.
I'm sorry if this is an incorrect place for my post, added it here just because I found that the change has been made in this PR.

PR #7129, opened 2025-05-24
Fix/show only possible moves

Scoppio:
# What it does?

This change makes so that only possible moves are shown in the "show possible moves" and "show best modifier" actions.

## Before

The hexes would light up as long as it is inside the max MP for the unit, even if such movement would be considered illegal (over water you cant run, only walk)

Screenshot 2025-05-24 at 11 41 23

## After

All illegal moves are now not shown anymore (manual tests made with fast and slow units, also with VTOL, conventional wing and ASF units to make sure all is as correct as it should be)
Screenshot 2025-05-24 at 11 33 13

The "Show best modifier" is also updated with this functionality. Just remember that it will show you the best possible, not how to obtain it ;)

Screenshot 2025-05-24 at 12 11 35


rjhancock (commented):
Only did a quick pass, but looks good.

PR #7133, opened 2025-05-25
Fix: regex for words of interest

Scoppio:
Added a few extra words and removed others to better comply with compliance


Sleet01 (approved):
LGTM

PR #7145, opened 2025-05-26
Restore previous i18n functionality

SJuliez:
Adds some empty i18n _en files back and reverts changes in some _de to restore i18n functionality for those with a system language other than English.


rjhancock (approved):
I'll be fixing I18N sometime this cycle (I hope).

PR #7151, opened 2025-05-27
optimizations for RawImagePanel

exeea:
This PR optimizes RawImagePanel by using a cached scaled image with H/W acceleration. Improves massively the repainting speed.

It also:
- add new common method ImageUtil.convertToVolatileImage()
- uses RawImagePanel for MM startup GUI


rjhancock (approved):
Sanity Check Complete.

PR #7164, opened 2025-05-29
Fix: Rewording of Tech Rating Summary, tooltips and introdates

Scoppio:
# What it does?

Better wording for the tech rating summary on the Mek View Summary which aligns with stablished canon and official language used for organized play.

It also resolves the confusion where units would show starting dates which were not capped by their intro dates.

Also adds a better view of the split in the timeline of production of units taking into consideration the extinction and reintroduction periods.

# Edit:

Added the word "and onwards" instead of the dash the era text had to signify that it is "from now and to the future"

## Before
Screenshot 2025-05-28 at 18 34 29

## After
Screenshot 2025-05-28 at 22 09 46

### Tooltips

Screenshot 2025-05-28 at 22 09 16

### VTL/Rules Level column

When changing between VTL (Variable Tech Level) and Rules Level mode of "unit limiting", the column will also change from VTL to Rules Level.

Screenshot 2025-05-28 at 22 16 03

Screenshot 2025-05-28 at 22 15 23

### DIFF
--- a/megamek/unittests/megamek/common/EntityTechTest.java
+++ b/megamek/unittests/megamek/common/EntityTechTest.java
@@ -49,7 +49,7 @@ class EntityTechTest {
     }
 
     @Test
-    public void testDateRanges() {
+    public void testDateRangesChippewa() {
         AeroSpaceFighter entity = EntityLoader.loadFromFile("Chippewa CHP-W7.blk", AeroSpaceFighter.class);
         printEntity(entity);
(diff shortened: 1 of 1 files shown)


Sleet01 (approved):
LGTM!

HammerGS:
@Scoppio From Co-pilot on the errors

_The failing job indicates errors in the EntityTechTest class, specifically with the methods testDateRanges() and testDateRanges2(). Unfortunately, the file containing these tests (src/test/java/megamek/common/EntityTechTest.java) could not be accessed for further inspection._

PR #7170, opened 2025-05-29
(Multimap) Key dispatcher adaptations, remove unit load failure dlg

SJuliez:
- Changes the boardview handler to make any boardviews that aren't shown (in a multimap game) to not listen to key presses. 
- Prevents the "unit loading failures" dialog from showing up everywhere (lobby, scenarios), instead adds a warning in the log (we have 4 errors atm)
- Removes a key listener from the minimap dialog class because it created exceptions in java's internal DefaultKeyDispatcher. I believe it's not necessary as the focus of minimaps is being managed elsewhere, but maybe we have to revisit that at some point.


rjhancock (approved):
Sanity Check Complete.

PR #7171, opened 2025-05-29
(Multimap) Princess

SJuliez:
This enables Princess to have baseline functionality in a multimap game. Each unit acts as if other maps and units thereon don't exist. Included is a test scenario of 1 Grasshopper and 1 Sai aerospace fighter on an atmospheric map (Human) against 2 Grasshoppers (Princess). The Sai cannot enter the ground map but can make flight paths and strafe/strike. 
This is again a bit of a bigger PR as it required changing many game.getboard() calls across various pathfinders and movepath/movestep among some other adaptations.


rjhancock (approved):
Sanity Check Complete.

PR #7204, opened 2025-06-03
PR: Keyboard Zoom towards Mouse Cursor or View Center + Toggle between two Zoom Levels.

GamesByFelix:
- When zooming with the keyboard shortcuts, the zoom now is centered on the mouse cursor (like when zooming with the mouse wheel). If the cursor is not on the map board, the center of the viewer is zoomed to instead.
- Added the Menubar Entry "View - Toggle Overview Zoom" (Shortcut: Z), which toggles between the current view zoom/position and a zoomed-out view of the map board. This lets you quickly get an overview of the battlefield and then jump back into the action.


Sleet01 (approved):
LGTM

PR #7209, opened 2025-06-05
Adv board search no tag

kuronekochomusuke:
- correct issue with advanced board search filter not returning boards with no tags

image


Sleet01 (approved):
LGTM

PR #7217, opened 2025-06-06
PR: Added separate Filter Keyword Combobox (incl. Settings, etc.)

GamesByFelix:
Instead of using the "keyword" drop-down box for the filter functionality, I added a separate drop-down box for the filter with its own keywords, allowing one to filter the report and still use the regular keywords (e.g. "hit") and the arrows to search and jump through the _filtered_ report. This wasn't conveniently possible before, so it's better to separate those two features.

Screenshot 2025-06-06 at 20 35 28

Just like with the regular keywords, the user may configure the filter keywords in Client Settings -> Report.

### DIFF
--- a/megamek/build.gradle
+++ b/megamek/build.gradle
@@ -38,6 +38,7 @@ sourceSets {
 dependencies {
     implementation 'com.fasterxml.jackson.core:jackson-core:2.19.0'
     implementation 'com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.19.0'
+    implementation 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310:2.19.0'
 
     implementation 'com.formdev:flatlaf:3.6'
     implementation 'com.formdev:flatlaf-extras:3.6'
@@ -51,7 +52,7 @@ dependencies {
     implementation 'jakarta.xml.bind:jakarta.xml.bind-api:4.0.2'
 
     implementation 'org.apache.commons:commons-collections4:4.5.0'
-    implementation 'org.apache.commons:commons-lang3:3.17.0'
+    implementation 'org.apache.commons:commons-lang3:3.18.0'
     implementation 'org.apache.commons:commons-text:1.13.1'
 
     implementation 'org.apache.logging.log4j:log4j-api:2.24.3'
--- a/megamek/mmconf/milestoneReleases.yml
+++ b/megamek/mmconf/milestoneReleases.yml
@@ -1,0 +1,15 @@
+# This file should NOT be edited by non-developers.
+# It WILL brick your game, and it won't be immediately obvious until it's too late.
+milestone_releases:
+  - label: "v0.48.0" # This is the user-facing label, it also corresponds to the tail-end of the GitHub download page
+    version: "0.48.0" # This is used to construct the version information used in MekHQ to verify compatibility.
+    useFallbackHyperlink: false # Set to 'true' in the event that a Web Page is unavailable for this release
+  - label: "v0.49.7"
+    version: "0.49.7"
+    useFallbackHyperlink: false
+  - label: "v0.49.19.1"
+    version: "0.49.19-01"
+    useFallbackHyperlink: false
+  - label: "v0.50.06"
+    version: "0.50.06"
+    useFallbackHyperlink: false
--- a/megamek/resources/Version.properties
+++ b/megamek/resources/Version.properties
@@ -1,3 +1,5 @@
+# Was the last release confirmed as a Milestone?
+# Please update mmconf/milestoneReleases.yml
 major=0
 minor=50
 patch=7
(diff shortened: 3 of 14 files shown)


megamek/src/megamek/common/preference/ClientPreferences.java
```diff
@@ -70,6 +70,8 @@ public class ClientPreferences extends PreferenceStoreProxy {
     public static final String MAP_HEIGHT = "MapHeight";
     public static final String REPORT_KEYWORDS = "ReportKeywords";
     private static final String REPORTKEYWORDSDEFAULTS = "Needs\nRolls\nTakes\nHit\nFalls\nSkill Roll\nPilot Skill\nPhase\nDestroyed\nDamage";
+    public static final String REPORT_FILTER_KEYWORDS = "ReportFilterKeywords";
+    private static final String REPORTFILTERKEYWORDSDEFAULTS = "Fire Hit Damage\nHit Damage";
     public static final String IP_ADDRESSES_IN_CHAT = "IPAddressesInChat";
     public static final String START_SEARCHLIGHTS_ON = "StartSearchlightsOn";
     public static final String ENABLE_EXPERIMENTAL_BOT_FEATURES = "EnableExperimentalBotFeatures";
```
rjhancock: Please adjust this for readability. Add `_` for spaces between words.

GamesByFelix:
> I'm unconfortable with the adition of even more dropdowns here. If there is need to add more ways to do complex searches then why not use a query system? I wont block this feature, but I'd prefer to see a simpler UI with more powerful functionalities instead of a more granular with single-effect
> 
> Currently only blocker is the naming of the variable not following the style

Personally, I feel the current system works really well and can be set up and also changed quickly and easily. It's also easy to understand (at least it was for me as a new player). I agree though, that another drop-down adds visual clutter and perceived complexity and the whole thing just doesn't look great to begin with.

I'm not sure how a better system would look like at the moment, but I'll think about it. Switching this PR to draft in the meantime.

Scoppio:
I'm sure it works as intended and is easy to use (but me, I never could get my head around to what thise buttons did or when to use them, so to me personally they look very confusing as they currently are), dont need to stop this PR because of my worry, we can figure out something different in the (near) future.
The thing is that this is starting to get close to a point where the UI is daunting, many drop downs and entry fields, too many non descriptive butons and gliphs...

megamek/src/megamek/common/preference/ClientPreferences.java
GamesByFelix: Fixed

GamesByFelix:
After chatting about this on Discord, the idea came up to make the display of those drop-downs optional, since it appears as if players use them quite differently (or not at all?).

I added checkboxes to the Client Settings, with the defaults being the way it has been (i.e. no filter drop-down and button):
Screenshot 2025-06-08 at 18 18 15


With the checkboxes, the visibility of the components can be customized, e.g. only keyword search and filter:
Screenshot 2025-06-08 at 17 57 22

What is your opinion on this @Scoppio ?

megamek/src/megamek/client/ui/dialogs/buttonDialogs/CommonSettingsDialog.java
```diff
@@ -3050,6 +3138,11 @@
         GUIP.setReportrInfoColo(csbReportInfoColor.getColour());
         GUIP.setReportFontType(fontTypeChooserReportFont.getSelectedItem().toString());
         GUIP.setMiniReportShowSprites(showReportSprites.isSelected());
+        GUIP.setMiniReportShowPlayers(chkReportShowPlayers.isSelected());
+        GUIP.setMiniReportShowUnits(chkReportShowUnits.isSelected());
+        GUIP.setMiniReportShowKeywords(chkReportShowKeywords.isSelected());
+        GUIP.setMiniReportShowFilter(chkReportShowFilter.isSelected());
+        clientgui.getMiniReportDisplay().refreshSearchPanel();
```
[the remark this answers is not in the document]
GamesByFelix: Fixed!

GamesByFelix:
Something I've been thinking about regarding UI clutter and new users: how about making the default search/filter dropdown only the keyword search list and up/down buttons? This is probably the most useful feature at the beginning and it would only be a single dropdown then next to the "switch location" button.

Screenshot 2025-06-12 at 20 54 17


When players dive into the client settings then after a few games, they can further explore the features.

EDIT: This is not in the code, just an idea. At the moment the UI is not changed from the default (i.e. the filter box and button are not shown, but can be enabled in the client settings)

GamesByFelix:
Added some tooltips and switched this back to "for review".

HammerGS:
@Scoppio @rjhancock Leaving this until you look at current state

IllianiBird:
@HammerGS this might need to be Admin merged now that it has approval. It definitely wants to be in before the data merge and I don't think it's fair to exclude it and then the contributor to have to deal with who knows how many conflicts.

HammerGS:
Closing as the conflicts where fixed here - https://github.com/MegaMek/megamek/pull/7477

@GamesByFelix Thank you for the contribution and sorry this took so long.

PR #7224, opened 2025-06-07
PR: Added functionality to remove specific keybinds by pressing Esc

GamesByFelix:
When binding a key to a command in the Client Settings, you can now press Esc to remove the keybind and leave the command unbound. This opens up the possibility to add commands without a default keybind (simply set the default keyCode to 0), which prevents overloading the keyboard with functions, letting the user configure only the shortcuts they need.

Screenshot 2025-06-07 at 23 40 06


Sleet01 (approved):
LGTM

PR #7231, opened 2025-06-08
Infantry Field Gun validity tests

SJuliez:
Merge together with MegaMek/MegaMekLab#1933

Adds field gun validity checks to TestInfantry.


Sleet01 (approved):
LGTM

PR #7245, opened 2025-06-15
Mek Validation

SJuliez:
Adds two validation checks for meks
Fixes MegaMek/MegaMekLab#1939
Fixes MegaMek/MegaMekLab#1922
I'm taking the stance that the MML GUI doesnt have to enforce rules if validation catches them; validation gives feedback which is more helpful than blocking construction


HammerGS (approved):
Tested and seems to work.

PR #7246, opened 2025-06-15
Faction data as yml

SJuliez:
See the description and discussion in #6735, this PR replaces the old one and does not include the data; comes in separate PR.
I wanted to have the data separate so it's easier to review


HammerGS:
Cut and Paste of the notes from #7246 

> This PR together with MML and MHQ PRs aims to unify the faction data of MHQ and MM on the MM side to make it available everywhere and use individual yaml files instead of a single huge xml file.
> 
> For now, the old classes are preserved and still used; they only get their data from the new Faction2 class (the best name I could come up with) instead of loading it from factions.xml (=2 separate and different files in MHQ and MM)
> FactionRecord also loads complicated era-based RAT data from other files, this is unchanged; it doesnt seem useful to place this inside the yamls as this would make the yamls unwieldy
> Factions2 reads its data from the data/universe/factions and data/universe/commands folders. These need to be replicated in MML and MHQ so long as we dont have a unified data repo.
> "commands" are the subfactions (those with a dot in the name), e.g. the Hesperus Guards; in other words, combat groups rather than political entities; sorting their yaml data in their own folder is purely for cleanliness, it doesnt matter where which yaml file is. In the future, it's possible that additional commands data will make separate classes (in addition to separate yamls) better, but for now it's still good
> data is also read from the user directory, if available, so [userdir]/data/universe/factions or /commands; factions there will be added; if they have the same key as a canon faction, they replace the original
> "startingplanet" has been renamed "capital" as it was used for that purpose only
> This is pretty hefty for review, but when filtering out the .yml files, it's only a handful of classes. Ideally, the whole thing should have no visible effect in MM, MML or MHQ except fix a bug with MML faction selection which omitted a bunch of factions.

SJuliez:
The failing CI is the Faction2Test which fails without its data. Other tests seem to work. I'll add the two factions back in so it can run

rjhancock (approved):

PR #7249, opened 2025-06-16
minor graphical fixes for Faction Availability

exeea:
properly calculated the row height for the faction column.
removed the "Era: " from the era columns.

image


HammerGS:
I think that's much better. I'll hold merging this as Luana had a comment around antialising.

PR #7250, opened 2025-06-16
Fix Bind errors in Damage Manager tests: make server port selection more resilient with retry on BindException

Sleet01:
This should prevent a reported BindException in these tests that was caused by using randomized server ports without first confirming that the ports were available.

Notes:
1. The test setup may still fail on IOExceptions (which could happen if the local host could not be determined or Java did not have permission to open the ports) but these have not yet been reported.
2. Ports now start at 1024, so should avoid the standard server ports entirely.
3. In the best case, the entire suite of tests should complete in under a second; in the worst case, with all ports in use, each test may take several seconds to fail.

Testing:
- Ran a stress test focusing on repeated runs of this test suite.

### DIFF
--- a/megamek/unittests/megamek/server/totalwarfare/TWDamageManagerTest.java
+++ b/megamek/unittests/megamek/server/totalwarfare/TWDamageManagerTest.java
@@ -52,6 +52,7 @@ import org.junit.jupiter.params.provider.ValueSource;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.net.BindException;
 import java.util.Random;
 
 import static org.junit.jupiter.api.Assertions.*;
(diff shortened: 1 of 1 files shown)


megamek/unittests/megamek/server/totalwarfare/TWDamageManagerTest.java
```diff
@@ -85,8 +86,15 @@ void setUp() throws IOException {
...
-              gameMan, false, "", null, true);
+        for (int port=MMConstants.MIN_PORT_FOR_QUICK_GAME; port<=MMConstants.MAX_PORT; port++) {
+            try {
+                server = new Server(null, random.nextInt(MMConstants.MIN_PORT_FOR_QUICK_GAME, MMConstants.MAX_PORT),
+                      gameMan, false, "", null, true);
+                break;
+            } catch (BindException e) {
+                // Try the next one
+            }
+        }
```
rjhancock, line 97: This can be replaced by just picking a random port out of say 50k then adding the `DEFAULT_PORT` to the int. No need to loop.

This would keep the ports between 2346 and 52346.

Or since we will probably have a need for testing server communication at some point, create a helper method under the testing utilities that is static and have it generate a random port within the range of 2346 - MAX_PORT.

Sleet01: The issue is that we have no guarantee, as far as I can tell, that any given random port will be open.  If multiple test suites that use servers are running in parallel, any port within the range can be bound already.
Meanwhile, in the best case, this loop only runs once, and in the degenerate case where every port in the range is already bound... well, we have worse problems.

I'll take a look at adding a utility method but no guarantees.

rjhancock: My thinking would be catch the exception then retry another random port. Looping works as well with same basic catch and retry. Either way, extracting to a method for other tests would not be a bad idea.

rjhancock (approved):

rjhancock:
updated to master and re-running tests. Java 17 failed due to a random testing error still.

PR #7264, opened 2025-06-20
Fix FactionRecord lookup (MM)

SJuliez:
This fixes a problem with the new faction data where the faction selection in the various random army generators in the lobby showed different factions from what they used to.


IllianiBird (approved):
LGTM

PR #7265, opened 2025-06-21
Improvement: Added Aggregate Faction Tag

IllianiBird:
This PR adds a new faction tag `AGGREGATE` that is used to flag that a faction is not a singular entity but an aggregate of different independent factions.

The example I used in the code was the pirate faction. This represents not a singular monolithic pirate organization, but an _aggregate_ of numerous pirate bands.

This was introduced to simplify MekHQ's faction standing so that we can easily exclude aggregate factions from Faction Standing tracking.

### DIFF
--- a/megamek/data/universe/factions/IS.yml
+++ b/megamek/data/universe/factions/IS.yml
@@ -1,5 +1,8 @@
 key: IS
 name: Inner Sphere - General
+tags:
+  - IS
+  - AGGREGATE
 ratingLevels:
   - F
   - D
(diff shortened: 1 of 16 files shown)


megamek/data/universe/factions/IS.yml
```diff
@@ -1,5 +1,8 @@
 key: IS
 name: Inner Sphere - General
+tags:
```
IllianiBird, line 3: @SJuliez can you verify that these are the right tags for this faction? It previously didn't have any tags.

SJuliez: Looks good to me

SJuliez (approved):

PR #7267, opened 2025-06-21
Replace existing randomized server approach with systematic port finder

Sleet01:
Replaces randomized server port selection (which can fail in a limited number of circumstances) with a more systematic approach that finds the next open port #.

Also adds fixes to a related test that has a low chance to fail on triple crit rolls.

Testing:
- Ran the failing tests ~ 50,000 times to show update resolved the random failure issues


Sleet01:
Someday I will remember to add a license header to every new file, but this is not that day -_-;

HammerGS (approved):

PR #7272, opened 2025-06-23
correct BLORD faction data

SJuliez:
Somehow the key and filename for BLORD got messed up. Strange, when I rewrote the factions by program, not by hand


Sleet01 (approved):
LGTM

PR #7279, opened 2025-06-26
Improvement: Added Faction Leadership Tracking to Faction Overhaul

IllianiBird:
This PR allows us to store information about a faction's leaders in their YAML entry. It doesn't do anything on its own, but I opted to include the data separately as it's liable to be fairly hefty. Below is an excerpt from a completed YAML entry:

image


Sleet01 (approved):
LGTM

PR #7283, opened 2025-06-28
WAA refactor after move

SJuliez:
This gives the classes that were recently created by splitting WeaponAttackAction a bit of a code renovation, removing unused parameters, loads of reformatting and making code more readable.
~As the only intentional change to functionality this removes a deadzone check from strafing fire; this was already marked as questionable and from reading the strafing rules recently I am fairly sure deadzone rules dont apply.~


gsparks3:
The "standard" dead zone rules don't apply to strafing, but TW page 243 is quite clear on a secondary variant: if the attacker is at Altitude 1, and Hex A (an arbitrary hex along the flight/attack path) is two or more levels higher than Hex B (the next hex along the flight path), then Hex B cannot be attacked via strafe. 

At a quick glance, lines 881-891 in ComputeToHitIsImpossible.java appear to have been implementing this correctly, and should be replaced with the FIXME removed. The preceding section on how to handle cases where there is no such adjacent hex is more arguable, but in the absence of an official ruling can likely be left unchanged for now.

SJuliez:
Quite correct. I reverted that change. Thanks!

Sleet01 (approved):
LGTM.  Nice work!

PR #7297, opened 2025-07-07
Force hull-mounted equipment to be forward-facing

pavelbraginskiy:
Under certain circumstances, the loader might try to place equipment in the "rear hull" location, which doesn't exist.
This should never happen except due to bugs in MML or hand-built units.
Since it _does_ happen, see MegaMek/megameklab#1953, and the correct location (forward hull) is obvious, we can just ignore that the equipment specified itself to be rear-mounted and move on with loading the unit.


Sleet01 (approved):
LGTM

PR #7316, opened 2025-07-15
Fix #7315: Fixed Error When Removing Column in Availability Panel if Total Columns Equaled 1

### LINKED ISSUE #7315: [Issue] Out of Index Error in Availability Display For Support Vehicles Loaded into MegaMek Via MekHQ
### Brief Description *

The new availability display seems to be running into issues with Support Vehicles loaded from MekHQ.

Load the attached save (no customs) and launch the only available scenario. This will result in an out of index error due to the availability display being empty for the included support vehicle (Debbie). This appears to be the case for all support vehicles. As loading the unit into MegaMek hits an error, the unit is then not loaded into the scenario and counts as being destroyed for the purposes of MekHQ integration

### 3. Steps to Reproduce

1. Load the save into MekHQ
2. Launch the only available scenario
3. Hit the error
4. Cry

### Attach Files

[Autosave-3-The 23rd Inception Strategies-31510120.cpnx.gz](https://github.com/user-attachments/files/21220833/Autosave-3-The.23rd.Inception.Strategies-31510120.cpnx.gz)

### Severity *

### MegaMek Suite Version *

50.07

### Operating System *

Mac

### Java Version *

J17

### Final Verification

IllianiBird:
Fix #7315

This is just a bandaid fix until Drake is able to come in and address the root cause. However, it's a decent fix and if it's never removed it'll continue to work fine with no developer interaction.

### DIFF
--- a/megamek/src/megamek/client/ui/dialogs/unitSelectorDialogs/AvailabilityPanel.java
+++ b/megamek/src/megamek/client/ui/dialogs/unitSelectorDialogs/AvailabilityPanel.java
@@ -49,7 +49,6 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import javax.swing.*;
 import javax.swing.border.EmptyBorder;
-import javax.swing.event.HyperlinkEvent;
 import javax.swing.event.TableModelEvent;
 import javax.swing.event.TableModelListener;
 import javax.swing.table.DefaultTableCellRenderer;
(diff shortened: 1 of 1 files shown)


Sleet01 (approved):
I'm not entirely sure what the code in question is trying to accomplish, nor why it would only hit an exception when column count == 1 (but not, say, if column count == 2 and i goes down to 0) but if this works for now, that seems fine.

IllianiBird:
Yeah, I couldn't fathom it myself, either. I know the fix works. And I know without the fix anything that is missing availability information (support vehicles predominantly it seems) will break.

A proper fix is absolutely going to require Drake, because I suspect they're the only one that knows what this code is meant to be doing.

PR #7331, opened 2025-07-27
Improvement: Added Milestone Version Tracking

IllianiBird:
This PR adds Milestone version tracking to assist https://github.com/MegaMek/mekhq/pull/7425

### DIFF
--- a/megamek/resources/Version.properties
+++ b/megamek/resources/Version.properties
@@ -1,3 +1,5 @@
+# Was the last release confirmed as a Milestone?
+# Please update SuiteConstants#LAST_MILESTONE and SuiteConstants#ALL_MILESTONE_RELEASES
 major=0
 minor=50
 patch=7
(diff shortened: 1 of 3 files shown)


Sleet01:
As someone who does some release build work, I would highly recommend that we store any release-specific info (milestone numbers, acceptable upgrade targets, supported releases, etc) in YAML or XML files so that updating and/or correcting the data can be accomplished without access to an IDE or GUI.  Perhaps add Tex as a required reviewer as well.

IllianiBird:
I think this idea has a lot of value

megamek/resources/Version.properties
```diff
@@ -1,3 +1,5 @@
+# Was the last release confirmed as a Milestone?
+# Please update SuiteConstants#LAST_MILESTONE and SuiteConstants#ALL_MILESTONE_RELEASES
```
Sleet01: Should probably mention `milestoneReleases.yml` now.

IllianiBird: Fudge

PR #7332, opened 2025-07-28
getter for the icon filename

exeea:
added a getter to retrieve the icon filename of an unit.
(required for other features)


IllianiBird (approved):
This was a really difficult review, but I think it should be safe to include ;)

PR #7340, opened 2025-07-30
Fix: Fixed Clans Incorrectly Being Considered Honor:None by Default

IllianiBird:
This is the MegaMek portion of https://github.com/MegaMek/mekhq/pull/7439

### DIFF
--- a/megamek/src/megamek/common/universe/Faction2.java
+++ b/megamek/src/megamek/common/universe/Faction2.java
@@ -159,12 +159,78 @@ public class Faction2 {
         return nameGenerator;
     }
 
-    public HonorRating getPreInvasionHonorRating() {
+    /**
+     * Returns the pre-invasion honor rating assigned to this faction, if any.
+     *
+     * <p>This method provides direct access to the stored {@link HonorRating} value representing the faction's honor
+     * rating before the invasion period. No default or fallback value is applied.</p>
+     *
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/universe/Faction2.java
```diff
@@ -411,12 +411,12 @@ public void saveToFile(FactionRecord updatedRecord) throws IOException {
 
     @JsonGetter("preInvasionHonorRating")
     private HonorRating preInvasionHonorRatingSerializer() {
-        return preInvasionHonorRating != HonorRating.NONE ? preInvasionHonorRating : null;
+        return preInvasionHonorRating;
```
IllianiBird, line 480: In the event the faction has no honor, we're going to return `HonorRating.NONE` (handled elsewhere) rather than `null`.

SJuliez: I suggest making any changes so that whatever Faction2 returns is the correct value. It seems that MHQ's Faction does post-processing on this value and I think it should be done here instead. Ultimately I'd like to remove MHQ's Faction class entirely. This was just out of scope of the yaml conversion.

IllianiBird: I think that makes sense, lemme see what I can do

PR #7343, opened 2025-07-31
Improvement: MHQ->MM Data Update

IllianiBird:
This PR just copies over data changes from mhq into mm so that it can remain the gold standard without risk of overwriting mhq changes.

This largely affects some recent faction data updates and faction images


Sleet01 (approved):
LGTM

PR #7378, opened 2025-08-19
Class rename codeql copyrights

rjhancock:
megamek.common changes.


SJuliez (commented):
I prefer not adding .0 to all bv values that happen to be integer as it only adds clutter for no value that I can see.

rjhancock:
> I prefer not adding .0 to all bv values that happen to be integer as it only adds clutter for no value that I can see.

The original defined type for BV was Object and there is math going on. BV during calculations can be Double so this is the easiest fix to deal with the other issues that crop up.

PR #7379, opened 2025-08-21
Prevent log message from empty lines in rat data

SJuliez:
Lots of log entries from empty lines after the license note in rat data now.


rjhancock:
Added to the larger PR. Thank you.

PR #7381, opened 2025-08-22
Log entry correction

SJuliez:
Prevents a log entry from BLKSmallCraftFile.java on rear hull equipment when the equipment is not actually saved as rear mounted.

@rjhancock Please incorporate as you see fit.


rjhancock:
Fixed in the other PR. Thank you.

PR #7407, opened 2025-09-13
Playtest errata

pavelbraginskiy:
- Quad kick location table is back to normal
- Case II now vents all armor on cap, even for very small explosions
- Armor damage from ammo explosion contributes to PSR


HoneySkull (approved):
LGTM

PR #7420, opened 2025-09-20
Implement packet type check exceptions

Sleet01:
Per our conversation on Discord, this implements throwing InvalidPacketDataException when packet data types do not match their respective expected type[s].
This should give us explicit feedback when unexpected data formats are passed through packets; previously this would cause packet ingest to fail silently, leading to many parts of the UI or network behaviour to fail silently and inexplicably.

Part of this work is to add `try...catch` clauses around most of the packet handling `switch/case` constructs, the majority of which had no exception handling whatsoever.
We have the opportunity in 0.50.07 to greatly increase our error handling and logging in these areas by expanding on this work, but I'm unsure of the possible performance impacts.

Testing:
- Ran all MM unit tests (MHQ was still broken)
- Ran multi-hour bot vs. bot fight and confirmed no new errors or exceptions

### DIFF
--- a/megamek/src/megamek/client/AbstractClient.java
+++ b/megamek/src/megamek/client/AbstractClient.java
@@ -63,11 +63,13 @@ import megamek.common.net.events.DisconnectedEvent;
 import megamek.common.net.events.PacketReceivedEvent;
 import megamek.common.net.factories.ConnectionFactory;
 import megamek.common.net.listeners.ConnectionListener;
+import megamek.common.net.packets.InvalidPacketDataException;
 import megamek.common.net.packets.Packet;
 import megamek.common.preference.PreferenceManager;
 import megamek.common.units.Entity;
 import megamek.common.units.UnitNameTracker;
 import megamek.logging.MMLogger;
+import megamek.server.ConnectionHandler;
 
 /**
  * AbstractClient that handles basic client features.
@@ -266,7 +268,7 @@ public abstract class AbstractClient implements IClient {
      *
      * @param packet The packet we received.
      */
-    protected void receivePlayerInfo(Packet packet) {
+    protected void receivePlayerInfo(Packet packet) throws InvalidPacketDataException {
         int packetIndex = packet.getIntValue(0);
 
         Player newPlayer = (Player) packet.getObject(1);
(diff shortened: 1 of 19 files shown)


megamek/src/megamek/client/AbstractClient.java
```diff
@@ -359,7 +361,7 @@ protected synchronized void checkDuplicateNamesDuringAdd(Entity entity) {
         }
     }
 
-    protected void receiveUnitReplace(Packet packet) {
+    protected void receiveUnitReplace(Packet packet) throws InvalidPacketDataException {
         @SuppressWarnings(value = "unchecked") List<Force> forces = (List<Force>) packet.getObject(1);
         forces.forEach(force -> getGame().getForces().replace(force.getId(), force));
```
rjhancock: There should be a getForceList method on packet to remove the unchecked suppression.

megamek/src/megamek/common/net/packets/InvalidPacketDataException.java
```diff
@@ -0,0 +1,70 @@
+/*
+ * Copyright (C) 2024-2025 The MegaMek Team. All Rights Reserved.
+ *
+ * This file is part of MegaMek.
```
rjhancock, line 4: Unless the exception already existed before, please fix the year.

rjhancock (changes requested):
The copyright header is the only requested change to ensure it's the correct year (if the exception is new, fix it. If not, adjust the start to when it was introduced.

rjhancock:
Only other thoughts on my mind would be to move all of the `getObject` calls to non generic typed checks (nothing to a Collections for example) and make the method private.

Remove all of the `@SupressWarning("unchecked")` calls as possible in relation to the Packet class.

Not in this PR, just as a path forward. And I'm not worried about the performance issues with this as the security side of things for clients is a much higher priority in this case.

megamek/src/megamek/client/AbstractClient.java
Sleet01: Fixed; same for getInGameObjectList() in the next line after this.

Sleet01:
> Only other thoughts on my mind would be to move all of the `getObject` calls to non generic typed checks (nothing to a Collections for example) and make the method private.
> 
I  can commit to this, but it will have to be a separate PR: there are too many external calls to `getObject()` (probably bad practice) that will require adding type-specific getters to Packet first.
Removing the Collections now should be possible, though.

> Remove all of the `@SupressWarning("unchecked")` calls as possible in relation to the Packet class.
> 
> Not in this PR, just as a path forward. 
I can do that in the same subsequent PR.

> And I'm not worried about the performance issues with this as the security side of things for clients is a much higher priority in this case.

Agreed.

PR #7422, opened 2025-09-21
Prevent multiple clan CASE per location

pavelbraginskiy:
Currently a new Clan CASE gets added every time MML makes a change to a unit's equipment, so if you add 5 ammo bins to a location you get 5 Clan CASEs.

If a location already has CASE, it doesn't need more CASE.


Sleet01:
Do we need this for Combat Vehicles as well?

exeea:
> Do we need this for Combat Vehicles as well?

I don't think so. Only Mek and Aero are overriding the addClanCase() method.

exeea (approved):
LGTM

PR #7435, opened 2025-09-25
Fix: Fixed Failing Tests Following Data Overhaul (MegaMek Edition)

IllianiBird:
Following the data overhaul we found ourselves with a number of failing tests. These were tests that relied on data being present when the GitHub Test Environment ran.

With the segregation of Data out of the individual program directories we now run in a dataless state.

This PR updates a handful of classes so that they can accept both test and non-test directories. This allows us to include test data for testing purposes.

This PR is required for https://github.com/MegaMek/mekhq/pull/7588

### DIFF
--- a/megamek/src/megamek/client/generator/RandomCallsignGenerator.java
+++ b/megamek/src/megamek/client/generator/RandomCallsignGenerator.java
@@ -49,23 +49,53 @@ import megamek.logging.MMLogger;
  * that is used during generation
  */
 public class RandomCallsignGenerator {
-    private final static RandomCallsignGenerator INSTANCE = new RandomCallsignGenerator();
     private final static MMLogger logger = MMLogger.create(RandomCallsignGenerator.class);
+
+    private final static String TEST_DIR = "testresources/data/names/callsigns_test.csv";
+
+    private final static RandomCallsignGenerator INSTANCE = new RandomCallsignGenerator(false);
+    private final static RandomCallsignGenerator TEST_INSTANCE = new RandomCallsignGenerator(true);
+
 
     private final WeightedIntMap<String> weightedCallsigns = new WeightedIntMap<>();
 
-    RandomCallsignGenerator() {
+    RandomCallsignGenerator(boolean useTestDirectory) {
         final Map<String, Integer> callsigns = new HashMap<>();
-        loadCallsignsFromFile(new File(MMConstants.CALLSIGN_FILE_PATH), callsigns);
-        loadCallsignsFromFile(new File(MMConstants.USER_CALLSIGN_FILE_PATH), callsigns);
+
+        if (useTestDirectory) {
+            loadCallsignsFromFile(new File(TEST_DIR), callsigns);
+        } else {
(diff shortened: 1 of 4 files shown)


megamek/src/megamek/client/generator/RandomNameGenerator.java
```diff
@@ -253,24 +253,26 @@
...
+                      : KEY_DEFAULT_FACTION);
+                ethnicCode = factionEthnicCodes.get(faction).randomItem();
+                final int givenNameEthnicCode = factionGivenNames.get(faction).get(ethnicCode).randomItem();
+
+                name[0] = (gender.isFemale() ? femaleGivenNames : maleGivenNames).get(givenNameEthnicCode).randomItem();
+                name[1] = clanPilot ? "" : surnames.get(ethnicCode).randomItem();
+            } catch (Exception e) {
+                // This should only fail when we have no name data passed in. For example, if we're running in a
+                // dataless environment such as the GitHub test environment.
+                LOGGER.error(e, "Error generating name for faction: {} and ethnic code: {}", faction, ethnicCode);
```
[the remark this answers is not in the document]
IllianiBird: It calls to two arguments and supplies 2 arguments -_-

megamek/src/megamek/common/units/Crew.java
```diff
@@ -324,7 +324,8 @@ public String getNickname() {
     }
 
     public String getNickname(final int pos) {
-        return getNicknames()[pos];
+        String nickname = getNicknames()[pos];
```
IllianiBird, line 327: This would return `null` if there was nobody in `pos` or if that character didn't have a callsign. Potentially leading to an NPE. This just adds null prevention so we don't have to add null safety checks every time we call `getNickname`

megamek/src/megamek/client/generator/RandomCallsignGenerator.java
```diff
@@ -49,23 +49,53 @@
...
 public class RandomCallsignGenerator {
-    private final static RandomCallsignGenerator INSTANCE = new RandomCallsignGenerator();
     private final static MMLogger logger = MMLogger.create(RandomCallsignGenerator.class);
 
+    private final static String TEST_DIR = "testresources/data/names/callsigns_test.csv";
+
+    private final static RandomCallsignGenerator INSTANCE = new RandomCallsignGenerator(false);
+    private final static RandomCallsignGenerator TEST_INSTANCE = new RandomCallsignGenerator(true);
+
+
```
Sleet01, line 59: This is fine, but another approach we can use is extending the main class with a Test<ClassName> class that lives in the unittest code dir.
As long as the namespace (e.g. `client/generator/`) matches, JUnit should find the derived class automatically.
Of course, that approach requires changing the protection level of some internal fields, and for singletons (like this, or Systems) allowing the user to set the instance directly.

Both approaches have benefits and drawbacks.

IllianiBird: That's very interesting and not something I'd have considered independently. Definitely an arrow in my quiver for next time.

PR #7440, opened 2025-09-27
Feature: Implemented 'G-Tolerance' A Time of War SPA

IllianiBird:
This implements the G-Tolerance Trait (SPA in our parlance) that reduces the TN for High-G Maneuvers by 1.


Sleet01 (approved):
This looks good to go; it's labeled "for new dev cycle" internally so I'd missed it previously.

PR #7442, opened 2025-09-28
Fix #7441: Restrict player "list" to Vector

### LINKED ISSUE #7441: [Issue] Unable to change bot teams
### Brief Description *

Unable to use the drop-down to change bot teams under the players section.

### 3. Steps to Reproduce

Launch megamek
Start a new game
Click add bot a few times to add bots to the players list
Click on the bot in the list to select it
Notice the drop-down changes to reflect their team
Use the drop-down to change the team number
Nothing happens they stay the same

### Attach Files

_No response_

### Severity *

### MegaMek Suite Version *

f912a6e493e

### Operating System *

Linux popos

### Java Version *

Openjdk 21.0.8

### Final Verification

Sleet01:
More fallout from implementing strict typing in Packet handling: some team change packets were being created with HashSets (for whatever reason) but we had been expecting Vectors based on other packet creation code.

Testing:
- Ran reporter's repro campaign and confirmed functionality is working correctly.

Close #7441

### DIFF
--- a/megamek/src/megamek/client/Client.java
+++ b/megamek/src/megamek/client/Client.java
@@ -1271,7 +1271,7 @@ public class Client extends AbstractClient {
     /**
      * Sends a packet containing multiple entity updates. Should only be used in the lobby phase.
      */
-    public void sendChangeTeam(Collection<Player> players, int newTeamId) {
+    public void sendChangeTeam(Vector<Player> players, int newTeamId) {
         send(new Packet(PacketCommand.PLAYER_TEAM_CHANGE, players, newTeamId));
     }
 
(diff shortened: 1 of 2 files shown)


SJuliez:
I always go by the assumption that Vector is a legacy collection type and newer ones are preferred. But in this case, of course, whatever makes it run :)

Sleet01:
> I always go by the assumption that Vector is a legacy collection type and newer ones are preferred. But in this case, of course, whatever makes it run :)

It should be easy to adjust if we decide a HashSet or ArrayList makes more sense, but it'll be easier to avoid mismatches if we have unit tests that confirm all the entry points use the correct types.
I think it'll be simple to set up if I can hijack the `send` functionality for unit tests...

Issue #7457, opened 2025-10-04
[Issue][Nightly] Erroneous Leg destructions


Hwatwasthat:
### Brief Description *

Sometimes I'm getting mechs reporting they are having legs destroyed and automatically falling, while having taken no leg damage and in the next round the legs are still there. I think it's related to torso damage.

<img width="756" height="373" alt="Image" src="https://github.com/user-attachments/assets/2ee79942-3d47-4d88-9c86-0600aeeb9886" />
<img width="906" height="706" alt="Image" src="https://github.com/user-attachments/assets/b4a87ec1-3cfb-4b15-8608-938eb9d4856c" />

### 3. Steps to Reproduce

1. Play game
2. Hit anywhere but the legs and notice it says leg destroyed, automatic fall.

### Attach Files

[logs.zip](https://github.com/user-attachments/files/22704563/logs.zip)

### Severity *

Medium (Gameplay Limitation): Non-core functionality is impaired, providing a suboptimal but playable experience.

### MegaMek Suite Version *

v0.50.07-nightly-2025-10-03

### Operating System *

Linux (Ubuntu)

### Java Version *

21.0.8

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have included all necessary information and files to help reproduce this issue
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

IllianiBird:
Do you have the play test options enabled?

Hwatwasthat:
I do yes, sorry for not mentioning that.

Hwatwasthat:
I have another example of this occurring.
[logs2.zip](https://github.com/user-attachments/files/22711466/logs2.zip)

<img width="505" height="563" alt="Image" src="https://github.com/user-attachments/assets/2331075b-bd26-4e03-85d0-4db55c19a699" />
<img width="744" height="428" alt="Image" src="https://github.com/user-attachments/assets/2653005c-4b33-4f42-b5da-1d7acd9ed7b4" />
<img width="819" height="891" alt="Image" src="https://github.com/user-attachments/assets/5df83818-09b2-4112-9684-965114604a72" />

TenkawaBC:
PR#7462 for this fix.

IllianiBird:
Closing as resolved

PR #7462, opened 2025-10-05
Fix bug #7457

TenkawaBC:
I missed the locationIsLeg and canFall statements in the If in the playtest. This should fix this problem.

### DIFF
--- a/megamek/src/megamek/common/units/Mek.java
+++ b/megamek/src/megamek/common/units/Mek.java
@@ -5305,7 +5305,7 @@ public abstract class Mek extends Entity {
         if (game != null && locationIsLeg(loc) && canFall() && !game.getOptions().booleanOption(OptionsConstants.PLAYTEST_2)) {
             game.addPSR(new PilotingRollData(getId(),
                   TargetRoll.AUTOMATIC_FAIL, 5, "leg destroyed"));
-        } else if (game !=null && game.getOptions().booleanOption(OptionsConstants.PLAYTEST_2)) {
+        } else if (game !=null && locationIsLeg(loc) && canFall() && game.getOptions().booleanOption(OptionsConstants.PLAYTEST_2)) {
             game.addPSR(new PilotingRollData(getId(), TargetRoll.AUTOMATIC_FAIL, 4, "leg destroyed"));
         }
         


IllianiBird:
When titling PRs can you put more descriptive titles. The title is directly imported into our change logs on version launch so we try to make them layman legible

megamek/src/megamek/common/units/Mek.java
```diff
@@ -5305,7 +5305,7 @@ public void destroyLocation(int loc, boolean blownOff) {
         if (game != null && locationIsLeg(loc) && canFall() && !game.getOptions().booleanOption(OptionsConstants.PLAYTEST_2)) {
```
IllianiBird: `game != null && locationIsLeg(loc) && canFall()` is checked in every branch of this conditional. We either want to wrap the inner conditional in a single check that evaluates or include a gatekeeper conditional first that early exits or otherwise bypasses the rest of the conditional.

megamek/src/megamek/common/units/Mek.java
```diff
@@ -5305,7 +5305,7 @@ public void destroyLocation(int loc, boolean blownOff) {
...
             game.addPSR(new PilotingRollData(getId(),
                   TargetRoll.AUTOMATIC_FAIL, 5, "leg destroyed"));
-        } else if (game !=null && game.getOptions().booleanOption(OptionsConstants.PLAYTEST_2)) {
+        } else if (game !=null && locationIsLeg(loc) && canFall() && game.getOptions().booleanOption(OptionsConstants.PLAYTEST_2)) {
```
IllianiBird: If we using the same data more than once it should be cached into a variable and that variable re-used instead.

IllianiBird (changes requested):
This conditional should be refactored. Something like:

```
if (game != null && locationIsLeg(loc) && canFall()) {
    if (game.getOptions().booleanOption(OptionsConstants.PLAYTEST_2)) {
        true action
    else {
        false action
    }
}
```

megamek/src/megamek/common/units/Mek.java
TenkawaBC: Refactored and commit added.

megamek/src/megamek/common/units/Mek.java
TenkawaBC: Now only a single check.

IllianiBird (approved):
Much cleaner. Good job jumping on this so fast.

PR #7466, opened 2025-10-06
Fix 7449: Replace entity ID with actual entity to avoid NPEs

### LINKED ISSUE #7449: [Issue] Java exception upon melee attacks when searchlight is also an option
### Brief Description *

When a mek that has an option to use searchlight uses melee attack in the physical attack phase, an error occurs. The error can be clicked through and the physical attack then can be used without an error if pressed again.

Image

### 3. Steps to Reproduce

1. Have a MM instance with a scenario taking place under dark light conditions,
2. Have meks with a searchlight being an option attack in melee either kick or punch

### Attach Files

_No response_

### Severity *

### MegaMek Suite Version *

MekHQ-0.50.07-nightly-2025-10-01

### Operating System *

Windows 11

### Java Version *

Adoptium 17

### Final Verification

Sleet01:
It looks like we were mistakenly passing the activated entity ID rather than the entity itself, but the function to return a BoardView using an `int` is expecting a Board ID, not an entity ID.
This would only fire for:
1. _Not_ unit ID 0,
2. _Not_ already using the searchlight,
3. _With_ `AutoDeclareSearchlight` enabled,
4. With light level set such that a searchlight would be beneficial,
so it was sort of a pain to repro - tip of the hat to the reporter for getting a good repro save.

Close #7449 

Testing:
- Confirmed fix with OP's savegame.

### DIFF
--- a/megamek/src/megamek/client/ui/panels/phaseDisplay/PhysicalDisplay.java
+++ b/megamek/src/megamek/client/ui/panels/phaseDisplay/PhysicalDisplay.java
@@ -816,7 +816,7 @@ public class PhysicalDisplay extends AttackPhaseDisplay {
 
         // and add it into the game, temporarily
         game.addAction(saa);
-        clientgui.getBoardView(currentEntity).addAttack(saa);
+        clientgui.getBoardView(ce()).addAttack(saa);
 
         // and prevent duplicates
         setSearchlightEnabled(false);
--- a/megamek/src/megamek/client/ui/panels/phaseDisplay/PhysicalDisplay.java
+++ b/megamek/src/megamek/client/ui/panels/phaseDisplay/PhysicalDisplay.java
@@ -257,7 +257,7 @@ public class PhysicalDisplay extends AttackPhaseDisplay {
     private void cacheVisibleTargets() {
         clearVisibleTargets();
 
-        List<Entity> vec = clientgui.getClient().getGame().getValidTargets(ce());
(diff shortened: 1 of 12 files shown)


megamek/src/megamek/client/ui/panels/phaseDisplay/PhysicalDisplay.java
```diff
@@ -816,7 +816,7 @@ private void doSearchlight() {
 
         // and add it into the game, temporarily
         game.addAction(saa);
-        clientgui.getBoardView(currentEntity).addAttack(saa);
+        clientgui.getBoardView(ce()).addAttack(saa);
```
IllianiBird: If it's not too much trouble can we rename this method. `ce` is a super ambiguous method name and even with good documentation is going to cause confusion. If nothing else it makes the code harder to parse 'at a glance'.

Sleet01: I have no confidence in my ability to cleanly replace `ce()` with `currentEntity()` without breaking stuff, but I'll give it a shot.

Sleet01:
@IllianiBird I've expanded `ce()` to `currentEntity()` using the IDE's refactoring tools, and it seems to be working fine:
- all tests passing
- no errors or exceptions in debug logs
- game runs fine

Unless you have any objections, I'll go ahead and merge.

IllianiBird:
I was about to say 'go for it' but then you already went for it :D

PR #7468, opened 2025-10-07
Fix: Added Missing 's' to /logs in Build Directory Construction

IllianiBird:
Fix https://github.com/MegaMek/mekhq/issues/7637

We were adding a log directory when building the project, when we should have been added logs. This resulted in players ending up with both a /log and logs directory.

Insofar as I can tell this bug is benign, but if we don't fix it I can guarantee we will see people asking about it throughout 07's lifecycle. :D

### DIFF
--- a/megamek/build.gradle
+++ b/megamek/build.gradle
@@ -248,7 +248,7 @@ tasks.register('stageFiles') {
     dependsOn deleteAtlasedImages
 
     doLast {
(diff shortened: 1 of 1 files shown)


IllianiBird:
Confirmed working via test build. Flipping to live.

Sleet01 (approved):
LGTM

PR #7471, opened 2025-10-08
Fix 7409: starting location only updating once via apply button

### LINKED ISSUE #7409: [Issue] 50.06 Deployment Area Apply button only works 1 time
### Brief Description *

Image

The Apply button here doesn't apply the selected NW/SE corner values if it has already been clicked before.

Correct behavior: the Apply button should work no matter how many times it has been clicked before

(You can work around this by reopening the player settings if you want to update these values again.)

### 3. Steps to Reproduce

- start new game
- right click player, bring up Player Settings
- set Deployment Any NW/SE Corner to some values
- click Apply.  The numbers will be properly applied to the player.
- update any of the corner values again
- click Apply.  The player's deployment will not be updated.  (clicking ok also doesn't update the area, but i don't think it's supposed to)

### Attach Files

_No response_

### Severity *

### MegaMek Suite Version *

0.50.06

### Operating System *

Windows 10 10.0 (amd64)

### Java Version *

Eclipse Adoptium 17.0.12

### Final Verification

Sleet01:
Avoids an issue with the dialog's Player info and the local client's version of the same info desynching while within the 
player settings dialog.
The initial change to the player's deployment area would be sent out, but further changes would not be applied... because the local client's game would not update until the dialog closed.
Fix is to send out the dialog's version of the Player instance, and let the AbstractClient and / or Server worry about keeping the local client's game in sync.

Testing:
- Confirmed fix now applies player starting zone updates every time Apply is pressed.

Close #7409

### DIFF
--- a/megamek/src/megamek/client/AbstractClient.java
+++ b/megamek/src/megamek/client/AbstractClient.java
@@ -227,7 +227,16 @@ public abstract class AbstractClient implements IClient {
      * Sends the info associated with the local player.
      */
     public void sendPlayerInfo() {
-        send(new Packet(PacketCommand.PLAYER_UPDATE, getGame().getPlayer(localPlayerNumber)));
+        sendPlayerInfo(getGame().getPlayer(localPlayerNumber));
+    }
+
+    /**
+     * Sends the info associated with a player, usually the local game's local player but sometimes the modified version
+     * of that player.
+     * @param player
+     */
+    public void sendPlayerInfo(Player player) {
+        send(new Packet(PacketCommand.PLAYER_UPDATE, player));
     }
 
     @Override
--- a/megamek/src/megamek/client/ui/panels/phaseDisplay/lobby/PlayerSettingsDialog.java
+++ b/megamek/src/megamek/client/ui/panels/phaseDisplay/lobby/PlayerSettingsDialog.java
@@ -637,7 +637,7 @@ public class PlayerSettingsDialog extends AbstractButtonDialog {
                 // Use sendUpdate because we want the Game to allow us to change on Bot's
                 // behalf.
                 clientgui.chatlounge.sendProxyUpdates(updateEntities, client.getLocalPlayer());
-                // clientGUI.chatlounge.sendUpdate(updateEntities);
+                clientgui.chatlounge.sendUpdate(updateEntities);
             }
         }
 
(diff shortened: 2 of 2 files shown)


megamek/src/megamek/client/AbstractClient.java
```diff
@@ -227,7 +227,16 @@ public void sendNextPlayer() {
...
      */
     public void sendPlayerInfo() {
-        send(new Packet(PacketCommand.PLAYER_UPDATE, getGame().getPlayer(localPlayerNumber)));
+        sendPlayerInfo(getGame().getPlayer(localPlayerNumber));
+    }
+
+    /**
+     * Sends the info associated with a player, usually the local game's local player but sometimes the modified version
+     * of that player.
+     * @param player
```
IllianiBird, line 236: If we're not going to include a description of the param, do we need to include it?

In this instance it's super obvious what `Player player` is, especially from the rest of the JavaDoc

megamek/src/megamek/client/ui/panels/phaseDisplay/lobby/PlayerSettingsDialog.java
```diff
@@ -637,7 +637,7 @@ private void apply() {
                 // Use sendUpdate because we want the Game to allow us to change on Bot's
                 // behalf.
                 clientgui.chatlounge.sendProxyUpdates(updateEntities, client.getLocalPlayer());
-                // clientGUI.chatlounge.sendUpdate(updateEntities);
```
IllianiBird, line 640: Do you know why this was commented out, originally? My concern is that it was disabled for a good reason and that re-enabling it is going to cause unexpected problems.

The lack of comment explaining why something was commented out is always gray-hair inducing :D

megamek/src/megamek/client/AbstractClient.java
Sleet01: I _believe_ the build tests will throw an error if there is a comment block without a matching @param for each argument to the function so I've been relying on Idea's auto-complete to fill those in, but to be honest I just forgot to explicitly describe it here.

megamek/src/megamek/client/ui/panels/phaseDisplay/lobby/PlayerSettingsDialog.java
Sleet01: That is totally valid.
Looking through the history, it looks like I commented it out to test replacing it with sendProxyUpdates(), which is the call that allows one player to make updates to another player's configuration (e.g. Player 1 changing a bot's entities).
Then it got renamed during Tex's name normalization pass.
So I uncommented and renamed it, thinking that it was left commented out as an interim fix.
Then I tested with it in place.

Leaving it in place may cause additional chat log spam as player munition configuration changes will get logged twice, but it doesn't do anything for bot changes (the local player can't make bot config changes without going through `sendProxyUpdate()`).
But it's fine to delete it.

PR #2002, opened 2020-06-13
Refactor StepSprite, BoardView

SJuliez:
Pure refactor.

PR #7478, opened 2025-10-09
MM side of fix for MML #2002: remove extraneous low-BAR weight values

Sleet01:
It looks like a bunch of extraneous values got extrapolated and added during the big data refactoring pass.
Currently we rely on  BAR armor _not_ having a weight for a given tech rating to determine if it is not applicable, or at least having a 0.0 value defined.
We somehow ended up with numerous incorrect weight values for the lower BAR armor types; removing them corrects the MML issue where e.g. Tech Rating A can equip BAR 4 armour.

Testing:
- Added a flipping buttload of new unit tests to validate every combination against the documentation listed in the OP.
- Ran a couple games and loaded several SVs to verify that no canon designs have been rendered invalid.

Closes: MegaMek/megameklab#2002

Note: this patch is independent from the MML patch and can be merged in any order.

### DIFF
--- a/megamek/src/megamek/common/equipment/ArmorType.java
+++ b/megamek/src/megamek/common/equipment/ArmorType.java
@@ -1798,7 +1798,6 @@ public class ArmorType extends MiscType {
         armor.bar = 4;
 
         armor.armorType = T_ARMOR_SV_BAR_4;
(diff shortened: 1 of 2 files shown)


Sleet01:
> I appreciate the tests

It felt like it was time :)

Issue #7377, opened 2025-08-18
[RFE] Laser Heatsinks using old extreme temperature rules.


x-Lanceman-x:
### Enhancement Type *

Implementation of Errata

### Brief Description of Enhancement *

As presented originally in MaxTech, laser heat sinks (LHS) were not effected by immersion in water or by extreme temperatures. However, this clause was eliminated from their game rules starting with the first printing of Tactical Operations and remains absent in the current printing.

MegaMek is still using the MaxTech version of LHS.

_While they should be brought in line with the current rules, since the MaxTech code is already there and functional, I think it would be a good idea to allow it to remain as an optional unofficial rule._

### Use Cases or Rationale

1. Brings equipment in line with current ruleset.

### MegaMek Suite Version *

MekHQ-0.50.07-nightly-2025-08-01

### Final Verification

- [x] I confirm this request hasn't already been submitted (checked the tracker)
- [x] I've discussed or asked about this enhancement on MegaMek Discord
- [x] I’m opening this on the correct repo (MegaMek, MegaMekLab, or MekHQ)

HammerGS:
Fixed

PR #7483, opened 2025-10-10
Laser Heat Sink Bug Fix Complete (Needs Player testing)

HammerGS:
I've successfully fixed both laser heat sink bugs to update from MaxTech rules to current Tactical Operations
  rules:

  Changes Made:

  1. Fixed Water Immersion Immunity - Mek.java:1588-1590
  - Removed the early return for laser heat sinks in getHeatCapacityWithWater()
  - Laser heat sinks now correctly receive the +6 heat capacity bonus when underwater
  - Changed from 4 lines to 2 lines, eliminating the special case handling

  2. Fixed Temperature Immunity - HeatResolver.java:1241-1251
  - Removed the laserHS variable and || laserHS check in adjustHeatExtremeTemp()
  - Laser heat sinks are now correctly affected by extreme temperatures
  - Removed unnecessary variable declaration and condition

  Build Status:

  ✓ Project compiles successfully with no errors

  What These Fixes Do:

  Laser heat sinks now follow the current Tactical Operations rules:
  - Water immersion: Meks with laser heat sinks get the standard +6 heat capacity bonus when underwater (same as regular heat sinks)
  - Extreme temperatures: Laser heat sinks are affected by extreme hot/cold temperatures (same as regular heat sinks)
  - Night/dusk modifiers: Already working correctly (reduces modifier by 1 when generating heat, removes completely when overheating)
  - Ammo explosion bonus: Already working correctly (+1 to avoid roll at heat 19+)

  The changes align with GitHub issue #7377 and bring MegaMek's laser heat sink implementation in line with
  current BattleTech rules.


Sleet01:
@HammerGS 
And so I'm back from ~~outer space~~ the Far East.
How are we looking on playtesting?

HammerGS:
> @HammerGS And so I'm back from ~outer space~ the Far East. How are we looking on playtesting?

Claude got some but not all of the issues. I need to revisit this with Claude. That said gonna leave the feedback so I don't lose it. 

> if laser heat sinks are supposed to give a bonus when submerged, they do not give any.
> Notos prime. Walk, +1, fire 5x med opulse at 4 each (20), end turn at depth 4 water = earn 21 heat, sink 20 due to 20 laser heat sinks, at 1 heat.
> image
> 
> Yeah, something seems a little wonky
> image
> 
> Extreme temps still don't seem to be effecting them one way or another too
> Or...it's not reporting that they are at least.
> 
> i started the next turn a +1 heat instead of netural (if submerged work)
> 
> Here's another example that's easy to calculate: Notos didn't walk, fired all weapons for 20 heat. 100 C planetary condition should add an additional 5 heat 
> image
> 
> Looking at the code changes, I'm wondering though if it's getting the right number behind the scenes, but still reporting the old number

HammerGS:
Taking off draft to test a few things.

HammerGS:
While we had some success, there are still some issues. I'm going to explore more but in a different way with Claude code on this specific bug.

PR #7484, opened 2025-10-10
Clear/Delete Icon

SJuliez:
Adds a clear or delete icon (an X in the GUIP warning color) for JButtons. Used in an MML PR but makes sense to have it in MM.


IllianiBird:
How will this handle gui scaling? Or is that a non-issue?

SJuliez:
It scales with GUI scaling, FlatLaf handles that
220%:
image
80%:
image

IllianiBird:
It suddenly occurs to me that you're the one who _wrote_ the scaling code, of course your stuff would be good! :D

PR #7510, opened 2025-10-18
Improvement: Added Additional Rounding Utility Methods to MathUtility

IllianiBird:
These are required for https://github.com/MegaMek/mekhq/pull/7750 - I figured I'd add them to MathUtility as I doubt this will be the last time they're needed, and they might as well live in MegaMek so there is global access to them.

### DIFF
--- a/megamek/src/megamek/codeUtilities/MathUtility.java
+++ b/megamek/src/megamek/codeUtilities/MathUtility.java
@@ -33,9 +33,6 @@
 
 package megamek.codeUtilities;
 
-import static java.lang.Math.ceil;
-import static java.lang.Math.floor;
-
 import megamek.logging.MMLogger;
 
 public class MathUtility {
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/codeUtilities/MathUtility.java
```diff
@@ -42,6 +45,41 @@ private MathUtility() {
...
+     *
+     * @param value The double value to round.
+     *
+     * @return The rounded integer value, away from zero.
+     *
+     * @author Illiani
+     * @since 0.50.10
+     */
+    public static int roundAwayFromZero(double value) {
+        return value > 0 ? (int) ceil(value) : (int) floor(value);
```
psikomonkie: Doubles can be bigger than the maximum value of an int so Math.ceil & floor returns longs, which is why you need to cast the result as an int. Consider using `Math.toIntExact` instead of casting it as an int, so it throws an exception if it overflows (and therefore we need to consider using a long instead of an int).

IllianiBird: This is super interesting and not something I've really had to think about before. I've updated the code with better protections. Lemme know what you think.

psikomonkie: I love it! Good idea adding the warning for if it does invoke it, that'll help with debugging those cases if/when it happens.

psikomonkie (approved):
Nice!

PR #7517, opened 2025-10-21
Fix: Issue 7506 clan mek engines get IS suffix when making mtfFiles

### LINKED ISSUE #7506: Some Clan Mechs have IS indicated on Engines and Structure
Current MTF files for some Clan Mechs (NOT ALL) contain:

(EXAMPLE FROM DAISHI PRIME MTF FILE) 
mm-data/data/mekfiles/meks/3050U/Daishi (Dire Wolf) Prime.mtf

Mass:100
Engine:300 XL (Clan) Engine(IS)
Structure:IS Standard
Myomer:Standard

Error seems to be from how Mek.Java creates the mtf files since MegaMekLab also has these if a custom mech MTF file is created
From Mek.Java's getMTF function, line 4336:

if (hasEngine()) {
            sb.append(getEngine().getEngineName())
                  .append(" Engine")
                  .append(!(getEngine().hasFlag(Engine.CLAN_ENGINE) && isMixedTech()) ? ("(IS)")
                        : "");

--->the hasFlag(Engine.CLAN_ENGINE) is probably returning false

AND in line 4346:

sb.append(EquipmentType.getStructureTypeName(getStructureType(),
              TechConstants.isClan(structureTechLevel)));

--->structureTechLevel was never assigned to (apart from its default value) within the Mek.Java file itself, but not sure if it might have been assigned elsewhere

GypsyFromTheSatelliteofLove:
Super minor fix #7506  for megamek mtf file creation where the clan engines would get (IS) appended to them

line 4339 kept returning true because the parenthesis was misplaced
append(!(getEngine().hasFlag(Engine.CLAN_ENGINE) && isMixedTech()) ? ("(IS)") : "");

changed to:
append((!getEngine().hasFlag(Engine.CLAN_ENGINE) || isMixedTech()) ? ("(IS)") : "");

### DIFF
--- a/megamek/src/megamek/common/units/Mek.java
+++ b/megamek/src/megamek/common/units/Mek.java
@@ -4336,7 +4336,7 @@ public abstract class Mek extends Entity {
         if (hasEngine()) {
             sb.append(getEngine().getEngineName())
                   .append(" Engine")
-                  .append(!(getEngine().hasFlag(Engine.CLAN_ENGINE) && isMixedTech()) ? ("(IS)")
+                  .append((!getEngine().hasFlag(Engine.CLAN_ENGINE) || isMixedTech()) ? ("(IS)")
                         : "");
         } else {
             sb.append("(none)");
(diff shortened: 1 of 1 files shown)


GypsyFromTheSatelliteofLove:
I tested it on custom files and it fixed the mtf as long as you overwrite the file. 

But there is another error in this issue where clan mechs have IS standard structure that i havent figured out yet :)

PR #7520, opened 2025-10-21
RFE: I Lift Things Up and Put Them Down

psikomonkie:
Implements Tactical Operations: Advanced Rules (6th edition) rules for carrying friendly units (p. 94) in a Mek's arms. 

Unfortunately, I had to cut back the scope a little - Make `isCarryableObject` return `true` for `Tank` and `AeroSpaceFighter` to enable picking up and carrying those objects. The rules for damaging carried tanks etc. is complicated and based on logic we haven't implemented yet. Once TO:AR p. 88 Grappling has the friendly fire functionality then tanks etc. should be made carryble again. 

At the moment this only supports picking up and carrying Battle Armor in a Mek's hands.

This is a stepping stone towards handheld weapons, which are damaged in the same manner as carried Battle Armor.

image

image
(pay no mind to the values not lining up for the failed roll, I used IntelliJ's debugger to force it to fail)

To-do: 
- [x] Clean up comments
- [x] Renaming (in many cases ground objects should be switched to carryable, not all though)
- [x] Mechs must go hull-down to pick up level-1 objects
- [x] **PSR to pickup**
- [x] Break VTOL rotors if picked up
- [x] Double check airborne vehicles incl. fighters can't be picked up - must be landed!
- [x] Damaging carried units (implementing `Entity.damage(double)`)

### DIFF
--- a/megamek/src/megamek/common/units/Entity.java
+++ b/megamek/src/megamek/common/units/Entity.java
@@ -2987,24 +2987,26 @@ public abstract class Entity extends TurnOrdered
         return 0.0;
     }
 
-    public boolean canPickupCarryableObject(ICarryable carryable) {
-        if (carryable == null || !canPickupGroundObject() || !carryable.isCarryableObject()) {
-            return false;
+    /**
+     * Returns true if the carryable object is able to be picked up. Units must be hull down to pick up other units,
+     * unless the unit is tall. Airborne aeros cannot be grabbed either.
+     *
+     * @param isCarrierHullDown is the unit that's picking this up hull down, or otherwise able to pick up
+     *                          ground-level objects
+     * @return true if the object can be picked up, false if it cannot
+     */
+    @Override
+    public boolean canBePickedUp(boolean isCarrierHullDown) {
+        if (height() >= 1 || isCarrierHullDown) {
+            return !isAirborneAeroOnGroundMap();
         }
+        return false;
+    }
 
-        // TO:AR 90: Units must be hull down to pick up objects. Briefcases, the dedicated objective object
-        // in MegaMek should be exempt from this rule.
-        if (!(carryable instanceof Briefcase)) {
-            if (carryable instanceof Entity carryableEntity) {
-                if ((carryableEntity.height() < 1) && !isHullDown()) {
-                    return false;
-                }
-
-                // Can't grab a flying aero!
-                if (carryableEntity.isAirborneAeroOnGroundMap()) {
-                    return false;
-                }
-            }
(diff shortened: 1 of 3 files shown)


psikomonkie:
This is finally ready for review!

megamek/src/megamek/common/game/Game.java
```diff
@@ -1278,6 +1278,7 @@ public synchronized void addEntity(Entity entity, boolean genEvent) {
         if (entity instanceof Mek entityMek) {
             entityMek.setBAGrabBars();
             entityMek.setProtoMekClampMounts();
+            entityMek.setMekArms();
```
Sleet01, line 1281: I think we have enough transport options now that there's a strong argument to add "setupTransporters" in Entity and make every subclass responsible for extending that.
Not in this PR, but soon.
ETA: doubly so since we are also duplicating this code block in TWGameManager!

megamek/src/megamek/common/units/Entity.java
```diff
@@ -2987,10 +2987,33 @@ public double maxGroundObjectTonnage() {
...
+                if ((carryableEntity.height() < 1) && !isHullDown()) {
+                    return false;
+                }
+
+                // Can't grab a flying aero!
+                if (carryableEntity.isAirborneAeroOnGroundMap()) {
+                    return false;
+                }
+            }
+        }
```
Sleet01: Lots of `instanceof` usually indicates an opportunity to move code to the target class.

Could we have `static bool canBeCarriedBy(Entity carrier)` in ICarryable, and have the Briefcase and AerospaceFighter implement their own versions?
That function could see if the prospective carrier is short enough, hull down, or standing, except Briefcase and ASF versions would just return true and false, respectively.
Then this check just becomes:

```
        if (carryable == null || !canPickupGroundObject() || !carryable.isCarryableObject()) {
            return false;
        }
        return carryable.canBeCarriedBy(this) && (carryable.getTonnage() <= maxGroundObjectTonnage());
```

psikomonkie: Thank you! Great suggestion! Done! ASF can be picked up as long as it isn't airborne, and that's a method under entity so I only need to implement the method for `Briefcase` and `Entity`.

megamek/src/megamek/common/game/Game.java
psikomonkie: I agree! I've also noticed a minor oddity that's probably related to this duplicate code, so I've opened an issue to make sure we don't forget about it: #7531

PR #7527, opened 2025-10-25
Return Unit Sprites to be Inline in `Report`

psikomonkie:
#7189 helped break up reports to make the reports more readable. 

Unit sprites were put onto their own line. However, this leaves a lot of white space to the right of the sprite. In addition, when sprites are disabled the new line is still added. 

This change puts units sprites back inline but leaves the other spacing and line break changes from #7189 

image

image

image

Please don't judge my friendly fire, it was a quick and shrimple test. Kamini Ekachakra, pilot of ID:1 was properly compensated for their service.


Sleet01 (approved):
LGTM

PR #7536, opened 2025-10-29
Fix 4879: zipline dropping using starting position rather than target step position

### LINKED ISSUE #4879: [0.49.15] [Bug] Zip Line Dismount uses starting hex to calculate modifiers
When dismounting infantry from VTOLs using the TacOps Zip Lines rule, the modifiers for the roll needed for the infantry unit to avoid damage are determined by the hex the VTOL starts the turn on rather than the hex the infantry is being unloaded into.

To illustrate: the screen shot I have attached the Karnov is unloading an infantry unit into a clear hex, but began the turn over a wooded hex. When you press done on the movement, the roll confirmation screen shows that the roll is being influenced not by the hex the infantry is going into, but by the hex the VTOL started its movement over. 

Screenshot 2023-11-08 091940

Sleet01:
Use the target hex coordinates and target move step elevation for generating ziplining PSR mods.
Previously we were using the VTOL unit's starting hex and elevation, which would cause incorrect mods to be applied to the zipline PSR.

Testing:
- Tested dropping non-jump MP infantry and BA units from a VTOL at various heights and over various terrains.

Close #4879

### DIFF
--- a/megamek/src/megamek/client/ui/SharedUtility.java
+++ b/megamek/src/megamek/client/ui/SharedUtility.java
@@ -562,10 +562,10 @@ public class SharedUtility {
                       && (((Entity) targ).getJumpMP() < 1)
                       && !((Infantry) targ).isMechanized()) {
                     rollTarget = TWGameManager.getEjectModifiers(game, (Entity) targ, 0,
-                          false, entity.getPosition(), "zip lining");
+                          false, md.getFinalCoords(), "zip lining");
                     // Factor in Elevation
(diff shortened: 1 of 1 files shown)


IllianiBird (approved):
For how much of a pain in the bum this issue has been, I hate how small the fix is

Sleet01:
> For how much of a pain in the bum this issue has been, I hate how small the fix is

Yeah; if I hadn't worked on other movement path and step stuff, I wouldn't have had any idea what to do here.

PR #7549, opened 2025-10-31
Fix #7546: No Air to Ground LOS if Aircraft Ended Really Far Away

### LINKED ISSUE #7546: [Issue] Return of the LoS/Sensor Aero Bug
### Brief Description *

@psikomonkie Requested this issue be posted. 

In a game I had last night [unfortunately no longer have the save cause im an idiot], I ran into the bug where an Aero unit cannot attack a unit on its flight path due to said unit being outside of its visual/sensor range. This bug was at one time fixed, but seems to have come back at some point. I will keep a keen eye for it happening again.

### 3. Steps to Reproduce

_No response_

### Attach Files

_No response_

### Severity *

### MegaMek Suite Version *

50.10 Nightly 10-27

### Operating System *

Window 11

### Java Version *

Java Adoptium 11

### Final Verification

psikomonkie:
Fixes #7546 

Looks like these unit tests were stale! 

Also, re: `target.isGround()` - I believe this is incorrect. `isGround` returns true if the unit is a ground unit, if it's not an `Aero`. But I believe an attack is still "air to ground" if the unit is `Aero` and the unit is on the ground / not flying. I believe it should be `!target.isAirborne` - this would also cover cases where a non-aero unit has an altitude, for whatever reason. Altitude=/=Elevation - elevation is what VTOL & WiGEs have, so they are not `isAirborne` - an attack from an aircraft on a VTOL/WiGE should be air-to-ground.

### DIFF
--- a/megamek/src/megamek/common/compute/Compute.java
+++ b/megamek/src/megamek/common/compute/Compute.java
@@ -2024,11 +2024,11 @@ public class Compute {
         if (finalPos != null) {
             distance = flyerPosition.distance(finalPos);
         }
-        for (Coords c : flyingEntity.getPassedThrough()) {
-            if ((c != null)
-                  && ((c.distance(targetPosition) < distance) || (distance == 0))) {
-                finalPos = c;
-                distance = c.distance(targetPosition);
+        for (Coords coord : flyingEntity.getPassedThrough()) {
+            if ((coord != null)
+                  && ((coord.distance(targetPosition) < distance) || (distance == 0))) {
+                finalPos = coord;
+                distance = coord.distance(targetPosition);
             }
         }
         return finalPos;


megamek/src/megamek/common/compute/Compute.java
```diff
@@ -2010,6 +2010,30 @@ public static int getClosestFlightPathFacing(int attackerId, Coords aPos, Entity
...
+     */
+    public static @Nullable Coords getClosestToFlightPath(Entity flyingEntity, Coords targetPosition) {
+        Coords flyerPosition = flyingEntity.getPosition();
+        Coords finalPos = flyerPosition;
+        int distance = Integer.MAX_VALUE;
+        if (finalPos != null) {
+            distance = flyerPosition.distance(finalPos);
+        }
+        for (Coords c : flyingEntity.getPassedThrough()) {
+            if ((c != null)
```
IllianiBird: I know we're more pedantic about this, over in MekHQ-topia, but if we're introducing new code it might be worth avoiding pesky single-letter and abbreviated variable names.

You know how I am about this, so feel free to ignore the request.

psikomonkie: Done!

PR #7556, opened 2025-11-01
RFE: Simple Cargo (Lift Hoists)

psikomonkie:
Lift Hoists and Roof Racks

For the new Salvage functionality in MHQ, people want to be able to use Lift Hoists. This provides just enough support for Lift Hoists that we can do that. They aren't usable in MM yet, but they can be considered in MHQ for cargo capacity, and in theory could even be stuffed full of `Cargo` before a scenario.

`Briefcase` has been refactored into `abstract GroundObject` which is extended by `Briefcase` and `Cargo` (Cargo is more-or-less placeholder at the moment, but will use much more prohibitive rules for picking up & putting down, while Briefcase is for simple gameplay objectives.)

`ExternalCargo` has been refactored into `abstract ExternalCargo`. It's extended by `MekArms`, which functions as `ExternalCargo` did prior to this; `LiftHoist` which is the transporter used for holding whatever's in a lift hoist - notably not really functional yet, the point was to have the Transporter code available to MekHQ so both MegaMek and MekHQ can share logic when possible; And `RoofRack`, a partial implementation of the cargo carrier rules from total war. None of the logic is there for it yet other than the capacity. 

Some other methods and classes were renamed to more it more clear what is being passed around.

Both `LiftHoist` and `RoofRack` should be inaccessible for players, they're placeholders so MekHQ can begin considering them while they're implemented concurrently for MegaMek gameplay.

### DIFF
--- a/megamek/src/megamek/common/weapons/FlamerHandlerHelper.java
+++ b/megamek/src/megamek/common/weapons/FlamerHandlerHelper.java
@@ -39,6 +39,7 @@ import megamek.common.Report;
 import megamek.common.equipment.ArmorType;
 import megamek.common.equipment.EquipmentType;
 import megamek.common.equipment.WeaponType;
+import megamek.common.options.OptionsConstants;
 import megamek.common.units.Entity;
 
 /**
@@ -50,16 +51,8 @@ public class FlamerHandlerHelper {
     /**
      * Handles flamer heat damage.
      */
-    // PLAYTEST3 call without playtest if it is not passed.
     public static void doHeatDamage(Entity entityTarget, Vector<Report> vPhaseReport, WeaponType weaponType,
           int subjectId, HitData hit) {
-        doHeatDamage(entityTarget, vPhaseReport, weaponType, subjectId, hit, false);
-    }
-
-
-    // PLAYTEST3 included. 
-    public static void doHeatDamage(Entity entityTarget, Vector<Report> vPhaseReport, WeaponType weaponType,
-          int subjectId, HitData hit, boolean playtestThree) {
         Report report = new Report(3400);
         report.subject = subjectId;
         report.indent(2);
@@ -72,6 +65,10 @@ public class FlamerHandlerHelper {
 
         boolean heatDamageReducedByArmor = false;
         int actualDamage = heatDamage;
+        boolean playtestThree = false;
+        if (entityTarget != null && entityTarget.getGame() != null) {
+            playtestThree = entityTarget.getGame().getOptions().booleanOption(OptionsConstants.PLAYTEST_3);
+        }
 
         // armor can't reduce damage if there isn't any
         if (entityTarget.getArmor(hit) > 0) {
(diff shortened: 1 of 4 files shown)


IllianiBird:
Have you tried rebuilding the project in IDEA? Sometimes it gets stuff stuck in memory from other branches

IllianiBird:
Oh god, Copilot is reviewing our definitely not janky princess code isn't it? My sympathies.

psikomonkie:
This is ready for review.

megamek/src/megamek/common/game/Game.java
```diff
@@ -1275,21 +1282,7 @@ public void addEntity(Entity entity) {
...
-            entityTank.setTrailerHitches();
-        }
-
-        // Add magnetic clamp mounts
-        if ((entity instanceof Mek) && !entity.isOmni() && !entity.hasBattleArmorHandles()) {
-            entity.addTransporter(new ClampMountMek());
-        } else if ((entity instanceof Tank entityTank) && !entityTank.isOmni() && !entityTank.hasBattleArmorHandles()) {
-            entityTank.addTransporter(new ClampMountTank());
-        }
+        entity.addIntrinsicTransporters();
```
Sleet01, line 1285: Glorious!

megamek/src/megamek/common/jacksonAdapters/CarryableDeserializer.java
```diff
@@ -67,20 +68,20 @@ public CarryableDeserializer(Class<?> vc) {
     }
 
     @Override
-    public CarryableInfo deserialize(JsonParser jp, DeserializationContext context) throws IOException {
+    public GroundObjectInfo deserialize(JsonParser jp, DeserializationContext context) throws IOException {
         JsonNode node = jp.getCodec().readTree(jp);
         requireFields("Carryable", node, NAME, WEIGHT);
 
-        Briefcase briefcase = new Briefcase();
+        GroundObject groundObject = new Briefcase(); //TODO: Differentiate between Briefcase and Cargo
```
Sleet01, line 75: Does this TODO still need addressing?

megamek/unittests/megamek/common/actions/WeaponAttackActionToHitTest.java
```diff
@@ -641,14 +641,14 @@ void basicTest() {
 
             MekFileParser mekFileParser;
             HandheldWeapon handheldWeapon;
-            Mek mek;
+            MekWithArms mek;
```
Sleet01, line 644: I'm starting to feel MekWithArms should go away, and we should use composition to add arms to units that have them.
Of course, then MekArms would need to be their own class, and the cargo space functions would need to iterate over them...
I will sit with this for a while.

Sleet01 (approved):
Just one question about a TODO comment, but overall looks good.

megamek/src/megamek/common/game/Game.java
psikomonkie: Glad you like it! Thanks for suggesting it the other day.

megamek/unittests/megamek/common/actions/WeaponAttackActionToHitTest.java
psikomonkie: So, I didn't realize for a while that `MekWithArms` isn't to differentiate a like Stalker from Atlas, but `BipedMek` and `TripodMek` from `QuadMek`. The prior have arms and use distinctly different code than the `QuadMek` and its funky arms for legs situation.

megamek/src/megamek/common/jacksonAdapters/CarryableDeserializer.java
psikomonkie: Eventually! It's a heads up that this is wrong and will horrifically break if it isn't updated alongside the implementation of Cargo. I hate when I'm investigating old code, and I can't tell if there's a reason something is the way it is, or if it's just a mistake/known shortfall. I want to make it clear it will need fixed.

There is nothing that utilizes `Cargo` in gameplay with this PR. The existing ground objects are still handled as `Briefcase`.

Issue #7557, opened 2025-11-01
[Issue]Playtest 3 Option available but not working


conservationstats:
### Brief Description *

In the Game Options tab, Playtest #3 has an option, but so far I have tried AP and Precision Autocannon ammo. Neither have the new rounds per ton correct, multiplier is still x.5 . AP doesn't have the new modifier for crit chance. Also tried MRMs, they still show a +1 to hit malus. Ultra AC's however don't seem to jam. 

### Severity *

Critical (Game-breaking/Crash): The game crashes or a core feature (like saving, loading, or network connection) is completely unusable.

### MegaMek Suite Version *

50.10-nightly-2025-11-01

### Operating System *

Windows II

### Java Version *

Java Vendor: Eclipse Adoptium   Java Version: 17.0.16

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have included all necessary information and files to help reproduce this issue
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

TenkawaBC:
Confirmed the playtest ammos were not showing in the selection drop-down correctly. Working on getting that fixed.

The MRMs and AP ammo +1 to hit should be negated in the code.  Can you provide a savefile of your game so I can review the round reports and the logs?

TenkawaBC:
AP ammo modifier for crit roll was reversed. fixing that one.

The +1 to hit removal logic is really simple, so I'm wondering what you are seeing:

// PLAYTEST3 AP ammo is no longer +1 to hit.
        if (!game.getOptions().booleanOption(OptionsConstants.PLAYTEST_3)) {
            if (((ammoType.getAmmoType() == AmmoType.AmmoTypeEnum.AC) ||
                  (ammoType.getAmmoType() == AmmoType.AmmoTypeEnum.LAC) ||
                  (ammoType.getAmmoType() == AmmoType.AmmoTypeEnum.AC_IMP) ||
                  (ammoType.getAmmoType() == AmmoType.AmmoTypeEnum.PAC)) &&
                  (munition.contains(AmmoType.Munitions.M_ARMOR_PIERCING)
                        || munition.contains(AmmoType.Munitions.M_ARMOR_PIERCING_PLAYTEST))) {
                toHit.addModifier(1, Messages.getString("WeaponAttackAction.ApAmmo"));
            }
        }

[...]

PR #7559, opened 2025-11-02
Playtest issues. AP ammo crit modifiers were flipped. Playtest ammos were not available.

TenkawaBC:
Fixing issues reported in bug ticket #7557
https://github.com/MegaMek/megamek/issues/7557

AP ammo crit modifiers were flipped between playtest and normal. Playtest ammos were not listed as available.

### DIFF
--- a/megamek/src/megamek/common/actions/compute/ComputeToHit.java
+++ b/megamek/src/megamek/common/actions/compute/ComputeToHit.java
@@ -1463,11 +1463,12 @@ public class ComputeToHit {
                     modifier += RangeType.RANGE_LONG;
                 }
             }
-            toHit.addModifier(modifier, Messages.getString("WeaponAttackAction.WeaponMod"));
             // PLAYTEST3 No more modifier for MRMs
             if (game.getOptions().booleanOption(OptionsConstants.PLAYTEST_3) && weaponType instanceof MRMWeapon) {
                 modifier--;
             }
+            toHit.addModifier(modifier, Messages.getString("WeaponAttackAction.WeaponMod"));
+
         }
 
         // Indirect fire (LRMs, mortars and the like) has a +1 mod


TenkawaBC:
Need to add a fix to this on Sunday morning. Do not merge yet

IllianiBird:
Thanks for jumping on this so quick

IllianiBird:
I saw your message after reviewing. I've moved the PR into draft so it can't be merged until you're ready

TenkawaBC:
This can now move forward.

megamek/src/megamek/common/actions/compute/ComputeToHit.java
```diff
@@ -1463,11 +1463,12 @@ private static ToHitData compileWeaponToHitMods(Game game, Entity ae, Entity spo
                     modifier += RangeType.RANGE_LONG;
                 }
             }
-            toHit.addModifier(modifier, Messages.getString("WeaponAttackAction.WeaponMod"));
             // PLAYTEST3 No more modifier for MRMs
             if (game.getOptions().booleanOption(OptionsConstants.PLAYTEST_3) && weaponType instanceof MRMWeapon) {
                 modifier--;
             }
+            toHit.addModifier(modifier, Messages.getString("WeaponAttackAction.WeaponMod"));
+
```
psikomonkie, line 1471: Could this be handled like other weapon modifiers - could this be in `MRMWeapon`? It looks like for weapons normally, including the MRM, the `toHitModifier` comes from here. I think you could put this check in `MRMWeapon#getCorrectHandler(ToHitData, WeaponAttackAction, Game, Manager)`.

TenkawaBC: I considered it, but unfortunately I can't do it in MRMweapon. the modifier is set on initialization, and has no access to the game object, which we need for the purposes of the option flag.

In future it potentially might be possible to make a get and set call into mrmweapon to set the modifier, and then during one of the various initializations change it over, but it is a LOT more complicated, so it is just best to override it during the calculation for what modifiers apply and just reduce it here.

TenkawaBC: the GetCorrectHandler refers back to MRMHandler, which doesn't touch the tohitmodifier.

It comes down to this call: int modifier = weaponType.getToHitModifier(weapon);. This call is what returns the 1 on MRMs.

Issue #7571, opened 2025-11-03
[Issue] Playtest3 if an empty ammo bin is hit, it still explodes for 2 damage


TenkawaBC:
### Brief Description *

Due to the code for explosive equipment, ammo bins that were empty could explode for 2 damage. Ammo bins with rounds left in them were handled normally.

### 3. Steps to Reproduce

1. enable playtest3
2. empty an ammo bin
3. take a critical to that ammo bin

### Severity *

Medium (Gameplay Limitation): Non-core functionality is impaired, providing a suboptimal but playable experience.

### MegaMek Suite Version *

0.50.10 nightly

### Operating System *

windows 11

### Java Version *

17.0.11

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have included all necessary information and files to help reproduce this issue
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

TenkawaBC:
This was fixed in a PR and can be closed.

PR #7572, opened 2025-11-03
Fix playtest3 ammo explosion with 0 rounds remaining bug #7571

TenkawaBC:
This fixed this bug https://github.com/MegaMek/megamek/issues/7571, initially reported by Daishi on the discord.

When a bin had 0 ammo, it was still flagged as potentially explosive, so triggered the code to be 2 damage / crit. I have added a check to ensure it is not an AmmoType class, which will avoid this.

### DIFF
--- a/megamek/src/megamek/server/totalWarfare/TWGameManager.java
+++ b/megamek/src/megamek/server/totalWarfare/TWGameManager.java
@@ -22617,7 +22617,7 @@ public class TWGameManager extends AbstractGameManager {
         int damage = mounted.getExplosionDamage();
 
         // PLAYTEST3 explosive equipment does 2 damage per crit slot. This overrides previous amounts
-        if (game.getOptions().booleanOption(OptionsConstants.PLAYTEST_3)) {
+        if (game.getOptions().booleanOption(OptionsConstants.PLAYTEST_3) && !(mounted.getType() instanceof AmmoType)) {
             damage = mounted.getNumCriticalSlots() * 2;
         }
 


megamek/src/megamek/common/weapons/handlers/FlamerHandler.java
```diff
@@ -87,15 +87,16 @@ protected void handleEntityDamage(Entity entityTarget, Vector<Report> vPhaseRepo
         if (bmmFlamerDamage || flamerDoesOnlyDamage || (flamerDoesHeatOnlyDamage && !entityTarget.tracksHeat())) {
             // PLAYTEST3 Heat-dissipating armor reduces damage
             if (game.getOptions().booleanOption(OptionsConstants.PLAYTEST_3)) {
-                hit.setHeatWeapon(true);
+                if (hit != null) {
```
Sleet01, line 90: I'm curious how `hit` _could_ be null, now.
Are we generating a lot of WeaponAttackActions with null toHit members?
We might want to investigate that at some point.

TenkawaBC: I have no idea why/how hit could be null, but it was and caused a crash during a victory declaration, so it remains that it CAN be null.

megamek/src/megamek/common/weapons/FlamerHandlerHelper.java
```diff
@@ -72,6 +65,10 @@ public static void doHeatDamage(Entity entityTarget, Vector<Report> vPhaseReport
 
         boolean heatDamageReducedByArmor = false;
         int actualDamage = heatDamage;
+        boolean playtestThree = false;
+        if (entityTarget != null) {
+            playtestThree = entityTarget.getGame().getOptions().booleanOption(OptionsConstants.PLAYTEST_3);
+        }
 
         // armor can't reduce damage if there isn't any
         if (entityTarget.getArmor(hit) > 0) {
```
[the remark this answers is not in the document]
IllianiBird: @TenkawaBC can I ask you just to verify that this isn't an issue, and if it is can you pop in a follow-up PR?

IllianiBird:
I went ahead and merged despite Copilot query. The value of getting this in the nightly and fixing a problem we _know_ occurs outweighs potential issues that _might_ arise in a patch of code not touched directly by this PR.

PR #7573, opened 2025-11-03
Fix: Corrected DropShips Being Incorrectly Able to Salvage on Ground

IllianiBird:
DropShips are not meant to be a ground salvage enabled unit, this fixes that by removing the override. The default returns `false`.

### DIFF
--- a/megamek/src/megamek/common/units/Dropship.java
+++ b/megamek/src/megamek/common/units/Dropship.java
@@ -821,11 +821,6 @@ public class Dropship extends SmallCraft {
     }
 
     @Override
-    public boolean canPerformGroundSalvageOperations() {
-        return true;
-    }
-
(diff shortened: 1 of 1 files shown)


psikomonkie:
Could a Mule help transport cargo on the ground with its cargo bays? (I assume it's landing & taking off)

gsparks3:
> Could a Mule help transport cargo on the ground with its cargo bays? (I assume it's landing & taking off)

DS are closer to the "final destination" of a salvaged unit. Logically, you'd still need other equipment to move it off the field and into the DS even if it landed as close as possible. Spheroids in particular have a minimum safe distance from their landing point if they don't want to damage whatever they're picking up

Sleet01 (approved):
LGTM

Issue #7576, opened 2025-11-04
[Issue] Playtest 3 ECM Still Fully Disabling C3 Network


stonewall072:
### Brief Description *

With Playtest 3 options turn on, ECM is still completely disabling C3 networks instead of just degrading them. 

### 3. Steps to Reproduce

1. Turn on playtest 3 option
2. Build a C3 force
3. Build OpFor with ECM
4. Get C3 Slave Unit within close range of ECM carrier
5. Observe that C3 is still totally disabled, instead of halving bonus

### Attach Files

<img width="2096" height="1031" alt="Image" src="https://github.com/user-attachments/assets/b744d07c-5fd1-48b3-ba13-b8330ab72ca6" />

### Severity *

Medium (Gameplay Limitation): Non-core functionality is impaired, providing a suboptimal but playable experience.

### MegaMek Suite Version *

MekHQ-0.50.10-nightly-2025-11-04

### Operating System *

Windows 11

### Java Version *

Adoptium 17.0.11

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have included all necessary information and files to help reproduce this issue
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

x-Lanceman-x:
Closed by #7577

PR #7577, opened 2025-11-04
Playtest3 C3 network fix for bug #7576. 

TenkawaBC:
Fixed C3 checks in Entity to no longer break C3 networks in the presence of ECM.


Sleet01 (approved):
LGTM

PR #7579, opened 2025-11-05
Simple Cargo - Improvements to support Lift Hoist in MHQ

psikomonkie:
Entities in MHQ won't have an id when initially created. It should have one when we're assigning the game, so let's try to assign it then.

### DIFF
--- a/megamek/src/megamek/common/loaders/MtfFile.java
+++ b/megamek/src/megamek/common/loaders/MtfFile.java
@@ -1010,9 +1010,8 @@ public class MtfFile implements IMekLoader {
                                 mount.setFacing(facing);
                             }
                         }
-                        if (etype instanceof MiscType && mount.getType().hasFlag(MiscType.F_LIFT_HOIST)) { // Cargo
-                            // Container
-                            // too?
+                        if (etype instanceof MiscType && mount.getType().hasFlag(MiscType.F_LIFT_HOIST)) {
+                            // Cargo container too?
                             mek.addTransporter(new LiftHoist(mount, mek.getWeight() / 2), isOmniPod);
                         }
                     }


megamek/src/megamek/common/equipment/LiftHoist.java
```diff
@@ -35,19 +35,26 @@
...
+import megamek.logging.MMLogger;
 
 /**
  * Transporter for Lift Hoists as described TW p. 136
  */
 public class LiftHoist extends ExternalCargo {
+    private final static MMLogger logger = MMLogger.create(LiftHoist.class);
 
+    private transient Entity entity;
     private int entityId;
```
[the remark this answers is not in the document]
IllianiBird: I'm not sure about this. I leave it to your discretion.

megamek/src/megamek/common/equipment/LiftHoist.java
```diff
@@ -64,6 +71,18 @@ public boolean canLoad(Entity unit) {
...
+    public void setGame(Game game) {
+        super.setGame(game);
+        if (entity != null) {
+            entityId = entity.getId();
+        } else if (entityId != Entity.NONE) {
+            entity = game.getEntity(entityId);
+        } else {
+            logger.warn("LiftHoist has no entity or entityId");
+        }
+    }
```
IllianiBird, line 84: I'm not sure about this logic. Lemme run through it to make sure I'm parsing it correctly

- If entity isn't `null`, assign entity ID
- If entity _is_ `null`, fetch the entity from the game using entityId

If entity is `null` where are we getting the entityId from? Furthermore, what happens if game.getEntity returns `null`?

megamek/src/megamek/common/loaders/MtfFile.java
```diff
@@ -1010,7 +1010,9 @@ private void parseCrits(Mek mek, int loc) throws EntityLoadingException {
                                 mount.setFacing(facing);
                             }
                         }
-                        if (mount.getType().hasFlag(MiscType.F_LIFT_HOIST)) { // Cargo Container too?
+                        if (etype instanceof MiscType && mount.getType().hasFlag(MiscType.F_LIFT_HOIST)) { // Cargo
+                            // Container
+                            // too?
```
IllianiBird: The worlds' tiniest point, let's move the comment above the conditional so it doesn't get gored across multiple lines.

megamek/src/megamek/common/equipment/LiftHoist.java
psikomonkie: Great question! This comes from the `private transient Entity`.

Entity may be null on clients because the Entity will not be serialized with this object.

By the time it's being sent, the entity should've been assigned an entity id. 

MekHQ doesn't give a new entity an entity id right away, it'll happen after one of the set games, so that's why it'll try to set the entity id when the game is updated. 

If both Entity and Entity ID are missing, then the transporter won't be able to work. But that shouldn't happen! It should always be able to have an entity or an entity id.

IllianiBird: Perfect

megamek/src/megamek/common/loaders/MtfFile.java
psikomonkie: The world's tiniest commit

PR #7584, opened 2025-11-07
PR: Updates to Princess autocannon rate-of-fire

SuperStucco:
The current code which controls whether Princess uses rapid fire on RACs, UACs, and other autocannons is lifted almost straight from ye olde TestBot code, and comes with a number of shortcomings.  This updates the code in a number of ways:

- change behavior setting from Aggression (controls how close to get to target) to Self Preservation, which also controls things like holding fire on long odds and conserving ammunition.  Moving the slider to the left means more likely to fire multiple shots, and moving the slider to the right means more likely to fire fewer or single shots.
- change filtering from testing ammo type to weapon type, so LAC/PAC will be rapid-fired if the option is enabled
- some tuning to the internal numbers so average performance works well.  Rapid firing standard autocannon have a slight debuff to account that things go badly when they fail, with an exception for infantry field guns which don't fail.
- RACs originally were limited to 2/4/6 rate of fire, they are now capable of the odd-numbered 3 and 5 rates as well
- some improvements to program flow for performance and readability

### DIFF
--- a/megamek/src/megamek/common/compute/Compute.java
+++ b/megamek/src/megamek/common/compute/Compute.java
@@ -3913,11 +3913,13 @@ public class Compute {
         boolean isUAC = false;
         boolean isRAC = false;
 
-        // Get the to-hit number for this attack
-        if (null == atk) {
-            LOGGER.warn("null provided as WeaponAttackAction in Compute.spinUpCannon");
+        // Basic protections against null values
+        if (null == atk || null == cgame || null == atk.toHit(cgame)) {
+            LOGGER.warn("null parameter passed to Compute.spinUpCannon");
             return final_spin;
         }
+
+        // Get the to-hit number for this attack
         to_hit = atk.toHit(cgame).getValue();
 
         // If weapon can't hit target, exit with the default mode setting


megamek/src/megamek/common/compute/Compute.java
```diff
@@ -1470,13 +1476,13 @@ public static ToHitData getRangeMods(Game game, Entity attackingEntity, WeaponMo
...
         int c3ecmDist = Compute.effectiveDistance(game, c3spotterWithECM, target, false);
-        
+
         // C3 can't benefit from LOS range.
         int c3range = RangeType.rangeBracketC3(c3dist, distance, weaponRanges, useExtremeRange, false);
         // PLAYTEST3 checking for ECM ranged member
         int c3ecmRange = RangeType.rangeBracketC3(c3ecmDist, distance, weaponRanges, useExtremeRange, false);
-        
-       
+
```
Sleet01, line 1484: This can be one line

megamek/src/megamek/common/compute/Compute.java
```diff
@@ -3882,82 +3888,110 @@ public static double getAmmoAdjDamage(Game game, WeaponAttackAction weaponAttack
...
+                    return final_spin;
+                }
 
-            // If random roll is >= to-hit + 3 then set to six-spin
-            if (spinupThreshold >= (threshold + 2)) {
-                final_spin = 3;
-                weapon.setMode(Weapon.MODE_RAC_SIX_SHOT);
+                if (to_hit <= (spinupThreshold - 2)) {
+                    final_spin = 4;
+                    weapon.setMode(Weapon.MODE_RAC_FIVE_SHOT);
```
Sleet01, line 3973: IIRC there is a significant advantage to 3/5 shots over 4/6 due to the RAC jam roll stepping.
Or rather, the expected damage output : jamming chance is maximized for 3 and 5 shots, since they use the same jam threshold as 2/4.

SuperStucco:
> Most of the conversations we've had on Discord have included the desire for Princess to use the "safe" firing modes of 3 shots or 5 shots by preference, as this maximizes her effectiveness while minimizing jam chances. If possible, please see about implementing that logic as well.
> 
> Also, the CI/CD tests are failing with an NPE.

That should be handled by the logic.  It will use 4 shots if the TN is 6 or less, which is a reasonable target; if the TN is 7 or higher it will switch to 3 shots instead.  Six shots vs. five is handled slightly differently, with six shots requiring a 1-point advantage between the provided threshold and attack TN; firing a full six shots will require a more aggressive setting.

Reasonably certain I haven't changed anything seriously enough to not pass the CI/CD tests.  From Discord, they may need reviewing which is beyond my capability.

Sleet01:
By downloading the CI/CD test output from the failed task, I got this:
```
java.lang.NullPointerException: Cannot invoke "megamek.common.ToHitData.getValue()" because the return value of "megamek.common.actions.WeaponAttackAction.toHit(megamek.common.game.Game)" is null
	at megamek.common.compute.Compute.spinUpCannon(Compute.java:3917)
	at megamek.client.bot.princess.WeaponFireInfo.initDamage(WeaponFireInfo.java:891)
	at megamek.client.bot.princess.WeaponFireInfo.initDamage(WeaponFireInfo.java:839)
	at megamek.client.bot.princess.WeaponFireInfoTest.testInitDamage(WeaponFireInfoTest.java:231)
	at java.base/java.lang.reflect.Method.invoke(Method.java:569)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.ArrayList.forEach(Unknown Source)
```
in test `all > megamek.client.bot.princess > WeaponFireInfoTest.testInitDamage()`

Sleet01:
Looks like you need to add a ToHit with a value to that test, so it's not passing in a null value (alternatively you can check for a null value in your changes, which wouldn't be a bad idea either).

Sleet01:
Looks like there's still an issue in one of the tests.  You should be able to run just that test, locally, and debug the issue.

PR #7590, opened 2025-11-08
Fix 7421: Replace incorrect artillery code with correct bomb code in BA Micro Bomb Rack handler

### LINKED ISSUE #7421: [Issue] [0.50.06] BA Micro Bomb Rack using artillery vertical damage falloff
### Brief Description *

Per TW bomb rules, damage should only be dealt vertically if they strike water or building hexes. The BA Micro Bomb Rack (Sylph BA used as example) appears to be using the TO:AR artillery vertical damage rules instead, where vertical damage is always dealt in the central hex, but at -10 damage per level. Combined with the Micro Bomb Rack's requirement of being at elevation 1, this causes 6-suit squads of BA to damage themselves when bombing regardless of hex type. Smaller squads do 10 damage or less, so the issue is not visible on hexes that are not water or buildings.

Image

Image

Selecting the Micro Bomb Rack for firing also appears to cause a NPE. I believe that is unrelated, although it would be good if that could be fixed as well.

### 3. Steps to Reproduce

1. Load the provided save file.
2. Declare bomb attacks against the ground with the three Sylph BA squads.
3. Observe that the 6-suit squad takes self damage from the bomb attack, while the others do not.
4. (optional) Observe that selecting the Micro Bomb Rack in the weapons menu produces a NPE.

### Attach Files

[sylph_bomb_test.sav.gz](https://github.com/user-attachments/files/22452760/sylph_bomb_test.sav.gz)

### Severity *

### MegaMek Suite Version *

0.50.06

### Operating System *

Windows 10

### Java Version *

     Java Vendor: Eclipse Adoptium     Java Version: 17.0.11

### Final Verification

Sleet01:
BA Micro Bomb Racks were dealing damage to the carrying units no matter what terrain they were over, which is not how bomb blast shaping should work.
I didn't notice this when expanding on the artillery handling, but the BA micro bomb handler ended up with artillery, not bomb, damage delivery code at some point.

This PR:
1. switches wrong code that caused artillery-type blast shape with correct bomb blast call.
2. adds missing bomb markings so that UI elements showing bomb hits and misses will be drawn.

Testing:
- Tested with Sylph BAs over a variety of targets and terrain

Fix #7421

### DIFF
--- a/megamek/src/megamek/common/weapons/handlers/MicroBombHandler.java
+++ b/megamek/src/megamek/common/weapons/handlers/MicroBombHandler.java
@@ -34,20 +34,22 @@
 
 package megamek.common.weapons.handlers;
 
-import static megamek.common.weapons.handlers.AreaEffectHelper.calculateDamageFallOff;
-
 import java.io.Serial;
 import java.util.Vector;
 
+import megamek.common.HexTarget;
+import megamek.common.Messages;
+import megamek.common.Player;
 import megamek.common.Report;
+import megamek.common.SpecialHexDisplay;
 import megamek.common.ToHitData;
 import megamek.common.actions.WeaponAttackAction;
 import megamek.common.board.Coords;
 import megamek.common.compute.Compute;
+import megamek.common.equipment.enums.BombType;
 import megamek.common.game.Game;
 import megamek.common.loaders.EntityLoadingException;
 import megamek.common.units.Entity;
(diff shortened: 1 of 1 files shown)


IllianiBird (approved):
I'm not seeing anything obviously wrong here. And I know you'll have tested it, so approving.

PR #7596, opened 2025-11-08
Fix 5234: vtols cannot load at end of turn

### LINKED ISSUE #5234: [0.49.18] Bug: VTOL's unable to pick up infantry at the end of movement
Bundled Megamek with MekHQ 0.49.18

TW pages 223-224

Currently VTOL's are behaving differently to ground vehicles when picking up infantry. 

A VTOL should be able to pick up an infantry unit at the end of it's movement when it lands and still has sufficient MP remaining. In MM this isn't working, resulting in a two turn process to load infantry (one to land, one to remain stationery and pick up at the beginning of movement).

Name: test

[save3.sav.gz](https://github.com/MegaMek/megamek/files/14547232/save3.sav.gz)

[save4.sav.gz](https://github.com/MegaMek/megamek/files/14547234/save4.sav.gz)

image

image

image

Option to load only available if VTOL starts landed.

Sleet01:
There were a number of issues with our Load step handling, including how we detect that a Load step option should be displayed.

This PR:
- Makes loading units into VTOL at the end of its movement possible.
- Allows loading multiple units, per TW 223-ish, but only at the end of the movement.
- Cleans up selecting from multiple valid candidates by filtering out units that already have a target transporter assigned.

NOTE: we need a new serializable "load/unload action data" object to help keep tracking the state of units that _should_ load, but have not yet loaded, clean and correct.
We currently use a mishmash of MovementPath state, FinalStep state, GUI state, and twiddling the Entity state to try to keep straight the list of _possible_ vs _selected_ loadable units.
But we can't currently differentiate units that won't be loaded, units that will be loaded by the current unit, and units that will be loaded by _another_ unit from the list of valid target entities.
This means that we might do the wrong thing when, e.g. a Large VTOL SV is trying to figure out if it can load an infantry unit from an adjacent hex despite a VTOL in that infantry unit's hex already planning to load it.
This will lead to unexpected state, if not outright incorrect state, in some edge cases.
However, this work is outside the scope of the bug fix.

Testing:
- Ran several repros with various numbers of infantry

Fix #5234


megamek/src/megamek/common/units/Entity.java
```diff
@@ -8583,9 +8599,9 @@ public boolean canLoad(Entity unit, boolean checkElev) {
                 boolean isLoungeOrUnknownPhase = false;
                 if (getGame() != null) {
                     isLoungeOrUnknownPhase = getGame().getPhase().isLounge() || getGame().getPhase().isUnknown();
-                }
+                } // getElevation() on this is the issue; see ce.canGoUp(cmd.getFinalElevation(), cmd.getFinalCoords(), cmd.getFinalBoardId())
```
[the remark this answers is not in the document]
Sleet01: Damn your sensors, you're right.  I'll fix this tonight.

megamek/src/megamek/client/ui/panels/phaseDisplay/MovementDisplay.java
```diff
@@ -3199,63 +3202,80 @@ private Entity getMountedUnit() {
...
 
-        if (!(choice instanceof Infantry)) {
-            List<Integer> bayChoices = new ArrayList<>();
+        // Handle canceled dialog
+        if (choice == null) {
+            return null;
+        }
+
+        // Safety set, apparently
+        choice.setTargetBay(-1);
```
IllianiBird: I feel like we should have a static constant we can pull in here, rather than magic numbers. Not necessarily an issue that needs to be fixed in this specific PR, but something we should keep in mind I think.

megamek/src/megamek/client/ui/panels/phaseDisplay/MovementDisplay.java
```diff
@@ -3199,63 +3202,80 @@ private Entity getMountedUnit() {
...
+
+        List<Integer> bayChoices = new ArrayList<>();
+        for (Transporter t : currentEntity().getTransports()) {
```
IllianiBird: If you're not against the idea, I'd really like us to move away from single letter and abbreviations in variable, class, and method names. This is something we've been doing in mhq and it's made the code substantially easier to work with and review.

megamek/src/megamek/client/ui/panels/phaseDisplay/MovementDisplay.java
```diff
@@ -3199,63 +3202,80 @@ private Entity getMountedUnit() {
...
+            if (t.canLoad(choice) && (t instanceof Bay)) {
+                bayChoices.add(((Bay) t).getBayNumber());
+            }
+        }
+
+        if (bayChoices.size() == 1) {
+            choice.setTargetBay(bayChoices.get(0));
+        } else if (bayChoices.size() > 1) {
+            String[] retVal = new String[bayChoices.size()];
```
IllianiBird: I know `retVal` is pretty common, but it doesn't really tell me what this String[] represents. I know I can just look right, but it would be far better (imo) to have a variable name that explains the variable. That way when it appears elsewhere I don't need to head back to where it was initialized to find the intent.

megamek/src/megamek/client/ui/panels/phaseDisplay/MovementDisplay.java
```diff
@@ -3199,63 +3202,80 @@ private Entity getMountedUnit() {
...
+            int i = 0;
+            for (Integer bn : bayChoices) {
```
IllianiBird: I think we can use a primitive here, unless we can reasonably expect bayChoices to contain `null` values

megamek/src/megamek/client/ui/panels/phaseDisplay/MovementDisplay.java
```diff
@@ -3199,63 +3202,80 @@ private Entity getMountedUnit() {
...
+                retVal[i++] = bn.toString() + " (Free Slots: " + (int) currentEntity().getBayById(bn).getUnused() + ")";
```
IllianiBird: Is this something that would benefit from i18n?

IllianiBird (approved):
I have some feedback and critiques but nothing that should stop this from being merged. I leave it to your discretion. It mostly involves my petty feud against single letter and abbreviated class, method, and variable names.

megamek/src/megamek/client/ui/panels/phaseDisplay/MovementDisplay.java
Sleet01: This is what I get for moving stuff inside braces :)

These are all good suggestions.
I didn't "touch" this code so much as "make the IDE move this code", so I wasn't even looking at much of this.
I'll take a stab at cleaning it up.

megamek/src/megamek/client/ui/panels/phaseDisplay/MovementDisplay.java
Sleet01: Amazingly, this is the most concise way to convert iterate over a collection of integers and convert them to strings for concatenation, without writing a helper function or completely changing the code.

megamek/src/megamek/client/ui/panels/phaseDisplay/MovementDisplay.java
Sleet01: I'm going to pass on creating the very first i18n strings for MovementDisplay in _this_ PR, although it does need to be done at some point.

PR #7605, opened 2025-11-11
Fix burst-fire MG menu incorrectly appearing for battle armor

### LINKED ISSUE #5202: [50.03] Burst-fire MGs incorrectly available in lobby context menu for battle armor
Reproduction steps: in a fresh copy of 49.18, start a game, change the Game Options to allow TacOps Burst Fire MGs, then add any BA with MGs to the lobby (I used IS Standard [MG]). The option to set their MGs to burst-fire mode will incorrectly be available in the right-click context menu:
image

Per errata, this option should not be available:
image

Using the option in the context menu does not appear to actually affect the weapon behavior in combat, but displaying the option is unexpected when it has no effect.

HammerGS:
Fixes #5202

 Battle armor units were incorrectly displaying a burst-fire machine gun option in the lobby context menu, despite errata prohibiting this feature for BA weapons.

Root cause: The hasRapidFireMG() and hasNormalFireMG() methods checked  all entities without excluding battle armor, causing the Equipment menu to appear when it shouldn't.

Fix: Add battle armor checks at the start of both methods to return false immediately, preventing BA units from triggering the burst-fire menu options.

Changes:
  - LobbyMekPopup.java: Add BattleArmor import
  - hasRapidFireMG(): Return false for battle armor entities
  - hasNormalFireMG(): Return false for battle armor entities

The burst-fire option now correctly appears only for non-BA units with MGs  when the TacOps Burst Fire MGs option is enabled.

Hammer Notes - Test in lobby and in game and no issues spotted.

### DIFF
--- a/megamek/src/megamek/client/ui/panels/phaseDisplay/lobby/LobbyMekPopup.java
+++ b/megamek/src/megamek/client/ui/panels/phaseDisplay/lobby/LobbyMekPopup.java
@@ -64,6 +64,7 @@ import megamek.client.ui.util.MenuScroller;
 import megamek.client.ui.util.ScalingPopup;
 import megamek.client.ui.util.UIUtil;
 import megamek.common.Player;
+import megamek.common.battleArmor.BattleArmor;
 import megamek.common.battleArmor.ProtoMekClampMount;
 import megamek.common.bays.Bay;
 import megamek.common.equipment.AmmoType;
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/client/ui/panels/phaseDisplay/lobby/LobbyMekPopup.java
```diff
@@ -1004,6 +1009,10 @@ private static boolean hasRapidFireMG(Entity entity) {
 
     /** Returns true when the entity has an MG set to normal (non-rapid) fire. */
     private static boolean hasNormalFireMG(Entity entity) {
+        // Battle armor cannot use burst-fire MGs per errata
```
SJuliez, line 1012: When wanting to be reeeeeal nitpicky, one might want to update the javadoc on the method by adding "*and* this entity is allowed to change it to rapid fire (i.e., not a BA)"

PR #7617, opened 2025-11-12
Adding override to Streak LRMs to prevent indirect mode

TenkawaBC:
Streak LRMs could select indirect mode (but the ammo would prevent usage). This adds an override of the LRM class that added the indirect mode, and runs a removal of it to be sure, as clearing the modes in the constructor did not prevent it getting added later.

### DIFF
--- a/megamek/src/megamek/common/weapons/lrms/StreakLRMWeapon.java
+++ b/megamek/src/megamek/common/weapons/lrms/StreakLRMWeapon.java
@@ -52,7 +52,6 @@ import megamek.common.equipment.Mounted;
 import megamek.common.game.Game;
 import megamek.common.loaders.EntityLoadingException;
 import megamek.common.options.IGameOptions;
-import megamek.common.options.OptionsConstants;
 import megamek.common.units.Entity;
 import megamek.common.weapons.handlers.AttackHandler;
 import megamek.common.weapons.handlers.lrm.StreakLRMHandler;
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/weapons/lrms/StreakLRMWeapon.java
```diff
@@ -87,6 +89,17 @@ public double getTonnage(Entity entity, int location, double size) {
...
 
+    @Override
+    // Streak LRMs could end up with an Indirect mode as a result of this game option selection. Overriding to 
+    // prevent this.
+    public void adaptToGameOptions(IGameOptions gameOptions) {
+        super.adaptToGameOptions(gameOptions);
+
+        // Make sure Indirect Fire is not present for Streak LRMs
+        removeMode("");
+        removeMode("Indirect");
```
IllianiBird: Is there a Constant we can use here, instead of a magic string?

TenkawaBC: Potentially calling clearModes() (EquipmentType), which should clear everything. I was just going off the code used in the parent call

TenkawaBC: Latest commit has this change.

IllianiBird (approved):
I'm not super familiar with this corner of the codebase, but the change is small enough that we can reverse it in the event there is an issue.

megamek/src/megamek/common/weapons/lrms/StreakLRMWeapon.java
```diff
@@ -51,6 +51,8 @@
 import megamek.common.equipment.Mounted;
 import megamek.common.game.Game;
 import megamek.common.loaders.EntityLoadingException;
+import megamek.common.options.IGameOptions;
+import megamek.common.options.OptionsConstants;
```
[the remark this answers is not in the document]
TenkawaBC: fixed in 1c7a940

HammerGS:
@TenkawaBC Friendly reminder to include a close or fix comment PR notes so it auto closes the issues associated.

PR #7619, opened 2025-11-13
Sentry: Added Sanity Check Around `Compute.isPilotingSkillNeeded()`

IllianiBird:
[Sentry Report](https://sentry.tapenvy.us/organizations/tapenvyus/issues/8014/events/dea7e288f0af43b5b5018df120851b16/)

`Compute.isPilotingSkillNeeded()` throws an error if it can't find the entity that's being referenced. However, there are valid reasons why this might occur - such as the unit no longer exists in the game.

This PR just adds a sanity check so that if it can't find the entity in game, it will check to see if it can find it out of game. Only if it can't find it out of game will it return an error.

### DIFF
--- a/megamek/src/megamek/common/compute/Compute.java
+++ b/megamek/src/megamek/common/compute/Compute.java
@@ -67,9 +67,6 @@ import megamek.common.weapons.artillery.ArtilleryCannonWeapon;
 import megamek.common.weapons.attacks.DiveBombAttack;
 import megamek.common.weapons.attacks.InfantryAttack;
 import megamek.common.weapons.autoCannons.ACWeapon;
-import megamek.common.weapons.autoCannons.HVACWeapon;
-import megamek.common.weapons.autoCannons.LACWeapon;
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/compute/Compute.java
```diff
@@ -658,11 +655,16 @@ public static boolean isPilotingSkillNeeded(Game game, int entityId,
...
         final Entity entity = game.getEntity(entityId);
         if (entity == null) {
-            throw new IllegalArgumentException("Entity invalid. ID " + entityId);
+            if (game.getEntityFromAllSources(entityId) == null) {
+                // We have no recollection of the entity anywhere. At this point an error will be thrown.
+                throw new IllegalArgumentException("Entity invalid. ID " + entityId);
+            }
+
+            // Otherwise, it's likely the unit has been destroyed prior to this point.
+            return false;
```
IllianiBird, line 667: Here's the only changes of substance, the rest is auto formatter.

Sleet01 (approved):
LGTM

PR #7628, opened 2025-11-14
Fix 7599: save game broken exiting simult deploy

### LINKED ISSUE #7599: [Issue] Nightly 11-9 - Loading saved megamek mission after crash (from briefing room) breaks
### Brief Description *

I had a mission in megamek go nonresponsive on physical attack phase of turn 12. Saved upon exiting. Went to reload (as I usually do after freezes of this nature) and the save doesn't load properly. Dumps me in to lobby with all of my units missing from list. 

### 3. Steps to Reproduce

as above

### Attach Files

[Autosave-5-The Smoke and Mirrors-30681222.cpnx.gz](https://github.com/user-attachments/files/23441994/Autosave-5-The.Smoke.and.Mirrors-30681222.cpnx.gz)

Round 12 autosave
[Round-12-autosave_2025-11-09_06-56-00.sav.gz](https://github.com/user-attachments/files/23441996/Round-12-autosave_2025-11-09_06-56-00.sav.gz)

The temp save I made when exiting after crash
[temp.sav.gz](https://github.com/user-attachments/files/23441999/temp.sav.gz)

Custom units
[customs.zip](https://github.com/user-attachments/files/23442008/customs.zip)

### Severity *

### MekHQ Suite Version *

50.10 nightly 11/9

### Operating System *

Windows 11

### Java Version *

Adoptium 17

### Final Verification

Sleet01:
This PR prevents two race conditions that could lead to the corruption of save games in specific situations:
1. Simultaneous Deploy / Attack options are enabled;
2. Round is in the End Phase;
3. Player exits the client via "Quit" or the [X] button and chooses to save the game;
4. Network or client computer is "slow".

First, by adding `GAME_LOCK.lock() / .unlock()` calls around game saving and in-game packet handling;
second, by using `SwingUtilities.invokeLater()` to reschedule `die()` handling until after packet handling if the client is expecting a savegame file back from the Server.

A brief note on the labyrinthine process of making a local save game file:
One might expect that a client makes a save game file based on its locally-maintained Game instance, bypassing any network traffic, but this is not the case!
We implement information hiding (from Double Blind or environmental effects) by selectively filtering out data sent to each client during updates, so a given client may not be privy to all Game state.
The process of making a local savegame file is thus:
1. Client signals desire for a savegame file via a Chat packet to the server, containing the `/localsave` command and the local path where the game should be saved on the client.
2. The server receives and processes this packet; if game options allow, it will send a savegame data packet back to that client.
3. The client receives the savegame packet and dumps its data contents to a new file on its local filesystem.
4. The server then sends a chat message to the client, telling the player that the savegame file has been sent to them.

If the client disconnects between 1 and 2, the server does not send the savegame packet.
If the client disconnects between 2 and 3, it will not process the savegame packet.
If the client dies (kills all threads, closes all connections, and exits) _during_ 3, it may not fully write the savegame file.

The two race conditions each can contribute to data corruption in different ways:

A. Simultaneous options allow Server to accept incoming packets that modify the Game state while in the process of packaging the savegame file.  
In-game entities may be removed (by fleeing or dying), map information may change, and the player count can be updated (e.g. by the quitting player's Disconnect packet being processed by the Game).
This can cause numerous errors in the savegame file such as missing or malformed entities, equipment, player, or map data.

B. ClientGUI may kill the Client while the Client is in the process of saving the savegame file.
I've only seen this while debugging but it's conceivable it could occur on a sufficiently languorous machine.
The end result is a savegame file that is truncated or empty, and does not parse as valid XML.

Testing:
- Quit and saved game at various phases using the endturntest2.sav.gz game provided by OP.  Reloading works on all.

Fix #7599

### DIFF
--- a/megamek/src/megamek/client/AbstractClient.java
+++ b/megamek/src/megamek/client/AbstractClient.java
@@ -69,9 +69,6 @@ import megamek.common.preference.PreferenceManager;
 import megamek.common.units.Entity;
 import megamek.common.units.UnitNameTracker;
 import megamek.logging.MMLogger;
-import megamek.server.ConnectionHandler;
-
-import static java.lang.Thread.sleep;
 
 /**
  * AbstractClient that handles basic client features.
--- a/megamek/src/megamek/client/ui/clientGUI/ClientGUI.java
+++ b/megamek/src/megamek/client/ui/clientGUI/ClientGUI.java
@@ -182,8 +182,6 @@ import megamek.common.util.StringUtil;
 import megamek.common.weapons.handlers.WeaponOrderHandler;
 import megamek.logging.MMLogger;
 
-import static java.lang.Thread.sleep;
-
 public class ClientGUI extends AbstractClientGUI
       implements BoardViewListener, ActionListener, IPreferenceChangeListener, MekDisplayListener, ILocalBots,
                  IDisconnectSilently, IHasUnitDisplay, IHasBoardView, IHasMenuBar, IHasCurrentPanel {
--- a/megamek/src/megamek/server/Server.java
+++ b/megamek/src/megamek/server/Server.java
@@ -233,7 +233,7 @@ public class Server implements Runnable {
          */
         @Override
         public void disconnected(DisconnectedEvent e) {
-            // Only do this if nobody is trying to sav
+            // Only do this if nobody is trying to save a game currently
             GAME_LOCK.lock();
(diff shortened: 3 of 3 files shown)


Sleet01:
AI comments look rational so I'll fix those up.

IllianiBird (approved):

PR #7648, opened 2025-11-19
ACAR complex force structure handling

Beornwulf:
Fixes https://github.com/MegaMek/mekhq/issues/8121

Previously, units that were in forces that had subforces were being ignored in ACAR simulations. This fix should rectify that.


Beornwulf:
@Sleet01 to clarify, the subforces weren't getting ignored, the units which were directly in a force that also had subforces were being ignored. I couldn't see any particular reason why they would be, other than "not thinking people might have units in non-leaf nodes".

PR #7656, opened 2025-11-22
Task: De-escalated Error Level of Incorrect Flag Check

IllianiBird:
We've been seeing a decent amount of user contact regarding incorrect flag checks. As these are benign errors I got permission from Hammer to de-escalate them from WARN to debug. This means the logs are still there in case anyone shows an interest in this, but they're not spamming up user logs and leading to contact.


IllianiBird:
No longer needed. Closing.

PR #7660, opened 2025-11-23
Cargo including units in Lift Hoists should always take damage if the unit with the lift hoist is damaged

psikomonkie:
The more I work on `HandheldWeapon`s, the more I reconsider its implementation - I'm starting to feel like when it's equipped it should add its `Mounted` so the carrying unit, this'd eliminate a lot of the weird code I added to support HHW in the first place. Anyway, that's outside the scope of this PR, just the reasoning for some TODOs. 

The `damageHandheldWeapon` method is partially implemented and wasn't meant to be part of this PR, but it was throwing errors without having something there; Glad I caught it I guess.


megamek/src/megamek/server/totalWarfare/TWDamageManagerModular.java
```diff
@@ -1105,6 +1107,35 @@ public void damageMek(Vector<Report> reportVec, Mek mek, HitData hit, int damage
         }
     }
 
+    private void damageCargo(Vector<Report> reportVec, Entity entity, ICarryable cargo, int damage, int entityId) {
```
Sleet01, line 1110: This looks really good, thanks for splitting the code out!

Issue #7644, opened 2025-11-18
[Issue] Right-click "View" shows unrevealed hidden units


gsparks3:
### Brief Description *

While testing #7642, I found _yet another_ issue. Right-clicking on a hex containing an unrevealed hidden enemy unit will allow you to "View" said unit, rather spoiling the surprise. 

<img width="600" height="335" alt="Image" src="https://github.com/user-attachments/assets/fa5943b4-889b-4c2d-bcc2-9fee5f0b3ccf" />

### Attach Files

[hidden_unit_test.sav.gz](https://github.com/user-attachments/files/23596105/hidden_unit_test.sav.gz)

### Severity *

Medium (Gameplay Limitation): Non-core functionality is impaired, providing a suboptimal but playable experience.

### MegaMek Suite Version *

0.50.07

### Operating System *

Windows 10

### Java Version *

Java 17

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have included all necessary information and files to help reproduce this issue
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

Sleet01:
As expected, we were not filtering out hidden units here.

Issue #7643, opened 2025-11-18
[Issue] Pathfinder is aware of unrevealed hidden units


gsparks3:
### Brief Description *

While testing #7642, I noticed another issue: the pathfinder knows if there's an enemy unit in the hex that would prevent you from ending your movement in that hex. This is most obvious with jumping movement, where the hex with a hidden mech in it appears as a "hole" in the available jump destinations for a jumping mech:

<img width="685" height="520" alt="Image" src="https://github.com/user-attachments/assets/ef0445fd-9500-476b-9a5e-3c4fd40d6c2b" />

However, it also applies to plotted ground movement:
<img width="653" height="447" alt="Image" src="https://github.com/user-attachments/assets/024f8942-ce9c-4474-8077-c2ad37c67fff" />

### 3. Steps to Reproduce

1. Load the provided save as "Test" in Megamek 50.07.
2. Connect a second client as "Test2".
3. Experiment with moving the units as "Test2" to demonstrate the issues described above.

### Attach Files

[hidden_unit_test.sav.gz](https://github.com/user-attachments/files/23596057/hidden_unit_test.sav.gz)

### Severity *

Medium (Gameplay Limitation): Non-core functionality is impaired, providing a suboptimal but playable experience.

### MegaMek Suite Version *

0.50.07

### Operating System *

Windows 10

### Java Version *

     Java Vendor: Eclipse Adoptium     Java Version: 17.0.11

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have included all necessary information and files to help reproduce this issue
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

Sleet01:
We actually need to plumb an "ignore hidden units" flag into a bunch of code, mostly involved with detecting stacking violations, for this to work.
On the one hand, I have the jump display working correctly.
On the other hand, I don't know why the walk/run display is still detecting hidden units.

Sleet01:
Multiple redundant checks for illegal stacking violations hidden deep in the move path compiler.

kuronekochomusuke:
these also may be related 

#413
#541
#3033
#7396

Issue #7642, opened 2025-11-18
[Issue] Point blank shots triggering when they shouldn't


gsparks3:
### Brief Description *

Rules for hidden units and point-blank shots from hidden units are found on page 259-260 of Total Warfare. In particular, it is called out that a point-blank shot can occur when a hidden unit would be revealed. Currently, in Megamek, point-blank shots will trigger in cases that would not normally reveal the hidden unit (although it _will_ be revealed if it chooses to take the shot).

A saved game is provided to demonstrate several instances of this. In this setup, we have a hidden unit in hex 0613.
<img width="510" height="406" alt="Image" src="https://github.com/user-attachments/assets/9af5c237-0eed-4fdf-a030-041dab961068" />

### Case 1
A unit that moves _past_ the hidden unit via ground movement, but does not end adjacent to the hidden unit, should not reveal the hidden unit or trigger a PBS. An example of such movement is shown below. Testing will show that this _does_, incorrectly, trigger a PBS.
<img width="684" height="478" alt="Image" src="https://github.com/user-attachments/assets/b83febeb-b3dd-46c5-83bb-9edb4a96d6a5" />

### Case 2
Similarly, a jumping unit with part of its movement path adjacent to a hidden unit will currently (incorrectly) trigger a PBS. An example of such movement is shown below. Jumping should not reveal hidden units even if the jumping unit passes through their hex partway through the movement; it should only check at the end of movement.
<img width="668" height="588" alt="Image" src="https://github.com/user-attachments/assets/da0f3628-e172-4fe9-93bb-dabcdb7b6a61" />

### Case 3
A [recent forum ruling](https://battletech.com/forums/index.php?topic=84054) stated that airborne ground units should never reveal hidden units while moving (at least without dedicated equipment such as an Active Probe). Currently, they trigger PBS just as non-airborne units do, with an example of such movement shown below. Unlike ground units, this lack of PBS should apply even if they move through the hex of a hidden unit, as long as they are airborne at a sufficient elevation to not collide with said hidden unit.
<img width="393" height="722" alt="Image" src="https://github.com/user-attachments/assets/41ad0a82-16ea-468f-b6aa-150f0541a0c3" />

### Case 4
[...]

Issue #7396, opened 2025-09-05
Hidden Deployment and Movement


strategoslevel3:
### Brief Description *

0.50.6

A jump 5 mech, jumped 5 hexes into a hex containing some BattleArmor in Hidden Deployment.

The BattleArmor got the opportunity to make a point blank attack (as per TW) BUT THEN the Mech got to make an additional walk movement with its full MP - this seems wrong. Opinions?

### Severity *

Critical (Game-breaking/Crash): The game crashes or a core feature (like saving, loading, or network connection) is completely unusable.

### MegaMek Suite Version *

0.50.6

### Operating System *

IOS Sequoia 

### Java Version *

17

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have included all necessary information and files to help reproduce this issue
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

PhoenixHeart512:
Probably related to https://github.com/MegaMek/megamek/issues/541

PR #7662, opened 2025-11-23
Fix 7642-7644: hidden units visible, Pointblank Shots not working properly, Aero non-BAP range was 1

Sleet01:
This fixes a swath of reported issues in Hidden units / Pointblank Shots:
1. Pathfinder is aware of hidden units, causing hexes containing hidden units to be avoided / un-highlighted as possible move positions;
2. Hidden units are viewable by right-clicking on a hex containing them and selecting 'View';
3. ASFs always get a 1-radius detection range, even if they don't have BAP;
4. Hidden unit owner is not given a Pointblank Shot prompt if the moving unit tries to enter the hidden unit's hex;
5. Hidden unit owner _is_ prompted to take a Pointblank Shot if the moving unit passes next to them but does not stop;
6. Moving units hit by PBS may, in some cases, be granted an unwarranted continuation of their movement, or even a full new turn's worth of MP/thrust points.
7. VTOLs detect Hidden units like ASFs, and take PBS as if at range 1 even if stopping movement directly above the Hidden unit (in this case there should be no PBS unless the hidden unit can attack in the same hex, like Infantry);
8. A unit jumping into a hex containing a Hidden unit should cause a domino effect with concomitant PSRs for both units.
9. Some PBS reports were doubled.

This PR fixes all these issues, adds correct handling of stacking violations (both moving and jumping), prevents extra turns being granted, and cleans up reporting.

As part of this fix, I replaced the Enumeration<MoveStep> object (an ancient configuration) with a ListIterator<MoveStep>; this has zero effect on legacy code as Iterator provides all functionality of Enumeration, it just simplifies keeping the previous step while handling PBS and domino effect setup here.

Fix MegaMek/megamek#413
Fix MegaMek/megamek#541
Fix MegaMek/megamek#3033
Fix MegaMek/megamek#7396 
Fix MegaMek/megamek#7642
Fix MegaMek/megamek#7643
Fix MegaMek/megamek#7644

NOTE:
I still think we should move all Double Blind + Hidden unit filtering into the same calls, remove all unfiltered calls, and rip out the hacky 'ignoreHidden' boolean I had to add to get this all working correctly.
That would take about this much work again, though, so not right now.

Testing:
- Ran various scenarios from OP issues, both with Bot and joined as second player

### DIFF
--- a/megamek/src/megamek/server/totalWarfare/MovePathHandler.java
+++ b/megamek/src/megamek/server/totalWarfare/MovePathHandler.java
@@ -1434,8 +1434,8 @@ class MovePathHandler extends AbstractTWRuleHandler {
                         // and the revealed hidden unit has not already made a pointblank shot this turn.
                     } else if (
                           (dist <= 1) && !hiddenEntity.madePointblankShot() &&
-                          ((!this.entity.isAero() && md.isEndStep(step)) ||
-                                (this.entity.isAero() && (dist == ((this.entity.getBAPRange() > 0) ? 1 : 0))))
+                          ((!this.entity.isAirborne() && md.isEndStep(step)) ||
+                                (this.entity.isAirborne() && (dist == ((this.entity.getBAPRange() > 0) ? 1 : 0))))
                     ) {
                         // Hidden unit should always be revealed as the PBS trigger _is_ getting revealed.
                         hiddenEntity.setHidden(false);
@@ -3652,6 +3652,12 @@ class MovePathHandler extends AbstractTWRuleHandler {
 
     }
 
+    /**
+     * Wrapper for processPointblankShotCFR with packet error handling and consolidated reports
+     * @param step          MoveStep to prompt for a PBS
+     * @param hiddenEntity  Candidate to fire PBS
+     * @return  Vector<Report> collection of reports; caller responsible for displaying these.
+     */
     protected Vector<Report> processPossiblePBS(MoveStep step, Entity hiddenEntity) {
         Vector<Report> pbsReports = new Vector<>();
         Vector<Report> attackReports = new Vector<>();
--- a/megamek/src/megamek/client/bot/princess/PathRanker.java
+++ b/megamek/src/megamek/client/bot/princess/PathRanker.java
@@ -41,7 +41,6 @@ import java.math.MathContext;
 import java.math.RoundingMode;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.Enumeration;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Optional;
--- a/megamek/src/megamek/client/ui/SharedUtility.java
+++ b/megamek/src/megamek/client/ui/SharedUtility.java
@@ -35,7 +35,6 @@
 package megamek.client.ui;
 
 import java.util.ArrayList;
-import java.util.Enumeration;
 import java.util.List;
 import java.util.ListIterator;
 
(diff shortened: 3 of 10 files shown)


Sleet01:
Double-rebased because my home repo ended up temporarily behind.

megamek/src/megamek/server/totalWarfare/MovePathHandler.java
```diff
@@ -1336,101 +1336,136 @@
...
+                    if ((dist == 0) && !continueTurnFromPBS &&
+                          (Compute.stackingViolation(getGame(), this.entity,
+                              step.getPosition(), null, this.entity.climbMode(), false
+                          ) != null)
+                    ) {
+                        // Attempting to move into hex of a hidden unit detects the unit
+                        hiddenEntity.setHidden(false);
+
+                        // Set location per previous step; this prevents destroyed entities appearing at move start loc.
+                        this.entity.setPosition(previousStep.getPosition());
```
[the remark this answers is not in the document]
Sleet01: Grr, fair.

megamek/src/megamek/server/totalWarfare/MovePathHandler.java
```diff
@@ -1336,101 +1336,136 @@ private void positionOnAtmosphericMap() {
...
+
+                        // Potential point-blank shot when not causing stacking violation, but only in some situations:
+                        // 1. mover is ground unit _and_ ends its movement adjacent to / in the hidden unit's hex;
+                        // 2. mover is Aerospace and hidden unit is within detection range of its flight path
+                        //    (with or without Active Probe).
+                        // and the revealed hidden unit has not already made a pointblank shot this turn.
+                    } else if (
+                          (dist <= 1) && !hiddenEntity.madePointblankShot() &&
+                          ((!this.entity.isAero() && md.isEndStep(step)) ||
+                                (this.entity.isAero() && (dist == ((this.entity.getBAPRange() > 0) ? 1 : 0))))
```
psikomonkie: An Aero can be taxiing around the ground and would return true for `isAero()` but we probably still want to consider it. Check out `isAirborneAeroOnGroundMap()` (or maybe `isAirborne()`).

Sleet01: I guess we can replace `isAero()` with `isAirborne()`, then taxiing aeros would be treated like other ground units and only detect on their end step...

megamek/src/megamek/server/totalWarfare/MovePathHandler.java
```diff
@@ -1336,101 +1336,140 @@ private void positionOnAtmosphericMap() {
...
                         return;
-                        // Potential point-blank shot
-                    } else if ((dist == 1) && !entity.madePointblankShot()) {
+
+
+                        // Potential point-blank shot when not causing stacking violation, but only in some situations:
+                        // 1. mover is ground unit _and_ ends its movement adjacent to / in the hidden unit's hex;
+                        // 2. mover is Aerospace and hidden unit is within detection range of its flight path
+                        //    (with or without Active Probe).
+                        // and the revealed hidden unit has not already made a pointblank shot this turn.
```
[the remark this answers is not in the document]
Sleet01: Stylistic choice held over from precious code.

megamek/src/megamek/server/totalWarfare/MovePathHandler.java
Sleet01: Works good, tested with flying and taxiing ASFs.

PR #7672, opened 2025-11-25
Two minor 0.50.10 fixes

Sleet01:
Fix one last glancing blow report format and make hidden units get revealed by AE hits.

1. There's a separate Glancing Blow string, for N/LP, that needed the explicit trailing space added.

2. Separately, I noticed that hidden units are not revealed by AE damage - contravening Total Warfare Pg. 259.  Since AEs are about the fastest way to root out Hidden units (after BAP), I figured this should get fixed.

Testing:
- Ran a big bot vs. bot fight with AE, hidden units, N/LP meks, the works.

### DIFF
--- a/megamek/resources/megamek/common/report-messages.properties
+++ b/megamek/resources/megamek/common/report-messages.properties
@@ -379,6 +379,7 @@
 3185=<newline>    (continuing hit report):
 # This trailing white space is important and shouldn't be removed
 3186=\ - Glancing Blow -\u0020
+# This trailing white space is important and shouldn't be removed
 9985=\ - Glancing Blow due to Narrow/Low Profile -\u0020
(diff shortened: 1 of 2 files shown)


megamek/resources/megamek/common/report-messages.properties
```diff
@@ -379,7 +379,7 @@
 3185=<newline>    (continuing hit report):
 # This trailing white space is important and shouldn't be removed
 3186=\ - Glancing Blow -\u0020
-9985=\ - Glancing Blow due to Narrow/Low Profile -
+9985=\ - Glancing Blow due to Narrow/Low Profile -\u0020
```
IllianiBird, line 383: Can we add the 'this trailing space is important' line above this, too?

I think we'll also need to update the foreign language bundles

PR #7675, opened 2025-11-27
Improvement: Added Two New Utility Methods to Reduce Potential Mistakes When Comparing Planetary Tech Levels

IllianiBird:
## Required by [Fix: Fixed Availability of Prosthetics Based on Planetary Tech Level](https://github.com/MegaMek/mekhq/pull/8362)

The planetary tech level scale is inverted (F->A) instead of the scale used by normal parts and units (A->F). That means the existing comparisons were not suitable when comparing planetary tech levels. This recently caused a bug to enter MekHQ due to the non-obvious nature of this inversion.

This PR adds two new utility methods explicitly for use when comparing planetary tech levels. It also added extensive JavaDocs to further reduce room for error.

### DIFF
--- a/megamek/src/megamek/common/enums/TechRating.java
+++ b/megamek/src/megamek/common/enums/TechRating.java
@@ -72,6 +72,9 @@ public enum TechRating {
      * <p>The comparison is based on the internal {@code index} value; a larger index indicates a more advanced
      * rating.</p>
      *
+     * <p><b>Warning:</b> Do not use when comparing planetary tech levels. Use
+     * {@link #isPlanetaryTechLevelBetterThan(TechRating)} instead.</p>
+     *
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/enums/TechRating.java
```diff
@@ -38,7 +38,12 @@
 
 // --- Tech Rating Enum ---
 public enum TechRating {
-    A(0, "A"), B(1, "B"), C(2, "C"), D(3, "D"), E(4, "E"), F(5, "F");
+    A(0, "A"),
+    B(1, "B"),
+    C(2, "C"),
+    D(3, "D"),
+    E(4, "E"),
+    F(5, "F");
```
IllianiBird, line 46: This was done to make this enum easier to read when using split-view

PR #7693, opened 2025-11-29
Task: Removed No Longer Used Methods

IllianiBird:
Removed two no longer in use methods. This were incorrectly implemented, regardless, because it turns out that planetary tech rating _isn't_ normal tech rating backwards. It's a whole separate thing and we already translate planetary tech _sophistication_ into normal tech rating so these additional methods were unnecessary.


Sleet01 (approved):
LGTM

PR #7694, opened 2025-11-30
Fix #7686 Firing Multiple TAGS.

### LINKED ISSUE #7686: [Issue] Firing multiple TAGs when there are homing munitions (Homing AIV, Copperhead, etc) causes the game to lock up
### Brief Description *

When there are homing artillery munitions in the air (i.e. have been fired) and are landing on the current turn, firing multiple TAGs causes the game to lock up at the end of the TAG round.

### 3. Steps to Reproduce

1. Start up game of Megamek with at least two units equipped with TAGs or C3 Masters and at least one unit equipped with homing artillery munitions.
2. Launch the homing munitions as normal.
3. On the turn that the homing munition will impact the target, hit at least two TAGs on valid targets.
4. Game locks up

### Attach Files

[megamek.zip](https://github.com/user-attachments/files/23829373/megamek.zip)

[MessingAround.sav.gz](https://github.com/user-attachments/files/23829428/MessingAround.sav.gz)

### Severity *

### MegaMek Suite Version *

0.50.10

### Operating System *

Windows 11

### Java Version *

Java Version: 19.0.1

### Final Verification

HammerGS:
Fix #7686 

Full write up in the bug report

  Changes made to TWGameManager.java:

  | Method                        | Lines Changed | Fix                                               |
  |-------------------------------|---------------|---------------------------------------------------|
  | processPointblankShotCFR()    | ~6175, ~6182  | Re-queue mismatched packets instead of discarding |
  | processTeleguidedMissileCFR() | ~6291, ~6297  | Re-queue mismatched packets instead of discarding |
  | processTAGTargetCFR()         | ~6324, ~6330  | Re-queue mismatched packets instead of discarding |

  What the fix does:
  - When a CFR packet arrives that doesn't match the expected type or player ID, it's now added back to the queue with cfrPacketQueue.add(rp) instead of being discarded
  - This allows other waiting handlers to eventually receive their packets
  - Removed the error logging since this is now expected behavior (packets arriving out of order)

  Build status: Successful

  The fix is ready for testing. You'll need to test with the save file from the bug report to verify the lockup no
  longer occurs.

### DIFF
--- a/megamek/src/megamek/common/compute/ComputeECM.java
+++ b/megamek/src/megamek/common/compute/ComputeECM.java
@@ -600,8 +600,9 @@ public class ComputeECM {
                     newInfo = new ECMInfo(6, 0, entity);
                     newInfo.setAngelECMStrength(1);
                 } else if (m.curMode().equals("ECM & ECCM") || m.curMode().equals("ECM & Ghost Targets")) {
+                    // In dual mode, Angel ECM operates as two standard ECM suites,
+                    // losing Angel abilities but gaining simultaneous ECM+ECCM (TO p.100)
                     newInfo = new ECMInfo(6, 1, entity);
-                    // Doesn't count as Angel ECM
                 }
                 // BA Angel ECM has a shorter range
                 if ((newInfo != null) && (entity instanceof BattleArmor)) {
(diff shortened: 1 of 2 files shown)


HammerGS:
image

Sleet01 (approved):
This looks reasonable to me, although I think there's now a possibility of an infinite loop if we get an unexpected packet with bad information.
We may want to keep the logging in cases where we re-queue the packet, but since the existing logging didn't seem to work I'm not sure that's all that helpful...

HammerGS:
The latest change seems to be working. Been raining Arrow IV's on a company of Urbies and can pick and choose the targets at will

HammerGS:
Going to merge, we have player confirmation its working.

Issue #7688, opened 2025-11-29
[Issue] Bridges with non-zero height not walkable


gsparks3:
### Brief Description *

I am seeing consistent issues where units cannot move onto bridges in generated maps with ground movement. Jumping onto bridges seems to work normally, and allow normal movement _off_ the bridge. 

<img width="1061" height="690" alt="Image" src="https://github.com/user-attachments/assets/d8fe5661-f0b2-4a65-a7fc-10a231b52cbf" />
(Also visible is a separate issue where generated bridges across depth 0 water hexes become roads instead. This technically shouldn't be an actual gameplay issue, as nominally such a bridge should work and connect to the road segment, but it does look a little weird when it happens as part of a longer bridge of relatively uniform height. That one is not critical for now.) Jumping onto the bridge leaves me on top of the bridge as expected, and allows me to move across the rest of the bridge without issues.
<img width="763" height="547" alt="Image" src="https://github.com/user-attachments/assets/6e9b60d8-3e0c-45e3-8eeb-e76f112ab995" />

### 3. Steps to Reproduce

1. Load provided savegame.
2. Attempt to walk onto the various bridge segments.
3. Jump onto the bridge, then attempt to walk off (this should work as expected).
4. Optional: generate further maps with "Road: High" and "Lakes: High" and try a few times to get freshly generated maps with bridges across water.

### Attach Files

[generated_bridge_test.sav.gz](https://github.com/user-attachments/files/23834270/generated_bridge_test.sav.gz)

### Severity *

High (Major Disruption): A major feature is broken or incorrect, but a workaround exists.

### MegaMek Suite Version *

0.50.10

### Operating System *

Windows 10

### Java Version *

     Java Vendor: Eclipse Adoptium     Java Version: 17.0.11

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have included all necessary information and files to help reproduce this issue
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

gsparks3:
Some further examples from premade maps: 

The bridges on the 16x17 Archipelago 2 map in the distribution work fine. These are height 0 bridges level with the adjacent non-water terrain.

<img width="390" height="325" alt="Image" src="https://github.com/user-attachments/assets/64cad6d0-90d8-4693-bbbe-31404ace7f41" />

[...]

PR #7701, opened 2025-12-01
Issue #7688: Bridges with Non-Zero Height Not Walkable

HammerGS:
Fix  Issue #7688: Bridges with Non-Zero Height Not Walkable
  
  Units cannot walk onto bridges with non-zero height on generated maps. Jumping onto bridges works, then walking on them
  works. Height-0 bridges work fine.

  Root Cause

  Commit 7819a8e04f (May 16, 2025) by Luana Coppio added building climb logic that inadvertently broke bridge movement. The fix used destHex.floor() for buildings in climbMode, but bridges are stored as buildings in MegaMek (e.g., Hardened Building Bridge #15028). This caused bridge hex elevation to be calculated as water bottom (-3) instead of bridge elevation (0), creating a 3-level change that exceeds allowed movement
  
 This PR is strictly for comparison and future testing.

### DIFF
--- a/megamek/src/megamek/client/ui/clientGUI/ClientGUI.java
+++ b/megamek/src/megamek/client/ui/clientGUI/ClientGUI.java
@@ -2771,10 +2771,9 @@ public class ClientGUI extends AbstractClientGUI
 
         @Override
         public void gameClientFeedbackRequest(GameCFREvent gameCFREvent) {
+            // Note: entity may be null for CFR types that don't use entityId (e.g., TAG_TARGET, TELEGUIDED_TARGET)
+            // Each case handles null checking as appropriate
             Entity entity = client.getGame().getEntity(gameCFREvent.getEntityId());
-            if (entity == null) {
-                return;
-            }
 
             Object result;
             String input;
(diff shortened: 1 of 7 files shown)


psikomonkie (commented):
I was making similar changes. I'd written the tests first, so have some free tests as a treat.

megamek/src/megamek/common/moves/MoveStep.java
```diff
@@ -2396,7 +2396,7 @@ private void compileIllegal(final Game game, final Entity entity, final MoveStep
         }
 
         // do not allow to move onto a bridge if there's no exit in lastPos's
```
psikomonkie, line 2406: We shouldn't allow units to move onto a bridge if there's no exit on a given side, even if above bridge level. I'm updating my unit tests.

EDIT: I've updated the unit tests to reflect how the logic should work.

psikomonkie: This was incorrect, you are allowed to move units onto a bridge even if the exits don't match. TO:AR 114-115

psikomonkie:
Looks like there are two failing tests I added for this, both I think are minor: 
- testMovePathBoardWalkUnderLowBridge
It shouldn't be possible to move under a height 1 bridge as a Mek without going over it

- testMovePathBoardWalkUnderBridgeButTryToClimb
When moving under a very high bridge (Mek at 0, bridge elevation 4) it shouldn't matter if climb mode is ON, this test is failing because right now it will be an illegal move.

psikomonkie:
I think this PR should be closed and we should continue with #7707.

HammerGS:
This PR superceded by https://github.com/MegaMek/megamek/pull/7707

PR #7714, opened 2025-12-05
PR: Do you want to build a Fortress? `BuildingEntity`

psikomonkie:
### What is this PR?
On its own? Nothing, this shouldn't have any game play impact on MegaMek unless the player creates, by hand, a `BuildingEntity`, and then decides to deploy it. Otherwise this is just an enabling PR so work can begin on a UI for designing `BuildingEntity`s in MegaMekLab.

### `Building Entity`
Introducing... `BuildingEntity`! The newest addition to our growing number of in game entities. The `BuildingEntity` follows the TO:AR 116 Advanced Building rules. As an `Entity`, they can be controlled by a player, deployed during deployment, and use weapons to defend itself. The goal for this PR isn't to be a feature complete implementation, it's to be a complete-enough implementation that the `MegaMekLab` crew can start working on the UI for letting players create their own, while the gameplay implementation of the `BuildingEntity` continues concurrently. 

I've got a busy weekend, and I'm eepy, so this might sit in limbo for a week.

To do before passing this off to MML:
- [x] Levels for everything! Mishra kindly confirmed that you can pintle mount a weapon at any level of an external hex of a building. That means locations for every level of every hex of a building, woohoo!
- [x] Some more unit tests!! 
- [x] Ensure copyright is completely correct for `Building`, `BuildingEntity`, `BuildingTerrain`, and `IBuilding`

Other to do:
- See how hard it is to unbreak Princess's brain when a `BuildingEntity` is placed on board, not a dealbreaker as this won't ship with any usable buildings (see #7737)

https://github.com/user-attachments/assets/53da50ca-716a-4c14-a733-fc723a785a15

https://github.com/user-attachments/assets/5dda51f5-a4bd-4fd4-81ad-67fbe1eb973b

### Developer Notes
Some big changes here!
Core logic for buildings that was common between `BuildingTerrain` and `BuildingEntity` has been moved to `Building`. 
`Building` is more than just a refactoring of where the logic lives - it also internally tracks the locations of its hexes differently. `Building` uses `CubeCoords` to store its relative positions. The `BuildingTerrain` and `BuildingEntity` are responsible for mapping those onto the board properly. `BuildingTerrain` and `BuildingEntity` are essentially wrappers for the internal `Building` class.

### DIFF
--- a/megamek/src/megamek/common/internationalization/I18n.java
+++ b/megamek/src/megamek/common/internationalization/I18n.java
@@ -76,11 +76,14 @@ public class I18n {
               throws IOException {
             // The below is one approach; there are multiple ways to do this
             String resourceName = toResourceName(toBundleName(baseName, locale), "properties");
-            try (InputStream is = loader.getResourceAsStream(resourceName)) {
-                assert is != null;
-                try (InputStreamReader isr = new InputStreamReader(is, StandardCharsets.UTF_8)) {
-                    return new PropertyResourceBundle(isr);
-                }
+            InputStream is = loader.getResourceAsStream(resourceName);
+            if (is == null) {
+                return null; // Resource not found, let ResourceBundle try fallback
+            }
+            try (InputStreamReader isr = new InputStreamReader(is, StandardCharsets.UTF_8)) {
+                return new PropertyResourceBundle(isr);
+            } finally {
+                is.close();
             }
         }
     }


megamek/src/megamek/common/units/Entity.java
```diff
@@ -3935,6 +3939,11 @@ public void refreshLocations() {
         for (int i = 0; i < locations(); i++) {
             crits[i] = new CriticalSlot[getNumberOfCriticalSlots(i)];
         }
+        armorType = new int[locations()];
+        armorTechLevel = new int[locations()];
+        hardenedArmorDamaged = new boolean[locations()];
+        locationBlownOff = new boolean[locations()];
+        locationBlownOffThisPhase = new boolean[locations()];
```
psikomonkie: This is why the tests are failing. Figured there was a good reason these weren't already here. Glad the unit tests caught the issues! I'll look into this later™️

psikomonkie:
Ready for review. Don't be intimidated by the +3500 lines, ~1500 of those are unit tests & some files for the tests.

psikomonkie:
Moved back to Draft.

Glad I provided the `blk` template to Raoz, they asked question that now that I'm looking closer at it, it means I should rework the locations of a `BuildingEntity`. It's for the best. 

Right now in this PR,  every floor of every hex is a location, starting at ground floor of origin. However, we need to differentiate the different sides of a hex for side mounted weapons... That's not good. That really balloons the number of locations. And we already aren't using the internal/cf for those locations, we just need places to put equipment. 

I'm going to refactor the locations of a `BuildingEntity` to just be the points of interest - locations that have equipment of some kind that we need to track. And probably a "Base" or something so `BuildingEntity`s with no equipment don't break.

EDIT: JK, I think I can use the various existing turret types, which'd better handle damaging anything in the same floor when hit...

psikomonkie:
Okay, this is ready for review again :) There should be no gameplay impact from this PR, this is just an enabling PR so we can start working on the MML UI while work on gameplay continues.

BLR-IIC:
This looks neat!  I take it that this is a stepping stone towards Bolos- er, mobile structures?  :)

megamek/src/megamek/client/ui/clientGUI/boardview/sprite/EntitySprite.java
```diff
@@ -542,7 +543,7 @@ public void prepare() {
         boolean isTank = (entity instanceof Tank);
         boolean isInfantry = (entity instanceof Infantry);
         boolean isAero = entity.isAero();
-        boolean isGunEmplacement = entity instanceof GunEmplacement;
+        boolean isGunEmplacement = entity instanceof GunEmplacement || entity instanceof HandheldWeapon || entity instanceof BuildingEntity;
```
Sleet01: Should probably rename this boolean, maybe something like "isStaticEntity"?

megamek/src/megamek/common/board/Board.java
```diff
@@ -2146,4 +2128,19 @@ public BoardType getBoardType() {
     public String toString() {
         return "[Board-%s] (%s) %dx%d".formatted(boardType, mapName, width, height);
     }
+
+    /**
+     * Add a building and all of its coordinates to the board. {@link BuildingTerrain} should be added when
+     * initializing, this method is public so {@link megamek.common.units.BuildingEntity} can register buildings when deploying buildings.
+     * @param bldg {@link IBuilding} to add to the board
+     */
+    public void addBuildingToBoard(IBuilding bldg) {
```
Sleet01, line 2138: This is a nice optimization.

megamek/src/megamek/common/loaders/BLKFile.java
```diff
@@ -296,13 +301,15 @@ protected void loadEquipment(Entity t, String sName, int nLoc) throws EntityLoad
                               isTurreted,
                               isPintleTurreted,
                               isOmniMounted);
-                        // Need to set facing for VGLs
+                        // Need to set facing for VGLs & BuildingEntitys
```
Sleet01: I think "BuildingEntities" is fine, or "BuildingEntity instances".

megamek/src/megamek/common/units/BuildingEntity.java
```diff
@@ -0,0 +1,803 @@
...
+            }
+        }
+    }
+
+    // FIXME: IDK if this is right, just needed something to pass tests
+    private static final TechAdvancement TA_BUILDING_ENTITY = new TechAdvancement(TechBase.ALL)
+          .setAdvancement(DATE_PS, DATE_PS, DATE_PS)
+          .setTechRating(TechRating.B)
+          .setAvailability(AvailabilityValue.A, AvailabilityValue.A, AvailabilityValue.A, AvailabilityValue.A)
+          .setStaticTechLevel(SimpleTechLevel.ADVANCED);
```
Sleet01: I believe this is because every Entity implements ITechnology so we have a common interface to make an entity total up its internals and tell us its overall rating and availability.

If we _have_ any TA data on mobile structures or advanced buildings, we should try to match that data, but I'm guessing it will only be important for the various subclasses - hardened buildings have existed since pre-industrial times, but within CBT we probably only care about TA for mobile structures, naval vessels, Lighter Than Air craft, etc.

megamek/src/megamek/common/units/BuildingEntity.java
```diff
@@ -0,0 +1,803 @@
...
+    }
+
+    @Override
+    public String[] getLocationNames() {
+        ArrayList<String> locationNames = new ArrayList<String>();
+        if (getInternalBuilding() == null || getInternalBuilding().getCoordsList() == null) {
+            return new String[] { LOCATION_NAMES_PREFIX + ' ' + LOC_BASE };
+        }
+        for (CubeCoords coord : getInternalBuilding().getCoordsList()) {
+            // Do we start with "1st floor" or "0 floor?" This is "1st floor"
```
Sleet01: I believe it's matched with level / elevation, so the ground floor should be level 0 / elevation 0; the next floor up should be level 1 / elevation 1, etc.

megamek/src/megamek/common/units/BuildingEntity.java
```diff
@@ -0,0 +1,803 @@
...
+    /**
+     * Returns the Rules.ARC that the weapon, specified by number, fires into.
+     *
+     * @param weaponNumber integer equipment number, index from equipment list
+     *
+     * @return arc the specified weapon is in
+     */
+    @Override
+    public int getWeaponArc(int weaponNumber) {
+        return 0;
```
Sleet01: Is this correct?

megamek/src/megamek/common/units/BuildingEntity.java
psikomonkie: Argh, okay, let me refactor some stuff.

megamek/src/megamek/common/units/BuildingEntity.java
psikomonkie: No but yes! This will not function, but gameplay functionality isn't part of this PR. This'll be implemented in #7737. It will be something like: 
```
@Override
public int getWeaponArc(int weaponNumber) {
    WeaponMounted weapon = getWeapon(weaponNumber);
    if (weapon.isTurret()) {
        return 0;
    }
    switch(weapon.getFacing()) {
        case 0:
            return 1;
        case 1:
            return 50;
        case 2:
            return 51;
        case 3:
            return 52;
        case 4:
            return 53;
        case 5:
            return 54;
        default: return 0;
    }
}
```

megamek/src/megamek/common/units/BuildingEntity.java
Sleet01: Before you start changing stuff up, try just looking at some scenarios from TW and see if your implementation _does_ the right thing.
If so, maybe nothing needs to change here.

psikomonkie: Oh, I've already looked, I know it doesn't do what we want...

PR #7726, opened 2025-12-06
Issue #7725: Equipment flag checks shouldn't be `if else` because equipment can have multiple flags

### LINKED ISSUE #7725: [Issue] Nova CEWS units can't be detected by the RAT generator/formation builder
### Brief Description *

When trying to generate Nova CEWS formations using either the formation builder or RAT generator, Megamek can't find any units that have it equipped. This is caused by ModelRecord checking if something has Nova on [line 683](https://github.com/MegaMek/megamek/blob/main/megamek/src/megamek/client/ratgenerator/ModelRecord.java#L683) as part of an else-if chain, with Nova also counting as ECM/BAP, which means the if statement that starts the chain on 657 returns true early and it never checks if Nova. Not sure of the best fix.
Image

### 3. Steps to Reproduce

1. Go into formation builder
2. Set year to 3072 and faction to The Society (so that Nova CEWS mechs are theoretically available), and Network to Nova CEWS
3. Click analyse and observe that there's 0 units that count as having Nova CEWS regardless of settings elsewhere

### Attach Files

_No response_

### Severity *

### MegaMek Suite Version *

0.50.10

### Operating System *

Window 11

### Java Version *

Java Vendor: Eclipse Adoptium, Java Version: 17.0.16

### Final Verification

psikomonkie:
Fixes #7725 

I'm only like 75% sure this is right, ideally one of the MegaMekLab side peoples could chime in, I feel like this is their jam.

When using the formation builder, there's an option to try and use all network units. Use that, it'll make testing quicker. If you, without this fix, use that functionality with NOVA CEWS, none of the meks will be Nova CEWS. With this fix, it works as expected.


SuperStucco:
Endless if's isn't my first choice, because most of those are never going to be overlapping flags - might as well stop checking for most of those.  However, it has grown to be more complex and the brute force approach will do until it can be properly refactored into something more elegant.

SuperStucco (approved):

PR #7729, opened 2025-12-06
Fix support vehicle engine weight rounding to use nearest half-ton

### LINKED ISSUE #7350: [Issue] Support Vehicle Engine Weight Rounding Improperly
### Brief Description *

Currently, support vehicle engine weights appear to be rounding up instead of rounding normally.

Tested with 5t 5/8 WiGe ICE with D equipment.  The math puts the correct engine weight before rounding at 1.0857t and the tech manual says "Determining the weight of a Support Vehicle’s engine is simply a matter of multiplying the vehicle’s total weight by its Base Engine Weight (a base multiplier of the vehicle’s total weight based on its size and motive type), Movement Factor (a value derived from the unit’s desired Cruise/Safe Thrust) and Engine Weight Multiplier (a value based on the type and Tech Rating of the engine chosen). This final weight (rounded to the nearest kilogram for Small Support Vehicles, **and to the nearest half-ton for those weighing 5 tons and up)** includes the engine, transmission and any integral cooling systems and shielding necessary to operate the engine." meaning it should be rounding down to 1t in the given example, but MML (and MM itself) are rounding up to put the engine weight at 1.5t.

Image

### 3. Steps to Reproduce

Make a 5t 5/8 movement support WiGe with D equipment and an ICE engine.
Weight of the engine will be 1.5t instead of the correct 1t weight.
Filling with 2.5t of equipment and 0.5t fuel (as if the engine weighed 1t) will cause the unit to be overweight when loaded into MegaMek (meaning it's probably not just a display issue in MML)

### Attach Files

MML log file empty

### Severity *

### MegaMeklab Suite Version *

0.50.06

### Operating System *

Windows 10

### Java Version *

Java Vendor: Eclipse Adoptium     Java Version: 17.0.12

### Final Verification

HammerGS:
Fixes #7350

  Root cause: Support vehicle engine weight was using RoundWeight.STANDARD
  which delegates to NEXT_HALF_TON (ceiling rounding). Per TM p.133, engine
  weight should round to the nearest half-ton for vehicles 5+ tons, or
  nearest kg for small support vehicles.

  Fix: Added new RoundWeight.SV_ENGINE enum that uses NEAREST_HALF_TON for
  ton-standard vehicles and NEAREST_KG for small SVs. Updated Engine.java
  to use this rounding mode specifically for support vehicle engines.

  Example: A 5-ton WiGE with 5/8 movement, ICE engine, and tech rating D
  calculates to 1.0875 tons raw weight. This now correctly rounds to 1.0
  ton instead of 1.5 tons.

### DIFF
--- a/megamek/src/megamek/common/equipment/Engine.java
+++ b/megamek/src/megamek/common/equipment/Engine.java
@@ -378,7 +378,8 @@ public class Engine implements Serializable, ITechnology {
               && isValidEngine()) {
             double weight = getWeight(entity);
 
-            return roundWeight.round(weight, entity);
+            // SV engine weight rounds to nearest half-ton (or kg for small SVs) per TM p.133
+            return RoundWeight.SV_ENGINE.round(weight, entity);
         } else if (entity.hasETypeFlag(Entity.ETYPE_PROTOMEK) && (engineRating < 40)) {
             // ProtoMek engines with rating < 40 use a special calculation
             return roundWeight.round(engineRating * 0.025, entity);
--- a/megamek/src/megamek/common/util/RoundWeight.java
+++ b/megamek/src/megamek/common/util/RoundWeight.java
@@ -65,6 +65,14 @@ public enum RoundWeight {
         } else {
             return RoundWeight.NEXT_HALF_TON.round(w, e);
         }
+    }),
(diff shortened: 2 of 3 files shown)


Sleet01 (approved):
LGTM

PR #7747, opened 2025-12-10
Fix MegaMek/mekhq#8013: FIx missing ammunition bins on units 

psikomonkie:
Fixes MegaMek/mekhq#8013

When parsing the `size` we should not use the value of "0" as default. If we do not get a value when we `parseDouble`, we should not `setAmmoCapacity`. Prior to an April change, this was wrapped in a try/catch block. If the value wasn't there, it would throw an exception and not call `setAmmoCapacity`. This brings that functionality back.


Sleet01 (approved):
LGTM

PR #7862, opened 2026-01-05
Fix C3M BV Bonus When Connected to C3MM

HammerGS:
Summary

  A C3M (Lance Master) connected to a C3MM (Company Commander) was not receiving the 5% C3 network BV bonus, even though this forms a valid 2-unit network per the rules.

  The Bug

  In Entity.getExtraC3BV(), the condition for C3M units only checked if they had C3S slaves connected:
  (hasC3M() && (calculateFreeC3Nodes() < 3))

  A C3M with no slaves has 3 free nodes, so 3 < 3 = false - the condition always failed.

  Root Cause

  This bug has existed since the original implementation in commit ba2b372ef6 (October 28, 2010). The C3S condition correctly checks c3Master > NONE (has a master connected), but the C3M condition only checked for slaves, not for being connected to a C3MM.

  The Fix

  Added a check for whether the C3M has a master (C3MM) connected:
  (hasC3M() && ((calculateFreeC3Nodes() < 3) || (getC3Master() != null)))

  Now a C3M receives the bonus if it either:
  1. Has C3S slaves connected (original behavior), OR
  2. Is connected to a C3MM as its company commander (new fix)

  Rules Reference

  From the BattleTech rules:
  "Two or more units in a battle force equipped with C3 systems can be designated as part of a C3 network. Add 5 percent of the total BV of all units in a C3 network to each of the units linked by each network."

  A C3MM + C3M is exactly 2 units with C3, forming a valid network.

  Files Changed

  - megamek/src/megamek/common/units/Entity.java - getExtraC3BV() method

  Testing

  Verified with Tai-sho TSH-7S (C3MM) + Shugenja SJA-7D (C3M, no slaves):
  - Before: Only C3MM got bonus; C3M got nothing
  - After: Both units correctly receive 5% of combined network BV

### DIFF
--- a/megamek/src/megamek/common/units/Entity.java
+++ b/megamek/src/megamek/common/units/Entity.java
@@ -13853,8 +13853,13 @@ public abstract class Entity extends TurnOrdered
             // PLAYTEST3 C3 BV changes. each unit is +30% BV, +35% for boosted
             boolean playtestThree = gameOptions().booleanOption(OptionsConstants.PLAYTEST_3);
 
+            // C3 network bonus requires at least 2 members. Check conditions:
+            // - C3MM: has at least one C3M connected
+            // - C3M: has C3S slaves connected OR is connected to a C3MM master
+            // - C3S: has a master (C3M or C3MM) connected
+            // - C3i/Naval C3: has at least one other network member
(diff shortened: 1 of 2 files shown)


exeea:
LGTM! (2010, wow!)

PR #7864, opened 2026-01-06
Fix MD Augmentation Tech Availability Check and Add Unit Tests

HammerGS:
## Summary

  Fixes a bug in `MDAugmentationType.isAvailableIn()` where the `isClan` parameter was incorrectly interpreted as `ignoreExtinction`, causing tech availability checks to ignore faction restrictions.

  ## The Bug

  The method was calling:
```
  java
  return techAdvancement.isAvailableIn(year, isClan);
```

  But ITechnology only has isAvailableIn(int year, boolean ignoreExtinction) - not a year/clan variant. Java matched to this signature, treating isClan as ignoreExtinction and ignoring faction distinctions entirely.

  The Fix

  Changed to explicitly call the 3-param version:
`  return techAdvancement.isAvailableIn(year, isClan, false);`

  Tests Added

  New TechAvailabilityTests in MDAugmentationCostTest.java:
  - Pain Shunt (TechBase.ALL) available for both factions
  - TSM Implant (TechBase.IS, 3060) year-based availability
  - Glider Wings (3069) not available before intro date
  - Extraneous Limbs (3068) not available before intro date
  - Improved Enhanced (2650) available early
  - All augmentations have valid tech data

  Test Plan

  - Unit tests pass
  - Verified with MegaMekLab loading Tau Zombie unit (year 3065)


Sleet01 (approved):
LGTM

Issue #2983, opened 2021-07-01
Errata: Punches with damaged vs non-present hand actuators should have identical to-hit.


xWowie:
### Environment
Standard Megamek version 0.48.0
Windows 7

### Description
Mechs built without hand actuators should have the same to-hit modifiers as mechs with damaged hand actuators when making punch attacks.
Total Warfare Errata 8.0 page 27, Under "punch attacks":
> 3)  ⑤ Under “Missing Actuators”, last sentence
> Likewise, ’Mechs that do not come equipped with a lower arm actuator on the punching arm must add a +2 modifier to the
> to-hit number (in place of the +1 modifier for punching without a hand actuator),
> Change to:
> Likewise, ’Mechs that do not come equipped with a lower arm actuator on the punching arm must add a +2 modifier to the
> to-hit number,

### Files
Saved Game: [punch.sav.gz](https://github.com/MegaMek/megamek/files/6744854/punch.sav.gz)
Playername: Wowie

(my first time subbing a bug or even using github dontkillmepls)

NickAragua:
![image](https://user-images.githubusercontent.com/29387808/126023084-56c956b5-de6f-4796-81da-198356a02b98.png)
Here we have an Archer with a damaged left hand actuator taking a +1 to the target number.

![image](https://user-images.githubusercontent.com/29387808/126023113-243194cf-578b-40bd-aab0-196f1d13aef3.png)
Here we have an Awesome which has no right hand actuator taking a +1 to the target number.

So they do both take the same penalty. Is the problem that they're both taking a +1 instead of +2?

NickAragua:
Based on clarification, MM currently implements these rules as written. If you punch with a lower arm actuator missing, that's a +2. Lower arm AND hand, that's a +3. Whether the actuator is damaged or never existed in the first place doesn't seem to affect the modifiers. 

Thanks for the report, though.

xWowie:
The problem is with mechs that don't have hands in the first place. Like a Rifleman doesn't factor in the penalty for not having hands even though an Awesome or an Archer does - the Awesome and Archer already handle missing/damaged hands correctly. That's what the errata touched on - previously, riflemen didn't suffer the missing hands penalty, but now they do, because the clause in parenthesis was removed. (IE, a Rifleman should suffer a +3 when punching. Everything else is already correct.)

NickAragua:
I think I see what you're talking about now. It's helpful to include that sort of thing in the initial bug report.

PR #7866, opened 2026-01-07
Fixes missing hand actuator being cumulative with

Dakkon42:
missing lower arm actuator in calculating to-hit


Dakkon42:
Proposed fix for issue https://github.com/MegaMek/megamek/issues/2983

PR #7900, opened 2026-01-20
Making the random army dialog usable everywhere

SJuliez:
This PR decouples the random army dialog (force generator) from ClientGUI, making it usable in places other than the lobby and game. Most of the former RandomArmyDialog is moved into AbstractRandomArmyDialog. Similar to the unit selector, subclasses can add a button panel that provides interaction while the main content comes from the abstract base class. As use cases, the dialog can now be called from MM's main menu and MML's force builder to save resp. add random forces.

I havent done much in the way of refactoring on AbstractRandomArmyDialog, it is almost 100% old code and whatever issues it has are for another day.


Sleet01 (approved):
LGTM

PR #7909, opened 2026-01-23
Merge damage managers

Sleet01:
Completes the work begun with the Modular Damage Manager development last year.
This PR removes all old damageEntity code and combines the old "TWDamageManager" class and the new "TWDamageManagerModular" class into one under the original name.
"TWDamageManagerModular" can now be removed.

While we are keeping the new paradigm, where a Game Manager gets assigned a Damage Manager to delegate some calls to, we no longer need the option to switch back to the old code so this option is also being removed.

Testing:
- Ran several bot-on-bot games


Sleet01:
I'll get to the code comments in a bit.

PR #7929, opened 2026-01-25
Remove incorrect validation limiting Mission Equipment Storage to one per unit

HammerGS:
Summary

  Removes erroneous validation that prevented BattleArmor units from mounting multiple Mission Equipment Storage items.

  Changes

  - Removed validation check in TestBattleArmor.hasIllegalEquipmentCombinations() that rejected units with more than one
   Mission Equipment Storage

  Root Cause

  The validation incorrectly treated Mission Equipment Storage like unique equipment (partial wings, magnetic clamps,
  parafoils) which are legitimately limited to one per unit. Mission Equipment Storage has no such restriction in the
  rules.

  Files Changed

  - megamek/src/megamek/common/verifier/TestBattleArmor.java


Sleet01 (approved):
LGTM

PR #7945, opened 2026-01-30
Memoize WeaponAttackAction expected damage

Sleet01:
With the addition of more logic depending on getExpectedDamage() comparisons, it makes sense to memoize the results from these calculations.
They won't exist outside of a single Ranged Attack Phase, so we don't need anything complicated.
But avoiding repeated toHit calcs (with all those involve) in short succession is desirable.

As a small additional bonus, this can reduce some overhead for Princess when comparing WAAs for heat reduction purposes.

Testing:
- Added a couple new unit tests
- Ran all 3 projects' existing unit tests
- Ran some long-ish bot fights with the new code

### DIFF
--- a/megamek/unittests/megamek/common/actions/WeaponAttackActionToHitTest.java
+++ b/megamek/unittests/megamek/common/actions/WeaponAttackActionToHitTest.java
@@ -57,8 +57,10 @@ import megamek.common.board.Board;
 import megamek.common.board.Coords;
 import megamek.common.compute.Compute;
 import megamek.common.enums.GamePhase;
+import megamek.common.enums.Gender;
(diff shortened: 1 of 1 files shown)


megamek/unittests/megamek/common/actions/WeaponAttackActionToHitTest.java
psikomonkie, line 2: License header should be updated to `2025-2026`.

megamek/src/megamek/common/actions/WeaponAttackAction.java
psikomonkie, line 3: License header should be updated to `2002-2026`

PR #7969, opened 2026-02-08
Fixes #7967. Tripod turns were not being handled correctly.

### LINKED ISSUE #7967: [Issue]Tripod Facing Change uses full MP cost.
### Brief Description *

Tripod facing change is not using their special rule that allows Tripods to make any number of Facing Changes for a single Movement Point and are instead paying full movement.

### 3. Steps to Reproduce

1. Start a game with at least one Tripod 'Mech under your control.
2. Deploy Tripod and any other 'Mechs.
3. Continue until movement phase.
4. Select the "Turn" command and spin your tripod around itself.
5. Notice the MP is climbing with each hex "turned", it should stay at 1 MP no matter how many turns you make.

### Attach Files

In the images belowmy Tripod is trying to spin around facing its rear, as you can see the MP cost is not correct.
In the second image it is also trying to enter the hex after it has turned by first using the turn command and then the walk command.

Image

Image

Below it is tyring to spin 360 degrees which once again should only cost 1 MP even if it wanted to spint around several more times, but as can be seen it is costing 1 Hex per facing change in the current version of the game.

Image

Below is an example of what it would look like if it was trying to perform a lateral motion just to show that that is not what's causing the issue in the first images, even here you can see it's costing 1 more MP than it should.

Image

### Severity *

### MegaMek Suite Version *

0.50.11

### Operating System *

Windows 10

### Java Version *

    Java Vendor: Eclipse Adoptium     Java Version: 17.0.13

### Final Verification

TenkawaBC:
This corrects Tripods to turn correctly (1 MP no matter how many facing changes, as long as it has all its legs)


megamek/src/megamek/common/moves/TurnStep.java
```diff
@@ -87,9 +87,10 @@ public PhasePassResult preCompilation(final MoveStep moveStep, final Game game,
         }
 
         // tripods with all their legs only pay for their first facing change
+        // Rules reference Interstellar Operations - Alternate Eras, page 158.
         if ((entity instanceof TripodMek mek) &&
-              mek.atLeastOneBadLeg() &&
-              getTypesOfInterest().contains(prev.getType())) {
+              !mek.atLeastOneBadLeg() &&
+              (prev.getType() == MoveStepType.TURN_LEFT || prev.getType() == MoveStepType.TURN_RIGHT)) {
```
SJuliez, line 93: It seems that l.93 does exactly the same as before. Is there a reason for replacing the previous logic?

SJuliez (commented):
Logic ok, just a question.

megamek/src/megamek/common/moves/TurnStep.java
TenkawaBC: Theoretically, no. I just replaced both conditions when I saw the badleg was not correct, to be sure it worked, and mirroring previous code that did work.

Would you like me to put the getTypesOfInterest part back? (Keeping the !mek.atLeastOneBadLeg())

SJuliez: Nah, I just looked again, doesn't look like it could break anything, maybe even more readable the new way :)

SJuliez (approved):

PR #7986, opened 2026-02-11
Gitignore update

SJuliez:
Updates gitignore so the recent advanced search files dont show as changes


rjhancock:
Unless you want them included in the package, update the megamek/gradle build file as well in the `copyFiles` section to exclude them.

rjhancock:
For all three repos.

PR #7992, opened 2026-02-12
Bridge collapse warning fix

### LINKED ISSUE #7878: [Issue] 'Overweight' warning indicator for bridges being incorrectly calculated if other unit in same hex but not on bridge
### Brief Description *

If a heavy unit is underneath a bridge hex that another heavy unit might cross, with their combined weight exceeding the bridge's CF, the yellow 'overweight warning' indicator is marked on the bridge, even though the first heavy unit is not actually on the bridge and thus the CF will not be exceeded.
See attached screenshot, where one Atlas (red) is on the CF120 bridge, and the other (blue) is on the ground beneath (despite the aforementioned rendering issue making it appear to be on the bridge. You can just see the edge of the yellow warning icon poking out from under the Atlas.

Image

It is slightly more visible in this second image, where again the red Atlas is on the bridge, while the blue Spider (at 30 tons enough to result in a combined weight of 130 tons, greater than the bridge's CF of 120) is on the ground beneath, yet the warning icon is still generated, even though there is no danger to the Atlas.

Image

The issue is solely with the warning symbol generation, as the Atlases are able to safely stroll along the bridge over the other unit.

### 3. Steps to Reproduce

1. Fire up a new MegaMek session.
2. Select a map with bridges high enough for 'mechs to pass under (e.g. '64x51 MesaCity1 N - Mesas').
3. Select at least two units of choice (preferably nice heavy ones like Atlases so that they'll trigger warnings even on the CF120 bridges).
4. Deploy units - one under a bridge, one on a bridge hex next to it.
5. In the movement phase, the bridge hex with the Atlas underneath it will still be incorrectly flagged with a 'weight warning' yellow exclamation marker (although due to a separate rendering order issue (#7877), this may be partly obscured by the sprite for the unit under the bridge).

### Attach Files

_No response_

### Severity *

### MegaMek Suite Version *

0.50.11

### Operating System *

Windows 10

### Java Version *

Java Eclipse Adoptium 17.0.13

### Final Verification

SJuliez:
Disregards units not actually on a bridge for the collapse warning indicator
Fixes #7878

### DIFF
--- a/megamek/src/megamek/common/Hex.java
+++ b/megamek/src/megamek/common/Hex.java
@@ -557,8 +557,8 @@ public class Hex implements Serializable {
     }
 
     /**
-     * @return True if there is a bridge in this hex (thereby excluding any other type of building and requiring
-     * Terrains.BRIDGE_ELEV and Terrains.BRIDGE_CF to be present in this hex).
+     * @return True if there is a bridge in this hex. Tests if the hex has Terrains.BRIDGE. When true, the hex will also
+     *       have Terrains.BRIDGE_ELEV and Terrains.BRIDGE_CF.
      */
     public boolean hasBridge() {
         return containsTerrain(Terrains.BRIDGE);


megamek/src/megamek/common/Hex.java
```diff
@@ -556,6 +556,14 @@ public boolean hasDepth1WaterOrDeeper() {
         return containsTerrain(Terrains.WATER) && (terrainLevel(Terrains.WATER) >= 1);
     }
 
+    /**
+     * @return True if there is a bridge in this hex (thereby excluding any other type of building and requiring
+     * Terrains.BRIDGE_ELEV and Terrains.BRIDGE_CF to be present in this hex).
```
psikomonkie: While this is currently true this isn't true per the rules. I believe a bridge can go over a building. I'm certain a bridge can go over an underground building. Should we acknowledge that in this comment?

EDIT: The fix for it will be a doozy, so that's why I'm mentioning just leaving a comment. Throughout the repo we constantly assume that a hex can only have one building, but because buildings can be underground, and because I believe a bridge can go over a building, it isn't actually true, and is something that will eventually™️ need refactored.

SJuliez: OK I removed that comment, it wasn't really to the point anyway. TO:AR, p.28 calls building base terrain, which makes me think only one can be in a hex and a bridge is a building. Otherwise there could be multiple buildings and stacked bridges, right?

PR #7995, opened 2026-02-13
Show network information dialog added

TenkawaBC:
This adds a dialog under Game->Network Information which then can display the local IP, public IP, and host connection address.
They are hidden by default, and require button presses to display, with each being a separate button.

NetworkInformation


Sleet01 (changes requested):
Please address the CodeQL warnings, specifically:
1. missing @Override notation,
2. string comparison using `==`,
3. Not closing the InputStreamReader on line 184.

For 3. I would recommend a `try-with-resources` approach, which will close the InputStreamReader for you automatically.

I'm ambivalent about using an AWS endpoint to find the user's external IP, but even with older or more established services, the possibility of the service being down is always there.

TenkawaBC:
Updates done and committed. I added a in.close(); on the stream.

megamek/src/megamek/client/ui/dialogs/buttonDialogs/NetworkInformationDialog.java
```diff
@@ -0,0 +1,197 @@
+/*
+ * Copyright (C) 2021-2025 The MegaMek Team. All Rights Reserved.
+ *
+ * This file is part of MegaMek.
```
rjhancock, line 4: These are new files. Make sure the correct copyright header is in place.

This should strictly be 2026.

rjhancock (dismissed):
Copyrights get everyone.

Sleet01 (approved):
Looks good to me!

PR #7999, opened 2026-02-13
Add upnp to network information

TenkawaBC:
This adds UPnP support, accessed through the Network Information dialog.

UPnP code from https://github.com/adolfintel/WaifUPnP/. Copyrights maintained. Thanks to Federico.
Package names updated.

Please test.


TenkawaBC:
taking another look at this one to see if I can update the security stuff properly without breaking things.

PR #8019, opened 2026-02-18
Silver Bullet Gauss HitMod moved from Ammo to Weapon.

exeea:
Silver Bullet Gauss HitMod moved from Ammo to Weapon.


exeea:
> Was this related to [MegaMek/mekbay#130](https://github.com/MegaMek/mekbay/issues/130) or is this changed for a different reason?

Yes, is exactly that issue. 
The issue was also in MML (but no issue opened for it)

PR #8060, opened 2026-03-04
Added tooltips to quirks and tooltip word wrapping for readout

capellancitizen:
Addresses https://github.com/MegaMek/mekhq/issues/8731, which is filed under MekHQ, though the changes to the readout have to live in MegaMek proper. I've also added a word wrapping utility function to allow for the sometimes excessively long quirk tooltips to be wrapped to a more sensible length, similar to `MultiLineTooltip` in `MekHQ`. I've got another set of changes for `UnitViewPanel` and related classes in MekHQ that depend on these changes (mainly making `MultiLineTooltip` use the new utility function).

Screenshots:
screen03
screen04
screen05
screen06

~Known issue: After hovering over any span, that tooltip "sticks" and will show hovering over the readout, even if you aren't hovering over the span itself. This is a pre-existing issue with the readout view itself (Indeed with JTextPane) with the readout that could be observed before I made these changes. Fixing it is probably out of scope, and TBH I'm not sure how you'd even fix it: It seems like the only built-in option for hover events on html elements in a JTextPane is [HyperlinkListener](https://docs.oracle.com/javase/8/docs/api/javax/swing/event/HyperlinkListener.html)?~ Fixed, was an issue with the mouse listener registered in EntityReadoutPanel.

screen07

### DIFF
--- a/megamek/src/megamek/client/ui/entityreadout/ViewElement.java
+++ b/megamek/src/megamek/client/ui/entityreadout/ViewElement.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2026 The MegaMek Team. All Rights Reserved.
+ * Copyright (C) 2025-2026 The MegaMek Team. All Rights Reserved.
  *
  * This file is part of MegaMek.
  *


capellancitizen:
Oh, I see the issue with the tooltips, looking deeper. I'll tag a change for that onto this MR later.

megamek/src/megamek/client/ui/entityreadout/ViewElement.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2025 The MegaMek Team. All Rights Reserved.
+ * Copyright (C) 2026 The MegaMek Team. All Rights Reserved.
```
psikomonkie: This should be `2025-2026`.

capellancitizen: Fixed. (Should I resolve the convo, or is that only for bot reviews?)

psikomonkie: I think just for the bots - I usually leave my convos with humans open for the reviewer to resolve.

psikomonkie (approved):
Nice! Thanks for getting the updated copyright year in quick. TBH thought you were gaslighting me for a second as it wasn't a new commit lol

capellancitizen:
That's... huh, I see the time is off in the VM I was using, that might be why?

PR #7635, opened 2025-11-16
Nova CEWS fixes

HammerGS:
# Complete Nova CEWS Implementation (Fixes #6754)

  ## Summary
  Comprehensive implementation of Nova Composite Electronic Warfare System (Nova CEWS) rules per Interstellar
  Operations: Alternate Eras (p.60 Game Rules, p.183 BV Rules).

  **TT Reference:**
  - IO: Alternate Eras p.60 (Network Formation)
  - IO: Alternate Eras p.183 (BV Bonus Rules)

  ---

  ## Key Features Implemented

  ### 1. BV Cap Fix (Issue #6754)
  **File:** `Entity.java` (lines 13310-13316)

  Fixed missing 35% BV bonus cap per TT rules:
  > "The maximum total increase a force's Battle Value can receive from the use of the Nova CEWS is equal to 35
  percent of the force's base Battle Value."

  ```java
  double rawBonus = totalForceBV * multiplier;
  if (hasNovaCEWS()) {
      double maxBonus = baseBV * 0.35;
      rawBonus = Math.min(rawBonus, maxBonus);
  }

  Testing: NovaCEWSBVTest.java - 6/6 tests passing

  ---
  2. In-Game Network Management UI

  File: NovaNetworkDialog.java (NEW - 460+ lines)

  End Phase dialog for Nova CEWS network management per TT rules:
  - Link/unlink units during End Phase
  - Max 3 units per network enforcement
  - Pending changes preview
  - Network status display (X/3 format)
  - 21 internationalized strings

  Integrated with: ReportDisplay.java (Nova Networks button)

  ---
  3. Lobby Configuration Persistence Fix

  File: Client.java (lines 531-537)

  Problem: Nova CEWS networks configured in lobby didn't persist to game.

  Root Cause: Client never called C3Util.wireC3() to reconstruct network IDs from UUID arrays after entity
  deserialization.

  Solution: Added wireC3() reconstruction loop for all C3 types (Standard C3, C3i, Naval C3, Nova CEWS).

  ---
  4. C3 Connection Lines Display Fix

  File: BoardView.java (lines 615-622)

  Problem: C3 connection lines disappeared during report phases.

  Root Cause: BoardView phase handler didn't call redrawAllEntities() for report phases.

  Solution: Added report phase case handlers ensuring connection lines persist through ALL phases.

  ---
  5. Network/ECM Independence Fix

  Files: Entity.java, BoardView.java

  Problem: C3 lines disappeared when Nova CEWS set to "Off" mode.

[...]

PR #8062, opened 2026-03-04
Fix NPE in /nova link command caused by null pending network

HammerGS:
## Root Cause

  PR #7635 changed `Entity.getNewRoundNovaNetworkString()` to return `null` when no pending network change exists (previously it returned the current network ID). This was intentional to fix phantom pending changes in the NovaNetworkDialog, but the `/nova` chat command handlers were not updated to handle the new null return value.

  When a player runs `/nova link 1 2 3`, the server-side handler calls `listNetwork()` which does
  `novaUnit.getNewRoundNovaNetworkString().equals(...)` - calling `.equals()` on null throws the NPE.

## Changes

  1. **AssignNovaNetServerCommand.java** - Added `getEffectiveNovaNetworkId()` helper that returns the pending network ID if set, otherwise falls back to `getOriginalNovaC3NetId()`. Updated `strLink3()`, `strLink2()`, and `listNetwork()` to use it. Also made `listNetwork()` null-safe with `Objects.equals()`.

  2. **AssignNovaNetworkCommand.java** - Same fix applied to the client-side command handler for consistency. The client's `listNetwork()` already used `Objects.equals()` but still passed raw null values from `getNewRoundNovaNetworkString()` into network ID comparisons.

## Files Changed

  - `megamek/src/megamek/server/commands/AssignNovaNetServerCommand.java` - NPE fix + null-safe comparisons
  - `megamek/src/megamek/client/commands/AssignNovaNetworkCommand.java` - Matching fix for client-side consistency

  ## Testing

  - Command `/nova link ID ID` and `/nova link ID ID ID` no longer throw NPE
  - Command `/nova print`, `/nova unlink` unaffected (don't use the changed code paths)
  - NovaNetworkDialog behavior unchanged (still uses `getNewRoundNovaNetworkString()` directly with its own null handling)

  ## Regression

  Introduced by #7635 (merged 2025-11-25). Working in 50.06, broken in 50.07+.

### DIFF
--- a/megamek/src/megamek/client/commands/AssignNovaNetworkCommand.java
+++ b/megamek/src/megamek/client/commands/AssignNovaNetworkCommand.java
@@ -299,12 +299,27 @@ public class AssignNovaNetworkCommand extends ClientCommand {
     }
 
     /**
-     * Returns the effective Nova network ID for the entity. If a pending change exists, returns that; otherwise falls
-     * back to the entity's original Nova C3 network ID.
+     * Returns the effective Nova network ID for the entity.
+     * <p>
+     * Precedence:
+     * <ol>
+     *     <li>Pending next-round network ID, if set.</li>
+     *     <li>Current C3 network ID (shared by linked units).</li>
+     *     <li>Original per-unit Nova C3 network ID as a last resort.</li>
+     * </ol>
      */
     private String getEffectiveNovaNetworkId(Entity entity) {
         String pendingId = entity.getNewRoundNovaNetworkString();
-        return pendingId != null ? pendingId : entity.getOriginalNovaC3NetId();
+        if (pendingId != null) {
+            return pendingId;
+        }
+
+        String currentId = entity.getC3NetId();
+        if (currentId != null) {
+            return currentId;
+        }
+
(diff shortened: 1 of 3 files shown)


HammerGS:
## Copilot Feedback - Accepted

Both comments are valid and have been applied. 

**The issue:** `getOriginalNovaC3NetId()` returns a per-unit unique ID (`"C3Nova.{entityId}"`), while `getC3NetId()` returns the shared network ID that linked units have in common. Falling back to the original ID would cause `listNetwork(..., planned=true)` to treat already-linked units (with no pending change) as unlinked, breaking `/nova unlink` and `/nova print` for active networks.

**The fix:** `getEffectiveNovaNetworkId()` now uses a 3-tier fallback:
1. Pending next-round network ID (if set)
2. Current C3 network ID (shared by linked units)
3. Original per-unit Nova C3 network ID (last resort, e.g., inactive Nova CEWS)

psikomonkie (commented):
Nice! Can you update the license on these two to end in `2026`, but I think the code's good.

rjhancock (approved):

PR #8120, opened 2026-03-24
tech advancement overrides + unit build year

exeea:
This PR adds the following:
1) possibility to set tech advancement overrides per-faction (example, the whole ComStar ordeal)
2) removes the "hacks" to make ComStar work making it fallback on TH or shortcircuiting many tech advancement flows by:
- properly set CS as prototype/production where TH has it too. 
- use the override system at point 1
3) introduces the possibility to set a build year for units (implemented over the originalBuildYear field): for tech evaluation both intro year and build year will be checked and tech from both years will be considered valid for the unit. This will solve situations where a mech has been introduced in yearr X while the tech to build it has been lost few years before (and the lore about it is that it was simply pulled out from a large stock and refitted, see Mongoose MON-69, Mongoose MON-70 and BattleMaster BLR-2C).


exeea:
deconstructed into multiple PRs

PR #8124, opened 2026-03-24
Fixes issue #6727 : Back to default WASD centered keybinds

### LINKED ISSUE #6727: [RFE] Funky default keybinds
### Enhancement Type *

Improvement to Existing Feature

Since 50 dev line the default keybinds changed. 
They used to be as per MekHq > docs > default keybinds but now are strangely using top and down buttons for weapon cycling, which is not very convenient when all the keys are basically around WASD.

With that said it is nothing too major, but I think it will be cleaner to change them for a milestoe update like 50.04.

### Use Cases or Rationale

1. Keybinds being spread widely accross the keyboard
2. By having more intuitive keybinds by default
3. No but the docs > default keybinds is not respected right now

### Attach Files

Image

### MegaMek Suite Version *

Any 50.x

### Final Verification

ghost:
Fixes #6727 

Mainly to bring back the key binds to the documented default keybinds.
For instance in current version next weapon is down instead of E.

Also this is a question, is that intended to enable a more balanced keyboard ?

In this pull request keyboard is following a classic "mouse in right hand" layout centered on the WASD.

Please give me any advice for the advancement of that request

### DIFF
--- a/megamek/mmconf/defaultKeyBinds.xml
+++ b/megamek/mmconf/defaultKeyBinds.xml
@@ -29,8 +29,8 @@
     </KeyBind>
 
     <KeyBind>
-        <command>toggleChat</command> <!-- Enter -->
-        <keyCode>10</keyCode>
+        <command>toggleChat</command> <!-- Backquote -->
+        <keyCode>192</keyCode>
         <modifier>0</modifier>
         <isRepeatable>false</isRepeatable>
     </KeyBind>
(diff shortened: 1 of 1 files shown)


ghost:
So about the conflicts I followed what was written in /doc/Default Key Binds.txt 
I think we should discuss in which case where we can keep it and in which a separate key should be used

kuronekochomusuke:
when updating key binds, you also need to update them in KeyCommandBind.java

PR #8183, opened 2026-04-10
Fix: Unstreamlined quirk now prevents atmospheric and ground operations

### LINKED ISSUE #6834: [Issue] Invalid Dropship assigned to planetary operations
### Brief Description *

The Behemoth dropship, and theoretically any dropship with Unstreamlined, should not be able to operate in atmosphere. MHQ still selects it for dropship defense type missions.

### 3. Steps to Reproduce

Have a behemoth.
Get a dropship defense mission.
Wonder how the heck you are going to launch

### Attach Files

_No response_

### Severity *

### MekHQ Suite Version *

50.05

### Operating System *

Windows 3.2

### Java Version *

No coffee here

### Final Verification

VicenteCartas:
Units with the `unstreamlined` negative quirk (e.g., Behemoth) are now correctly marked as doomed in atmosphere and on ground maps via `Aero.doomedInAtmosphere()` and `Aero.doomedOnGround()`. Previously the quirk existed but had no gameplay effect.

Fixes MegaMek/mekhq#6834 (along with companion PRs in mekhq and mm-data).

### DIFF
--- a/megamek/src/megamek/common/units/Aero.java
+++ b/megamek/src/megamek/common/units/Aero.java
@@ -1791,7 +1791,12 @@ public abstract class Aero extends Entity implements IAero, IBomber {
 
     @Override
     public boolean doomedInAtmosphere() {
-        return false;
+        return hasQuirk(OptionsConstants.QUIRK_NEG_UNSTREAMLINED);
+    }
+
+    @Override
+    public boolean doomedOnGround() {
+        return hasQuirk(OptionsConstants.QUIRK_NEG_UNSTREAMLINED);
     }
 
     @Override


SJuliez:
The mentioned issue states that MHQ selects such a DS and it doesn't work out since it cant start. But the Unstreamlined quirk had no gameplay effect so far. Therefore I'm not sure where the actual issue was (until now). MHQ does not seem to check units for being suitable; wouldnt then this PR actually introduce that precise issue with MHQ selecting unfit DS?
Also, since this quirk is about an interaction with atmosphere (not e.g. gravity), boards at vacuum conditions would likely be ok for those units, would they?

VicenteCartas:
> The mentioned issue states that MHQ selects such a DS and it doesn't work out since it cant start. But the Unstreamlined quirk had no gameplay effect so far. Therefore I'm not sure where the actual issue was (until now). MHQ does not seem to check units for being suitable; wouldnt then this PR actually introduce that precise issue with MHQ selecting unfit DS? Also, since this quirk is about an interaction with atmosphere (not e.g. gravity), boards at vacuum conditions would likely be ok for those units, would they?

I will have the MekHQ PR up tomorrow, I need to do some more playing with it before pushing it and today has been crazy.

As I understand the bug from MekHQ, the user is saying that the unit should not have appeared as the dropship in the defense mission, not that they can't use it (I take their 3rd point as a retorical question, no that the game is stopping them from anything). In theory, when selecting those missions the quirk should be taken into account to avoid using such units as realistically they couldn't be in the ground to be defended.

About vacuum, I think you are right. Grounded units in planets with no atmosphere should work (I think? :D). I need to check how to find that information when checking with isDoomed, or if I need then to move the check somewhere else.

SJuliez:
> About vacuum, I think you are right. Grounded units in planets with no atmosphere should work (I think? :D). I need to check how to find that information when checking with isDoomed, or if I need then to move the check somewhere else.

Would probably have to check in another way, the isDoomed checks are really single-minded. I didn't want to suggest it as it will make this much more complicated. It's an upgrade the way you did it and vacuum conditions and unstreamlined units sounds like a fringe case. I'd say, unless you want to dig into this, add a "// FIXME: vacuum conditions should allow unstreamlined units" line and leave for posterity :)

VicenteCartas:
I will take of this on StratCon instead:
- In StratCon the check is just about not using the unit, which is less punishing than dooming the units in megamek
- In StratCon I can check atmosphere conditions.

PR #8257, opened 2026-05-06
Move conventional infantry to its own Infantry subclass

SJuliez:
Merge together with the MML and MekHQ PRs!!

As the title says, this PR moves CI to its own Infantry subclass ConvInfantry (borrowing from ConvFighter), making Infantry abstract. I think this better reflects that while there is some overlap between BA and CI and they're sometimes treated together in the rules, there is also substantial non-overlap and many concepts of CI don't apply to BA at all (field guns, specializations, MD and more).

This way, it's no longer necessary to write 
`entity instanceof Infantry && !(entity instanceof BattleArmor)`
instead ConvInfantry can be directly checked and used as a pattern variable.
Present instanceof Infantry checks are still valid and correct, although some can be simplified. In the interest of finishing this and not making it even larger, I have not searched for those checks that can be simplified.

I have done my best to sort CI-concepts into ConvInfantry, leaving only generic infantry stuff in Infantry but I suppose a few things may still be sorted out. As a result, Infantry is now at 700 lines with both ConvInfantry and BA at around 2000 lines which I suppose is a good reflection of the rules overlap.


SJuliez:
@HammerGS please add Copilot review as no one else is going to look at those 100 file changes :)

Sleet01:
> @HammerGS please add Copilot review as no one else is going to look at those 100 file changes :)

Don't threaten me with a good time!

SuperStucco:
Ejected pilots and vehicle crews are peculiar cases of conventional infantry.  These should get extra testing.

SJuliez:
> > @HammerGS please add Copilot review as no one else is going to look at those 100 file changes :)
> 
> Don't threaten me with a good time!

Copilot is going to give me hell already!

Just to note, I ran the MM tests and ran a botvbot battle with BA and CI and things didnt explode, at least.

Sleet01:
> Ejected pilots and vehicle crews are peculiar cases of conventional infantry. These should get extra testing.

I actually think this would give us a clean way to avoid some of the current weird bot-vs-ejected-pilot behaviour, although I'll wait until I've read more of the code to say for sure.

megamek/src/megamek/common/units/Infantry.java
```diff
@@ -177,56 +87,19 @@ public class Infantry extends Entity {
...
-    }
+    private static final int[] NUM_OF_SLOTS = { 20, 20 };
 
     /**
      * Generate a new, blank, infantry platoon. Hopefully, we'll be loaded from somewhere.
      */
-    public Infantry() {
-        super();
+    protected Infantry() {
         // Create a "dead" leg rifle platoon.
```
Sleet01, line 103: Can we really skip running the Entity constructor here?
It seems unlikely on the face of it, but if so, noice!

NM, it's because it's Abstract now -_-;

Sleet01 (approved):
Looking really good so far!

megamek/src/megamek/common/units/Infantry.java
SJuliez: you're confusing me :) The entity constructor cannot be skipped, but writing super(); is redundant so I removed it

megamek/src/megamek/common/compute/MarinePointsScoreCalculator.java
```diff
@@ -165,11 +165,11 @@
      *
      * @return base MPS before building modifier
      */
-    private static int calculateInfantryMPS(Infantry inf) {
+    private static int calculateInfantryMPS(ConvInfantry inf) {
         int mps = 0;
 
         // Check if marines (specialized infantry)
-        boolean areMarines = (inf.getSpecializations() & Infantry.MARINES) != 0;
+        boolean areMarines = (inf.getSpecializations() & ConvInfantry.MARINES) != 0;
```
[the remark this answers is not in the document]
SJuliez: unchanged from before

HammerGS (approved):

PR #8261, opened 2026-05-08
  Fix FormationType discrepancies with Campaign Operations rules

HammerGS:
## Root Cause
  Several `FormationType` definitions in `ratgenerator/FormationType.java` drifted from the Campaign Operations rules
  (2024 5th Print, pp. 60–68). Two ideal-role assignments were missing, one armor threshold was significantly off, one
  variant approximated an "OR" requirement instead of using the paired-constraint mechanism the parent already used, and
   one grouping constraint required twice as many pairs as the rules state.

## Changes
  1. `createAnvilLance` — Armor threshold corrected from `40` to `105` per CamOps; `mainDescription` updated to `"Armor 105+"`.
  2. `createPursuitLance` — Added missing `idealRole = STRIKER` per CamOps.
  3. `createRangerLance` — Added missing `idealRole = SKIRMISHER` per CamOps.
  4. `createFastAssaultLance` — Replaced the flagged FIXME approximation (`CountConstraint(2, JUGGERNAUT or SNIPER)`) with paired `Juggernaut` (1, pairedWithNext) / `Sniper` (2, pairedWithPrevious) constraints, mirroring the parent `createAssaultLance` pattern. This now correctly enforces "1 Juggernaut **or** 2 Snipers" rather than "any 2 of either".
  5. `createVehicleCommandLance` — `GroupingConstraint` `numGroups` reduced from `2` to `1`. CamOps states "*only one pair* of vehicles needs to have the Sniper, Missile Boat, Skirmisher, or Juggernaut Unit Roles"; the previous code required two such pairs based on a speculative reading. Comment updated to reference the rules text.
  6. Added `getIdealRole()` accessor on `FormationType` for test access (consistent with existing getters).
  7. Copyright header updated to span 2016–2026.

  ## Files Changed
  - `megamek/src/megamek/client/ratgenerator/FormationType.java` — five formation rule corrections, one new accessor, copyright bump.
  - `megamek/unittests/megamek/client/ratgenerator/FormationTypeTest.java` — new regression test class with five tests covering each fix.

  ## Testing
  - `./gradlew :megamek:compileTestJava` — `BUILD SUCCESSFUL`.
  - `./gradlew :megamek:test --tests megamek.client.ratgenerator.FormationTypeTest` — `tests=5, failures=0, errors=0`.
  - Tests cover:
    - Anvil `mainCriteria` rejects armor 40/104 and accepts 105/200; description string is `"Armor 105+"`.
    - Pursuit `idealRole` is `STRIKER`.
    - Ranger `idealRole` is `SKIRMISHER`.
    - Fast Assault Juggernaut and Sniper constraints exist, are mutually paired (`pairedWithNext`/`pairedWithPrevious`), and have minimums 1 and 2 respectively.
    - Vehicle Command `GroupingConstraint` has `groupSize=2` and `numGroups=1`.

  ## Reference
  Campaign Operations (2024 5th Print), pp. 60–68 — formation type requirements.


megamek/src/megamek/client/ratgenerator/FormationType.java
```diff
@@ -1516,11 +1520,12 @@ private static void createFastAssaultLance() {
...
-        //  constraints with.
-        ft.otherCriteria.add(new CountConstraint(2,
-              ms -> ms.getRole().isAnyOf(JUGGERNAUT, SNIPER),
-              "Juggernaut or Sniper"));
+        Constraint c = new CountConstraint(1, ms -> ms.getRole() == JUGGERNAUT, "Juggernaut");
+        c.setPairedWithNext(true);
+        ft.otherCriteria.add(c);
+        c = new CountConstraint(2, ms -> ms.getRole() == SNIPER, "Sniper");
+        c.setPairedWithPrevious(true);
+        ft.otherCriteria.add(c);
```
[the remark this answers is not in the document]
HammerGS: Good catch — addressed. `qualifies()` now evaluates the alternative when the first paired constraint falls short and returns `false` only if both alternatives are short. Added regression test `qualifies_pairedConstraint_failsWhenNeitherAlternativeIsMet` that fails under the previous logic.

megamek/src/megamek/client/ratgenerator/FormationType.java
```diff
@@ -1911,6 +1913,7 @@ private static void createLightStrikerCavalryLance() {
     private static void createRangerLance() {
         FormationType ft = new FormationType("Ranger", "Striker/Cavalry");
         ft.allowedUnitTypes = FLAG_GROUND;
+        ft.idealRole = SKIRMISHER;
         ft.maxWeightClass = EntityWeightClass.WEIGHT_HEAVY;
         allFormationTypes.put(ft.name, ft);
```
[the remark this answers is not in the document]
HammerGS: Good catch — addressed. Added the missing `retVal.addAll(units)` so generated units actually accumulate. Did not add a dedicated unit test because exercising `tryIdealRole` requires `UnitTable.findTable(...)` with an initialized RAT registry, which is out of scope for a focused unit test; the fix is a one-line accumulation correction.

megamek/unittests/megamek/client/ratgenerator/FormationTypeTest.java
```diff
@@ -0,0 +1,200 @@
...
+ * Print, pp. 60-68).
+ */
+class FormationTypeTest {
+
+    @Test
+    void anvilLance_armorThresholdIs105_perCamOps() {
+        FormationType anvil = FormationType.getFormationType("Anvil");
+        assertNotNull(anvil, "Anvil formation should be registered");
+
+        MekSummary armor40 = mock(MekSummary.class);
```
SJuliez, line 67: Imho these values are trivialities that dont break anything if wrong and therefore not really worth testing. What would be worth testing is if the system of using paired OR constraints as well as AND constraints still works; that isnt so trivial to see from the code and rather more easy to break. This is (more or less) done below testing a vehicle command. It would be even better to create a complex formation type as a test (so it's not dependent on the real formation types) and then test the constraints. Just my 2cents

SJuliez: Better add rule source info as a comment to the real code (as we all should, wherever we change code) to prevent people from changing simple values

megamek/src/megamek/client/ratgenerator/FormationType.java
```diff
@@ -1516,11 +1527,12 @@ private static void createFastAssaultLance() {
         ft.otherCriteria.add(new CountConstraint(3,
               ms -> ms.getWeightClass() >= EntityWeightClass.WEIGHT_HEAVY,
               "Heavy+"));
-        // FIXME: The actual requirement is one juggernaut or two snipers; there needs to be a way to combine
```
SJuliez, line 1519: Nice if this can actually be done :)

SJuliez (approved):
Good update, especially on that piece that seemed to miss some technical change according to the comment. What this class, yet another monolithic juggernaut without comments or usability help, also needs is breaking out the lance types into their own classes, giving it some javadoc so it's actually usable by other people and maybe separate logic and UI (break out UI related stuff)

megamek/unittests/megamek/client/ratgenerator/FormationTypeTest.java
HammerGS: Fair point. I've kept the value tests since they pin specific CamOps numbers (so a future revert to e.g. armor 40 would fail under the regression test), but added an independent synthetic complex-formation test that exercises paired-OR alongside an AND-style `CountConstraint` — see `qualifies_complexFormation_*` (3 cases: both met, AND fails despite OR met, OR fails despite AND met). The synthetic formation is built via reflection so it doesn't depend on the real formation registry.

HammerGS: Done. Added Campaign Operations rule citations as comments in `FormationType.java` at each modified call site (Anvil armor threshold, Pursuit/Ranger ideal roles, Fast Assault paired-OR, Vehicle Command grouping). Cited the formation by name rather than page number to keep the references edition-resilient.

megamek/src/megamek/client/ratgenerator/FormationType.java
HammerGS: The paired-with-previous/next mechanism was already in place for the parent Assault Lance — Fast Assault just hadn't picked it up. The FIXME called it out, and the new `qualifies_pairedConstraint_failsWhenNeitherAlternativeIsMet` regression test exercises the algorithm path that was previously broken, so any future regression will be caught.

PR #8274, opened 2026-05-15
Fix Entity.NONE id assignment in Game.addEntity

VicenteCartas:
## Summary

Fixes `Game.addEntity` so entities whose id is still `Entity.NONE` are assigned a real game id before being inserted into the in-game object map.

A unit added under `Entity.NONE` can collide with other sentinel-based lookups. In the reported failure, a newly constructed mek was stored under `-1`; later `Entity.getElevation()` resolved `getTransportId() == Entity.NONE` back to that same entity and recursed until `StackOverflowError`.

This also adds a focused regression test that verifies `Game.addEntity` does not leave an entity registered under `Entity.NONE`.

## Alternative Considered

The narrower test-only fix would be to assign an id in `TWGameManagerTest.testCriticalEntityAllReactiveSlots()` before calling `game.addEntity(mek)`, for example:

```java
mek.setId(game.getNextEntityId());
game.addEntity(mek);
```

That would fix the failing test, but it would still allow other callers to add a default-id entity under the `Entity.NONE` sentinel. The production fix prevents the sentinel collision at the source.

## Testing

- `./gradlew test --rerun-tasks --tests megamek.common.GameTest --tests megamek.server.totalWarfare.TWGameManagerTest.testCriticalEntityAllReactiveSlots -x jacocoTestReport`
- `./gradlew :megamek:checkstyleTest --rerun-tasks`

### DIFF
--- a/megamek/src/megamek/common/game/Game.java
+++ b/megamek/src/megamek/common/game/Game.java
@@ -1401,7 +1401,7 @@ public final class Game extends AbstractGame implements Serializable, PlanetaryC
         }
         // Add this Entity, ensuring that its id is unique
         int id = entity.getId();
-        if (isIdUsed(id)) {
+        if ((id == Entity.NONE) || isIdUsed(id)) {
             id = getNextEntityId();
             entity.setId(id);
         }
--- a/megamek/unittests/megamek/common/GameTest.java
+++ b/megamek/unittests/megamek/common/GameTest.java
@@ -33,14 +33,43 @@
 package megamek.common;
 
 import static org.junit.jupiter.api.Assertions.assertFalse;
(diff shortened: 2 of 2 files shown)


Sleet01:
This _seems_ reasonable, but I'd like to poke around a bit to confirm that this won't break anything.
I can't shake the feeling that there's a reason the unit ID doesn't get set as soon as it's added to the game...

Sleet01:
Apologies, I'll have an up/down on this today.

Sleet01 (approved):
This should work well, thanks for taking care of this!
It looks like every user of `.addEntity()` just does its _own_ id checks and workarounds, so this change is pure positive.

VicenteCartas:
Set the PR ready for review.

PR #8275, opened 2026-05-15
Frankenmek support

exeea:
This PR adds FrankenMek support

### DIFF
--- a/megamek/src/megamek/common/units/Entity.java
+++ b/megamek/src/megamek/common/units/Entity.java
@@ -1455,8 +1455,46 @@ public abstract class Entity extends TurnOrdered
         return compositeTechLevel.getTechRating();
     }
 
+    private @Nullable SimpleTechLevel getRulesLevelOverride() {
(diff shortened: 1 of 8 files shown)


megamek/src/megamek/common/units/Entity.java
```diff
@@ -1455,8 +1455,46 @@ public TechRating getTechRating() {
         return compositeTechLevel.getTechRating();
     }
 
+    private @Nullable SimpleTechLevel getRulesLevelOverride() {
```
pavelbraginskiy: This seems pretty complicated, with a bunch of extra methods. Why aren't these changes to Mek's TechAdvancement methods?

pavelbraginskiy (changes requested):

megamek/src/megamek/common/units/Entity.java
exeea: Ok, I directly corrected it in the part2

pavelbraginskiy (approved):

PR #8280, opened 2026-05-20
FrankenMek leg-mismatch fix

exeea:
This PR drops the use of the `model` to identify mismatching legs.
Now we do it using: chassis/tonnage/structure/omni/BM-IM

This is simplified by:
`donor:` property now stores only the chassis + clanname
`donor type:` IndustrialMek/BattleMek/OmniMek

tonnage/structure is derived directly by the location itself.

### DIFF
--- a/megamek/src/megamek/common/loaders/MtfFile.java
+++ b/megamek/src/megamek/common/loaders/MtfFile.java
@@ -115,6 +115,7 @@ public class MtfFile implements IMekLoader {
     private final String[] armorValues = new String[12];
     private final Map<Integer, String> frankenMekStructureValues = new HashMap<>();
     private final Map<Integer, String> frankenMekLocationSources = new HashMap<>();
+    private final Map<Integer, String> frankenMekLocationSourceMetadata = new HashMap<>();
     private boolean mismatchedFrankenMekLegs = false;
 
     private final String[][] critData;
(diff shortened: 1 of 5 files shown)


megamek/src/megamek/common/loaders/MtfFile.java
```diff
@@ -1634,11 +1641,6 @@ private boolean isProcessedComponent(String line) {
             return true;
         }
```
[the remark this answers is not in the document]
exeea: was never released, was only in the nightly build for few days. Makes no sense.

megamek/src/megamek/common/ui/EnhancedTabbedPane.java
```diff
@@ -1205,6 +1205,70 @@ public boolean hasDetachedTabs() {
         return !detachedTabs.isEmpty();
     }
 
+    /**
```
Sleet01, line 1208: This file's changes seem like they are from a different PR; are they meant to be included here?

megamek/unittests/megamek/common/units/FrankenMekTest.java
```diff
@@ -47,6 +52,142 @@ static void beforeAll() {
         EquipmentType.initializeTypes();
     }
 
+    private static Mek newFrankenMek() {
```
Sleet01, line 55: Ignore previous comment about deleting tests; I see the tests have been migrated.

Sleet01 (changes requested):
I'd prefer that the seemingly-unrelated GUI file change get moved to its own PR if possible.

Otherwise, looks good!

exeea:
> I'd prefer that the seemingly-unrelated GUI file change get moved to its own PR if possible.
> 
> Otherwise, looks good!

is not unrelated, is the only way to make the FrankenMek Structure floating tab disappear. 
Without that update you could:
1) set an unit to frankenmek
2) float the frankenmek structure tab-view to a window
3) set the unit back to non-frankenmek
4) the frankenmek structure window remains visible and configurable (!)

Sleet01 (approved):
LGTM

PR #8289, opened 2026-05-25
Unit selector G/P option

SJuliez:
This moves the unit selector gunnery/piloting toggle directly into the unit selector where it is used exclusively and removes it from the GUI settings, see below. The G/P toggle still saves its state to the client preferences, so the unit selector will remember the latest setting.

Other changes:
- moved the G/P fields down and into a single line that is hidden when G/P is turned off
- internal changes to how they are read; when the value is illegal, they show a red error border
- when G/P is turned off, the values reset to 4/5 so that there is no hidden BV multiplier
- the shown unit count is now updated also when first showing the unit selector
- bit of cleanup of code repetition

image


megamek/src/megamek/client/ui/dialogs/buttonDialogs/CommonSettingsDialog.java
psikomonkie, line 3: Copyright

psikomonkie (dismissed):
One change needed: Copyright on `CommonSettingsDialog`. Everything code wise looks great & I'm otherwise ready to approve.

I love learning about a feature by seeing a PR where it's replaced by a better version!! 

This looks fantastic. The caveats about the old implementation just make it seem so... Unhelpful. This is such a good change that I can see myself using. Thank you!

HammerGS (approved):

PR #8291, opened 2026-05-27
PR: Begin Deprecating Gun Emplacements

psikomonkie:
image

Publicly marks the old Gun Emplacements as deprecated, Advanced Buildings which follow the TO:AR rules more closely will be supported going forward.

I did not mark the class with the deprecated annotation yet, just indicating that we no longer intend to support this unit type.


SJuliez:
I'm not sure I would do that. It does not explain anything to the users. Also, afaik we have no construction in MML in place to replace GE, so what would a player do? Maybe we can do the following: introduce a conversion that loads GE as advanced buildings, so that the GE class can actually be removed at some point while to users this is transparent (and works for customs!). Also, we can start showing GE as advanced buildings in the unit selector (again, nothing the player needs to know). Once this really works and we have at least some MML construction in place, we can take the step to remove the class and the unit type entry without any loss and with not too much puzzling on the player part. Ideally, custom GE should still work at that point.

psikomonkie:
> I'm not sure I would do that. It does not explain anything to the users. Also, afaik we have no construction in MML in place to replace GE, so what would a player do? Maybe we can do the following: introduce a conversion that loads GE as advanced buildings, so that the GE class can actually be removed at some point while to users this is transparent (and works for customs!). Also, we can start showing GE as advanced buildings in the unit selector (again, nothing the player needs to know). Once this really works and we have at least some MML construction in place, we can take the step to remove the class and the unit type entry without any loss and with not too much puzzling on the player part. Ideally, custom GE should still work at that point.

Correct, there's no construction for this. 

I don't think a fully automatic converter would work well, the current gun emplacements lack structure / armor so when I converted the existing Gun Emplacements to Advanced Buildings I assigned that based on the name - existing ones were already named Light, Heavy, Assault, etc. so I used that to assign some values. 

I do think I could add a simplified editor for Advanced Buildings and let players _load_ Gun Emplacements into the new Advanced Building Editor. The new editor would be a copy of the Gun Emplacement one essentially and add the fields necessary: CF and Armor. The player would need to set those and save them as Advanced Buildings. There's a lot more validation needed if we let players build anything other than Gun Emplacements, but if we just want to do Gun Emplacements then I think I could figure something out on the MML side.

I've set this to draft in the meantime.

SJuliez:
> I don't think a fully automatic converter would work well, the current gun emplacements lack structure / armor so when I converted the existing Gun Emplacements to Advanced Buildings I assigned that based on the name - existing ones were already named Light, Heavy, Assault, etc. so I used that to assign some values.

Please correct me if I'm misreading things... they lack CF/armor, true, but in hand-converting some GE you assigned values (as I read what you wrote). Autoconverting could do just that. Since there are no canon values (GE inherit from their building), any sensible formula would do. It seems there's no way around fixed values if we go towards AB (barring creating customs in MML). *Unless* - we add configuration for CF/armor to the lobby which would bridge this gap. So long as CF and/or armor can be configured without in turn breaking other construction details this may be a good idea.

SJuliez:
Well thinking about this a bit more, it's true that auto-converting can't account for the freedom of putting a GE on a simple house or, alternatively, on a heavy armored fortress. That would require configuration in the lobby

PR #8296, opened 2026-05-28
Fixed Mek Bays Being Able to Load LAM in Air Mek Mode

IllianiBird:
As per https://battletech.com/forums/index.php/topic,69442.msg1845814.html#msg1845814 LAM in Air-Mek mode cannot use MekBays


Sleet01 (approved):
Looks good to me.  Sad, but good.

PR #8300, opened 2026-05-30
Fix #8293: Fix terrain elevation not being accounted for when infantry are jumping into a building

### LINKED ISSUE #8293: [Issue] [50.12] Problems with infantry jumping into buildings and Go Up/Go Down buttons
### Brief Description *

Infantry are at height 0 and have 3 jump. Building is height 4.
Image

Declare Jump movement and click on the building hex. Observe that both the Go Down and Go Up buttons are available. 
Image

Executing the movement with no further actions causes the infantry to jump into the building at the highest level they can reach, a reasonable default. This correctly requires a PSR to not take damage. No issues here. Similarly, pressing Go Down will cause them to jump to a lower level as expected (pressing it once jumps them to level 2, and twice jumps them to level 1). 
Image

However, if you have already pressed Go Down twice (targeting level 1 of the building), the Go Down button will remain available. Pressing it a third time will cause the planned movement to gray out as an invalid move (although it is only barely visible against the building itself in this case). Attempting to execute this movement will jump you to level 1 of the building. This appears to be correct per page 168 of Total Warfare (9th printing), which specifically states that you cannot jump into the ground level of a building; I assume the movement path is being truncated to the last valid step. However, it might be better if the Go Down button was simply grayed out/unavailable once you are targeting level 1 of a building hex.
Image

Alternately, pressing the Go Up button will indicate a valid movement, and will in fact cause them to reach a higher level. Interestingly, this will still require a PSR, despite technically moving them onto the roof (although such a movement shouldn't be possible anyway).
Image

Image

More extreme cases are possible, such as jumping into the nearby height-8 building on a level-2 hex and hitting Go Up as many times as you can. Become Superman and clear 10 levels in a single bound!
Image

Oddly, on a different map, I find myself unable to target level 1 of a height-2 building, only the roof. This may be related to the fact that the building itself rests on height-1 hexes? That seems to be the primary difference between this new map and the previous map. 

First click on target hex:
Image

First Go Down press:
Image

After executing move above:
Image

### 3. Steps to Reproduce

1. Load the provided savegame and attempt to jump into various buildings per the instructions above.

### Attach Files

[plaza_infantry_jump_test.sav.gz](https://github.com/user-attachments/files/28325025/plaza_infantry_jump_test.sav.gz)

[height_2_jump_test.sav.gz](https://github.com/user-attachments/files/28325514/height_2_jump_test.sav.gz)

### Severity *

### MegaMek Suite Version *

0.50.12

### Operating System *

Windows 10

### Java Version *

Java 17.0.11

### Final Verification

psikomonkie:
Fixes #8293 - again
There was a second bug described in #8293 where infantry were unable to jump onto the 2nd floor of a 2 story building, but could jump onto the roof. The issue was because the terrain elevation of the final hex was not being considered when comparing it to the height of the adjacent hex - the logic used to make sure we aren't jumping in through an obstructed window.

### DIFF
--- a/megamek/src/megamek/common/moves/MoveStep.java
+++ b/megamek/src/megamek/common/moves/MoveStep.java
@@ -1737,7 +1737,7 @@ public class MoveStep implements Serializable {
 
         if (isInfantry && isJumping() && stepType == MoveStepType.DOWN) {
             Hex curHex = game.getBoard(boardId).getHex(curPos);
-            if (curHex != null && curHex.containsTerrain(Terrains.BUILDING)) {
+            if (curHex.containsTerrain(Terrains.BUILDING)) {
                 Coords startingPosition = entity.getPosition();
                 Coords adjacentCoords = curPos.translated(curPos.direction(startingPosition));
                 Hex adjacentHex = game.getHex(adjacentCoords, boardId);
--- a/megamek/src/megamek/common/moves/MoveStep.java
+++ b/megamek/src/megamek/common/moves/MoveStep.java
@@ -1737,7 +1737,7 @@ public class MoveStep implements Serializable {
 
         if (isInfantry && isJumping() && stepType == MoveStepType.DOWN) {
             Hex curHex = game.getBoard(boardId).getHex(curPos);
-            if (curHex.containsTerrain(Terrains.BUILDING)) {
+            if (curHex != null && curHex.containsTerrain(Terrains.BUILDING)) {
                 Coords startingPosition = entity.getPosition();
                 Coords adjacentCoords = curPos.translated(curPos.direction(startingPosition));
                 Hex adjacentHex = game.getHex(adjacentCoords, boardId);
--- a/megamek/src/megamek/common/moves/MoveStep.java
+++ b/megamek/src/megamek/common/moves/MoveStep.java
@@ -1736,7 +1736,8 @@ public class MoveStep implements Serializable {
         } // end AERO stuff
 
         if (isInfantry && isJumping() && stepType == MoveStepType.DOWN) {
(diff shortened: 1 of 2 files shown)


megamek/src/megamek/common/moves/MoveStep.java
```diff
@@ -1736,7 +1736,8 @@ private void compileIllegal(final Game game, final Entity entity, final MoveStep
         } // end AERO stuff
 
         if (isInfantry && isJumping() && stepType == MoveStepType.DOWN) {
-            if (game.getBoard(boardId).getHex(curPos).containsTerrain(Terrains.BUILDING)) {
+            Hex curHex = game.getBoard(boardId).getHex(curPos);
```
IllianiBird, line 1739: `getHex` has the `Nullable` tag, and looking at the method itself it can definitely return `null`, which means we'd get an NPE on line 1740

IllianiBird (changes requested):
Spotted a potential NPE. @ me on Discord once it's fixed and we can easily get this merged pronto.

megamek/src/megamek/common/moves/MoveStep.java
psikomonkie: The only time this method'll be used it's inside a try-catch block that ignores exceptions and has the comment:
`// N.B. the pathfinding will try steps off the map.`

The code expects invalid hexes at this point to throw an exception because it means the Step / Path is invalid.

This issue was present prior to this change, this is just splitting it in two instead of having it all in one if statement. I'd like to request not to make this change as part of this PR.

IllianiBird (approved):

PR #4240, opened 2023-03-11
Issue 4213: Allow indirect fire for BA mortar weapons

SJuliez:
(Not CI)

PR #8323, opened 2026-06-08
Allow indirect fire for conventional infantry mortars

HammerGS:
## Summary
  Per TO:AUE (the same rule reference as the 'Mech Mortar / BA Mortar indirect-fire rules), conventional
  infantry whose **Light or Heavy Mortar defines the platoon's final range value** may fire indirectly
  like 'Mech Mortars. Battle Armor mortars already had this (PR #4240); this extends the identical
  mechanism to conventional-infantry mortars, which were previously unable to use mortar-style indirect
  fire (they still required a spotter).

  ## Background
  Indirect mortar fire is gated entirely by the `WeaponType.F_MORTAR_TYPE_INDIRECT` flag plus
  `hasIndirectFire()`. The to-hit code only consults these on `weapon.getType()`, which for a conventional
  platoon is the **range-defining** weapon (the first constructor argument of `InfantryWeaponMounted`).
  Flagging the mortar weapon classes therefore enables mortar-style indirect fire **only when the mortar
  defines the platoon's range** — exactly the rule's "for whom Light or Heavy Mortars define the final
  range value" clause — with no new conditional logic. No engine/gating changes were needed; that
  infrastructure already exists from PR #4240.

  ## Changes
  1. Added `F_MORTAR_TYPE_INDIRECT` and overrode `hasIndirectFire()` → `true` on the four conventional infantry mortar weapons (Light, Heavy, Light Inferno, Heavy Inferno), mirroring the BA mortar pattern.
  2. Overrode `isAlphaStrikeIndirectFire()` → `false` on those weapons. `hasIndirectFire()` defaults this to true, but the conventional-infantry AlphaStrike converter computes damage per-trooper and does not grant the IF ability, so this keeps AlphaStrike output unchanged.
  3. Added `EquipmentTypeLookup` constants for the four mortars and switched their `setInternalName(...)` calls to use them (parity with how PR #4240 handled the BA mortars; also gives `@EquipmentName` verification coverage).

  ## Files Changed
  - `megamek/src/megamek/common/weapons/infantry/support/mortar/InfantrySupportMortarLightWeapon.java` - flag + overrides + lookup constant
  - `megamek/src/megamek/common/weapons/infantry/support/mortar/InfantrySupportMortarHeavyWeapon.java` - flag + overrides + lookup constant
  - `megamek/src/megamek/common/weapons/infantry/support/mortar/InfantrySupportMortarLightInfernoWeapon.java` - flag + overrides + lookup constant
  - `megamek/src/megamek/common/weapons/infantry/support/mortar/InfantrySupportMortarHeavyInfernoWeapon.java` - flag + overrides + lookup constant
  - `megamek/src/megamek/common/equipment/EquipmentTypeLookup.java` - four `@EquipmentName` mortar constants
  - `megamek/unittests/megamek/common/weapons/infantry/support/mortar/InfantryMortarIndirectFireTest.java` - new regression/behavior tests

  ## Testing
  - New `InfantryMortarIndirectFireTest`:
    - Capability granted on all four mortars (flag + `hasIndirectFire`), AS-IF stays off; non-mortar control.
    - Range-defining dispatch: the flag is exposed only when the mortar is the platoon's range-defining weapon.
    - Behavioral (real engine gate): `Compute.indirectAttackImpossible` allows a mortar to fire indirectly even with clear LOS, while an ordinary infantry weapon is blocked — proving the flag drives the gate.
  - `EquipmentTypeLookupTest` passes (new constants resolve to real equipment).
  - Manual in-game verification: a mortar-range-defining platoon fires indirectly without a spotter and with LOS; +2 (no spotter) / +3 (direct) modifiers apply; non-mortar platoons still require a spotter.

  ## Notes / Out of scope
  - Does not change the "infantry cannot clear woods" rule or the airborne/submerged firing restriction (the latter is not flag-gated and behaves identically for all mortar types).

### DIFF
--- a/megamek/src/megamek/common/equipment/EquipmentTypeLookup.java
+++ b/megamek/src/megamek/common/equipment/EquipmentTypeLookup.java
@@ -224,6 +224,14 @@ public class EquipmentTypeLookup {
     @EquipmentName
     public static final String INFANTRY_TAG = "InfantryTAG";
     @EquipmentName
+    public static final String INFANTRY_MORTAR_LIGHT = "InfantryLightMortar";
+    @EquipmentName
+    public static final String INFANTRY_MORTAR_HEAVY = "InfantryHeavyMortar";
+    @EquipmentName
+    public static final String INFANTRY_MORTAR_LIGHT_INFERNO = "InfantryLightMortarInferno";
+    @EquipmentName
+    public static final String INFANTRY_MORTAR_HEAVY_INFERNO = "InfantryHeavyMortarInferno";
+    @EquipmentName
     public static final String VIBRO_SHOVEL = "Vibro-Shovel";
     @EquipmentName
     public static final String DEMOLITION_CHARGE = "Demolition Charge";
--- a/megamek/src/megamek/common/weapons/infantry/support/mortar/InfantrySupportMortarHeavyInfernoWeapon.java
+++ b/megamek/src/megamek/common/weapons/infantry/support/mortar/InfantrySupportMortarHeavyInfernoWeapon.java
@@ -1,6 +1,6 @@
 /*
  * Copyright (C) 2004,2005 Ben Mazur (bmazur@sev.org)
- * Copyright (C) 2010-2025 The MegaMek Team. All Rights Reserved.
+ * Copyright (C) 2010-2026 The MegaMek Team. All Rights Reserved.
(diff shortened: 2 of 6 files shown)


IllianiBird (approved):
Looks good to me

PR #8324, opened 2026-06-08
Add infantry Hitting the Deck rule (TO:AR p.106)

HammerGS:
---
  Summary

  Implements the Hitting the Deck infantry rule (TO:AR p.106), a movement-phase alternative to Digging In: an infantry
  unit may spend its movement to drop prone for a smaller defensive bonus while remaining able to fire. Includes all
  sub-rules (idle→dig-in conversion, Anti-Mek skill penalty, and the field-weapon front-arc restriction).

  Changes

  1. New posture state on Infantry (hitTheDeck + idle/fire tracking), modeled independently from dugIn because an on-deck unit stays able to fire and still takes double damage in the open — the two ways it inverts dig-in.
  2. New action plumbing: MoveStepType.HIT_THE_DECK, MoveCommand.MOVE_HIT_DECK button (auto-generated, infantry-only), MoveStep legality (first/only action, any terrain, can't move after), and server application in MovePathHandler.
  3. Combat effects: +1 to-hit vs. on-deck units (ComputeTerrainMods, flamer-exempt), +1 Anti-Mek Skill Roll on the attacker's leg/swarm attacks (Compute.getAntiMekMods); double damage in the open deliberately preserved.
  4. Sub-rules: 2-idle-turn conversion to dug-in (canDigInFromDeck(), applied in MovePathHandler); field-weapon units restricted to their front arc while on the deck (ConvInfantry.getWeaponArc), with an INFO toast hinting the player to turn-in-place to aim.
  5. UX/discoverability: expanded the tacops_dig_in game option to "TacOps Dig In / Hit the Deck (Infantry)" and rewrote its description to cover both rules; board sprite shows a Deck status; report 5301 ("hits the deck").
  6. Refactors (behavior-preserving): extracted Infantry.clearGroundPostures() (de-duplicates the "movement/loading ends posture" logic) and MovementDisplay.infantryGroundPostureAvailable() (shared Dig In / Hit the Deck button gate).

  Files Changed

  Core / rules
  - common/units/Infantry.java — hit-the-deck state, idle tracking, canDigInFromDeck(), clearGroundPostures()
  - common/units/ConvInfantry.java — front-arc for on-deck field-weapon units
  - common/enums/MoveStepType.java — HIT_THE_DECK
  - common/moves/MoveStep.java — legality + can't-move-after-deck guard
  - common/actions/compute/ComputeTerrainMods.java — +1 to-hit modifier
  - common/compute/Compute.java — +1 Anti-Mek skill roll

  Server
  - server/totalWarfare/MovePathHandler.java — apply/clear posture, dig-in-from-deck conversion
  - server/totalWarfare/TWGameManager.java — report 5301, fired-while-on-deck tracking

  Client / UI
  - client/ui/panels/phaseDisplay/MovementDisplay.java — button gate, action, toast hint, shared helper
  - client/ui/panels/phaseDisplay/commands/MoveCommand.java — MOVE_HIT_DECK
  - client/ui/clientGUI/boardview/sprite/StepSprite.java, .../EntitySprite.java — step icon + Deck status

  Resources (i18n)
  - resources/megamek/client/messages.properties — button label, tooltip, toast, to-hit reason
  - resources/megamek/common/messages.properties — Anti-Mek modifier label
  - resources/megamek/common/options/messages.properties — option rename/description
  - resources/megamek/common/report-messages.properties — report 5301

  Tests
  - unittests/.../common/units/HitTheDeckTest.java (new) — movement legality + state machine
  - unittests/.../common/compute/AntiMekAttackTest.java — +1 Anti-Mek section

  Testing

  - New unit tests: HitTheDeckTest (legal as sole action, illegal to move after, idle counter, fire resets streak, clear-on-move, clear-on-transport) and a hit-the-deck section in AntiMekAttackTest (+1 on leg & swarm, none when not on deck).
  - Full regression sweep of common.units, common.moves, common.compute, common.actions, and server.totalWarfare packages — all passing; clean compile.
  - Manual in-game verification via the TacOps Dig In / Hit the Deck option (button enable, +1 to-hit, field-gun front-arc toast).


megamek/src/megamek/common/moves/MoveStep.java
```diff
@@ -1783,6 +1785,16 @@ private void compileIllegal(final Game game, final Entity entity, final MoveStep
...
+            // Hitting the deck (TO:AR p.106) may only be the unit's sole action and is allowed in any terrain.
+            if (!isInfantry || !isFirstStep()) {
+                return; // only infantry can hit the deck, and only as their first/only action
+            }
+            if (((Infantry) entity).isHitTheDeck()) {
+                return; // already on the deck
+            }
+            isHittingDeck = true;
+            movementType = EntityMovementType.MOVE_NONE;
         }
```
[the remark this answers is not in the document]
HammerGS: Good catch, confirmed. A `DUG_IN_COMPLETE` unit is still eligible to act, so it could take a `HIT_THE_DECK` step and end up both dug in and on the deck. Fixed: `HIT_THE_DECK` is now rejected when `getDugIn() != DUG_IN_NONE` (in addition to the existing `isHitTheDeck()` check), the comment is updated, and a regression test was added (`HitTheDeckTest#cannotHitDeckWhileDugIn`).

megamek/src/megamek/client/ui/panels/phaseDisplay/MovementDisplay.java
```diff
@@ -734,16 +751,21 @@ private void updateButtons() {
...
-              && (infantry.getDugIn() == Infantry.DUG_IN_NONE);
+              && ((infantry.getDugIn() == Infantry.DUG_IN_NONE) || infantry.canDigInFromDeck());
         getBtn(MoveCommand.MOVE_DIG_IN).setEnabled(canDigIn);
 
+        // Infantry - Hitting the deck, TO:AR p.106. Allowed for any infantry (including mechanized) that has not
+        // already hit the deck and has not yet moved this turn (enforced as a first/only action in MoveStep).
+        boolean canHitDeck = infantryGroundPostureAvailable(selectedUnit, gameOptions)
+              && (selectedUnit instanceof Infantry deckInfantry)
+              && !deckInfantry.isHitTheDeck();
+        getBtn(MoveCommand.MOVE_HIT_DECK).setEnabled(canHitDeck);
```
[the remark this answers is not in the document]
HammerGS: Agreed, mirrored the same restriction on the button: `canHitDeck` now also requires `getDugIn() == DUG_IN_NONE`, so a dug-in unit no longer sees an enabled Hit the Deck button. The `MoveStep` guard remains the authoritative check; this keeps the UI consistent with it.

megamek/resources/megamek/common/messages.properties
```diff
@@ -829,6 +829,7 @@ BV.DermalArmor=Dermal Armor:
 Compute.VariableRangeTargetingShort=Variable Range Targeting (Short mode)
 Compute.VariableRangeTargetingLong=Variable Range Targeting (Long mode)
 Compute.EnhancedImaging=Enhanced Imaging
+Compute.HitTheDeck=hit the deck
 Compute.MountainTroops=Mountain Troops
```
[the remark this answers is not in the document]
HammerGS: Done, changed `Compute.HitTheDeck` to title case (Hit the Deck) to match the neighboring `Compute.*` modifier labels (Enhanced Imaging, Mountain Troops).

megamek/src/megamek/client/ui/panels/phaseDisplay/MovementDisplay.java
```diff
@@ -766,6 +766,23 @@ private boolean hasJumpMP() {
...
+     * @param gameOptions the active game options
+     *
+     * @return true if the unit is infantry on a ground map at ground level with the dig in / hit the deck option on
+     */
+    private boolean infantryGroundPostureAvailable(Entity unit, GameOptions gameOptions) {
+        return (unit instanceof Infantry)
+              && gameOptions.booleanOption(OptionsConstants.ADVANCED_TAC_OPS_DIG_IN)
+              && game.isOnGroundMap(unit)
+              && (unit.getAltitude() == 0)
+              && (unit.getElevation() == 0);
```
IllianiBird, line 783: This can only be done at elevation 0? What about on hills?

megamek/src/megamek/client/ui/panels/phaseDisplay/MovementDisplay.java
```diff
@@ -881,16 +898,23 @@ private void updateButtons() {
...
-        // Infantry - Digging in, TO:AR p.106; could add terrain checking and restrict to first action here
-        boolean canDigIn = (selectedUnit instanceof Infantry infantry)
-              && gameOptions.booleanOption(OptionsConstants.ADVANCED_TAC_OPS_DIG_IN)
-              && game.isOnGroundMap(selectedUnit)
-              && (selectedUnit.getAltitude() == 0)
-              && (selectedUnit.getElevation() == 0)
+        // Infantry - Digging in, TO:AR p.106; could add terrain checking and restrict to first action here.
+        // A unit that has hit the deck and stayed idle long enough may also convert directly to dug in.
+        boolean canDigIn = infantryGroundPostureAvailable(selectedUnit, gameOptions)
+              && (selectedUnit instanceof Infantry infantry)
```
IllianiBird, line 904: This will return `true` for BattleArmor is this intentional?

We probably want to return `true` for all infantry types except BA.

megamek/src/megamek/client/ui/panels/phaseDisplay/MovementDisplay.java
```diff
@@ -881,16 +898,23 @@ private void updateButtons() {
...
               && !infantry.isMechanized()
-              && (infantry.getDugIn() == Infantry.DUG_IN_NONE);
+              && ((infantry.getDugIn() == Infantry.DUG_IN_NONE) || infantry.canDigInFromDeck());
         getBtn(MoveCommand.MOVE_DIG_IN).setEnabled(canDigIn);
 
+        // Infantry - Hitting the deck, TO:AR p.106. Allowed for any infantry (including mechanized) that is not
+        // already in a ground posture (dug in or on the deck) and has not yet moved this turn (the first/only action
+        // restriction is enforced in MoveStep).
+        boolean canHitDeck = infantryGroundPostureAvailable(selectedUnit, gameOptions)
+              && (selectedUnit instanceof Infantry deckInfantry)
```
IllianiBird, line 913: Same as last comment re: BA

megamek/src/megamek/common/units/Infantry.java
```diff
@@ -448,16 +477,71 @@ public int getDugIn() {
         return dugIn;
     }
 
+    /**
+     * @return True if this infantry is currently "hitting the deck", TO:AR p.106.
```
IllianiBird: Would be cleaner to write `{@code true}`

megamek/src/megamek/common/units/Infantry.java
```diff
@@ -448,16 +477,71 @@ public int getDugIn() {
...
+     */
+    public boolean isHitTheDeck() {
+        return hitTheDeck;
+    }
+
+    /**
+     * Sets the "hitting the deck" status. Clearing the status (e.g. when the unit moves or is loaded) also resets the
+     * idle-turn counter used for the dig-in conversion.
+     *
+     * @param hitTheDeck True to mark the unit as hitting the deck
```
IllianiBird: And here

IllianiBird (changes requested):
Not sure if the BA thing is an actual issue or no, hitting this with request changes just in case it is. Easy enough to approve if I'm wrong :)

The `True` vs. `{@code true}` is stylistic and not super important. Ignore to apply, no issue either way.

megamek/src/megamek/client/ui/panels/phaseDisplay/MovementDisplay.java
HammerGS: Hills are fine here. For infantry, `getElevation()` is measured relative to the hex surface (which already accounts for the hill's terrain level), so a unit standing on a hill is still at elevation 0. The `getElevation() == 0` gate only filters out infantry that are *above* the surface -- upper building floors, bridges, or airborne (mid-jump / carried VTOL) -- which is the intent. `getAltitude() == 0` covers the aero-altitude case.

megamek/src/megamek/client/ui/panels/phaseDisplay/MovementDisplay.java
HammerGS: This `!isMechanized()` gate is unchanged from `main` -- the PR only hoisted the shared prefix into `infantryGroundPostureAvailable()`; it didn't change who can dig in. Standard (non-mechanized) BA could already reach the Dig In button before this PR, so that's pre-existing behavior rather than something introduced here, and I'd keep it out of scope for the Hit the Deck work. If we ever do want to restrict digging in to conventional infantry, `infantry.isConventionalInfantry()` is the clean gate (BA returns false), but that's better as its own change.

megamek/src/megamek/client/ui/panels/phaseDisplay/MovementDisplay.java
HammerGS: For Hitting the Deck this is intentional. Battle Armor is infantry (`BattleArmor extends Infantry`) and is treated as such for these posture rules, so it's deliberately included. Hitting the deck models the unit dropping prone -- something any infantry type can do -- and the resulting effects (+1 to-hit, +1 Anti-Mek skill roll, double damage in the open) all apply regardless of infantry type. So returning true for BA here is by design rather than an oversight.

megamek/src/megamek/common/units/Infantry.java
HammerGS: Good catch -- will switch to `{@code true}`.

megamek/src/megamek/common/units/Infantry.java
HammerGS: Same here -- will switch to `{@code true}`.

megamek/src/megamek/client/ui/panels/phaseDisplay/MovementDisplay.java
Sleet01: You're going to love this:

Elevation: distance above the underlying hex level in standard 6m increments.
Level: how high the terrain is compared to baseline level of "0" in standard 6m increments.
Altitude: how high Aerospace units are above the baseline altitude of "0" in arbitrary numbers that get bigger as you go higher, which is different from Elevation (but on Low-Altitude maps, level = altitude and Elevation doesn't really exist).

So a hex has a Level, anything in the hex has an elevation (often 0), and any flying Aerospace has an Altitude which is not comparable with Elevation.

So a ground unit that is at the level of the ground on a hex has a Level == hex's level, Elevation == 0, and Altitude == 0, no matter how high up or low down the hex is.

IllianiBird: I hate it here =_=

megamek/src/megamek/client/ui/panels/phaseDisplay/MovementDisplay.java
IllianiBird: `infantry.isConventionalInfantry()` would be a bad call, I think, as it'd exclude other types of infantry such as dismounted pilots or ejected crew

IllianiBird (approved):

PR #8336, opened 2026-06-12
Force Generator Improvements Part 1 - Clans and Tools

HammerGS:
This is the first of a few PR's to improve the Force Generator. Most of the code changes here are to help with generation and fill some holes. The data side focuses on the Clans but future fixes will introduce some new functions and tweaks for Inner Sphere factions. I'll add some of these tools will be gated or removed before the next release.

## Summary

Adds a weight-budget allocator to the ratGenerator Force Generator so faction rules can hit a target weight mix (Assault/Heavy/Medium/Light) per cluster type, instead of relying on random per-star rolls. Also adds Clan naval organization, optional carried-fighter complements, chassis-only WarShip support, a faction-rules XML viewer, and diagnostic CSV output. The faction-rules data that uses these features ships in the companion mm-data PR.

## Changes Made

  - **Weight-budget allocator** (new): a `<weightTarget>` element on a `%CLUSTER%`  force node sets a target weight distribution per unit type (e.g. `A:40,H:30,M:20,L:10`, with `spread` jitter). The allocator reshapes element weight classes before units are picked, so a cluster hits its target on average across many generations. 
  
  Data-gated: a no-op for any cluster without a `<weightTarget>`. Replaces the old weight-emphasis profile mechanism (removed).

  - **Chassis-only WarShips**: a ship referenced by chassis with no model and no RAT availability now resolves by full name from the unit cache, instead of failing to generate (symptom: a named flagship that produced no unit/sprite). Reusable for any hull a faction owns by name without a RAT entry.

  - **Naval organization**: large craft group into Naval Stars; the heaviest hull sorts to commander. Tonnage is read from ModelRecord, since the Entity is still null at the commander-assignment step.

  - **Carried fighter complement** (optional toggle, off by default): fills each large craft's fighter bays, nested Ship -> Star/Squadron -> Point/Flight -> matched fighter pair. Bay capacity from TransportCalculator; degrades gracefully on odd bays.

  - **Diagnostic CSVs**: optional per-cluster weight (`forcegen_weights.csv`) and per-ship naval (`forcegen_warships.csv`) output for tuning faction rules against canon.

  - **Ruleset XML viewer**: a read-only viewer for the faction_rules XML.

  - **Clan rank system** support and a Force Generator command-line flag.

  ## Changes

  1. **WeightTarget.java** (NEW) - weight-target value record (unitType, spread, weight map);
     transient, not serialized.
  2. **WeightBudgetAllocator.java** (NEW) - the allocator: integer budget by largest-remainder
     + soft spread, role-aware distribution to stars. Single source for star-role weight bias.
  3. **WeightBudgetAllocatorTest.java** (NEW) - integer-budget invariants.
  4. **ForceDescriptor.java** - `weightTargets` field; chassis-only resolution
     (getModelName/loadEntities by name); heaviest-hull commander sort; carried-fighter
     complement; null-unitType guard.
  5. **ForceNode.java** - parse `<weightTarget>`; resolve after the rule loop.
  6. **Ruleset.java** - run the allocator at generation step 2; per-cluster weight CSV walk;
     warship CSV append.
  7. **TransportCalculator.java** - fighter bay capacity from MekSummary.
  8. **ForceGenWeightCsv.java / ForceGenWarshipCsv.java** (NEW) - diagnostic output.
  9. **RulesetXmlViewer.java** (NEW) - faction_rules XML viewer.
  10. **RankSystem.java / Ranks.java** (NEW) - rank system support.
  11. **ForceGeneratorOptionsView.java / ForceGeneratorViewUi.java / RandomArmyDialog.java** -
      weight-class binding fix; ship-first ToE render for large craft; carried-fighter checkbox.
  12. **FormationType.java / OptionGroupNode.java / SubForcesNode.java** - supporting fixes
      (formation role/weight handling, element-slot weight spread, subforce ordering).
  13. **MegaMekCommandLineFlag.java / MegaMekCommandLineParser.java / MMConstants.java /
      MegaMek.java** - Force Generator command-line flag.
  14. **messages.properties** - new UI labels and tooltips.

  ## Files Changed

  - `megamek/src/megamek/client/ratgenerator/WeightTarget.java` - NEW
  - `megamek/src/megamek/client/ratgenerator/WeightBudgetAllocator.java` - NEW
  - `megamek/src/megamek/client/ratgenerator/ForceGenWeightCsv.java` - NEW
  - `megamek/src/megamek/client/ratgenerator/ForceGenWarshipCsv.java` - NEW
  - `megamek/src/megamek/client/ratgenerator/ForceDescriptor.java`
  - `megamek/src/megamek/client/ratgenerator/ForceNode.java`
  - `megamek/src/megamek/client/ratgenerator/Ruleset.java`
  - `megamek/src/megamek/client/ratgenerator/FormationType.java`
  - `megamek/src/megamek/client/ratgenerator/OptionGroupNode.java`
  - `megamek/src/megamek/client/ratgenerator/SubForcesNode.java`
  - `megamek/src/megamek/client/ratgenerator/TransportCalculator.java`
  - `megamek/src/megamek/client/ui/dialogs/randomArmy/ForceGeneratorOptionsView.java`
  - `megamek/src/megamek/client/ui/dialogs/randomArmy/ForceGeneratorViewUi.java`
  - `megamek/src/megamek/client/ui/dialogs/randomArmy/RandomArmyDialog.java`
  - `megamek/src/megamek/common/ui/RulesetXmlViewer.java` - NEW
  - `megamek/src/megamek/common/universe/RankSystem.java` - NEW
  - `megamek/src/megamek/common/universe/Ranks.java` - NEW
  - `megamek/src/megamek/common/commandLine/MegaMekCommandLineFlag.java`
  - `megamek/src/megamek/common/commandLine/MegaMekCommandLineParser.java`
  - `megamek/src/megamek/MMConstants.java`
  - `megamek/src/megamek/MegaMek.java`
  - `megamek/src/megamek/resources/megamek/client/messages.properties`
  - `megamek/unittests/megamek/client/ratgenerator/WeightBudgetAllocatorTest.java` - NEW

  ## Testing

  - Unit tests: WeightBudgetAllocatorTest (integer-budget invariants). Full suite passes.
  - Manual: generated multi-touman batches across factions and eras, read the diagnostic
    CSVs, and compared the weight mix to canon:
    - Steel Viper (3060), Clan Wolf (3050), Wolf-in-Exile (3062), Wolf Empire (3145),
      Widowmaker (2830)
    - Verified naval rosters generate with correct hulls, date gates, and flagship by tonnage
    - Verified chassis-only flagships resolve and render (e.g. a faction's single named WarShip)
    - Verified carried-fighter complement nests as Star/Squadron of paired fighters
    - Save/load a generated force


megamek/src/megamek/client/ratgenerator/FormationType.java
```diff
@@ -470,6 +473,16 @@ public List<MekSummary> generateFormation(List<Parameters> params, List<Integer>
...
 
+        LOGGER.info("[ForceGen][Formation] ENTER formation='{}' minWC={} maxWC={} idealRole={} bestEffort={}"
+                    + " networkMask={} paramSets={} totalUnits={}",
+              name, minWeightClass, maxWeightClass, idealRole, bestEffort, networkMask, params.size(),
+              numUnits.stream().mapToInt(Integer::intValue).sum());
+        for (int pi = 0; pi < params.size(); pi++) {
+            Parameters p = params.get(pi);
+            LOGGER.info("[ForceGen][Formation]   param[{}] unitType={} requestedWC={} roles={} moves={} numUnits={}",
+                  pi, p.getUnitType(), p.getWeightClasses(), p.getRoles(), p.getMovementModes(), numUnits.get(pi));
+        }
```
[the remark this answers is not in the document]
HammerGS: Done — downgraded this per-formation trace to DEBUG, so normal runs stay quiet.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -876,21 +1098,33 @@ public ModelRecord generate() {
...
                     entity.setCrew(getCo().createCrew(entity.defaultCrewType()));
                     entity.setExternalIdAsString(UUID.randomUUID().toString());
                     String forceString = getForceString();
                     entity.setForceString(forceString);
+                    LOGGER.info("[ForceGen][ToE] leaf '{}' forceString='{}'{}",
+                          entity.getShortName(), forceString,
+                          forceString.isBlank()
+                                ? " (BLANK - this unit will lose its ToE position; parent="
+                                      + (parent == null ? "null" : "set") + ")"
+                                : "");
```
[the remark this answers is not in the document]
HammerGS: Done — split this: a blank force string now logs at WARN (the actionable case), and the normal per-unit trace is DEBUG.

megamek/src/megamek/client/ui/dialogs/randomArmy/ForceGeneratorViewUi.java
```diff
@@ -247,6 +302,9 @@ public void addChosenUnits(String playerName, ClientGUI clientGui) {
                 e.setDone(true);
                 e.setUnloaded(true);
             }
+            logger.info("[ForceGen][ToE] add-to-game '{}' forceString='{}'{}",
+                  e.getShortName(), e.getForceString(),
+                  e.getForceString().isBlank() ? " (BLANK - ToE structure will be lost)" : "");
```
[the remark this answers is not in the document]
HammerGS: Done — same split here: a blank force string logs at WARN, and the per-unit add-to-game trace is DEBUG.

megamek/src/megamek/client/ratgenerator/Ruleset.java
```diff
@@ -185,38 +187,139 @@ public interface ProgressListener {
...
+                ForceGenWeightCsv.append(cluster.getFaction(), cyear, cluster.getRating(), cwc,
+                      String.join(";", cluster.getFlags()), cluster.tallyWeightClassesByType());
+            }
+        }
+
+        // Large craft (WarShips/DropShips/JumpShips/Space Stations) are absent from the weight CSV
+        // (no L/M/H/A class), so record them separately with their structural path for naval
+        // verification (logs/forcegen_warships.csv): correct galaxy/reserve nesting, no duplicates,
+        // and EMPTY-point detection.
+        ForceGenWarshipCsv.append(fd);
```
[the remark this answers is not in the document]
HammerGS: Done — both CSV writers now early-return unless a new silent `ForceGeneratorDiagnostics` client preference is set (default off, no UI), so no diagnostic files are written in normal use. The verbose [ForceGen] logs were also downgraded to DEBUG.

megamek/src/megamek/client/ratgenerator/WeightBudgetAllocator.java
```diff
@@ -0,0 +1,264 @@
...
+        }
+        Integer echelon = root.getEchelon();
+        Map<Integer, WeightTarget> targets = root.getWeightTargets();
+        if ((echelon != null) && (echelon == CLUSTER_ECHELON) && (targets != null) && !targets.isEmpty()) {
+            allocateCluster(root, targets);
+        }
+        for (ForceDescriptor sub : root.getSubForces()) {
+            allocate(sub);
+        }
+    }
```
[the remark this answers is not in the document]
HammerGS: Good catch — fixed. `allocate()` now also recurses through attached forces (mirroring `collectClusters`), so `<weightTarget>` clusters attached to a galaxy/touman are reshaped.

megamek/src/megamek/client/ratgenerator/TransportCalculator.java
```diff
@@ -253,6 +253,27 @@ private boolean hasBayFor(MekSummary ms, int unitType) {
...
+     */
+    public static int fighterBayCapacity(MekSummary ms) {
+        if (!bayTypeCache.containsKey(ms)) {
+            try {
+                Entity entity = new MekFileParser(ms.getSourceFile(), ms.getEntryName()).getEntity();
+                bayTypeCache.put(ms, countBays(entity));
+            } catch (EntityLoadingException ex) {
+                return 0;
+            }
+        }
```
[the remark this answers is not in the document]
HammerGS: Done — on `EntityLoadingException` we now cache an empty bay map, so a unit that fails to load isn't re-parsed and re-thrown on subsequent calls.

megamek/src/megamek/common/ui/RulesetXmlViewer.java
```diff
@@ -0,0 +1,1169 @@
...
+        rulesetFiles.sort(Comparator.comparing(File::getName));
+        for (File file : rulesetFiles) {
+            if (!documentCache.containsKey(file)) {
+                try {
+                    DocumentBuilder builder = MMXMLUtility.newSafeDocumentBuilder();
+                    Document document = builder.parse(file);
+                    stripWhitespace(document.getDocumentElement());
+                    documentCache.put(file, document);
+                } catch (Exception ex) {
+                    logger.error(ex, "Failed to parse ruleset file {}", file.getName());
```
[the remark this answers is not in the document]
HammerGS: The ruleset XML viewer was a dead end and has been removed entirely, so this no longer applies.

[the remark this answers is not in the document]
HammerGS: Removed — `RulesetXmlViewer` was deleted, so this finding no longer applies.

megamek/resources/megamek/client/messages.properties
```diff
@@ -5269,6 +5276,11 @@ ForceGeneratorDialog.formation=Formation:
 ForceGeneratorDialog.formation.tooltip=<html>The echelon level (size) of the generated force.<br/>Available echelons depend on the selected faction and unit type.<br/>Suffix codes: <b>^</b> = augmented (Clan Nova/Supernova, Blakist Choir, Capellan augmented lance); <b>+</b> = reinforced; <b>-</b> = understrength.</html>
 ForceGeneratorDialog.rating=Equipment Rating:
 ForceGeneratorDialog.rating.tooltip=<html>The unit-equipment quality rating for the force.<br/>Available codes depend on the faction's rating system:<br/><b>Inner Sphere:</b> A (best), B, C, D, F (worst)<br/><b>Star League:</b> A, B, C<br/><b>Clan:</b> Keshik (command/elite), FL (Front Line), SL (Second Line), PG (Provisional Garrison), Sol (Solahma)<br/><b>ComStar/WoB:</b> SB (Shadow Brigade), HS, PG, TP<br/>Some echelons restrict which ratings are valid (e.g., a Clan Galaxy only allows FL/SL).</html>
+ForceGeneratorDialog.search=Search ToE:
```
IllianiBird: Should be `TO&E`. We call it the 'TOE' colloquially but formally it's the TO&E

megamek/resources/megamek/client/messages.properties
```diff
@@ -5286,6 +5298,8 @@ ForceGeneratorDialog.warshipPercentage=WarShip Percentage:
 ForceGeneratorDialog.warshipPercentage.tooltip=<html>Generate WarShip docking-collar capacity for the designated portion of the DropShip fleet.<br/>Independent of JumpShip percentage — combine to allocate fleet composition (e.g., 50% WarShip + 50% JumpShip for split coverage).<br/>Only available for factions with WarShip production in the selected era.</html>
 ForceGeneratorDialog.cargo=Cargo Tonnage:
 ForceGeneratorDialog.cargo.tooltip=<html>Additional cargo tonnage to provision when sizing transport.<br/>Setting a non-zero value forces transport assignment to fire even when all percentages are 0 — useful for generating standalone supply lifts.<br/>0 is the default and adds no cargo requirement.</html>
+ForceGeneratorDialog.fighterComplement=Carried Fighter Complement
+ForceGeneratorDialog.fighterComplement.tooltip=<html>For each WarShip, DropShip, JumpShip, and Space Station in the force, generate the Aerospace Fighters it carries (filled to the ship's fighter bays) and nest them under the ship.<br/>Off by default.</html>
```
IllianiBird, line 5326: Is it `Aerospace` or `AeroSpace` in BattleTech?

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -90,6 +90,12 @@ public class ForceDescriptor {
                                            "XIII", "XIV", "XV", "XVI", "XVII", "XVIII", "XIX", "XX" };
 
     private int index;
+    /**
+     * Unique id of this node within the generated force, assigned by {@link #assignForceIds(int)} and
+     * emitted by {@link #getForceString()}. Must be unique across the whole generated force so the
+     * server does not merge distinct forces that would otherwise share an id. -1 means unassigned.
+     */
+    private int forceStringId = -1;
```
IllianiBird: Having 'String' in a variable assigned to an int is confusing.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -196,6 +207,33 @@ public void generateUnits(ProgressListener l, double progress) {
...
+            roles.removeIf(r -> (r != MissionRole.ARTILLERY)
+                  && (r != MissionRole.MISSILE_ARTILLERY)
+                  && (r != MissionRole.MIXED_ARTILLERY));
+            // Battery uniformity: an artillery formation fields one gun type. The first artillery
+            // node with children picks a single artillery unit and pins it via setUnit, which on a
+            // non-leaf node propagates the model to every descendant. Each element then resolves
+            // that one model by name (generateUnits' getModelRecord rescue), so the whole battery
+            // comes out identical even when the Mek->vehicle fallback changes the unit type.
+            if (models.isEmpty() && chassis.isEmpty() && !subForces.isEmpty()) {
+                ModelRecord arty = generateArtilleryPreferred();
```
IllianiBird: artillery

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -349,13 +406,23 @@ private boolean generateAndAssignFormation(List<ForceDescriptor> subs, boolean c
      *       the given parameters.
      */
     private List<ModelRecord> generateFormation(List<ForceDescriptor> subs, int networkMask, int numGroups) {
+        // Collect the weight classes the force tree assigned to this formation's elements. Passing
+        // them to the formation builder keeps it within the lance's intended weight profile; left
+        // null it would pick any weight the FormationType itself allows (e.g. a light Mek in a
+        // Heavy/Assault Hunter lance).
+        Set<Integer> formationWeightClasses = new TreeSet<>();
+        for (ForceDescriptor sub : subs) {
+            if (sub.useWeightClass() && (null != sub.getWeightClass()) && (sub.getWeightClass() >= 0)) {
```
IllianiBird: I feel like there must be a const somewhere for Weight Class `0`, it so it'd be best to use that here

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -810,12 +889,135 @@ public void generate(String level) {
     }
 
     public ModelRecord generate() {
```
IllianiBird: Please add the `@Nullable` tag here. We should be adding that tag to any method that can return `null` as part of normal operations. Please also verify that **all** calls to `generate()` include `null` protection, this is especially vital as this method is public.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -810,12 +889,135 @@ public void generate(String level) {
...
+        // Artillery preference: before the rating ladder below relaxes the mission role and
+        // backfills a non-artillery unit, try a real artillery unit. Front-line (Mek) prefers an
+        // artillery BattleMek then drops to an artillery combat vehicle; second-line (Tank) stays
+        // vehicle. Only if no artillery unit exists for this faction and year do we fall through and
+        // let the normal ladder field a non-artillery unit of the original type as a last resort.
+        // Skipped when a model is already pinned (battery uniformity), so the pinned gun wins and
+        // every element resolves to the same unit instead of re-picking its own.
+        if (models.isEmpty() && ((unitType == UnitType.MEK) || (unitType == UnitType.TANK))
+              && roles.contains(MissionRole.ARTILLERY)) {
+            ModelRecord arty = generateArtilleryPreferred();
```
IllianiBird: artillery

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -810,12 +889,135 @@ public void generate(String level) {
...
+        }
+        // Artillery preference: before the rating ladder below relaxes the mission role and
+        // backfills a non-artillery unit, try a real artillery unit. Front-line (Mek) prefers an
+        // artillery BattleMek then drops to an artillery combat vehicle; second-line (Tank) stays
+        // vehicle. Only if no artillery unit exists for this faction and year do we fall through and
+        // let the normal ladder field a non-artillery unit of the original type as a last resort.
+        // Skipped when a model is already pinned (battery uniformity), so the pinned gun wins and
+        // every element resolves to the same unit instead of re-picking its own.
+        if (models.isEmpty() && ((unitType == UnitType.MEK) || (unitType == UnitType.TANK))
+              && roles.contains(MissionRole.ARTILLERY)) {
```
IllianiBird, line 907: It's worth checking that missile artillery and mixed artillery still have the artillery role. One to ask Dgorsman.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -810,12 +889,135 @@ public void generate(String level) {
...
+                return arty;
+            }
+        }
+        // Equipment-rating fallback ladder: try the force's own rating first and, only when
+        // generation comes up empty, step down to progressively worse ratings (never better).
+        // A rating-C force may field C/D/F equipment when nothing matches at its own rating,
+        // but never the A/B grades reserved for better-equipped commands.
+        List<String> failureTrace = new ArrayList<>();
+        for (String ratGenRating : ratingFallbackList()) {
+            ModelRecord mr = generateAtRating(ratGenRating, failureTrace);
```
IllianiBird: modelRecord, please

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -205,6 +243,16 @@ public void generateUnits(ProgressListener l, double progress) {
             }
             if (null != mr) {
```
IllianiBird: Can we rename `mr` to `modelRecord`, I know it's not your code but I went through most of this class review before discovering `mr` wasn't `missionRole`

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -810,12 +889,135 @@ public void generate(String level) {
...
+        if (unitType != null && unitType == UnitType.MEK) {
+            if (models.isEmpty()) {
+                // Genuine failure: no pinned model to fall back on, so the caller's
+                // getModelRecord(getModelName()) rescue (generateUnits) cannot recover. Log the full trace.
+                LOGGER.debug("[ForceGen][Weight] generate() FAILED requestedWeight={} -> no unit found."
+                            + " element: faction={} unitType={} year={} echelon={} roles={} movementModes={}"
+                            + " models={} chassis={}",
+                      weightClass, faction, unitType, year, echelon, roles, movementModes, models, chassis);
+                for (String line : failureTrace) {
+                    LOGGER.debug("[ForceGen][Weight]   attempt: {}", line);
```
IllianiBird, line 935: A recommendation: while debug _is_ the right log level for this class (and others), something I've used to great effect is to have debug logs set as `info` initially. Then, once we've had a couple of releases and we're happy everything is working, going back and lowering from info to debug.

The benefit here is that players won't have debug enabled, so by temporarily setting it as 'info' we can more easily tell what's going on when issues are reported.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -810,12 +889,135 @@ public void generate(String level) {
...
+     */
+    private ModelRecord generateArtilleryPreferred() {
+        // Front-line (Mek) prefers an artillery Mek, then an artillery vehicle. Second-line (Tank)
+        // stays vehicle, honoring the "front line = Mek, otherwise = vehicle" rule.
+        int[] order = (unitType == UnitType.TANK)
+              ? new int[] { UnitType.TANK }
+              : new int[] { UnitType.MEK, UnitType.TANK };
+        for (int candidateType : order) {
+            for (String ratGenRating : ratingFallbackList()) {
+                ModelRecord mr = generateArtilleryUnit(candidateType, ratGenRating);
```
IllianiBird: modelRecord (or better, something more descriptive such as `artilleryUnitRecord`

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -810,12 +889,135 @@ public void generate(String level) {
...
+    /**
+     * Front-line artillery fallback used by {@link #generate()}: looks for a true artillery unit, preferring an
+     * artillery BattleMek and dropping to an artillery combat vehicle, across the equipment-rating ladder. Weight class
+     * is intentionally NOT constrained - artillery hulls have fixed tonnages, so the artillery role takes priority over
+     * the star's rolled weight. Returns {@code null} when no artillery unit of either type exists for this faction and
+     * year, leaving {@link #generate()} to relax the role and field a non-artillery Mek as a last resort.
+     *
+     * @return an artillery unit of a preferred type, or {@code null} if none is available
+     */
+    private ModelRecord generateArtilleryPreferred() {
```
IllianiBird: Needs the `@Nullable` tag and please triple-check that all calls to this method include null safeguards

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -810,12 +889,135 @@ public void generate(String level) {
...
+        UnitTable table = UnitTable.findTable(getFactionRec(),
+              candidateType,
+              getYear(),
+              ratGenRating,
+              new ArrayList<>(),
+              ModelRecord.NETWORK_NONE,
+              movementModes,
+              EnumSet.of(MissionRole.ARTILLERY),
+              2);
+        MekSummary ms = table.generateUnit();
```
IllianiBird: mekSummary

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -810,12 +889,135 @@ public void generate(String level) {
...
+
+    /**
+     * Builds the equipment-rating fallback ladder for {@link #generate()}: the force's own resolved rating followed by
+     * each progressively worse rating in the faction's rating system. Generation tries each in order and stops at the
+     * first that yields a unit, so worse ratings act only as a safety net - the force never fields equipment better
+     * than its assigned rating.
+     */
+    private List<String> ratingFallbackList() {
+        String startRating = ratGeneratorRating();
+        Ruleset rs = Ruleset.findRuleset(this);
```
IllianiBird: ruleset

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
IllianiBird: Would you mind renaming this to forceDescriptor?

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -916,15 +1158,66 @@ private String getForceString() {
...
+            forceStringId = nextId++;
+        }
+        for (ForceDescriptor sub : subForces) {
+            nextId = sub.assignForceIds(nextId);
+        }
+        for (ForceDescriptor att : attached) {
+            nextId = att.assignForceIds(nextId);
+        }
+        return nextId;
+    }
```
IllianiBird, line 1222: Let's lose the truncated variable names

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -1136,6 +1429,19 @@ private int rank(ForceDescriptor fd) {
...
             }
+            // Large craft (WarShips/DropShips/JumpShips/Space Stations) have no L/M/H/A weight
+            // class, so rank them by tonnage: the heaviest vessel in a naval star becomes its
+            // command vessel (assignCommanders assigns the CO to forces[0]). The entity is not
+            // loaded yet when commanders are assigned, so read tonnage from the model record.
+            Integer largeCraftType = fd.getUnitType();
+            if ((largeCraftType != null) && ((largeCraftType == UnitType.WARSHIP)
+                  || (largeCraftType == UnitType.DROPSHIP) || (largeCraftType == UnitType.JUMPSHIP)
+                  || (largeCraftType == UnitType.SPACE_STATION))) {
+                ModelRecord mr = RATGenerator.getInstance().getModelRecord(fd.getModelName());
```
IllianiBird: modeRecord

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -1136,6 +1429,19 @@ private int rank(ForceDescriptor fd) {
...
+                if ((mr != null) && (mr.getMekSummary() != null)) {
+                    retVal += (int) (mr.getMekSummary().getTons() / 1000);
```
IllianiBird: This is going to return 0 for most units:

- 100 / 1000 = 0.1
- 10 / 1000 = 0.01 

However, because we're casting to `int` anything after the decimal point is discarded, so we're just going to return `0` always, unless the unit is a thousand tons or more, and even then it'd need to be 2000 tons to return 2, 3000 for 3, etc.

That could be what you're wanting here, but it's unclear and if it _is_ intentional it needs comments explaining so and why we're doing it this way. Elsewise someone is going to come in, think this is a bug, and 'correct' it by mistake. Like what happened last year (2024?).

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -1461,6 +1778,75 @@ public ArrayList<Object> getAllChildren() {
...
+     * <p>LandAirMeks are counted as Meks ({@code Entity.isMek()} is true for them). Non-Mek
+     * elements (vehicles, infantry, fighters) are ignored.</p>
+     *
+     * @return an int array indexed by {@link EntityWeightClass} constant
+     *       ({@code 0 = WEIGHT_ULTRA_LIGHT} … {@code 5 = WEIGHT_SUPER_HEAVY}); each slot holds the
+     *       number of Mek elements at that weight class
+     */
+    public int[] tallyMekWeightClasses() {
+        int[] counts = new int[EntityWeightClass.WEIGHT_SUPER_HEAVY + 1];
+        tallyMekWeightClasses(counts);
```
IllianiBird, line 1807: I'm not sure about this, this feels like we're just waiting on an out of index error

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -1461,6 +1778,75 @@ public ArrayList<Object> getAllChildren() {
...
+        return counts;
+    }
+
+    private void tallyMekWeightClasses(int[] counts) {
+        Entity leafEntity = getEntity();
+        if (leafEntity != null && leafEntity.isMek()) {
+            int wc = leafEntity.getWeightClass();
```
IllianiBird: weightClass

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -1461,6 +1778,75 @@ public ArrayList<Object> getAllChildren() {
...
+            if (wc >= 0 && wc < counts.length) {
+                counts[wc]++;
+            }
+        }
+        for (ForceDescriptor sub : subForces) {
+            sub.tallyMekWeightClasses(counts);
+        }
+        for (ForceDescriptor att : attached) {
```
IllianiBird: More truncated variable names

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -1461,6 +1778,75 @@ public ArrayList<Object> getAllChildren() {
...
+    public Map<Integer, int[]> tallyWeightClassesByType() {
+        Map<Integer, int[]> byType = new HashMap<>();
+        tallyWeightClassesByType(byType);
+        return byType;
+    }
+
+    private void tallyWeightClassesByType(Map<Integer, int[]> byType) {
+        Entity leafEntity = getEntity();
+        if (leafEntity != null) {
+            int ut = leafEntity.getUnitType();
```
IllianiBird: unitType

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -1461,6 +1778,75 @@ public ArrayList<Object> getAllChildren() {
...
+                    byType.computeIfAbsent(ut, k -> new int[EntityWeightClass.WEIGHT_SUPER_HEAVY + 1])[wc]++;
+                }
+            }
+        }
+        for (ForceDescriptor sub : subForces) {
+            sub.tallyWeightClassesByType(byType);
+        }
+        for (ForceDescriptor att : attached) {
+            att.tallyWeightClassesByType(byType);
+        }
```
IllianiBird, line 1859: truncated variable names

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -1521,6 +1909,29 @@ public String parseName() {
...
+        } else {
+            suffix = switch (n % 10) {
+                case 1 -> "st";
+                case 2 -> "nd";
+                case 3 -> "rd";
+                default -> "th";
+            };
+        }
+        return n + suffix;
+    }
```
IllianiBird, line 1945: These should be in a resource bundle, as they may differ based on language

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -1825,6 +2252,144 @@ public void setAttached(ArrayList<ForceDescriptor> attached) {
 
     public void addAttached(ForceDescriptor fd) {
```
IllianiBird: Let's rename this param

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -1825,6 +2252,144 @@ public void setAttached(ArrayList<ForceDescriptor> attached) {
...
+            // attached under the carrier so the ToE reads: Ship -> Star/Squadron -> Point/Flight ->
+            // fighters. The two fighters in a Point (Clan) or Flight (IS) are the SAME model: a Point
+            // is a matched pair. Different Points within a Star may differ. A Clan aero Star is 5
+            // Points of 2 = 10; an IS aero Squadron is 3 Flights of 2 = 6.
+            boolean clan = (carrier.getFactionRec() != null) && carrier.getFactionRec().isClan();
+            int pointSize = 2;
+            int pointsPerGroup = clan ? 5 : 3;
+            int groupSize = pointsPerGroup * pointSize;
+            String groupLabel = clan ? "Star" : "Squadron";
+            String pointLabel = clan ? "Point" : "Flight";
```
IllianiBird, line 2322: Player-facing hardcoded Strings

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -1825,6 +2252,144 @@ public void setAttached(ArrayList<ForceDescriptor> attached) {
...
+                    // (assignCommanders sets a force's CO from its lead subForce). Attach the rest:
+                    // the CO-reorder only sorts SUBFORCES, so with one subForce nothing scrambles, and
+                    // getAllChildren() (subForces + attached) still nests every Point in creation order.
+                    ForceDescriptor point = group.createChild(pointIndex);
+                    point.getModels().clear();
+                    point.getChassis().clear();
+                    point.setUnitType(UnitType.AEROSPACE_FIGHTER);
+                    point.setName(pointLabel + " " + (pointIndex + 1));
+                    point.setEchelon(pointEchelon);
+                    point.setCoRank(16);
```
IllianiBird, line 2362: I worry this is going to break for player-made ranks or any of the _weird_ ranking systems we have that don't have all the normal ranks.

Basically, unless the rank system explicitly has a rank 16 this will break. You can take a look at `AutoAssignRankForCompanyGenerator` in MekHQ for how I got around this issue.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
```diff
@@ -1825,6 +2252,144 @@ public void setAttached(ArrayList<ForceDescriptor> attached) {
...
+                    }
+                    pointIndex++;
+
+                    int pointTarget = Math.min(pointSize, groupTarget - producedInGroup);
+                    for (int k = 0; k < pointTarget; k++) {
+                        ForceDescriptor fighter = point.createChild(point.getSubForces().size());
+                        fighter.setUnitType(UnitType.AEROSPACE_FIGHTER);
+                        fighter.setUnit(RATGenerator.getInstance().getModelRecord(fighterSummary.getName()));
+                        fighter.setEchelon(1);
+                        fighter.setCoRank(31);
```
IllianiBird, line 2376: Similar issue here, I'm not going to comment on any more instances

megamek/src/megamek/client/ratgenerator/ForceGenWarshipCsv.java
```diff
@@ -0,0 +1,159 @@
...
+            }
+            String out = writeHeader ? (HEADER + "\n" + rows) : rows.toString();
+            Files.writeString(FILE, out, StandardCharsets.UTF_8,
+                  StandardOpenOption.CREATE, StandardOpenOption.APPEND);
+        } catch (IOException ex) {
+            LOGGER.warn("Could not append force-gen warship rows to {}: {}", FILE, ex.getMessage());
+        }
+    }
+
+    private static void collect(ForceDescriptor fd, String path, StringBuilder rows) {
```
IllianiBird: I'm not going to comment on any more abbreviated, truncated, or single letter variable or param names. Just imagine I did :D

megamek/src/megamek/client/ratgenerator/ForceGenWeightCsv.java
```diff
@@ -0,0 +1,132 @@
...
+                  safe(rolledWeight),
+                  safe(clusterFlags),
+                  safe(UnitType.getTypeName(entry.getKey())),
+                  Integer.toString(total),
+                  cell(w, EntityWeightClass.WEIGHT_ULTRA_LIGHT),
+                  cell(w, EntityWeightClass.WEIGHT_LIGHT),
+                  cell(w, EntityWeightClass.WEIGHT_MEDIUM),
+                  cell(w, EntityWeightClass.WEIGHT_HEAVY),
+                  cell(w, EntityWeightClass.WEIGHT_ASSAULT),
+                  cell(w, EntityWeightClass.WEIGHT_SUPER_HEAVY))).append("\n");
```
IllianiBird: What about units that don't fall into these categories? Such as AeroSpace and Support Vees.

megamek/src/megamek/client/ratgenerator/FormationType.java
```diff
@@ -493,7 +506,25 @@ public List<MekSummary> generateFormation(List<Parameters> params, List<Integer>
...
+                LOGGER.debug("[ForceGen][Formation]   weightIntersect: requested=[] formationRange={} -> final={}"
+                      + " (no caller weight; using formation range)", formationRange, p.getWeightClasses());
+            } else {
+                List<Integer> intersection = requested.stream()
+                      .filter(formationRange::contains)
+                      .collect(Collectors.toList());
+                p.setWeightClasses(intersection.isEmpty() ? formationRange : intersection);
+                LOGGER.debug("[ForceGen][Formation]   weightIntersect: requested={} formationRange={} -> final={}{}",
+                      requested, formationRange, p.getWeightClasses(),
+                      intersection.isEmpty() ? " (EMPTY intersection; fell back to formation range)" : "");
```
IllianiBird: Any lambda more than a couple lines long really should be in a utility method. It makes for far more easier to read code

megamek/src/megamek/client/ratgenerator/FormationType.java
```diff
@@ -1772,15 +1847,36 @@ private static void createFastAssaultLance() {
...
+              "Heavy+"));
+        // Campaign Operations (Assault Lance base, inherited by Hunter): at least one Juggernaut OR
+        // two Snipers. Encoded as a paired-OR pair of CountConstraints.
+        Constraint c = new CountConstraint(1, ms -> ms.getRole() == JUGGERNAUT, "Juggernaut");
+        c.setPairedWithNext(true);
+        ft.otherCriteria.add(c);
+        c = new CountConstraint(2, ms -> ms.getRole() == SNIPER, "Sniper");
+        c.setPairedWithPrevious(true);
+        ft.otherCriteria.add(c);
+        // Hunter-specific rule: at least 50% of the units must be Ambushers or Juggernauts.
```
IllianiBird, line 1884: I know I said I wouldn't comment on any more variable names, but it's _really_ bad here

megamek/src/megamek/client/ratgenerator/Ruleset.java
```diff
@@ -185,38 +187,139 @@ public interface ProgressListener {
...
+        logger.debug("[ForceGen][Weight] processRoot ENTER: faction={} echelon={} unitType={} rating={} " +
+                    "weightClass={} ({})",
+              fd.getFaction(), fd.getEchelon(), fd.getUnitType(), fd.getRating(),
+              fd.getWeightClass(),
+              fd.getWeightClassCode().isEmpty() ? "RANDOM - ruleset will roll one" : fd.getWeightClassCode());
         defaults.apply(fd);
         // save the setting so it can be restored after assigning names
         String rngFaction = RandomNameGenerator.getInstance().getChosenFaction();
 
         buildForceTree(fd, l, 0.05);
```
IllianiBird: There's a few instances of magic numbers in this block of code, can we more them to use consts?

megamek/src/megamek/common/universe/RankSystem.java
IllianiBird, line 1: Do we want to create a new RankSystem class unique to MegaMek, or do we want to just port the MekHQ RankSystem class into MegaMek and have both programs use the same class? I worry having two RankSystem classes is going to lead to confusion and error.

megamek/src/megamek/client/ratgenerator/WeightBudgetAllocator.java
```diff
@@ -0,0 +1,269 @@
...
+        m.put("Striker/Cavalry", 2.0);
+        m.put("Light Battle", 2.0);
+        m.put("Ranger", 2.0);
+        m.put("Sweep", 1.8);
+        m.put("Probe", 1.8);
+        m.put("Pursuit", 1.6);
+        m.put("Light Fire", 1.6);
+        m.put("Recon", 1.5);
+        m.put("Light Striker/Cavalry", 1.4);
+        m.put("Light Recon", 1.2);
```
IllianiBird: Are these player facing Strings?

megamek/src/megamek/client/ui/dialogs/randomArmy/ForceGeneratorOptionsView.java
IllianiBird, line 1: There's a lot of hardcoded size values in this class. Instead we should be using `UIUtil.scaleForGUI` so that the gui scales with the players' gui scaling settings.

megamek/src/megamek/client/ui/dialogs/randomArmy/ForceGeneratorViewUi.java
```diff
@@ -519,10 +668,61 @@ public void valueForPathChanged(TreePath arg0, Object arg1) {
...
+              Map.entry(34, "Captain"),
+              Map.entry(35, "Major"),
+              Map.entry(37, "Lt. Colonel"),
+              Map.entry(38, "Colonel"),
+              Map.entry(39, "Lt. General"),
+              Map.entry(42, "Maj. General"),
+              Map.entry(43, "General"),
+              Map.entry(46, "Loremaster"),
+              Map.entry(47, "saKhan"),
+              Map.entry(48, "Khan"));
```
IllianiBird, line 690: Hardcoded player-facing strings

megamek/src/megamek/common/universe/Ranks.java
IllianiBird, line 1: Similar to RankSystems, we should probably try to reuse the existing Ranks class from mhq

megamek/src/megamek/common/universe/Ranks.java
```diff
@@ -0,0 +1,211 @@
...
+          "DCMS",  // 8  RANK_DC    (Draconis Combine Mustered Soldiery)
+          "CLAN",  // 9  RANK_CLAN
+          "CG",    // 10 RANK_CS    (ComStar Guards)
+          "WOBM",  // 11 RANK_WOB   (Word of Blake Militia)
+          null,    // 12 RANK_CUSTOM — MekHQ-side user customs, no XML entry
+          "MAF",   // 13 RANK_MOC   (Magistracy Armed Forces)
+          "TDF",   // 14 RANK_TC    (Taurian Defense Force)
+          "MHAF",  // 15 RANK_MH    (Marian Hegemony Armed Forces)
+          "AMC",   // 16 RANK_OA    (Outworlds Alliance Militia Corps)
+          "KA"     // 17 RANK_FRR   (Kungsarme, Free Rasalhague Republic)
```
IllianiBird, line 94: I'd prefer it if we added default rank system to the faction entry in their xml, otherwise we're going to need to remember to maintain this list any time we add a new major faction, and we are **not** going to remember to do that.

IllianiBird (changes requested):
I went ahead and reviewed this as requested, for the most part it's good. There are a few questions I have that definitely need answering before we can merge.

The other issue is my continuing holy war against variable names. There's a few instances here where the code reads like this:

```
c = a * d;
d = 6.0;
Person z = force.getZPerson()
```

That's an exaggeration, but I definitely had a harder time reviewing because of chained single letter and abbreviated variable names. The Ruleset class was a particularly bad offender.

I also feel like, core functionality like this, could do with a few more tests. But that's optional.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: Was confirmed in development and testing.

megamek/resources/megamek/client/messages.properties
HammerGS: Fixed -> "Search TO&E:".

megamek/resources/megamek/client/messages.properties
HammerGS: It's "Aerospace" (one word, lowercase s) in BattleTech, which is what the string already uses. No change needed.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: Renamed the field to `forceId` so the name no longer implies a String.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: Renamed `arty` -> `artilleryUnitRecord`.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: Good catch - switched the `>= 0` checks in the new code to `>= EntityWeightClass.WEIGHT_ULTRA_LIGHT` (the 0 constant).

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: Added `@Nullable`. Verified all 7 internal `generate()` call sites null-check the result before use (generateUnits also falls back to the pinned model on null).

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: Renamed `arty` -> `artilleryUnitRecord`.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: Done -> `modelRecord`.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: Done - renamed `mr` -> `modelRecord` here and in the other generate paths in this class.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: Fair point on info-then-lower. I kept these at DEBUG because they fire per-unit/per-formation in a hot path (Copilot flagged the volume too), but I split out the actionable case: a blank force string now logs at WARN so it surfaces without debug. Happy to bump specific lines to INFO for a release or two if you'd prefer.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: Renamed to `artilleryUnitRecord`.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: Added `@Nullable`. Both callers (`generate()` and `generateUnits`) null-check before use.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: Renamed to `mekSummary`.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: Renamed to `ruleset`.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: Renamed `fd` -> `workingCopy` (it is the working copy of `this` used for the fallback attempts). Also cleaned the inner `ms`/`wcs`/`wtIndex`/`altWeights` in the same method.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: Done -> `attachedForce`.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: Renamed to `modelRecord`.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: It is intentional, and I made that explicit: extracted a `TONS_PER_NAVAL_RANK_POINT = 1000` constant with a comment. Large craft span hundreds to millions of tons, so raw tonnage would swamp the single-digit experience/weight terms; bucketing by thousands keeps them comparable while still ordering heaviest-first. Sub-1000-ton DropShips collapse to 0 on purpose - they are never the command vessel.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: The write is bounds-guarded: it only does `counts[leafWeightClass]++` inside `if (leafWeightClass >= 0 && leafWeightClass < counts.length)`, and the array is sized `WEIGHT_SUPER_HEAVY + 1`. An out-of-range weight class is skipped, not written. The by-type tally guards the same way.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: Renamed to `leafWeightClass` (avoids shadowing the `weightClass` field).

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: Done -> `leafWeightClass`, `attachedForce`.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: Renamed to `leafUnitType` (avoids shadowing the `unitType` field).

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: Done -> `leafUnitType`, `leafWeightClass`, `attachedForce`.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: These are English ordinal suffixes (st/nd/rd/th) for Clan unit designations, which are English in canon. A proper i18n of ordinals is a language-specific job; I would rather scope it into a dedicated localization pass than half-do it here. Flagging for follow-up.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: Renamed to `forceDescriptor`.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: Agreed these are player-facing (Star/Squadron/Point/Flight). They are BattleTech doctrine terms; I would group all the Force Generator player strings into one localization pass rather than piecemeal. Noting for follow-up.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: The rank resolution is null-safe end to end, so a missing rank degrades to a label rather than breaking: `RankSystem.nameAt` bounds-checks the index and returns null, `Ranks.resolveRankName` returns an `Optional`, and `commanderPrefix` falls back DEFAULT_RANK_TITLES -> a generic role label. A rank system without rank 16 just renders the role name (e.g. "Point: ") instead of a rank title. Thanks for the AutoAssignRankForCompanyGenerator pointer - good reference if we ever want true clamping.

megamek/src/megamek/client/ratgenerator/ForceDescriptor.java
HammerGS: Same as above - resolution is null-safe, so a missing rank 31 degrades to the role label, no break.

megamek/src/megamek/client/ratgenerator/ForceGenWarshipCsv.java
HammerGS: Took the hint :) Did a rename pass across the touched ratgen files: `fd`->`forceDescriptor`, `en`->`entity`, `ut`->`unitType`, `ms`->`mekSummary`, `wc`->`weightClass`, `vn`->`valueNode`, etc. Left the package-wide `fd`/`wn`/`ms` idioms only where the whole file already uses them.

megamek/src/megamek/client/ratgenerator/ForceGenWeightCsv.java
HammerGS: This is a diagnostic CSV (gated behind the silent ForceGeneratorDiagnostics pref, off by default). Aerospace fighters ARE tallied - `tallyWeightClassesByType` includes AEROSPACE_FIGHTER, so they get their own row. Support vehicles and other non-L/M/H/A types are out of scope for the weight-mix tuning this file exists for; large craft are recorded separately in forcegen_warships.csv.

megamek/src/megamek/client/ratgenerator/FormationType.java
HammerGS: Extracted the lambda body into `applyFormationWeightClasses(parameters, groundRange, airRange)` - the forEach is now a one-liner.

megamek/src/megamek/client/ratgenerator/FormationType.java
HammerGS: Fixed the real smell - the reused `Constraint c` is now `juggernautConstraint` / `sniperConstraint`. Left `ft`/`ms` as-is since all ~40 lance builders in this file share that idiom.

megamek/src/megamek/client/ratgenerator/Ruleset.java
HammerGS: Pulled the progress-bar fractions into named constants (`PROGRESS_BUILD_TREE`, `PROGRESS_GENERATE_UNITS`, `PROGRESS_LOAD_ENTITIES`, `PROGRESS_FINALIZE`).

megamek/src/megamek/common/universe/RankSystem.java
HammerGS: Fair concern. I made a lightweight MM-side RankSystem because MekHQ's pulls in MekHQ-only dependencies that MegaMek cannot depend on. A single shared class (living in megamek, used by both) is the right end state, but that is a bigger cross-module refactor I would rather do as its own PR. Your call on whether to block this on it.

megamek/src/megamek/client/ratgenerator/WeightBudgetAllocator.java
HammerGS: No - these keys are internal `FormationType.getName()` values used only as a lookup into the preference map; they are never displayed. They have to match the formation names verbatim, which is why they are literals (with a comment that they mirror the sim's CATEGORY_WEIGHT_PREF).

megamek/src/megamek/client/ui/dialogs/randomArmy/ForceGeneratorOptionsView.java
HammerGS: Switched the hardcoded sizes to `UIUtil.scaleForGUI` (the summary scroll pane preferred size and the GridBag insets).

megamek/src/megamek/client/ui/dialogs/randomArmy/ForceGeneratorViewUi.java
HammerGS: This map is a last-resort fallback used only when the localizable ranks.xml has no entry for a rank; the primary path is data-driven via Ranks. I will fold these fallbacks into the localization follow-up with the other Force Gen strings.

megamek/src/megamek/common/universe/Ranks.java
HammerGS: Same as RankSystem - MekHQ's Ranks has MekHQ-side dependencies, so I added a minimal read-only loader here. Consolidating into one shared class is the right end state but is a separate cross-module change. Deferring to you on sequencing.

megamek/src/megamek/common/universe/Ranks.java
HammerGS: Agreed - a per-faction default in the faction XML is the maintainable answer; a hardcoded index list will rot. I kept the list for this PR to avoid a data-schema change mid-feature, but I am on board with moving it into faction data as a follow-up.

megamek/src/megamek/client/ratgenerator/SubForcesNode.java
```diff
@@ -95,42 +98,57 @@ public ArrayList<ForceDescriptor> generateSubForces(ForceDescriptor forceDescrip
...
+                              && i == valueNode.getNum() / 2) {
                             continue;
                         }
                         ForceDescriptor sub = forceDescriptor.createChild(i);
-                        if (vn.getContent().endsWith("+")) {
+                        if (valueNode.getContent().endsWith("+")) {
                             sub.setSizeMod(ForceDescriptor.REINFORCED);
-                            sub.setEchelon(Integer.parseInt(vn.getContent().replace("+", "")));
-                        } else if (vn.getContent().endsWith("-")) {
+                            sub.setEchelon(Integer.parseInt(valueNode.getContent().replace("+", "")));
```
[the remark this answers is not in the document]
HammerGS: Addressed - the echelon parses in this method now go through a `parseEchelon(...)` helper that uses `MathUtility.parseInt` with a fallback (logs a warning and defaults to 0) instead of `Integer.parseInt`, so malformed ruleset content no longer throws. The +/- size-suffix handling is unchanged.

megamek/src/megamek/client/ratgenerator/SubForcesNode.java
```diff
@@ -95,42 +98,57 @@ public ArrayList<ForceDescriptor> generateSubForces(ForceDescriptor forceDescrip
...
+                        } else if (valueNode.getContent().endsWith("-")) {
                             sub.setSizeMod(ForceDescriptor.UNDERSTRENGTH);
-                            sub.setEchelon(Integer.parseInt(vn.getContent().replace("-", "")));
+                            sub.setEchelon(Integer.parseInt(valueNode.getContent().replace("-", "")));
```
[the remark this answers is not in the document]
HammerGS: Addressed - now routed through `parseEchelon(...)` (MathUtility.parseInt with a logged fallback) instead of `Integer.parseInt`, so a bad value is logged and defaulted rather than thrown.

megamek/src/megamek/client/ratgenerator/SubForcesNode.java
```diff
@@ -95,42 +98,57 @@ public ArrayList<ForceDescriptor> generateSubForces(ForceDescriptor forceDescrip
...
                         } else {
-                            sub.setEchelon(Integer.parseInt(vn.getContent()));
+                            sub.setEchelon(Integer.parseInt(valueNode.getContent()));
```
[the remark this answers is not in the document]
HammerGS: Addressed - `parseEchelon(...)` handles this parse safely now (MathUtility.parseInt + warn-and-default).

megamek/src/megamek/client/ratgenerator/SubForcesNode.java
```diff
@@ -95,42 +98,57 @@ public ArrayList<ForceDescriptor> generateSubForces(ForceDescriptor forceDescrip
...
+                                  sub.getWeightClass());
+                        }
                         subs.add(sub);
                     }
                     if (forceDescriptor.getSizeMod() == ForceDescriptor.REINFORCED) {
                         ForceDescriptor sub = forceDescriptor.createChild(subs.size());
-                        sub.setEchelon(Integer.parseInt(vn.getContent()));
-                        apply(sub, vn.getNum() / 2);
-                        n.apply(sub, vn.getNum() / 2);
+                        sub.setEchelon(Integer.parseInt(valueNode.getContent()));
```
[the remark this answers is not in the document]
HammerGS: Addressed - the REINFORCED-branch parse also goes through `parseEchelon(...)` now, so it can't throw on malformed content.

megamek/src/megamek/client/ratgenerator/SubForcesNode.java
```diff
@@ -95,42 +98,57 @@
...
+                              && i == valueNode.getNum() / 2) {
                             continue;
                         }
                         ForceDescriptor sub = forceDescriptor.createChild(i);
-                        if (vn.getContent().endsWith("+")) {
+                        if (valueNode.getContent().endsWith("+")) {
                             sub.setSizeMod(ForceDescriptor.REINFORCED);
-                            sub.setEchelon(Integer.parseInt(vn.getContent().replace("+", "")));
-                        } else if (vn.getContent().endsWith("-")) {
+                            sub.setEchelon(Integer.parseInt(valueNode.getContent().replace("+", "")));
```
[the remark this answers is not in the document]
HammerGS: Addressed - replaced the raw `Integer.parseInt` here with a `parseEchelon(...)` helper using `MathUtility.parseInt` (logs a warning and defaults to 0 on malformed input), so it no longer throws NumberFormatException.

megamek/src/megamek/client/ratgenerator/SubForcesNode.java
```diff
@@ -95,42 +98,57 @@
...
+                        } else if (valueNode.getContent().endsWith("-")) {
                             sub.setSizeMod(ForceDescriptor.UNDERSTRENGTH);
-                            sub.setEchelon(Integer.parseInt(vn.getContent().replace("-", "")));
+                            sub.setEchelon(Integer.parseInt(valueNode.getContent().replace("-", "")));
```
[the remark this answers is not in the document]
HammerGS: Addressed - same fix: `parseEchelon(...)` now does a safe parse with a logged fallback.

megamek/src/megamek/client/ratgenerator/SubForcesNode.java
```diff
@@ -95,42 +98,57 @@
...
                         } else {
-                            sub.setEchelon(Integer.parseInt(vn.getContent()));
+                            sub.setEchelon(Integer.parseInt(valueNode.getContent()));
```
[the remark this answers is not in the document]
HammerGS: Addressed - now parsed via `parseEchelon(...)` (MathUtility.parseInt with warn-and-default), no uncaught NumberFormatException.

megamek/src/megamek/client/ratgenerator/SubForcesNode.java
```diff
@@ -95,42 +98,57 @@
...
+                                  sub.getWeightClass());
+                        }
                         subs.add(sub);
                     }
                     if (forceDescriptor.getSizeMod() == ForceDescriptor.REINFORCED) {
                         ForceDescriptor sub = forceDescriptor.createChild(subs.size());
-                        sub.setEchelon(Integer.parseInt(vn.getContent()));
-                        apply(sub, vn.getNum() / 2);
-                        n.apply(sub, vn.getNum() / 2);
+                        sub.setEchelon(Integer.parseInt(valueNode.getContent()));
```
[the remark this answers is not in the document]
HammerGS: Addressed - the REINFORCED-branch parse uses `parseEchelon(...)` now, so malformed content is logged and defaulted instead of thrown.

megamek/src/megamek/common/ui/RulesetXmlViewer.java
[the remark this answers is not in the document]
HammerGS: This is in `RulesetXmlViewer`, which was removed from the PR (the XML viewer was a dead end), so the finding no longer applies.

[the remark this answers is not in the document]
HammerGS: This is in `RulesetXmlViewer`, which was removed from the PR, so the finding no longer applies.

HammerGS:
Pushed the CodeQL fix (commit `68ebdb7`). The SubForcesNode echelon parsing now goes through a `parseEchelon(...)` helper using `MathUtility.parseInt` with a logged fallback instead of `Integer.parseInt`, so malformed ruleset content is logged and defaulted to 0 rather than throwing. This covers all six parse sites in `generateSubForces`, so the four CodeQL "Missing catch of NumberFormatException" findings (10348-10351) should clear on the next scan. The RulesetXmlViewer findings are moot since that file was removed.

IllianiBird (approved):

PR #8343, opened 2026-06-14
  Implement Bridge-Building Engineers per TO:AuE p.152 (Not the AVLB)

HammerGS:
Add the BRIDGE_ENGINEERS infantry rule: a platoon can spend 6 turns adjacent to a hex to raise a single-hex bridge, with full pause / dismantle / abandon control. Reuses a new shared BridgeConstruction helper designed for future vehicle Bridge-Layers.

  ---
  ## Summary

  Implements the Bridge-Building Engineers specialized-infantry rule (TO:AuE p.152): a `BRIDGE_ENGINEERS` Conventional Infantry platoon with an Infantry Bridge Kit spends 6 full turns adjacent to a target hex - taking no other action - to raise a single-hex **Light Bridge** (CF 15) or **Medium Bridge** (CF 40). CF doubles over water. Taking trooper casualties extends the work by 1 turn. A per-scenario budget of 2 points allows two Light Bridges or one Medium.

  A build in progress can be **paused** (the platoon is freed to act and may return to resume), **dismantled** (over the turns spent, refunding the points), or **abandoned** (instantly, forfeiting the points). Gated behind a new advanced game option, **TO Bridge-Building Engineers (Infantry)** (default off).

  ## Rules Implemented

  - **Raise a bridge** (TO:AuE p.152): 6 full turns adjacent to the target hex, engaging in no other action.
  - **Two bridge types**: Light Bridge (CF 15) or Medium Bridge (CF 40); CF doubled when raised over water (depth 1+).
  - **Per-scenario budget**: 2 points - two Light Bridges (1 each) or one Medium (2). Spent at the start of a build.
  - **Casualties extend the work**: a turn in which the platoon loses troopers adds 1 turn (once per turn, regardless of the number of attacks).
  - **Destroyed/displaced before completion**: the work is lost and no bridge appears.
  - **Placement** (adopting TM p.242 Bridge-Layer constraints): a water hex is legal only if it connects to land or an existing bridge; the two banks may differ by at most 1 level; the deck sits at the lower anchoring rim. Bridges may be straight or curved (any two distinct hexsides). Canyons of any depth may be bridged.
  
   ## Player experience: managing a bridge in progress

  A build no longer has to run to completion. While a bridge is under way, the **Cancel Bridge** button (it reads **Resume Building** once paused or dismantling) opens a small chooser with the actions valid for the current state. All three end the platoon's turn.

  - **Pause** - stop and come back later. The partial progress is held on the platoon and the platoon is *freed* to move and fight normally. _Criteria:_ only while actively building; progress persists indefinitely; only that same platoon, standing adjacent to the site again, can **Resume**; one bridge in progress per platoon.

  - **Dismantle** - take the partial bridge back down and recover the cost. _Criteria:_ only while actively building; takes as many turns as were already spent (the indicator counts back down, e.g. 3/6 -> 2/6 -> 1/6); the spent bridge-building point(s) are **refunded** once it finishes.

  - **Abandon** - give up immediately. _Criteria:_ available while building, paused, or dismantling; instant (no turns spent); the partial work is lost and the spent point(s) are **forfeit** (no refund).

  - **Resume** - continue a paused build, or reverse a dismantle from where it stopped. _Criteria:_ a paused platoon must be adjacent to its site; a dismantling platoon resumes from the structure still standing.

  Note: a destroyed (or displaced) platoon loses any in-progress work with no refund, and an in-progress hex is reserved so a second engineer can't start another bridge in it.

  ## Changes

  1. **ConvInfantry.java** - bridge build/pause/dismantle state machine (turns, casualty-extended required turns, target/exits/type,  2-point budget, trooper snapshot); `pauseBridgeBuild()`/`resumeBridgeBuild()`/`abandonBridge()`/`startBridgeDismantle()`;
  `isBuildingBridge()`/`isBridgePaused()`/`isDismantlingBridge()`/`isBusyWithBridge()`/`hasBridgeInProgress()`; static
  `isBridgeTargetClaimed()`; `newRound()`/`isEligibleFor()` overrides; auto-equips the Infantry Bridge Kit for `BRIDGE_ENGINEERS`.
  2. **BridgeConstruction.java** (NEW) - shared site validation and bridge placement (terrains, deck elevation = lower rim, `IBuilding` registration). CF is a parameter, so future vehicle Bridge-Layers (TM p.242, CF 8/20/45, no water doubling) can reuse it.
  3. **MoveStep.java / MoveStepType.java** - `BUILD_BRIDGE`, `CANCEL_BRIDGE`, `RESUME_BRIDGE`, `PAUSE_BRIDGE`, `ABANDON_BRIDGE` steps; legality checks; movement lockdown while building/dismantling (no move/turn); bridge-deck crossing fix and exit constraint.
  4. **MovePathHandler.java** - processes the bridge steps: start build (spends budget, rejects a hex another platoon already claims), pause, abandon, dismantle, resume; reports.
  5. **TWGameManager.java** - END-phase `checkBuildBridges()`: banks one turn of work per round, applies casualty extension, abandons displaced/destroyed builds, completes finished bridges (places terrain, registers the `IBuilding`, updates clients), and advances dismantling with refund-on-finish.
  6. **TWPhasePreparationManager.java** - END-phase call to `checkBuildBridges()`.
  7. **MovementDisplay.java** - `Build Bridge` button + 2-click placement (bridge hex, then far end; origin fixed at the engineer); action chooser dialog (Pause / Dismantle / Abandon, or Resume / Abandon); ribbon lockdown while building; per-condition DEBUG logging of why the button is disabled.
  8. **BridgeBuildSprite.java / BridgeBuildSpriteHandler.java** (NEW) - in-hex ghost indicator showing the standing structure (`N/6`), counting up while building and back down while dismantling.
  9. **HexTooltip.java** - bridge hex tooltip line for building / paused / dismantling, naming the engineer (name + ID).
  10. **ClientGUI.java** - sprite-handler registration and per-round progress toasts for the local player's bridge platoons.
  11. **SystemPanel.java** - shows the remaining bridge budget ("2 Light / 1 Medium bridges left") on the Infantry Bridge Kit line in the Unit Viewer.
  12. **MoveCommand.java** - `MOVE_BUILD_BRIDGE` command, hidden when the game option is off.
  13. **OptionsConstants.java / GameOptions.java** - new `bridge_building_engineers` advanced game option.
  14. **EquipmentTypeLookup.java / MiscType.java** - `INFANTRY_BRIDGE_KIT` lookup constant.
  15. **messages.properties / report-messages.properties** + option/report strings - all user-facing text.

  ## Files Changed

  - `megamek/src/megamek/common/units/ConvInfantry.java` - build/pause/dismantle state machine, budget, eligibility, auto-equip
  - `megamek/src/megamek/common/board/BridgeConstruction.java` - NEW: shared validation + placement
  - `megamek/src/megamek/common/moves/MoveStep.java` - bridge steps, lockdown, crossing fix
  - `megamek/src/megamek/common/enums/MoveStepType.java` - BUILD/CANCEL/RESUME/PAUSE/ABANDON_BRIDGE
  - `megamek/src/megamek/server/totalWarfare/MovePathHandler.java` - step processing
  - `megamek/src/megamek/server/totalWarfare/TWGameManager.java` - END-phase progression/completion/dismantle
  - `megamek/src/megamek/server/totalWarfare/TWPhasePreparationManager.java` - END-phase call
  - `megamek/src/megamek/client/ui/panels/phaseDisplay/MovementDisplay.java` - placement, chooser, lockdown
  - `megamek/src/megamek/client/ui/panels/phaseDisplay/commands/MoveCommand.java` - Build Bridge command
  - `megamek/src/megamek/client/ui/clientGUI/boardview/sprite/BridgeBuildSprite.java` - NEW: hex indicator
  - `megamek/src/megamek/client/ui/clientGUI/boardview/spriteHandler/BridgeBuildSpriteHandler.java` - NEW: indicator handler
  - `megamek/src/megamek/client/ui/clientGUI/tooltip/HexTooltip.java` - bridge hex tooltip (with builder name/ID)
  - `megamek/src/megamek/client/ui/clientGUI/ClientGUI.java` - handler registration, progress toasts
  - `megamek/src/megamek/client/ui/dialogs/unitDisplay/SystemPanel.java` - Systems-tab budget display
  - `megamek/src/megamek/common/options/OptionsConstants.java`, `.../options/GameOptions.java` - new game option
  - `megamek/src/megamek/common/equipment/EquipmentTypeLookup.java`, `.../equipment/MiscType.java` - Infantry Bridge Kit lookup
  - `megamek/resources/megamek/client/messages.properties` - UI strings
  - `megamek/resources/megamek/common/report-messages.properties` - report strings (4274-4287)
  - `megamek/resources/megamek/common/options/messages.properties` - game option text
  - `megamek/unittests/megamek/server/totalWarfare/BridgeBuildingTest.java` - NEW: 53 tests
  
  **PLEASE NOTE-The MD file is still in this but will remove before I merge.** 

  ## Testing

  - **Unit tests**: 53 tests in `BridgeBuildingTest` - site validation (water, canyon, level limits, occupied, off-board), placement
  (terrain, CF, deck at lower rim, structure registration), build lifecycle (banked progress, casualty extension, displacement/destruction abandon, budget), cancel/dismantle (countback + refund), pause/resume, abandon, hex reservation against a second build, weight collapse, crossing, and serialization (build and paused state). Full suite green.
  
  - **Manual testing** (all passing):
    - Placement: origin-from-engineer 2-click flow, straight and curved spans.
    - Build progress counts 0/6 -> 6/6; bridge appears at completion; casualties push 6 -> 7.
    - While building, the ribbon is locked to Next Unit / Done / the bridge button; the unit cannot move or turn.
    - Action chooser offers Pause / Dismantle / Abandon (and Resume / Abandon while paused or dismantling).
    - Pause frees the platoon; progress persists and shows at the hex; returning adjacent and Resuming continues it.
    - Abandon is instant with no refund; Dismantle counts back down and refunds the points on completion.
    - A hex with an in-progress bridge cannot be targeted by a second engineer.
    - Unit Viewer Systems tab shows and updates the remaining bridge budget; hex tooltip names the builder.
    - Save/reload mid-build and mid-pause preserves all state and resumes correctly.
    - Units cross a finished bridge (climb mode as needed), entering/leaving only via its connected hexsides.
    - Still to verify: weight collapse and destroyed-engineer cases in live play (both covered by unit tests), and
  double-blind/multiplayer visibility of in-progress bridges.


megamek/src/megamek/server/totalWarfare/TWGameManager.java
```diff
@@ -15787,6 +15788,157 @@ void checkLayExplosives() {
...
+            // before completing its task, the bridge is destroyed as well"). This also covers a platoon wiped out
+            // during the combat of its final turn that is still doomed in the entity list when the END phase runs. A
+            // platoon destroyed while dismantling loses the partial bridge and is not refunded.
+            if (entity.isDestroyed() || entity.isDoomed()) {
+                LOGGER.info("[BuildBridge] {} was destroyed before finishing its bridge work at {}; abandoned",
+                      convInfantry.getShortName(), convInfantry.getBridgeTargetCoords());
+                convInfantry.cancelBridgeBuild();
+                continue;
+            }
+            if (!convInfantry.isAdjacentToBridgeSite()) {
```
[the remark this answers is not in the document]
HammerGS: Good catch — real leak. A paused platoon is `hasBridgeInProgress()` but not `isBusyWithBridge()`, so a destroyed/doomed paused platoon was skipped here and its state (and the reservation/sprite/tooltip that read it) persisted. Fixed: `checkBuildBridges()` now iterates `hasBridgeInProgress()` and clears any destroyed/doomed platoon (paused included); the adjacency-abandon and per-round progression are gated behind `isBusyWithBridge()` so a paused platoon still isn't advanced or penalized for moving away. Added a regression test (`destroyedPausedPlatoonClearsItsBridgeState`).

megamek/src/megamek/common/board/BridgeConstruction.java
```diff
@@ -0,0 +1,283 @@
...
+     * @return The newly registered bridge structure, for sending to the clients.
+     */
+    public static IBuilding placeBridge(Board board, Coords target, int exits, int bridgeType, int cf) {
+        Hex targetHex = board.getHex(target);
+        int[] exitDirections = exitDirections(exits);
+        Hex firstBank = board.getHex(target.translated(exitDirections[0]));
+        Hex secondBank = board.getHex(target.translated(exitDirections[1]));
+
+        // The deck sits at the highest anchoring bank (land, rim, or previous span) so units cross at grade and a
+        // chain of spans stays level, spanning the water or canyon below.
```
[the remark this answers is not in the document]
HammerGS: Correct, the inline comment was wrong. `bridgeDeckLevel()` deliberately picks the *lowest* anchoring bank (lower rim), as the method Javadoc states. Updated the comment in `placeBridge()` to describe the lowest anchoring bank / lower rim, matching the behavior.

megamek/mmconf/log4j2.xml
```diff
@@ -100,6 +100,22 @@
...
+        -->
+        <Logger name="megamek.client.ui.panels.phaseDisplay.MovementDisplay" level="debug" additivity="false">
+            <AppenderRef ref="MegaMekLog" />
+        </Logger>
+        <Logger name="megamek.common.units.ConvInfantry" level="debug" additivity="false">
+            <AppenderRef ref="MegaMekLog" />
+        </Logger>
+        <Logger name="megamek.common.moves.MoveStep" level="debug" additivity="false">
+            <AppenderRef ref="MegaMekLog" />
+        </Logger>
```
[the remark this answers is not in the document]
HammerGS: Agreed — these shouldn't ship on, especially `MoveStep` (it also emits climb/facing trace on every move). Commented the three loggers out in the default config and reworded the note to "OFF by default; uncomment to diagnose." The `LOGGER.debug(...)` calls stay in code so they can be re-enabled when needed.

docs/issues/features/bridge-building-engineers.md
```diff
@@ -0,0 +1,237 @@
...
+6. **Bridge elevation**: the bridge surface sits level with the LOWER of the two connected banks (a bridge hex has
+   a single `BRIDGE_ELEV`; two ends cannot differ in height).
+
+## Mechanics
+
+- **Declaration** (movement phase): `Build Bridge` button -> Light/Medium dialog -> click target hex (highlighted
+  valid sites) -> click a hex adjacent to the target to set the axis -> move committed. The declaring turn counts
+  as the first full turn of work.
+- **Progress**: `ConvInfantry.newRound()` increments the completed-turn counter; `isEligibleFor()` returns false
+  for all phases while building, so the platoon takes no turns.
```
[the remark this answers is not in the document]
HammerGS: Right, that section described the original design. Updated it: progress is now banked in the END phase (`checkBuildBridges()` -> `bankBridgeBuildTurn()`/`bankBridgeDismantleTurn()`), `newRound()` only handles displacement-abandon (no longer increments), and a platoon actively building/dismantling is Movement-phase eligible only (a paused platoon is freed in all phases).

HammerGS:
image

megamek/src/megamek/client/ui/clientGUI/boardview/sprite/BridgeBuildSprite.java
```diff
@@ -0,0 +1,190 @@
...
+ */
+public class BridgeBuildSprite extends HexSprite {
+
+    private static final MMLogger LOGGER = MMLogger.create(BridgeBuildSprite.class);
+
+    private static final Color TEXT_OUTLINE_COLOR = new Color(40, 40, 50);
+
+    private static final int HEX_CENTER_X = HexTileset.HEX_W / 2;
+    private static final int FONT_SIZE = 11;
+    private static final int BOTTOM_OFFSET = 24;
```
IllianiBird, line 71: Should this be scaled for gui?

megamek/src/megamek/client/ui/clientGUI/tooltip/HexTooltip.java
```diff
@@ -252,9 +254,56 @@ public static String getHexTip(Hex mhex, @Nullable Client client, int boardId) {
...
+    /**
+     * Appends a "bridge under construction (turn X of Y)" line when a Bridge-Building Engineer platoon is raising a
+     * bridge in the given hex, TO:AUE. The build state is read from the synced ConvInfantry entities.
+     *
+     * @param result  the tooltip text being built
+     * @param game    the current game
+     * @param mcoords the hex the tooltip describes
+     * @param boardId the board of the hex
+     */
+    private static void appendBridgeBuildInfo(StringBuilder result, Game game, Coords mcoords, int boardId) {
```
IllianiBird: `mcoords`? Not sure what this means, which likely means it needs a more descriptive param name

megamek/src/megamek/client/ui/clientGUI/tooltip/HexTooltip.java
```diff
@@ -252,9 +254,56 @@ public static String getHexTip(Hex mhex, @Nullable Client client, int boardId) {
...
+        for (Entity entity : game.getEntitiesVector()) {
+            boolean isWorkingHere = (entity instanceof ConvInfantry convInfantry)
+                  && convInfantry.hasBridgeInProgress()
+                  && (entity.getBoardId() == boardId)
+                  && mcoords.equals(convInfantry.getBridgeTargetCoords());
+            if (!isWorkingHere) {
+                continue;
+            }
+            ConvInfantry convInfantry = (ConvInfantry) entity;
```
IllianiBird: I really am not a fan of how this is being handled. It just asking for a class cast exception

megamek/src/megamek/client/ui/clientGUI/ClientGUI.java
```diff
@@ -582,6 +583,37 @@ public void addToast(ToastLevel level, String text, @Nullable Entity entity) {
...
+    private void showBridgeBuildProgressToasts() {
+        for (Entity entity : getClient().getGame().getEntitiesVector()) {
+            boolean isOwnBridgePlatoon = (entity instanceof ConvInfantry convInfantry)
+                  && convInfantry.isBusyWithBridge()
+                  && (convInfantry.getBridgeTargetCoords() != null)
+                  && (entity.getOwnerId() == getClient().getLocalPlayer().getId());
+            if (!isOwnBridgePlatoon) {
+                continue;
+            }
+            ConvInfantry convInfantry = (ConvInfantry) entity;
```
IllianiBird: Similar to my earlier comment. This is really fragile and it would be very easy for someone to edit and we end up flooded in class cast exceptions

megamek/src/megamek/client/ui/dialogs/unitDisplay/SystemPanel.java
```diff
@@ -448,11 +449,37 @@ private String getMountedDisplay(Mounted<?> m, int loc) {
...
+    /**
+     * Appends the remaining bridge building budget to the Infantry Bridge Kit line so a Bridge-Building Engineer
+     * platoon can see, in the systems tab, how many bridges it can still raise this scenario (TO:AUE p.152). The budget
+     * of 2 points covers either 2 Light Bridges (1 point each) or 1 Medium Bridge (2 points); the counts drop as
+     * bridges are built.
+     *
+     * @param sb the display text being built
+     * @param m  the mounted equipment for this row
+     */
+    private void appendBridgeKitBudget(StringBuilder sb, Mounted<?> m) {
```
IllianiBird: abbreviated params. `sb` isn't too bad, but far better to call it something like `report` or something similarly descriptive.

megamek/src/megamek/common/moves/MoveStep.java
```diff
@@ -89,6 +91,15 @@ public class MoveStep implements Serializable {
...
     public static final int CARGO_LOCATION_KEY = 1;
 
+    /**
+     * Additional int data keys for a BUILD_BRIDGE step (keys are scoped per step type): the target hex coordinates, the
+     * exits bitmask of the two connected hexsides, and the bridge type (light/medium).
+     */
+    public static final int BRIDGE_TARGET_X_KEY = 0;
+    public static final int BRIDGE_TARGET_Y_KEY = 1;
+    public static final int BRIDGE_EXITS_KEY = 2;
+    public static final int BRIDGE_TYPE_KEY = 3;
```
IllianiBird, line 102: Would this be better as an enum? Not super important, 4 consts is nothing, but if we ever add more it'd be better to use an end,.

megamek/src/megamek/common/moves/MoveStep.java
```diff
@@ -2640,6 +2758,32 @@ stepType, movementType, isProne(), isClimbing, entity.isClimbing(),
...
+                  && (prev.getElevation() == bridgeSrcHex.terrainLevel(Terrains.BRIDGE_ELEV))
+                  && entity.isLocationProhibited(lastPos, boardId, prev.getElevation())
+                  && !bridgeSrcHex.containsTerrainExit(Terrains.BRIDGE, lastPos.direction(curPos));
+            boolean enteringBridgeDeckOffExit = isRealBridgeSpan(bridgeDestHex)
+                  && (getElevation() == bridgeDestHex.terrainLevel(Terrains.BRIDGE_ELEV))
+                  && entity.isLocationProhibited(curPos, boardId, getElevation())
+                  && !bridgeDestHex.containsTerrainExit(Terrains.BRIDGE, curPos.direction(lastPos));
+            if (leavingBridgeDeckOffExit || enteringBridgeDeckOffExit) {
+                movementType = EntityMovementType.MOVE_ILLEGAL;
+            }
```
IllianiBird, line 2809: There are a lot of chained conditionals here. Might be worth breaking them out into composite booleans to aid readability.

megamek/src/megamek/common/units/ConvInfantry.java
```diff
@@ -522,6 +511,416 @@ public static String getSpecializationTooltip(int spec) {
...
+     */
+    public static boolean isBridgeTargetClaimed(Game game, int boardId, Coords target, @Nullable Entity excluded) {
+        if (target == null) {
+            return false;
+        }
+        for (Entity entity : game.getEntitiesVector()) {
+            if ((entity != excluded) && (entity instanceof ConvInfantry convInfantry)
+                  && convInfantry.hasBridgeInProgress()
+                  && (convInfantry.getBoardId() == boardId)
+                  && target.equals(convInfantry.getBridgeTargetCoords())) {
```
IllianiBird, line 706: Is this `null` safe? I worry that if we compare an object to `null` we might end up throwing an NPE

megamek/src/megamek/server/totalWarfare/MovePathHandler.java
```diff
@@ -4643,4 +4658,155 @@ private Coords nudgeOntoBoard(Coords position, int facing) {
...
+     */
+    private void processResumeBridgeStep(Infantry infantry) {
+        boolean canResume = (infantry instanceof ConvInfantry ci)
+              && (ci.isDismantlingBridge() || ci.isBridgePaused());
+        if (!canResume) {
+            logger.warn("[BuildBridge] RESUME_BRIDGE step ignored for {}: not a ConvInfantry platoon currently "
+                  + "dismantling or paused", infantry.getShortName());
+            return;
+        }
+        ConvInfantry convInfantry = (ConvInfantry) infantry;
```
IllianiBird: Similar to my earlier comments this is really fragile. Far better to wrap the entire thing in a conditional `infantry instanceof ConvInfantry convInfantry`. This allows you to use `convInfantry` instead of needing to create a new variable on line `4747` and it will be more robust.

megamek/src/megamek/server/totalWarfare/MovePathHandler.java
```diff
@@ -4643,4 +4658,155 @@ private Coords nudgeOntoBoard(Coords position, int facing) {
...
+    }
+
+    /**
+     * Pauses an active bridge build: the platoon holds its progress and is freed to act normally until it returns to
+     * resume. TO:AUE p.152.
+     *
+     * @param infantry the platoon pausing its build
+     */
+    private void processPauseBridgeStep(Infantry infantry) {
+        if (!(infantry instanceof ConvInfantry convInfantry) || !convInfantry.isBuildingBridge()) {
```
IllianiBird: This will give us false-positives if we pass in BattleArmor (or another type of Infantry) instead of ConvInfantry. In that case the first clause will return `true` and then we will report to the log that the unit is 'not a ConvInfantry platoon currently building a bridge' when really it should be reporting `not a ConvInfantry platoon cannot build a bridge` or something similar.

megamek/src/megamek/server/totalWarfare/TWGameManager.java
```diff
@@ -15787,6 +15788,162 @@ void checkLayExplosives() {
         resolveExplodingCharges();
     }
 
+    /**
```
IllianiBird, line 15790: This comment relates to all new code added to this class.

Does this code need to live in `TMGameManager` or can it be moved to a utility class? TWGameManager is one of (if not _the_) largest classes in the codebase. We should only be expanding it further when we have absolutely no other choice.

IllianiBird (changes requested):
Some relatively minor feedback, mostly good :)

megamek/src/megamek/client/ui/clientGUI/tooltip/HexTooltip.java
HammerGS: Good call — fixed. Replaced the `(ConvInfantry) entity` re-cast with a single `instanceof` pattern binding in a guard clause, so the bound `convInfantry` is used throughout and there's no second cast to drift from the check:

```java
if (!(entity instanceof ConvInfantry convInfantry) || !convInfantry.hasBridgeInProgress()
      || (entity.getBoardId() != boardId)
      || !coords.equals(convInfantry.getBridgeTargetCoords())) {
    continue;
}
```

megamek/src/megamek/client/ui/clientGUI/tooltip/HexTooltip.java
HammerGS: Agreed — renamed the parameter from `mcoords` to `coords`.

megamek/src/megamek/client/ui/clientGUI/ClientGUI.java
HammerGS: Fixed the same way — folded the `instanceof` check and the early-out into one guard so the pattern-bound `convInfantry` is used directly, no separate cast:

```java
for (Entity entity : getClient().getGame().getEntitiesVector()) {
    if (!(entity instanceof ConvInfantry convInfantry) || !convInfantry.isBusyWithBridge()
          || (convInfantry.getBridgeTargetCoords() == null)
          || (entity.getOwnerId() != getClient().getLocalPlayer().getId())) {
        continue;
    }
    ...
}
```

megamek/src/megamek/client/ui/dialogs/unitDisplay/SystemPanel.java
HammerGS: Renamed for clarity: `sb` -> `displayText` and `m` -> `mounted`.

megamek/src/megamek/common/moves/MoveStep.java
HammerGS: Left as int constants for now. These are keys into the generic per-step data map used by `addStepToMovePath(MoveStepType, Map<Integer, Integer>)`, which other step types (e.g. minefield steps) populate with the same `*_KEY` int-constant convention. An enum here would diverge from that shared mechanism (the map is `Map<Integer, Integer>`), so I kept it consistent — happy to revisit if the data-key map is ever reworked.

megamek/src/megamek/common/moves/MoveStep.java
HammerGS: Good point — extracted the two near-identical four-condition chains into a single named helper, so the call site reads as intent:

```java
boolean leavingBridgeDeckOffExit  = crossesBridgeDeckOffExit(entity, bridgeSrcHex,  prev.getElevation(), lastPos, curPos);
boolean enteringBridgeDeckOffExit = crossesBridgeDeckOffExit(entity, bridgeDestHex, getElevation(),      curPos, lastPos);
if (leavingBridgeDeckOffExit || enteringBridgeDeckOffExit) {
    movementType = EntityMovementType.MOVE_ILLEGAL;
}
```
`crossesBridgeDeckOffExit(...)` holds the (real-span && at-deck-elevation && terrain-prohibited && off-exit) test with Javadoc.

megamek/src/megamek/common/units/ConvInfantry.java
HammerGS: This one is safe as written: `target` is guard-checked non-null at the top of the method (`if (target == null) { return false; }`), and `Coords.equals(null)` returns `false` per the `equals` contract — so `target.equals(convInfantry.getBridgeTargetCoords())` can't NPE even when the platoon's target is null. Left as-is; let me know if you'd still prefer a restructure.

megamek/src/megamek/client/ui/clientGUI/boardview/sprite/BridgeBuildSprite.java
HammerGS: These constants are in unscaled hex-pixel space (relative to `HexTileset.HEX_W`/`HEX_H`), and `prepare()` applies the board zoom before any drawing:

```java
graph.scale(bv.getScale(), bv.getScale());
```
So everything drawn with these offsets scales with the board zoom level — same pattern the other hex sprites use. No separate scaling needed here.

megamek/src/megamek/server/totalWarfare/MovePathHandler.java
HammerGS: Refactored as suggested — one `instanceof ConvInfantry convInfantry` binding, the state check split out, and the later `(ConvInfantry) infantry` re-cast removed:

```java
private void processResumeBridgeStep(Infantry infantry) {
    if (!(infantry instanceof ConvInfantry convInfantry)) {
        logger.warn("[BuildBridge] RESUME_BRIDGE step ignored for {}: not a conventional infantry platoon",
              infantry.getShortName());
        return;
    }
    if (!convInfantry.isDismantlingBridge() && !convInfantry.isBridgePaused()) {
        logger.warn("[BuildBridge] RESUME_BRIDGE step ignored for {}: not currently dismantling or paused",
              convInfantry.getShortName());
        return;
    }
    Coords target = convInfantry.getBridgeTargetCoords();
    ...
}
```

megamek/src/megamek/server/totalWarfare/MovePathHandler.java
HammerGS: Good catch on the false positive. Split the combined guard so the type-mismatch and wrong-state cases log distinct, accurate reasons — applied to all five bridge step processors (build / cancel / resume / pause / abandon):

```java
if (!(infantry instanceof ConvInfantry convInfantry)) {
    logger.warn("[BuildBridge] PAUSE_BRIDGE step ignored for {}: not a conventional infantry platoon",
          infantry.getShortName());
    return;
}
if (!convInfantry.isBuildingBridge()) {
    logger.warn("[BuildBridge] PAUSE_BRIDGE step ignored for {}: not currently building a bridge",
          convInfantry.getShortName());
    return;
}
```
Now a BattleArmor (or any other Infantry) reports "not a conventional infantry platoon" rather than the misleading "not currently building a bridge".

megamek/src/megamek/server/totalWarfare/TWGameManager.java
HammerGS: Agreed — moved it out. Extracted the entire bridge END-phase block (`checkBuildBridges` / `progressBridgeBuild` / `progressBridgeDismantle` / `finishBridgeBuild`, ~155 lines) into a new `BridgeBuildPhaseHandler extends AbstractTWRuleHandler`. `TWGameManager` keeps only a one-line delegator so existing callers/tests are unaffected:

```java
void checkBuildBridges() {
    new BridgeBuildPhaseHandler(this).checkBuildBridges();
}
```
The handler uses the base `getGame()`/`addReport()` and broadcasts through the public `gameManager.sendChangedHex(...)` / `sendNewBuildings(...)` (so the private packet builder stays private). Same pattern as `BuildingCollapseHandler`. Net effect on `TWGameManager`: ~155 lines removed, one unused import dropped.

megamek/src/megamek/client/ui/clientGUI/ClientGUI.java
```diff
@@ -582,6 +583,39 @@ public void addToast(ToastLevel level, String text, @Nullable Entity entity) {
         }
     }
 
+    /**
+     * Shows a progress toast for each of the local player's platoons that is raising a bridge (TO:AUE). Called once per
+     * round at the start of the movement phase: a building platoon is ineligible for all phases, so no phase display
+     * ever selects it and this is its only per-turn feedback besides the hex indicator and END phase report.
+     */
```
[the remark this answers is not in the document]
HammerGS: You're right, the comment was stale. Updated it to the current eligibility model: a busy bridge platoon is eligible only in the movement phase (movement-only, so it can continue/cancel/pause/resume) and takes no other action — it's no longer "ineligible for all phases". The toast remains its main per-turn feedback alongside the hex indicator and the END-phase report.

megamek/resources/megamek/common/options/messages.properties
```diff
@@ -231,6 +231,10 @@ GameOptionsInfo.option.assault_drop.displayableName=(Legacy) Assault Drop
...
+GameOptionsInfo.option.bridge_building_engineers.description=If checked, infantry with the Bridge-Building Engineers specialization can spend 6 full turns adjacent to a target hex to raise a single-hex Light Bridge (CF 15, two per scenario) or Medium Bridge (CF 40, one per scenario). CF is doubled when the bridge is built over water. Taking casualties during construction extends the work by 1 turn. TO:AUE.
+GameOptionsInfo.option.unofficial_bridge_repair_engineers.displayableName=(Unofficial) Bridge-Building Engineers May Repair Destroyed Sections
+GameOptionsInfo.option.unofficial_bridge_repair_engineers.description=Requires the TO Bridge-Building Engineers option. If checked, a Bridge-Building Engineer platoon may also rebuild a single destroyed section (a missing hex) of an existing bridge by working from an adjacent surviving span or bank, using the normal build rules. The player chooses a Light (CF 15) or Medium (CF 40) section, doubled over water, costing one bridge-building point; the new section's deck matches the surviving span so the bridge reconnects. Rubble left under the collapsed section remains. This is a house rule, not part of any published BattleTech ruleset.
```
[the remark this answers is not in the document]
HammerGS: Fixed. The description now reflects the per-type cost rather than a flat "one point":

> ...costing bridge-building points by type (Light 1, Medium both)...

matching `spendBridgeBuildPoints(step.getBridgeType())` (Light = 1, Medium = 2 of the 2-point budget).

IllianiBird (approved):

HammerGS:
Merged latest `main` to clear the conflict (Firefighting Engineers had landed on main and touched the same `ConvInfantry` regions).

**Only conflict:** `ConvInfantry.setSpecializations()`. Resolved by keeping both features:
- This branch's refactor — the `updateEngineerEquipment(...)` helper calls for Trench / Demo / Bridge kits.
- main's Firefighting Engineers block — the Fire Extinguisher add/remove. It's a *weapon* (not a `MiscType` tool), so it stays inline rather than going through the tool helper.

Everything else auto-merged. Verified after the merge: `:megamek:compileJava` clean, and the relevant suites green — `BridgeBuildingTest` 63/63, `FirefightingEngineerTest` 9/9, `FirefightingSupportTest` 7/7, `ConvInfantryDisposableWeaponTest` 4/4.

PR #8350, opened 2026-06-16
  Add End-Phase activation toggle for vehicle minesweepers

HammerGS:
## Summary

  Vehicle minesweepers always swept mines automatically — the player could never turn one off. The rules let a player
  activate or deactivate a minesweeper in the End Phase, and only an activated sweeper clears mines (TO:AUE p.138). This
  adds that toggle.  It was determined the Minesweeping Engineers are coded correctly and are complete. 

 ## Changes Made

  - Minesweeper On/Off mode (TO:AUE p.138): the Mine Sweeper now has "On"/"Off" modes. Default is On, so existing behavior is unchanged. The switch is deferred (instantModeSwitch = false), so it takes effect next turn — matching the End-Phase timing.
  - End-Phase control: a new "Minesweepers" button on the End-Phase report screen opens a dialog to toggle each minesweeper unit. This mirrors the existing Nova Network and Variable Range Targeting controls.
  - Sweep check honors the mode: a sweeper set to Off no longer clears mines; its vehicle trips mines normally.

 ## Changes

  1. MiscType.java - createISMineSweeper adds setModes("On", "Off") and instantModeSwitch = false.
  2. TWGameManager.java - new getActiveMinesweeper(Entity) helper checks the mode; replaces two duplicated detection loops in enterMinefield and checkVibraBombs.
  3. Entity.java - new hasMinesweeper() helper (mirrors hasNovaCEWS).
  4. ReportDisplay.java - new REPORT_MINESWEEPER button, enabled in the End Phase when the player owns a minesweeper unit.
  5. MinesweeperActivationDialog.java (NEW) - per-unit On/Off dialog. Reuses the existing sendModeChange packet.
  6. messages.properties - button and dialog strings.

  No server packet was added. ENTITY_MODE_CHANGE already applies the mode and is not phase-gated.

## Files Changed

  - megamek/src/megamek/common/equipment/MiscType.java - Mine Sweeper gets On/Off modes
  - megamek/src/megamek/server/totalWarfare/TWGameManager.java - mode-aware sweeper lookup, shared helper
  - megamek/src/megamek/common/units/Entity.java - hasMinesweeper() helper
  - megamek/src/megamek/client/ui/panels/phaseDisplay/ReportDisplay.java - End-Phase button
  - megamek/src/megamek/client/ui/dialogs/phaseDisplay/MinesweeperActivationDialog.java - NEW: activation dialog
  - megamek/resources/megamek/client/messages.properties - i18n strings

##  Testing

  - Compiles clean (:megamek:compileJava).
  - Manual testing:
    - Minesweeper On clears mines on entry (unchanged behavior).
    - Toggle Off in the End Phase → next turn the vehicle takes mine damage instead of sweeping.
    - Unit with no minesweeper → button stays disabled.
    - Save/load with a minesweeper set to Off → state persists.


megamek/src/megamek/client/ui/dialogs/phaseDisplay/MinesweeperActivationDialog.java
```diff
@@ -0,0 +1,339 @@
...
+    @Serial
+    private static final long serialVersionUID = 1L;
+    private static final MMLogger logger = MMLogger.create(MinesweeperActivationDialog.class);
+    private static final int PADDING = UIUtil.scaleForGUI(10);
+    private static final int PADDING_SMALL = UIUtil.scaleForGUI(5);
+
+    // Mode indexes follow the order declared in MiscType.createISMineSweeper(): setModes("On", "Off")
+    private static final String MODE_ON = "On";
+    private static final String MODE_OFF = "Off";
+    private static final String MODE_NONE = "None";
```
IllianiBird, line 88: Hardcoded player-facing Strings

megamek/src/megamek/client/ui/dialogs/phaseDisplay/MinesweeperActivationDialog.java
```diff
@@ -0,0 +1,339 @@
...
+
+        JPanel unitPanel = new JPanel(new GridBagLayout());
+        GridBagConstraints gbc = new GridBagConstraints();
+        gbc.insets = new Insets(PADDING_SMALL, PADDING_SMALL, PADDING_SMALL, PADDING_SMALL);
+        gbc.anchor = GridBagConstraints.WEST;
+
+        // Header row
+        gbc.gridx = 0;
+        gbc.gridy = 0;
+        unitPanel.add(new JLabel("<html><b>"
```
IllianiBird: Should these `html` and `b` tags not live inside the resource bundle, rather than being manually added in code? I won't mark the other instances of these, but there's a couple.

megamek/src/megamek/client/ui/panels/phaseDisplay/ReportDisplay.java
```diff
@@ -442,4 +450,38 @@ private void setDetonateChargesEnabled(boolean enabled) {
...
+        if (button != null) {
+            button.transferFocus();
+        }
+    }
+
+    /**
+     * Checks if the local player has any units mounting a minesweeper.
+     */
+    private boolean hasMinesweeperUnits() {
+        int localPlayerId = clientgui.getClient().getLocalPlayer().getId();
```
IllianiBird: This will throw an NPE if `getLocalPlayer` returns `null`, which it can

megamek/src/megamek/client/ui/dialogs/phaseDisplay/MinesweeperActivationDialog.java
```diff
@@ -0,0 +1,340 @@
...
+        initializeUI();
+
+        pack();
+        setLocationRelativeTo(parent);
+
+        // Clear highlighting when dialog is closed via X button
+        addWindowListener(new WindowAdapter() {
+            @Override
+            public void windowClosing(WindowEvent e) {
+                clearHighlighting();
```
IllianiBird, line 118: Should this not also include dispose? I don't believe window close action defaults to dispose, though could be wrong.

megamek/src/megamek/client/ui/dialogs/phaseDisplay/MinesweeperActivationDialog.java
```diff
@@ -0,0 +1,340 @@
...
+    private JButton btnCancel;
+
+    private List<Entity> playerUnits;
+    private final Map<Integer, JRadioButton> onModeButtons = new HashMap<>();
+
+    public MinesweeperActivationDialog(JFrame parent, ClientGUI clientGUI) {
+        super(parent, Messages.getString("MinesweeperActivationDialog.title"), true);
+        this.clientGUI = clientGUI;
+        this.game = clientGUI.getClient().getGame();
+        this.localPlayerId = clientGUI.getClient().getLocalPlayer().getId();
```
IllianiBird: We're calling clientGUI.getClient() twice, let's add a client variable that calls that and then passes the cached variable to game and localPlayerId. Generally, any time we're using a call chain more than once it should probably be cached.

megamek/src/megamek/server/totalWarfare/TWGameManager.java
```diff
@@ -7429,6 +7429,23 @@ boolean enterMinefield(Entity entity, Coords c, int curElev, boolean isOnGround,
      *
      * @return - <code>true</code> if the entity set off any mines
      */
+    /**
+     * Returns the entity's active (switched-on) minesweeper, or {@code null} if it has none, the sweeper is not ready,
+     * its armor is depleted, or the player has deactivated it in the End Phase (TO:AUE p.138). A unit may mount only
+     * one minesweeper.
+     */
```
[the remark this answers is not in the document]
HammerGS: Good catch. Moved `getActiveMinesweeper()` above the `enterMinefield()` Javadoc so the Javadoc documents its method again (f3b4b9d).

megamek/src/megamek/server/totalWarfare/TWGameManager.java
```diff
@@ -7429,6 +7429,23 @@ boolean enterMinefield(Entity entity, Coords c, int curElev, boolean isOnGround,
...
+    private static Mounted<?> getActiveMinesweeper(Entity entity) {
+        for (Mounted<?> mounted : entity.getMisc()) {
+            if (mounted.getType().hasFlag(MiscType.F_MINESWEEPER)
+                  && mounted.isReady()
+                  && (mounted.getArmorValue() > 0)
+                  && !mounted.curMode().equals("Off")) {
```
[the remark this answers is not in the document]
HammerGS: Fixed in f3b4b9d — now uses `!mounted.curMode().isOff()` instead of `equals("Off")`, matching the dialog and other mode-aware code.

megamek/src/megamek/client/ui/dialogs/phaseDisplay/MinesweeperActivationDialog.java
IllianiBird: Regarding `getLocalPlayer`, while it isn't marked as `Nullable` I can see cases in the codebase where we expect it to return `null`. It might be worth adding a `null` safety here and updating the `getLocalPlayer` method signature to include `@Nullable`

megamek/src/megamek/client/ui/dialogs/phaseDisplay/MinesweeperActivationDialog.java
```diff
@@ -0,0 +1,340 @@
...
+            if ((entity.getOwnerId() == localPlayerId) && entity.hasMinesweeper()) {
+                playerUnits.add(entity);
+            }
+        }
+    }
+
+    /**
+     * Returns the first minesweeper mounted on the entity, or {@code null} if it has none.
+     */
+    private static MiscMounted getMinesweeper(Entity entity) {
```
IllianiBird: Missing `@Nullable` tag. Any method that can return `null` as part of its normal operation should include the nullable tag.

megamek/src/megamek/common/units/Entity.java
```diff
@@ -6724,6 +6724,19 @@ public boolean hasNavalC3() {
         return false;
     }
 
+    /**
+     * @return True if this unit mounts a minesweeper (TO:AUE p.138), whether or not it is currently activated. Used to
```
IllianiBird: `{@code true}` not `True`

IllianiBird: We should also include printing when adding page references

megamek/src/megamek/server/totalWarfare/TWGameManager.java
```diff
@@ -7415,6 +7415,23 @@ boolean enterMinefield(Entity entity, Coords c, int curElev, boolean isOnGround,
...
+    /**
+     * Returns the entity's active (switched-on) minesweeper, or {@code null} if it has none, the sweeper is not ready,
+     * its armor is depleted, or the player has deactivated it in the End Phase (TO:AUE p.138). A unit may mount only
+     * one minesweeper.
+     */
+    private static Mounted<?> getActiveMinesweeper(Entity entity) {
+        for (Mounted<?> mounted : entity.getMisc()) {
+            if (mounted.getType().hasFlag(MiscType.F_MINESWEEPER)
+                  && mounted.isReady()
+                  && (mounted.getArmorValue() > 0)
```
IllianiBird, line 7427: I don't have easy access to my manuals right now, but a location needing armor for a component to work seems odd. Can I just grab a quick confirmation that this is RAW?

megamek/src/megamek/server/totalWarfare/TWGameManager.java
```diff
@@ -7415,6 +7415,23 @@ boolean enterMinefield(Entity entity, Coords c, int curElev, boolean isOnGround,
         return enterMinefield(entity, c, curElev, isOnGround, vMineReport, -1);
     }
 
+    /**
+     * Returns the entity's active (switched-on) minesweeper, or {@code null} if it has none, the sweeper is not ready,
+     * its armor is depleted, or the player has deactivated it in the End Phase (TO:AUE p.138). A unit may mount only
+     * one minesweeper.
+     */
+    private static Mounted<?> getActiveMinesweeper(Entity entity) {
```
IllianiBird: Missing `@Nullable` tag

IllianiBird (changes requested):
A few items of feedback. We seem to be missing `@Nullable` tags from methods that return `null`. There's also a logic instance that doesn't 'smell' right.

megamek/src/megamek/client/ui/dialogs/phaseDisplay/MinesweeperActivationDialog.java
HammerGS: These aren't player-facing text. `MODE_ON`/`MODE_OFF` are the equipment mode identifiers that must match `setModes("On", "Off")` in `MiscType.createISMineSweeper()`, and `MODE_NONE` ("None") is the sentinel `EquipmentMode` name returned by `pendingMode()`. The player-visible labels are separate bundle keys (`stateOn`/`stateOff`). Localizing these would break the `setMode()` lookups, so they can't move to the resource bundle. Happy to rename them if the worry is magic strings.

megamek/src/megamek/client/ui/dialogs/phaseDisplay/MinesweeperActivationDialog.java
HammerGS: Good call. Moved the `<html><b>...</b></html>` markup for all three headers into the resource bundle values; the code now just reads the key.

megamek/src/megamek/client/ui/panels/phaseDisplay/ReportDisplay.java
HammerGS: Fixed. `hasMinesweeperUnits()` now reads `getLocalPlayer()` first and returns false if it's null before touching `getId()`.

megamek/src/megamek/client/ui/dialogs/phaseDisplay/MinesweeperActivationDialog.java
HammerGS: You're right - the default close op is HIDE, not DISPOSE. Added `dispose()` to the `windowClosing` handler so the X button releases the dialog like the Cancel/Apply paths do.

megamek/src/megamek/client/ui/dialogs/phaseDisplay/MinesweeperActivationDialog.java
HammerGS: Done - cached `Client client = clientGUI.getClient();` and use it for both `getGame()` and `getLocalPlayer()`.

HammerGS: Added a null guard here (falls back to `Player.PLAYER_NONE`). Agreed that `getLocalPlayer()` itself should carry `@Nullable`, but since it's called across the codebase I'd rather make that signature change in its own focused PR than fold it into this one. Flagging for follow-up.

megamek/src/megamek/client/ui/dialogs/phaseDisplay/MinesweeperActivationDialog.java
HammerGS: Added `@Nullable` to `getMinesweeper()`.

megamek/src/megamek/common/units/Entity.java
HammerGS: Fixed - `{@code true}`.

HammerGS: Updated to match the existing TO:AuE citations: `TO:AuE p.138, Corrected Sixth Printing` (same style as the Disposable Weapon refs). Applied to the `getActiveMinesweeper` helper Javadoc too.

megamek/src/megamek/server/totalWarfare/TWGameManager.java
HammerGS: Confirmed RAW. TO:AuE p.138: "Minesweeper equipment remains fully effective until its 'armor' falls to 0 (or less), at which point the Minesweeper is treated as a destroyed item." So `getArmorValue() > 0` is just the not-yet-destroyed check. Note this predates the PR - it was in the original detection loops in `enterMinefield`/`checkVibraBombs`; I only centralized it into the shared helper.

megamek/src/megamek/server/totalWarfare/TWGameManager.java
HammerGS: Added `@Nullable` to `getActiveMinesweeper()`.

megamek/src/megamek/common/units/Entity.java
```diff
@@ -6724,6 +6724,19 @@ public boolean hasNavalC3() {
...
+     *       is currently activated. Used to offer the End-Phase activation toggle.
+     */
+    public boolean hasMinesweeper() {
+        for (MiscMounted mounted : getMisc()) {
+            if (mounted.getType().hasFlag(MiscType.F_MINESWEEPER)) {
+                return true;
+            }
+        }
+        return false;
+    }
```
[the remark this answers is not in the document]
HammerGS: Good catch. Added `&& mounted.isOperable()` so `hasMinesweeper()` - and therefore the End-Phase button and the dialog list - only counts operable sweepers. A destroyed/breached/missing one is no longer offered. Matches the `hasNovaCEWS()` pattern.

megamek/src/megamek/client/ui/dialogs/phaseDisplay/MinesweeperActivationDialog.java
```diff
@@ -0,0 +1,340 @@
...
+     * Returns the first minesweeper mounted on the entity, or {@code null} if it has none.
+     */
+    private static @Nullable MiscMounted getMinesweeper(Entity entity) {
+        for (MiscMounted mounted : entity.getMisc()) {
+            if (mounted.getType().hasFlag(MiscType.F_MINESWEEPER)) {
+                return mounted;
+            }
+        }
+        return null;
+    }
```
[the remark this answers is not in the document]
HammerGS: Done - `getMinesweeper()` now also requires `mounted.isOperable()`, so the dialog won't display or send a mode change for an inoperable sweeper. Defense-in-depth alongside the `hasMinesweeper()` filter.

IllianiBird (approved):

Issue #416, opened 2016-11-17
Kill Count Victory Condition resets mid-game


VampireSeraphin:
[autosave_2016-11-17_12-36-48.sav.gz](https://github.com/MegaMek/megamek/files/598076/autosave_2016-11-17_12-36-48.sav.gz)

During my game I scored the 8 kills vic condition fairly early in the game. Later, when the units that scored those kills died, the kill count victory condition was no longer met.

arlith:
I changed `KillCountVictory.updateKillTables` to use `Game.getEntityFromAllSources` to get the "killer" instead of `Game.getEntity`.  `Game.getEntity` will only return units that aren't removed from the game, whereas `Game.getEntityFromAllSources` does as it says and gets Entity's from all sources.

Your saved game happens after the victory condition is achieved and so isn't very useful for testing.  I've done a bad thing and committed this without testing, but the issue and fix are very obvious.

HammerGS:
related https://github.com/MegaMek/megamek/issues/417

PR #8355, opened 2026-06-18
Fix swapped VTOL MP for Microlite/Micro-Copter

### LINKED ISSUE #8354: [Issue] Data - VTOL Infantry's VTOL MP
Image

According to TacOps 3rd printing, p.324, the Mechanized VTOL Infantry (Microlite) receives 6 VTOL MP, and the Micro-copter receives 5 VTOL MP.

Image

Image

However, the MML gives the Microlite 5 MP and the Micro-copter 6 MP during build.

HammerGS:
Fixes #8354

Microlite VTOL infantry should have 6 VTOL MP and Micro-Copter platoons 5 (TO p.324). They were built with the values swapped.

Root cause: ConvInfantry.setMovementMode() picks the VTOL MP from the microlite flag, but MegaMekLab sets the flag after the movement mode, so the MP was read from a stale flag.

Fix: Added MICROLITE_VTOL_MP (6) and MICRO_COPTER_VTOL_MP (5) constants used by setMovementMode(). setMicrolite() now re-syncs the VTOL MP when the unit is already in VTOL mode, so the MP is correct regardless of the order the flag and movement mode are set. Adds ConvInfantryVtolMovementTest covering both call orders.

### DIFF
--- a/megamek/src/megamek/common/units/ConvInfantry.java
+++ b/megamek/src/megamek/common/units/ConvInfantry.java
@@ -169,6 +169,11 @@ public class ConvInfantry extends Infantry {
      */
     private boolean isMicrolite = false;
 
+    /** Base VTOL MP for a Microlite VTOL platoon (TO p.324). */
+    private static final int MICROLITE_VTOL_MP = 6;
+    /** Base VTOL MP for a Micro-Copter VTOL platoon (TO p.324). */
+    private static final int MICRO_COPTER_VTOL_MP = 5;
+
     private boolean pheromoneImpaired = false;
 
     public static final int ANTI_MEK_SKILL_NO_GEAR = 8;
@@ -1769,11 +1774,7 @@ public class ConvInfantry extends Infantry {
                 setSpecializations(getSpecializations() | SCUBA);
                 break;
             case VTOL:
-                if (hasMicrolite()) {
-                    setOriginalJumpMP(6);
-                } else {
-                    setOriginalJumpMP(5);
-                }
+                setOriginalJumpMP(hasMicrolite() ? MICROLITE_VTOL_MP : MICRO_COPTER_VTOL_MP);
                 setOriginalWalkMP(1);
                 break;
             case INF_UMU:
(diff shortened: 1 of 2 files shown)


megamek/unittests/megamek/common/units/ConvInfantryVtolMovementTest.java
```diff
@@ -0,0 +1,100 @@
...
+import megamek.common.equipment.EquipmentType;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Tests the base VTOL movement points of mechanized VTOL conventional infantry: Microlite platoons get 6 VTOL MP and
+ * Micro-Copter platoons get 5 VTOL MP (TO p.324). The MP must come out correct regardless of whether the microlite flag
+ * is set before or after the movement mode, because the two construction paths set them in different orders (the .blk
+ * loader sets microlite first; the MegaMekLab structure tab set the mode first). Regression test for mm-data issue #416.
```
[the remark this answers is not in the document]
HammerGS: Fixed in 0d366f32a3 - changed "set" to "sets". Also updated the stale mm-data #416 reference in the same Javadoc to MegaMek #8354.

megamek/src/megamek/common/units/ConvInfantry.java
```diff
@@ -169,6 +169,11 @@ public class ConvInfantry extends Infantry {
      */
     private boolean isMicrolite = false;
 
+    /** Base VTOL MP for a Microlite VTOL platoon (TO p.324). */
+    private static final int MICROLITE_VTOL_MP = 6;
+    /** Base VTOL MP for a Micro-Copter VTOL platoon (TO p.324). */
+    private static final int MICRO_COPTER_VTOL_MP = 5;
```
psikomonkie: "TO" is the old book. I believe this should reference TO:AUE p. 136.

psikomonkie (approved):
I'd like it if the rulebook references can be replaced with a newer printing. 

Something about `setMicrolite` automatically setting the MP seems off - are we certain there no cases where this will be called on vtol infantry & we don't want to modify the original MP? But I looked at how it's used and don't think it'll immediately cause a problem.

megamek/src/megamek/common/units/ConvInfantry.java
HammerGS: Done in a1d363ce4a - updated the constants, the setMicrolite comment, and the test Javadoc to reference TO:AUE p.136.

HammerGS:
Thanks for the approval. On the rulebook reference - done in a1d363ce4a (TO:AUE p.136).

On setMicrolite re-syncing the MP: the re-sync only fires when the unit is already in VTOL mode, and for VTOL infantry the original jump MP is fully defined by the platoon type - 6 for Microlite, 5 for Micro-Copter (TO:AUE p.136). There is no VTOL case where we'd want a different original jump MP, so there's nothing for it to clobber. The microlite flag is only meaningful on VTOL infantry, so the guard never touches non-VTOL units. The two callers (the .blk loader and the MML structure tab) both set it while building a VTOL platoon - the re-sync just makes the result order-independent, which was the root cause of #8354.

PR #8371, opened 2026-06-21
Correct Document Linking

heckindoc:
When I reorganized documents I failed to update some of the property files, I went through and tried to make sure I updated all of the property files which were affected. 

I also had to restore one of the previously archived documents as it is still being pointed to within a property file. This should actually be resolved when the MegaMek Manual is updated and added.


HammerGS:
@copilot resolve the merge conflicts in this pull request

IllianiBird (approved):
Needs to have merge conflicts fixed, otherwise no issue

heckindoc:
Sorry, I did not mean to do that -- I opened a new PR with these changes.

PR #8373, opened 2026-06-22
Hopefully Fixed Unit Test Faction Race Condition

IllianiBird:
MekHQ has an annoying Unit Test race condition where if another test initializes factions before a test that relies on faction data the stale data can persist. Polluting all future tests. Generally this isn't an issue unless you are specifically testing factional behavior.

This PR adds a method to explicitly set the `Faction2` instance. I cannot be sure if this will fix the issue due to the nature of a race condition, however I'm hopeful.

### DIFF
--- a/megamek/src/megamek/common/universe/Factions2.java
+++ b/megamek/src/megamek/common/universe/Factions2.java
@@ -105,7 +105,7 @@ public final class Factions2 {
     private static final MMLogger LOGGER = MMLogger.create(Factions2.class);
     private static Factions2 instance;
 
-    private static final String TEST_DIR = "testresources/data/universe/factions";
+    public static final String FACTIONS2_TEST_DIRECTORY = "testresources/data/universe/factions";
 
     private final Map<String, Faction2> factions = new HashMap<>();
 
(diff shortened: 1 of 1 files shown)


megamek/src/megamek/common/universe/Factions2.java
psikomonkie, line 131: `getInstance()` is `synchronized`. We need to make `setInstance` synchronized, at the very least, to prevent threading errors. 

I'm also not sure exposing `setInstance()` will fix the error - I worry setting the instance could break something using the original instance.

psikomonkie (changes requested):
Synchronization issue needs resolved at the very least.

megamek/src/megamek/common/universe/Factions2.java
IllianiBird: Do you have an alternative suggestion? I'm out of ideas here.

psikomonkie: With `Factions2` being a singleton, and trying to use two different test data resources, I don't think we can avoid it while using `getInstance()`. That said, I do have a suggestion for the other PR.

rjhancock (approved):

PR #8387, opened 2026-06-23
Add itemized heat breakdown tooltips to the Heat Phase report

HammerGS:
## Summary
Makes both numbers on the Heat Phase line — the heat **gained** and the heat **sunk** — hover/click links
that show an itemized breakdown of what produced them, using the same `Report.addDataWithTooltip(...)`
mechanism that already backs the to-hit "needs" value. Previously these were opaque totals: a unit that
fired two PPCs read "gains 20 heat" with no indication it was two weapons, and "sinks 15 heat" gave no
breakdown at all.

https://github.com/user-attachments/assets/5325536a-ce4a-4bab-91ad-c7f62ba576c2

The white line on the video is an artifact of the recording. 

## Changes
1. **New `HeatBreakdown` domain class** (kept *out* of the `Entity` / `HeatResolver` mega-classes). It holds
   the itemized buildup (`label -> HeatContribution(count, totalHeat)`, a serializable record) and
   dissipation (`label -> heat`) maps and the tooltip formatters. `Entity` keeps only a single
   `HeatBreakdown` field, a thin `getHeatBreakdown()` accessor, and `addHeatBuildup(int heat, String reason)`
   (which must stay on `Entity` because it also adjusts the existing `heatBuildup` total). Both breakdowns are
   insertion-ordered, null-guarded for old saves, serializable, and cleared when heat resolves.
2. **Routed the heat sources through the trackers.** Converted the `heatBuildup +=` sites for movement
   (standing/walking/running/jumping/sprinting/UMU/damaged-RHS), weapons fired (per-weapon name, large-craft
   arc, rapid-fire MG), and the Heat Phase effects (engine hits, stealth, Nova CEWS, void/null signature,
   chameleon, vibroblades, capacitors, flawed cooling, external heat, cooling, combat computer) to
   `addHeatBuildup(..., reason)`. Recorded the dissipation sources (heat sinks, submerged/water bonus,
   radical heat sink, coolant pod) as `toSink` is computed.
3. **Wired the tooltips into report 5035** (Mek + Aero paths). `report.add(heatBuildup)` /
   `report.add(toSink)` become `addHeatBuildupValue(...)` / `addHeatDissipationValue(...)`, which attach a
   single-line tooltip when a breakdown exists and otherwise fall back to a plain value.
4. **Readable multi-weapon display.** Several of the same weapon read e.g. `PPC x2: +20 (10 each)` instead of
   a single misleading `+20`. A trailing `Other: +N` reconciles any heat from sources not individually
   itemized so the listed lines always sum to the reported total.

## Files Changed
- `megamek/src/megamek/common/units/HeatBreakdown.java` - **new** serializable class holding the itemized
  buildup/dissipation maps, the `HeatContribution` record, and the tooltip formatters.
- `megamek/src/megamek/common/units/Entity.java` - one `HeatBreakdown` field, `addHeatBuildup(...)`
  delegator, `getHeatBreakdown()` / `clearHeatBuildupBreakdown()` accessors.
- `megamek/src/megamek/server/totalWarfare/HeatResolver.java` - record dissipation sources, route Heat-Phase
  heat through the tracker, report 5035 wiring (Mek + Aero), thin value helpers calling the formatters.
- `megamek/src/megamek/server/totalWarfare/TWGameManager.java` - movement-heat sites labelled by move type.
- `megamek/src/megamek/common/weapons/handlers/WeaponHandler.java` - weapon-heat site labelled by weapon /
  firing arc.
- `megamek/src/megamek/common/weapons/handlers/MGHandler.java` - rapid-fire MG heat labelled by weapon.
- `megamek/unittests/megamek/common/units/HeatBuildupBreakdownTest.java` - new tests (totals, per-source
  counts, dissipation summing/order, clearing both breakdowns, serialization round-trip).
- `docs/issues/features/heat-gain-breakdown-tooltip.md` - design/implementation notes.

## Notes / scope
- **External heat** is shown as one pooled "External heat" line in the gains tooltip rather than split by
  source (inferno / plasma / flamer / magma). Each of those already prints its own existing report line next
  to the total, and itemizing them individually would touch ~29 scattered handler sites; left for a possible
  follow-up.
- Dissipation figures are the unit's dissipation **capacity** by source, so a cool unit may show capacity
  greater than the heat actually sunk that turn.
- No new strings are user-facing via resource bundles here as the heat report (report 5035) is unchanged;
  only the value fields gain tooltips. Behavior of heat totals/sinking is unchanged.

## Testing
- New `HeatBuildupBreakdownTest`: running total, per-source counts (PPC x2), dissipation summing + order,
  clearing both breakdowns, and an entity serialization round-trip.
- Regression: heat / weapons / `server.totalWarfare` suites pass; `compileJava` clean.
- Manual: moved a Mek and fired weapons, then hovered/clicked the "gains N heat" and "sinks N heat" values in
  the Heat Phase report to confirm the breakdowns (including cooling fluids showing as negative lines).


megamek/src/megamek/common/units/HeatBreakdown.java
```diff
@@ -0,0 +1,184 @@
...
+    public void addDissipation(int heat, String reason) {
+        if ((heat != 0) && (reason != null) && !reason.isBlank()) {
+            dissipation.merge(reason, heat, Integer::sum);
+        }
+    }
+
+    /** @return the itemized heat-buildup contributions for this turn (source label -> count and signed total) */
+    public Map<String, HeatContribution> buildup() {
+        return buildup;
+    }
```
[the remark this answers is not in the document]
HammerGS: Good catch - changed to return Collections.unmodifiableMap(buildup) in 0ad8093ec2. The formatters iterate the field directly, so insertion order is preserved while callers can no longer mutate the breakdown.

megamek/src/megamek/common/units/HeatBreakdown.java
```diff
@@ -0,0 +1,184 @@
...
+
+    /** @return the itemized heat-dissipation contributions for this turn (source label -> heat removed) */
+    public Map<String, Integer> dissipation() {
+        return dissipation;
+    }
```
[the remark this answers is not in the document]
HammerGS: Done in 0ad8093ec2 - dissipation() now returns Collections.unmodifiableMap(dissipation) for the same reason.

megamek/src/megamek/common/units/Entity.java
```diff
@@ -5562,6 +5567,38 @@ public int getEngineCritHeat() {
...
+        }
+        return heatBreakdown;
+    }
+
+    /**
+     * Clears the itemized heat breakdown. Called when heat resolves and {@link #heatBuildup} is reset, so the next turn
+     * starts fresh.
+     */
+    public void clearHeatBuildupBreakdown() {
+        getHeatBreakdown().clear();
```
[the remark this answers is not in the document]
HammerGS: Agreed - renamed to clearHeatBreakdown() in 0ad8093ec2 (it clears both the buildup and dissipation breakdowns); updated the two HeatResolver call sites and the test.

IllianiBird:
Not had a chance to review this yet, but do we want a dialog, or would a tooltip be better? I can imagine a situation where I want to view multiple units in quick succession and a hyperlink would offer a smoother experience

HammerGS:
> Not had a chance to review this yet, but do we want a dialog, or would a tooltip be better? I can imagine a situation where I want to view multiple units in quick succession and a hyperlink would offer a smoother experience

I just based it off the existing system used for to hits etc

IllianiBird:
Ah, I always thought those were tooltips, not that I've ever bothered to check.

Sounds good

megamek/src/megamek/common/units/Entity.java
```diff
@@ -5562,6 +5567,38 @@ public int getEngineCritHeat() {
...
+        heatBuildup += heat;
+        getHeatBreakdown().addBuildup(heat, reason);
+    }
+
+    /**
+     * @return this unit's itemized heat buildup/dissipation record for the current turn (never null). Used to build the
+     *       Heat Phase report's "gains N heat" / "sinks N heat" breakdown tooltips.
+     */
+    public HeatBreakdown getHeatBreakdown() {
+        if (heatBreakdown == null) {
```
IllianiBird: You're initializing heatBreakdown with a default value. Unless you're setting it to null elsewhere I don't think we need to check for null here

megamek/src/megamek/common/units/HeatBreakdown.java
```diff
@@ -0,0 +1,191 @@
...
+            if ((count > 1) && (heat % count == 0)) {
+                tooltip.append(" (").append(heat / count).append(" each)");
+            }
+        }
+        int remainder = reportedTotal - tracked;
+        if ((remainder != 0) && (tracked >= 0) && (reportedTotal >= 0)) {
+            if (tooltip.length() > 0) {
+                tooltip.append(", ");
+            }
+            tooltip.append("Other: ").append(remainder > 0 ? "+" : "").append(remainder);
```
IllianiBird: Hardcoded string

megamek/src/megamek/common/units/HeatBreakdown.java
```diff
@@ -0,0 +1,191 @@
...
+     * Formats the buildup breakdown into a single-line tooltip for the "gains N heat" value (e.g. "Movement (Running):
+     * +2, PPC x2: +20 (10 each), Cooling (water/coolant/etc.): -9"). Any heat not itemized here is reconciled into a
+     * trailing "Other" entry so the listed contributions still sum to the reported total - skipped when the net is
+     * negative, since the reported buildup is then clamped to 0.
+     *
+     * @param reportedTotal the heat-buildup total shown on the report (used to reconcile un-itemized heat)
+     *
+     * @return the tooltip text, or null if nothing was itemized this turn
+     */
+    public String buildupTooltip(int reportedTotal) {
```
IllianiBird, line 135: Nullable tag

megamek/src/megamek/common/units/HeatBreakdown.java
```diff
@@ -0,0 +1,191 @@
...
+        if (buildup.isEmpty()) {
+            return null;
```
IllianiBird: Why are we returning null here? Might be better to return an empty string instead

megamek/src/megamek/common/units/HeatBreakdown.java
```diff
@@ -0,0 +1,191 @@
...
+    /**
+     * Formats the dissipation breakdown into a single-line tooltip for the "sinks N heat" value (e.g. "Heat sinks: 10,
+     * Submerged: 2, Radical heat sink: 3"). The figures are the unit's dissipation capacity by source; a cool unit may
+     * not use all of it.
+     *
+     * @return the tooltip text, or null if nothing recorded dissipation this turn
+     */
+    public String dissipationTooltip() {
+        if (dissipation.isEmpty()) {
+            return null;
```
IllianiBird: Same here

megamek/src/megamek/common/weapons/handlers/WeaponHandler.java
```diff
@@ -1949,11 +1949,12 @@ protected void addHeat() {
                 int loc = weapon.getLocation();
                 boolean rearMount = weapon.isRearMounted();
                 if (!attackingEntity.hasArcFired(loc, rearMount)) {
-                    attackingEntity.heatBuildup += attackingEntity.getHeatInArc(loc, rearMount);
+                    attackingEntity.addHeatBuildup(attackingEntity.getHeatInArc(loc, rearMount),
+                          "Weapons fired (" + attackingEntity.getLocationName(loc) + " arc)");
```
IllianiBird: Hardcoded string

megamek/src/megamek/server/totalWarfare/HeatResolver.java
```diff
@@ -318,43 +318,43 @@ void resolveHeat() {
             }
 
             // engine hits add a lot of heat, provided the engine is on
-            entity.heatBuildup += entity.getEngineCritHeat();
+            entity.addHeatBuildup(entity.getEngineCritHeat(), "Engine hits");
 
             // If a Mek had an active Stealth suite, add 10 heat.
             if (entity.isStealthOn()) {
-                entity.heatBuildup += ArmorType.STEALTH_ARMOR_HEAT;
+                entity.addHeatBuildup(ArmorType.STEALTH_ARMOR_HEAT, "Stealth armor");
```
IllianiBird: We're adding a bunch of hardcoded strings here

megamek/src/megamek/server/totalWarfare/HeatResolver.java
```diff
@@ -1054,19 +1069,30 @@ void resolveAeroHeat(Entity entity, Vector<Report> vPhaseReport, Vector<Report>
         if (max_ext_heat < 0) {
             max_ext_heat = 15; // standard value specified in TW p.159
         }
-        entity.heatBuildup += Math.min(max_ext_heat, entity.heatFromExternal);
+        entity.addHeatBuildup(Math.min(max_ext_heat, entity.heatFromExternal),
+              "External heat (fire/inferno/plasma)");
         entity.heatFromExternal = 0;
         // remove heat we cooled down
-        entity.heatBuildup -= Math.min(9, entity.coolFromExternal);
+        entity.addHeatBuildup(-Math.min(9, entity.coolFromExternal), "Cooling (water/coolant/etc.)");
```
IllianiBird: If we're accepting both positive and negative values into this method it probably should be renamed to something like 'changeHeatBuildup'. That said, wouldn't a negative head buildup be a dissipation? I thought I saw that had its own method

megamek/src/megamek/server/totalWarfare/HeatResolver.java
```diff
@@ -1425,4 +1455,39 @@ private Report createTCPShutdownAvoidanceReport(Entity entity) {
...
+     * Adds the unit's net heat-buildup value to the heat-phase report. When an itemized breakdown was tracked this
+     * turn, the value is made a hover/click link whose tooltip lists every contribution; otherwise it is added as a
+     * plain value.
+     *
+     * @param report the heat-phase report being assembled (report 5035)
+     * @param entity the unit whose heat buildup is being reported
+     */
+    private static void addHeatBuildupValue(Report report, Entity entity) {
+        String tooltip = entity.getHeatBreakdown().buildupTooltip(entity.heatBuildup);
+        if (tooltip != null) {
```
IllianiBird: Better to use StringUtility.nullOrBlank (I think it's called) then you done need to return null

megamek/src/megamek/server/totalWarfare/HeatResolver.java
```diff
@@ -1425,4 +1455,39 @@ private Report createTCPShutdownAvoidanceReport(Entity entity) {
...
+     * dissipation breakdown was recorded this turn, the value is made a hover/click link whose tooltip lists each
+     * dissipation source; otherwise it is added as a plain value.
+     *
+     * @param report the heat-phase report being assembled (report 5035)
+     * @param entity the unit whose dissipation is being reported
+     * @param toSink the heat actually dissipated this turn
+     */
+    private static void addHeatDissipationValue(Report report, Entity entity, int toSink) {
+        String tooltip = entity.getHeatBreakdown().dissipationTooltip();
+        if (tooltip != null) {
```
IllianiBird: And here

megamek/src/megamek/server/totalWarfare/TWGameManager.java
```diff
@@ -8189,32 +8189,32 @@ void washInferno(Entity entity, Coords coords) {
     public void addMovementHeat() {
         for (Entity entity : game.inGameTWEntities()) {
             if (entity.hasDamagedRHS()) {
-                entity.heatBuildup += 1;
```
IllianiBird, line 8192: You know the drill :)

IllianiBird (changes requested):

megamek/src/megamek/common/units/Entity.java
HammerGS: Good call - removed the guard; the field is initialized inline so getHeatBreakdown() just returns it now. (6b48d8a78f)

megamek/src/megamek/common/units/HeatBreakdown.java
HammerGS: Moved to the resource bundle - the source labels now resolve via Messages.getString("HeatBreakdown.*") with keys in messages.properties. (6b48d8a78f)

megamek/src/megamek/common/units/HeatBreakdown.java
HammerGS: Resolved by the empty-string change below - buildupTooltip returns "" instead of null now, so no @Nullable needed. (6b48d8a78f)

megamek/src/megamek/common/units/HeatBreakdown.java
HammerGS: Done - both tooltip formatters return "" now, and the HeatResolver helpers use StringUtility.isNullOrBlank to decide plain-vs-tooltip. (6b48d8a78f)

megamek/src/megamek/common/units/HeatBreakdown.java
HammerGS: Same fix - dissipationTooltip returns "" too. (6b48d8a78f)

megamek/src/megamek/common/weapons/handlers/WeaponHandler.java
HammerGS: i18n-ed - the arc label now uses Messages.getString("HeatBreakdown.weaponsFiredArc", arcName); the weapon.getName() label was already localized. (6b48d8a78f)

megamek/src/megamek/server/totalWarfare/HeatResolver.java
HammerGS: All moved to the bundle (HeatBreakdown.* keys in messages.properties). (6b48d8a78f)

megamek/src/megamek/server/totalWarfare/HeatResolver.java
HammerGS: Done - addHeatBuildupValue now uses !StringUtility.isNullOrBlank(tooltip) and the tooltip returns "" rather than null. (6b48d8a78f)

megamek/src/megamek/server/totalWarfare/HeatResolver.java
HammerGS: Same for addHeatDissipationValue. (6b48d8a78f)

megamek/src/megamek/server/totalWarfare/TWGameManager.java
HammerGS: Movement labels i18n-ed too (HeatBreakdown.movement* keys). (6b48d8a78f)

megamek/src/megamek/server/totalWarfare/HeatResolver.java
HammerGS: Renamed to changeHeatBuildup. On the dissipation question - they are deliberately two different report numbers:

- changeHeatBuildup feeds the "gains N heat" value. External cooling (water/coolant) and the combat-computer reduction are negative *buildup* because they reduce that gain before sinking - they are part of the net heat gained, not heat-sink output.
- addHeatDissipation feeds the separate "sinks N heat" value (heat sinks, submerged/water bonus, radical heat sink, coolant pod).

So a negative buildup is not the same thing as dissipation; putting cooling through addHeatDissipation would wrongly show it on the "sinks" side. The rename makes the signed semantics explicit. (6b48d8a78f)

IllianiBird (approved):

PR #8400, opened 2026-06-26
Added Overload to `parseBoolean` that Uses a Default Default-Value

IllianiBird:
As per title. Streamlines calls.


Sleet01 (approved):
LGTM

PR #8420, opened 2026-06-30
removed precision PAC ammo, alias added to standard

exeea:
This PR removes the Precision PAC ammo mutator and adds a fallback to load units saved with it to standard PAC ammo.

Caseless/Flak mutators got their standalone mutator definition without IS advancement.


IllianiBird (approved):

IllianiBird:
Approved pending tests passing

exeea:
> Approved pending tests passing

il will never pass :D 
is a checkstyle, it doesn't like the "ProtoMech" word but I can't change it either.

PR #8463, opened 2026-07-07
Stop WeaponType/AmmoType.hasFlag warning spam when saving BattleArmor

HammerGS:
## Root Cause
`EntityListFile.writeEntityList` iterates every BattleArmor equipment mount and calls `mounted.getType().hasFlag(MiscType.F_BA_MEA / F_AP_MOUNT)`. For weapon and ammo mounts, `getType()` is a
`WeaponType`/`AmmoType`, and the typed `hasFlag` logs a warning with a full stack trace on every cross-type flag check — producing dozens to hundreds of stack traces (and the wasted `Throwable` construction) each time a force containing BattleArmor is saved.

## Changes
1. `EntityListFile.writeEntityList` — guard the BA equipment loop with an `instanceof MiscType` check (MEA and AP mounts only exist on MiscType equipment), so the MiscType flags are only tested on MiscType mounts. Renamed the loop variable `m` → `mounted` and `mountIdx` → `mountIndex` while there.

## Files Changed
- `megamek/src/megamek/common/units/EntityListFile.java` — `instanceof MiscType` guard in the BA save loop.

## Testing
- Behavior-preserving (weapon/ammo mounts already produced no MEA/AP output and returned false; only the per-call warning and stack construction are removed). Existing `EntityListFileDisposableWeaponTest` stays green.
- Runtime A/B on a real BattleArmor (Achileus, 15 mounts) saved to MUL: `main` logs 12 `Incorrect flag check`
  warnings, this branch logs 0; the save succeeds in both cases.
- `./gradlew :megamek:compileJava` + the save test green.

### DIFF
--- a/megamek/src/megamek/common/units/EntityListFile.java
+++ b/megamek/src/megamek/common/units/EntityListFile.java
@@ -1264,16 +1264,21 @@ public class EntityListFile {
             }
 
             if (entity instanceof BattleArmor ba) {
-                for (Mounted<?> m : entity.getEquipment()) {
-                    if (m.getType().hasFlag(MiscType.F_BA_MEA)) {
+                for (Mounted<?> mounted : entity.getEquipment()) {
+                    // Only MiscType equipment carries the BA MEA / AP-mount flags; guarding on the type avoids
+                    // calling WeaponType/AmmoType.hasFlag with a MiscType flag, which logs a warning per call.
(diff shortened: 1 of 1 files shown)


HammerGS:
**Merge order:** Independent — no dependency on the other open switchable-AI / perf PRs; can be merged in any order.

IllianiBird (approved):

PR #8493, opened 2026-07-12
store uuid into units

exeea:
This PR adds to units the UUID property. 
We will finally have a stable identifier.


HammerGS (approved):

NickAragua:
Is it possible to get some background for the motivation for this change? Not necessarily opposed, just want to understand what problem it's solving. If more convenient, this can be done over discord.

PR #8516, opened 2026-07-15
Add a Composite Tech Level report (and fix prototype mislabelling)

HammerGS:
Summary

Adds a Composite Tech Level report that shows where a unit's tech level and availability dates come from — which part makes the unit Prototype / Production / Common, and which part takes it extinct. It is opened from MegaMekLab's Reports menu (companion MegaMekLab PR).

While building the report I found and fixed a related bug in the tech-date calculation that was labelling a broad set of units as Prototype when they should have been in normal Production (details below). The report is what made the bug easy to see, so the two ship together here.

https://github.com/user-attachments/assets/5950b0d6-db93-4805-b366-1a309fe40216

---
Part 1 — the report

A unit's tech level (Introductory / Standard / Advanced / Experimental) and its availability dates are the combined result of every part it is built from, but only the final answer was ever shown. When a unit came out Experimental, there was no way to see which part made it so.

The report shows, for a unit:

- Component Tech Progression — every part with its own prototype / production / common / extinct / returns dates, plus its Static and Variable tech level.
- Unit Tech Progression — the running unit dates as each part is folded in, with the part that drives each date highlighted and labelled (e.g. "Becomes Common (3045)").
- Unit result — the final Static and Variable tech level and the full date ranges.

Worked example (Devastator DVS-2-EC)

The build-up table makes it obvious that the unit's 300 XL engine is what pushes the unit's common date out to 3045 (labelled "Becomes Common (3045)") and takes the whole unit extinct in 2865 (shown in its own Extinct / Returns columns). Later parts — the Improved Gauss Rifle and Enhanced PPC — tighten the extinction window further, each flagged on its own row.

How it works (the important design point)

The report does not re-derive the part list — it records the real calculation as it happens, so the report can never drift from the actual tech level. CompositeTechLevel gained an opt-in recording hook that is off during normal play; a RecordingCompositeTechLevel subclass captures each part as it is folded in. Because recording is off normally and the records live in a transient field, game state and saved games are completely unaffected. Verified across all 10,989 units in the cache that the recorded result matches each unit's real tech level exactly (0 mismatches) and that every part gets a name (0 blanks).

The unit classes now pass a name for their system parts (armor, structure, gyro, cockpit, turrets, docking collar, infantry specializations) so every report row is named — these are one-line, no-behaviour-change edits. Infantry / battle armor anti-Mek attack pack, etc.) are dropped from the report as noise; their dates still fold into the calculation. The report's own text is localized through messages.properties.

---
Part 2 — the prototype bug fix

What players see

Some units were labelled Prototype for a long stretch of their life when they should have been in Production. The clearest example is the Elemental Battle Armor: introduced in 2868, but shown as Prototype 2868-3053 instead of Production 2868-3053, Common 3054+.

Root cause

The code that folds each part into a unit's dates (CompositeTechLevel.addComponent) handled a part entering production earlier than the unit and later than the unit, but not the case where it enters production in the exact same year the unit already does. That into a branch written for parts that never went into production, which wiped out the unit's production date and relabelled its whole production span as "prototype". The Elemental trips this because its chassis and almost all of its Clan gear enter production in 2868 — the suit's own introduction year.

The fix and its blast radius

A part entering production in the same year the unit already is now changes no the cache (10,989 units) under the old and new code and compared:

- 656 units change. 646 lose a bogus prototype phase entirely; 10 have one shortened.
- 0 units newly become prototype, and 0 common dates move — the change only everimental, never more, so nothing correct becomes wrong.

Affected units span Age of War gun emplacements and turrets, field guns, a wide range of Clan battle armor, and 399 units introduced in 3050 or later.

---
Files changed

New:
- common/CompositeTechLevelReport.java, common/RecordingCompositeTechLevel.java, client/ui/dialogs/abstractDialogs/TechLevelDisplayDialog.java
- unittests/.../CompositeTechLevelReportTest.java, unittests/.../EntityTechTeson tests)

Modified:
- common/CompositeTechLevel.java — the prototype fix plus the recording hook
- common/units/{Entity,Mek,Tank,VTOL,Aero,Dropship,ProtoMek,ConvInfantry}.java — name the system parts (no behaviour change)
- resources/megamek/client/messages.properties (dialog title) and resources/megamek/common/messages.properties (report text)

Test fixtures: Elemental Battle Armor [Laser](Sqd5).blk, Devastator DVS-2-EC.mtf


megamek/src/megamek/common/units/ProtoMek.java
```diff
@@ -420,9 +420,9 @@ public TechAdvancement getConstructionTechAdvancement() {
     }
 
     @Override
-    protected void addSystemTechAdvancement(CompositeTechLevel ctl) {
+    protected void addSystemTechAdvancement(CompositeTechLevel techLevel) {
         if (interfaceCockpit) {
-            ctl.addComponent(TA_INTERFACE_COCKPIT);
+            techLevel.addComponent(TA_INTERFACE_COCKPIT, "Interface cockpit");
         }
     }
```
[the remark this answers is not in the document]
HammerGS: Good catch that ProtoMek is the only `Entity` subclass whose `addSystemTechAdvancement` doesn't call `super`. This predates this PR, though: on `main`, `ProtoMek.addSystemTechAdvancement` already only folds in the interface cockpit and skips `super` — this PR just added the `"Interface cockpit"` display name to that existing call.

Adding `super` here would pull armor, internal structure and engine into every ProtoMek's composite tech calculation, which is a game-behaviour change that needs validating against ProtoMek construction rules and is out of scope for this PR. I'd rather raise it as a separate follow-up than slip a behaviour change in here. Happy to open an issue for it.

megamek/src/megamek/common/CompositeTechLevelReport.java
```diff
@@ -0,0 +1,634 @@
...
+              useVariableTechLevel ? message("groupVariable") : message("groupStatic"),
+              (techFaction == Faction.NONE) ? null : techFaction.getCodeMM(),
+              rows,
+              formatLevel(staticTechLevel),
+              formatLevel(variableTechLevel),
+              techLevel.getPrototypeDateRange(),
+              techLevel.getProductionDateRange(),
+              techLevel.getCommonDateRange(),
+              normalizeRange(techLevel.getExtinctionRange()),
+              formatLevel(useVariableTechLevel ? variableTechLevel : staticTechLevel));
```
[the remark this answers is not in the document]
HammerGS: Fixed in 897684fc90. All four ranges in the Unit result section (prototype / production / common / extinct) now go through `normalizeRange`, so an empty range consistently shows `--` like every other empty cell in the report.

megamek/src/megamek/client/ui/dialogs/abstractDialogs/TechLevelDisplayDialog.java
```diff
@@ -0,0 +1,182 @@
...
+        super.finalizeInitialization();
+        if (entity != null) {
+            setTitle(getTitle() + " (" + entity.getShortName() + ")");
+        }
+        updateDialogSize();
+    }
+
+    @Override
+    protected Container createCenterPane() {
+        JButton exportText = new JButton("Copy as Text");
```
IllianiBird: Hardcoded player-facing strings. I'll not note any others, but this should be addressed throughout the PR

megamek/src/megamek/client/ui/dialogs/abstractDialogs/TechLevelDisplayDialog.java
```diff
@@ -0,0 +1,182 @@
...
+
+    private String plainTextReport() {
+        return CompositeTechLevelReport.toPlainText(entity, techFaction, evaluationYear, useVariableTechLevel);
+    }
+
+    /** Does gui-scaling, packs the dialog and reduces the height if it's too big. */
+    private void updateDialogSize() {
+        pack();
+        Dimension screenSize = UIUtil.getScaledScreenSize(this);
+        setSize(new Dimension(getSize().width, Math.min(getHeight(), (int) (screenSize.getHeight() * 0.8))));
```
IllianiBird: Due to integer rounding this can result in a 0 height element which would then cause all kinds of errors

IllianiBird (dismissed):
Minor changes necessary then this can be merged - I'll be streaming all day tomorrow, so once my requests have been actioned feel free to Admin merge.

megamek/src/megamek/client/ui/dialogs/abstractDialogs/TechLevelDisplayDialog.java
HammerGS: Fixed in 75ace6149a, swept across the whole PR:

- **Dialog** (`TechLevelDisplayDialog`): the two copy buttons and the no-unit error label now come from the client resource bundle (`TechLevelDisplayDialog.copyAsText` / `.copyAsHTML` / `.noUnit`).
- **Component names**: every literal passed to `addComponent(...)` in `Entity`, `Mek`, `Dropship`, `ProtoMek`, `Tank` and `VTOL` (Docking collar, Chin turret, Patchwork armor, Internal structure, etc.), plus the Chassis/Unnamed component names in `RecordingCompositeTechLevel` and the "x2" count suffix in the report, now come from `CompositeTechLevel.component.*` / `CompositeTechLevelReport.*` keys in the common bundle.

Names that come from existing APIs (cockpit/gyro type strings, infantry specialization names, equipment names) were left as-is, since those aren't localized anywhere in the codebase. One note: `BVDisplayDialog` hardcodes the same "Copy as Text"/"Copy as HTML" buttons (this dialog was modeled on it) - I left it untouched as out of scope, but happy to fix it in a follow-up.

megamek/src/megamek/client/ui/dialogs/abstractDialogs/TechLevelDisplayDialog.java
HammerGS: Fixed in 701246cc31. The 0.8-of-screen-height cap is now clamped to a 200-pixel minimum before it is applied, so a zero or near-zero reported screen height can no longer collapse the dialog into a zero-height window:

```java
int maximumHeight = Math.max(MINIMUM_DIALOG_HEIGHT, (int) (screenSize.getHeight() * 0.8));
setSize(new Dimension(getSize().width, Math.min(getHeight(), maximumHeight)));
```

PR #8519, opened 2026-07-15
Fix armor and structure name collisions and implement direct loaders

exeea:
This PR refactors how equipment, armor, and structure types are looked up by name to improve reliability, reduce naming collisions, and provide better debugging information. 
It adds explicit lookup methods for armor and structure types.
Adds collision detection and reporting.
Autogenerates lookup names with ` Structure` or ` Armor` suffixes for keys that don't have it in the name.

Additionally, on load of armor/structure, if they are identified as TechBase.ALL we cross-reference it with the unit's techbase.
If the unit techbase is strictly Clan or IS and the armor/structure definition in the file is mismatching, we ignore it.
Example:
an unit with `techbase:clan` and `armor:Standard(Inner Sphere)`. 
We now correctly load it as a clan armor and not IS.
a full resave of units with this procedure can be found here: https://github.com/MegaMek/mm-data/pull/478


megamek/src/megamek/common/templates/BattleArmorTROView.java
psikomonkie: Copyright needs updated!

psikomonkie (commented):
Other than the copyrights needing update, and the tests being fixed, I think this looks good. 

I poked around MekHQ's parts logic and I think it will be fine with this. 

I would like it if @pavelbraginskiy could take a quick peek, I feel like this information is used more on the MML-side - I don't often need to mess with the `EquipementType` at this level for the work I do, so while I think it looks sensible I'm also not super familiar with how we use it.

HammerGS (approved):

PR #8535, opened 2026-07-16
feat: allow bot to activate hidden units during firing phase

NickAragua:
A few changes:

- multiple units can take point blank shots (before, it was limited to just one for some reason)
- "maximum damage at range" estimate for bot-controlled infantry was always previously treated as below 0, probably causing weird bot infantry behavior
- After a hostile unit moves, Princess will look at her hidden units and evaluate whether it's worth it to activate them (based on her aggression level), doing so if the expected damage output passes the aggression threshold.


IllianiBird (approved):
Yessssssssss

PR #8541, opened 2026-07-17
Implement gm2.0 tools

### LINKED ISSUE #8497: [Issue] Add Unit, Add Army and Load Unit List sometimes disabled in the lobby
## Problem

The lobby sometimes opens with the Add Unit, Add Army, and Load Unit List buttons greyed out. Nothing re-enables them, so the only fix is to leave and re-enter the lobby. It happens intermittently, which points at a timing problem rather than a game-state one.

## Example

Start MegaMek and host a new game. Roughly one launch in several, the Unit Setup panel opens with Add Unit, Add Army, and Load Unit List all disabled. Other lobby controls work normally. Switching tabs or selecting players does not restore them.

More likely on a cold unit cache (delete `units.cache` to force a rebuild), because that changes when the cache finishes loading.

## Proposed fix

- In `ChatLounge.setupListeners()`, re-check `MekSummaryCache.isInitialized()` right after registering the done-loading listener, and enable the three buttons if the cache finished loading in the meantime.
- Enable the buttons on the event dispatch thread.

## Implementation notes

- Root cause is a startup race. The `ChatLounge` constructor calls `setupUnitConfig()`, which enables the buttons only if `MekSummaryCache.isInitialized()` is already true. The listener that would enable them later is not registered until `setupListeners()`, four calls afterwards. If the cache finishes loading in that window, the notification is missed and the buttons stay disabled.
- Separate defect in the same code: `MekSummaryCache.done()` notifies its listeners from its loader thread, so the listener was calling `setEnabled` on Swing components off the EDT.
- Affects `megamek/src/megamek/client/ui/panels/phaseDisplay/lobby/ChatLounge.java` only.
- Hard to unit test: reproducing it needs a constructed `ChatLounge`, which needs a full `ClientGUI`. Verify by launching the lobby several times, on both a cold and a warm unit cache.

## Background

Found during play testing of the gamemaster fixes, 2026-07-12.

HammerGS:
Title: Rework the GM damage editor around the armor diagram (Fixes #8497)

## Summary

Replaces PR #8505 (closed), with the fixes and rework found in testing since. The Edit Damage dialog listed armor, systems and equipment in three long columns, and a gamemaster had no good tools to run a game. This reworks the dialog around the unit's own armor diagram and makes it the gamemaster's console: damage, heat, crew hits, ammo, unit conditions, temporary skill modifiers, changing a unit's owner, and destroying a unit. In play, edits now travel as plain values that the server applies tovits own copy of the unit - not as a replacement unit carrying stale state.

https://github.com/user-attachments/assets/5d77aef7-fc15-4c3e-9f2d-1b5c95aebd3b

1. **The Game Master role** (`GameMasterCommand`, `TWGameManager`, `ChatLounge`) - one GM at a time,
   granted without a vote in the lobby and by vote in game, revoked when the game option turns off.
2. **The reusable armor diagram** (`picmap`,e diagram becomes clickable console: damage, heat, crew hits, ammo, unit conditions, temporary skill modifiers, changing a unit's owner, and destroying a unit. In play, edits now travel as plain values that the server applies to its own copy of the unit - not as a replacement unit carrying stale state.

1. **The Game Master role** (`GameMasterCommand`, `TWGameManager`, `ChatLounge`) - one GM at a time,
   granted without a vote in the lobby and by vote in game, revoked when the game option turns off.
2. **The reusable armor diagram** (`picmap`, `mapset`, `ArmorPanel`) - the diagram becomes clickable and scalable outside the unit display. Me
3. **The damage editor** (`unitEditor` package, `UnitEditorDialog`) - the dialog itself, split out of a 2000-line class.
4. **Server-side edits** (`DamageEditSpec`, `DamageEditApplier`, `ENTITY_DAMAGE_EDIT`) - in play the
   dialog sends values; the server applies them to its own authoritative unit.
5. **GM skill modifiers** (`TemporarySkillModifiers`, `Crew`, `/skillMod`, `ComputeToHit`) - a temporary gunnery/piloting/initiative change, timed or permanent.
6. **The lobby** (`ChatLounge`, `LobbyActions`, settings) - the Game Master button and its gates.

## Changes Made

- **The dialog is the armor diagram.** The suses, enlarged and clickable. Clicking a location shows that location's armor, structure, systems and equipment.  Systems with no location and the unit's coel below.
- **Critical hits are striped on the diagram**, over the damage color they already carry.
- **Heat, crew hits and ammo.** Crew hits stop one short of the six that kill; lowering a dead pilot's hits revives them, which `Crew.setHits` cannot do on its own. Critting an ammo bin empties it; taking the crit off restores the bin and its shots.
- **Unit conditions** moved here from Configure: shut down, prone, hull down, hidden, stealth, dug in, fuel. The gamemaster Configure men
- **In play, edits go to the server as values.** The server applies them to its own copy of the unit, checks whether the unit survives the Nothing the editor does not touch can be clobbered by a stale clieMekHQ the same values are applied locally, so an edit means the same thing on both paths.
- **GM skill modifiers.** A temporary gunnery/piloting delta and, under individual initiative, an initiative modifier - "+1 for 3 rounds" orog or the /skillMod command; both write the same state. Applied in the effective-skill getters, so combat, PSRs and Princess all follow it, and it reverses itself completely on expiry (with a GM toast). The to-hit breakdown shows it as its own "GM Modifiercustomization read the raw
  skills, so a mid-game save never bakes a temporary modifier in.
- **Change owner.** An Owner chooser in the her player through  /changeOwner; the switch resolves in the END phase. The Movement phase Traitor button uses the /traitor command the same way. A unit that changes hands recolors to its new side.
- **Destroy Unit** asks first, then has the server destroy the unit through the gamemaster kill command, and can eject the crew first. **Restore Unit** sets the unit fully repaired.
- **GM mode in the Lobby** It takes and gives up the role through the same /gm command the chat uses, so the rules live in one place. Taking it is put to a vote of the human players — a lone player's own yes passes at once — and the host's vote-threshold option decides unanimous or majority.-**
- **Two gates on the role**: a game option, Allow Game Master (on by default), which the host sets; and a client setting, Allow GM mode, which decides whether this player wants the role at all.
- **Only the owner or the gamemaster may edihy when it drops an update.

## Since PR #8505

- **Edits are sent as a `DamageEditSpec` instead of the edited Entity.** The old path let the caller push the whole unit back after the dialog closed; the stale copy erased server-side state - it is how a pending /changeOwner switch was silently cancelled. The push is gone.
- **Traitor works end to end**: the pending switch survives in-play entity updates, the unit recolors to its new side, and ownership changes go through server commands everywhere.
- **GM skill modifiers** (above) are new since the original notes, including the "GM Modifier" to-hit line and the always-visible column in the dialog's General panel.
- **The GM-appointed popup** now names the rthe GM Mode button or /gm), not just the host's game option.
- **[Traitor] log chain**: the command, the ND phase resolution all log, so a switch that goes missing can be traced to the hop that lost it.

## Files Changed

- `megamek/src/megamek/common/units/DamageEdg's edits as plain values
- `megamek/src/megamek/common/units/DamageEditApplier.java` - NEW: writes a spec onto a unit, on
  the server in play and locally otherwise
- `megamek/src/megamek/common/units/TemporarySkillModifiers.java` - NEW: the GM skill modifier state
- `megamek/src/megamek/common/units/Crew.java` - effective-skill getters, appliedGunneryModifier
- `megamek/src/megamek/common/units/LAMPilotthe modifier once
- `megamek/src/megamek/common/actions/compute/ComputeToHit.java` - the "GM Modifier" line
- `megamek/src/megamek/common/net/enums/PacketCommand.java` - ENTITY_DAMAGE_EDIT
- `megamek/src/megamek/server/totalWarfare/TWGameManager.java` - receiveDamageEdit, single Game
  Master, turn-state and traitor guards on elly edited unit
- `megamek/src/megamek/server/commands/SkillModifierCommand.java` - NEW: /skillMod
- `megamek/src/megamek/server/commands/GameMasterCommand.java` - the game option decides whether a
  game has a gamemaster
- `megamek/src/megamek/server/commands/KillCommand.java` - eject the crew before destroying a unit
- `megamek/src/megamek/client/ui/widget/picmap/*`, `mapset/*` - scale, a location listener, crit
  striping
- `megamek/src/megamek/client/ui/dialogs/UnitEditorDialog.java` - the dialog: 2020 lines down to 840
- `megamek/src/megamek/client/ui/dialogs/unitEditor/*` - NEW package: controls, panel builder, spec
  builder, crit panel, damage roller, diagra
- `megamek/src/megamek/client/ui/panels/phaseDisplay/lobby/ChatLounge.java` - the Game Master button
- `megamek/src/megamek/client/ui/panels/phaseDisplay/MovementDisplay.java` - Traitor button sends
  /traitor
- `megamek/src/megamek/client/ui/clientGUI/boardview/BoardView.java` - icon reload on owner change
- `megamek/unittests/...` - UnitEditorDialogTest, TemporarySkillModifiersTest,
  DamageEditApplierTest, TWGameManagerDamageEditTest, TWGameManagerTraitorUpdateTest,
  TWGameManagerFatalDamageEditTest, GameMasterCommandTest

## Testing

- Unit tests: 55 across the feature suites - UnitEditorDialogTest (17), TemporarySkillModifiersTest (11), DamageEditApplierTest (10), TWGameMa GameMasterCommandTest (5),
  TWGameManagerDamageEditTest (4), TWGameManagerTraitorUpdateTest (3). Full suite passes. The GUI
  tests skip in a headless environment.
- The tests cover what would silently break: a non-GM's edit packet is dropped, a fatal edit destroys the unit, a pending traitor switch survives an edit, zeroed modifiers clear an active one, Restore Unit resets the modifier spinners, and the applied GM modifier stays exact at the skill clamp.

## Manual testing (three clients: host GM, (Illiani and I have been playtesting this)
  - Edit a bot's damage in game; the edit lands and every client follows
  - Hand a unit to the observer via the dialog's Owner row; it switches at the END phase, recolors,
    and the observer stops being one
  - Set "+2 gunnery for 3 rounds" on a bot unit; its to-hit shows the GM Modifier line and the
    rounds count down
  - Take and give up the Game Master role wiecond client following


megamek/src/megamek/client/ui/dialogs/unitDisplay/ArmorPanel.java
```diff
@@ -255,6 +317,21 @@ public void displayMek(Entity en) {
                 minBottomMargin = minAeroTopMargin;
                 minRightMargin = minAeroLeftMargin;
             }
+            // Units with no drawn figure of their own get the plain box-per-location diagram.
+            case HandheldWeapon ignored -> {
```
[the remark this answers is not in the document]
HammerGS: Fixed in bbd3f56bd5. Both cases had identical bodies, so they are folded into the default branch behind an instanceof check, which removes the unread bindings. The suggested type-only case label (`case HandheldWeapon ->`) does not compile on Java 21 - a type pattern must bind a variable, and unnamed patterns only arrive in Java 22.

megamek/src/megamek/client/ui/dialogs/unitDisplay/ArmorPanel.java
```diff
@@ -255,6 +317,21 @@ public void displayMek(Entity en) {
...
+                ams = simpleUnit;
+                minLeftMargin = minInfLeftMargin;
+                minTopMargin = minInfTopMargin;
+                minBottomMargin = minInfTopMargin;
+                minRightMargin = minInfLeftMargin;
+            }
+            case AbstractBuildingEntity ignored -> {
```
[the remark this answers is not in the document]
HammerGS: Fixed in bbd3f56bd5. Both cases had identical bodies, so they are folded into the default branch behind an instanceof check, which removes the unread bindings. The suggested type-only case label (`case HandheldWeapon ->`) does not compile on Java 21 - a type pattern must bind a variable, and unnamed patterns only arrive in Java 22.

megamek/src/megamek/client/ui/dialogs/unitDisplay/ArmorPanel.java
```diff
@@ -255,6 +317,21 @@
                 minBottomMargin = minAeroTopMargin;
                 minRightMargin = minAeroLeftMargin;
             }
+            // Units with no drawn figure of their own get the plain box-per-location diagram.
+            case HandheldWeapon ignored -> {
```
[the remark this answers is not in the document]
HammerGS: Fixed in bbd3f56bd5. Both cases had identical bodies, so they are folded into the default branch behind an instanceof check, which removes the unread bindings. The suggested type-only case label (`case HandheldWeapon ->`) does not compile on Java 21 - a type pattern must bind a variable, and unnamed patterns only arrive in Java 22.

megamek/src/megamek/client/ui/dialogs/unitDisplay/ArmorPanel.java
```diff
@@ -255,6 +317,21 @@
...
+                ams = simpleUnit;
+                minLeftMargin = minInfLeftMargin;
+                minTopMargin = minInfTopMargin;
+                minBottomMargin = minInfTopMargin;
+                minRightMargin = minInfLeftMargin;
+            }
+            case AbstractBuildingEntity ignored -> {
```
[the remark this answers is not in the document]
HammerGS: Fixed in bbd3f56bd5. Both cases had identical bodies, so they are folded into the default branch behind an instanceof check, which removes the unread bindings. The suggested type-only case label (`case HandheldWeapon ->`) does not compile on Java 21 - a type pattern must bind a variable, and unnamed patterns only arrive in Java 22.

megamek/src/megamek/client/ui/panels/phaseDisplay/lobby/LobbyActions.java
```diff
@@ -245,7 +245,9 @@ void configureDamage(Collection<Entity> entities) {
             LobbyErrors.showCannotEditDamage(frame());
             return;
         }
-        UnitEditorDialog med = new UnitEditorDialog(frame(), entity, localPlayer().isGameMaster());
+        // Reaching here means the local player may edit this unit: the gamemaster, or the owner when no gamemaster
+        // is present. Either way the dialog offers the full editing tools without re-checking the gamemaster role.
+        UnitEditorDialog med = new UnitEditorDialog(frame(), entity, true);
```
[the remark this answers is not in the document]
HammerGS: As designed. The gate is `canEditDamage`: when any player holds the Game Master role, only the GM may edit damage at all. Without one, players may edit only their own units (and their local bots'), and get the full tools - the lobby campaign-prep case: rolling FSW pre-existing damage on your own unit, setting ammo. The `offerGameMasterTools` name describes who the extra tools are for in play; in a GM-less lobby the unit's owner is the closest thing there is.

megamek/src/megamek/server/commands/SkillModifierCommand.java
```diff
@@ -0,0 +1,126 @@
...
+import megamek.server.commands.arguments.UnitArgument;
+import megamek.server.totalWarfare.TWGameManager;
+
+/**
+ * The gamemaster command "/skillMod" that gives a unit's crew a temporary skill change: a delta to gunnery and
+ * piloting, and a delta to the unit's individual initiative roll, lasting a set number of rounds or until cleared.
+ * <p>
+ * Skill deltas are added to the skill number, so a negative delta improves the crew (a -1 makes a 4 gunner a 3) and
+ * a positive delta worsens it; the initiative delta is added to the roll, so there positive is the improvement. A
+ * duration of {@link TemporarySkillModifiers#PERMANENT} lasts until cleared, and a duration of 0 clears whatever is
```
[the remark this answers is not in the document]
HammerGS: Fixed in bbd3f56bd5.

megamek/src/megamek/server/commands/GameMasterCommand.java
```diff
@@ -73,35 +72,18 @@ public void run(int connId, String[] args) {
...
-            server.requestGameMaster(player);
-
-            for (Player p : server.getGame().getPlayersList()) {
-                p.setVotedToAllowGameMaster(false);
-            }
-
-            // requester automatically votes yes
-            AllowGameMasterCommand.voteYes(server, player);
+            gameManager.startGameMasterVote(player);
         }
```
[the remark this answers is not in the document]
HammerGS: Valid - the comment was stale, not the code. The command deliberately puts the role to a vote in every phase; a lone player's own yes passes at once, which is why the lobby usually feels vote-free with one human present. The ChatLounge comment now describes that (bbd3f56bd5), and the PR description will be updated to match.

megamek/src/megamek/common/actions/compute/ComputeToHit.java
```diff
@@ -588,22 +588,33 @@ public static ToHitData toHitCalc(Game game, int attackerId, Targetable target,
...
-        // Start with the attacker's weapon skill
-        toHit = new ToHitData(ae.getCrew().getGunnery(), Messages.getString("WeaponAttackAction.GunSkill"));
+        // Start with the attacker's weapon skill. A gamemaster's temporary modifier is taken back out of the
+        // skill number and shown as a line of its own below, so a shifted to-hit number can be traced to the
+        // gamemaster's intervention instead of looking like a different gunner.
+        int gamemasterModifier = ae.getCrew().appliedGunneryModifier();
+        toHit = new ToHitData(ae.getCrew().getGunnery() - gamemasterModifier,
+              Messages.getString("WeaponAttackAction.GunSkill"));
         if (game.getOptions().booleanOption(OptionsConstants.RPG_RPG_GUNNERY)) {
             if (weaponType.hasFlag(WeaponType.F_ENERGY)) {
```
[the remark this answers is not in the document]
HammerGS: Fixed in bbd3f56bd5. Each gunnery variant (RPG L/M/B, artillery) now computes its own applied modifier from its own raw skill, so the clamp to the skill range is exact per variant. The raw skill each effective getter adjusts is a hook on Crew that LAMPilot overrides with its mode-dependent skills - which also fixes the applied modifier for LAMs, whose raw base is not the crew array at all.

megamek/src/megamek/client/ui/clientGUI/boardview/BoardView.java
```diff
@@ -516,13 +516,16 @@ public void gameEntityChange(GameEntityChangeEvent gameEntityChangeEvent) {
...
+                // For units that have been blown up, damaged, ejected or handed to another player (a traitor
+                // switch means the new owner's camouflage), force a reload. Without the old state we cannot tell
+                // whether the damage changed, so reload to be safe; the reload reuses the cached image when
+                // nothing shown in fact changed, so it costs nothing in the common case.
+                final Entity oldEntity = gameEntityChangeEvent.getOldEntity();
+                if ((oldEntity == null)
+                      || (entity.getDamageLevel() != oldEntity.getDamageLevel())
+                      || (entity.isDestroyed() != oldEntity.isDestroyed())
+                      || (entity.getOwnerId() != oldEntity.getOwnerId())
+                      || (entity.getCrew().isEjected() != oldEntity.getCrew().isEjected())) {
```
IllianiBird: This is a chonky conditional chain. Let's break these out into booleans and pass those in

megamek/src/megamek/client/ui/dialogs/unitEditor/UnitDamagePanelBuilder.java
```diff
@@ -0,0 +1,1090 @@
...
+     */
+    private JSpinner skillDeltaSpinner(int appliedDelta, int rawSkill, String tooltipKey) {
+        JSpinner spinner = new JSpinner(new SpinnerNumberModel(appliedDelta, -rawSkill, Crew.MAX_SKILL - rawSkill, 1));
+        spinner.setToolTipText(UIUtil.formatSideTooltip(Messages.getString(tooltipKey)));
+        return spinner;
+    }
+
+    /** A spinner for the initiative delta, which is added to a roll rather than a skill and has no skill range. */
+    private JSpinner initiativeDeltaSpinner(int delta) {
+        JSpinner spinner = new JSpinner(new SpinnerNumberModel(delta, -MAX_SKILL_DELTA, MAX_SKILL_DELTA, 1));
```
IllianiBird: This can cause issues if delta is > or < the min/max values. As delta is passed in unchecked (insofar as I can tell) this could result in an error.

megamek/src/megamek/client/ui/dialogs/unitEditor/UnitDamagePanelBuilder.java
```diff
@@ -0,0 +1,1090 @@
...
+        // a LAM in fighter mode, or an airborne unit, is neither prone nor hull down
+        boolean canLieDown = !entity.isAero() && !entity.isAirborne() && !entity.isAirborneVTOLorWIGE();
+        if (canLieDown && entity.isMek()) {
+            controls.chkProne = addStatusRow("UnitEditorDialog.status.prone", entity.isProne());
+        }
+        if (canLieDown && (entity.isMek() || entity.isVehicle())) {
+            controls.chkHullDown = addStatusRow("UnitEditorDialog.status.hullDown", entity.isHullDown());
+        }
+
+        boolean canHide = !(entity instanceof Dropship) && !entity.isAirborne() && !entity.isAirborneVTOLorWIGE();
```
IllianiBird: I feel like we should add `canHide` as a method to `Entity` and then override it for the specific instances where a unit can't hide. That'll avoid class casting as it becomes `entity.canHide()` and also allows us to reuse the check anywhere we're verifying whether a unit can be deployed hidden.

megamek/src/megamek/client/ui/dialogs/unitEditor/UnitDamageSpecBuilder.java
```diff
@@ -0,0 +1,201 @@
...
+    private Integer spinnerValue(JSpinner spinner) {
+        return (null == spinner) ? null : (Integer) spinner.getValue();
+    }
+
+    /** A spinner's value as an integer whatever its model's number type, for the models that hold doubles. */
+    private Integer numberValue(JSpinner spinner) {
+        return (null == spinner) ? null : ((Number) spinner.getValue()).intValue();
+    }
+
+    private Boolean checkboxValue(JCheckBox checkbox) {
```
IllianiBird: This and several other methods need the @Nullable Jakarta tag

megamek/src/megamek/client/ui/dialogs/UnitEditorDialog.java
```diff
@@ -223,1742 +206,299 @@ private void initComponents() {
...
-            }
-        }
-        if ((entity instanceof Aero aero) && (null != structuralIntegrityLabel) && (null != spnInternal[0])) {
-            Color siColor = colorSpinner(spnInternal[0], aero.getOSI(), null);
-            if (null != siColor) {
-                structuralIntegrityLabel.setForeground(siColor);
-            }
-        }
+    /** A crew can only leave a unit that has one, that can eject, and that they have not already left. */
+    private boolean canEjectCrew() {
```
IllianiBird: Crews can also abandon their vehicles. Should that be accessible through the GM dialog?

I also think this logic could be moved to Entity and then overridden by those entity types that can eject crew.

Something like canEjectCrew false, on Entity. Then Mek and Aero would override it with the 'isEjected', 'isDead', etc logic. This makes it much more reusable.

megamek/src/megamek/client/ui/dialogs/unitEditor/UnitDamageControls.java
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    /** The locations that currently have at least one critical hit set on them. */
+    public Set<Integer> locationsWithCrits() {
+        return critsByLocation.entrySet()
+              .stream()
+              .filter(entry -> entry.getValue().stream().anyMatch(crit -> crit.getHits() > 0))
+              .map(Map.Entry::getKey)
+              .collect(Collectors.toSet());
+    }
```
IllianiBird: You can get a tiny performance boost by converting this to a traditional loop. Non-essential.

megamek/src/megamek/client/ui/widget/mapset/SimpleUnitMapSet.java
```diff
@@ -0,0 +1,231 @@
...
+        }
+    }
+
+    private void setBackGround() {
+        UnitDisplaySkinSpecification udSpec = SkinXMLHandler.getUnitDisplaySkin();
+
+        Image tile = jComponent.getToolkit()
+              .getImage(new MegaMekFile(Configuration.widgetsDir(), udSpec.getBackgroundTile()).toString());
+        PMUtil.setImage(tile, jComponent);
+        int b = BackGroundDrawer.TILING_BOTH;
```
IllianiBird: Let's nix the single letter variable name. Took me a few moments to figure out what `b` was in reference to

megamek/src/megamek/client/ui/widget/picmap/PicMap.java
```diff
@@ -244,26 +250,87 @@ public void paintComponent(Graphics g) {
...
     private void drawInto(Graphics g) {
         int w = Math.max(getSize().width, minWidth);
         int h = Math.max(getSize().height, minHeight);
+
+        // Everything is laid out in the content's own coordinates, including the background images that frame the
+        // diagram, so it is all drawn through one scaled graphics context and stays aligned.
+        Graphics2D contentGraphics = (Graphics2D) g.create();
+        contentGraphics.scale(displayScale, displayScale);
+        int contentWidth = (int) (w / displayScale);
+        int contentHeight = (int) (h / displayScale);
```
IllianiBird: This can truncate down to 0 which will then throw an error

megamek/src/megamek/client/ui/widget/picmap/PicMap.java
IllianiBird, line 1: This class has a lot of single letter or abbreviated variable names which is making it hard to parse.

megamek/src/megamek/common/units/Crew.java
IllianiBird: Not your code but this should be Math.clamp now

megamek/src/megamek/server/commands/CancelGameMasterCommand.java
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * This command withdraws the player's own request to assume the elevated Game Master role. There is no time limit on
+ * the vote, so cancelling is how a requester takes back a request that is not going to be answered.
+ */
+public class CancelGameMasterCommand extends ServerCommand {
+
+    private final TWGameManager gameManager;
+
+    public CancelGameMasterCommand(Server server, TWGameManager gameManager) {
+        super(server, "cancelGM", "Withdraws your own request to become Game Master. "
```
IllianiBird: Player facing hardcoded strings

IllianiBird (changes requested):
Some minor points of feedback

megamek/src/megamek/client/ui/clientGUI/boardview/BoardView.java
HammerGS: Fixed in 5f189037fe - the chain is broken out into named booleans (damageChanged, destructionChanged, ownerChanged, ejectionChanged), with the null-old-state case reading as its own default.

megamek/src/megamek/client/ui/dialogs/unitEditor/UnitDamagePanelBuilder.java
HammerGS: Fixed in 5f189037fe. The applied modifier fits the bounds by construction (it is effective minus stored, and the effective skill is clamped), but the model does throw on an out-of-bounds start, so the starting value is now held to the bounds with Math.clamp rather than trusting every caller forever. Same guard on the initiative and rounds spinners.

megamek/src/megamek/client/ui/dialogs/unitEditor/UnitDamagePanelBuilder.java
HammerGS: Done in 5f189037fe: Entity.canHide() covers the airborne checks and Dropship overrides it to false; the editor now reads entity.canHide(). Ready for reuse wherever hidden deployment is decided.

megamek/src/megamek/client/ui/dialogs/unitEditor/UnitDamageSpecBuilder.java
HammerGS: Fixed in 5f189037fe - @Nullable on the returns and parameters of all six value-reading helpers.

megamek/src/megamek/client/ui/dialogs/UnitEditorDialog.java
HammerGS: Done in 5f189037fe, and yes to vehicles: canEjectCrew lives on Entity now with a shared crewCanLeave() helper. Mek overrides it (always allowed while alive and aboard), Aero allows it only on the ground - which fixed a real bug, the old checkbox offered ejection to airborne fighters but abandonEntity has no airborne path, so the crew just died - and Tank allows it behind the vehicles-can-eject game option. So a vehicle crew can now abandon from the destroy dialog, through the same /kill eject path.

megamek/src/megamek/client/ui/dialogs/unitEditor/UnitDamageControls.java
HammerGS: Fixed in 5f189037fe - plain nested loops.

megamek/src/megamek/client/ui/widget/mapset/SimpleUnitMapSet.java
HammerGS: Fixed in 5f189037fe - it is tilingType now, in all nine uses.

megamek/src/megamek/client/ui/widget/picmap/PicMap.java
HammerGS: Fixed in 5f189037fe - the content size is held to at least one pixel, so a component smaller than one scaled pixel cannot produce a zero-size clip or image.

megamek/src/megamek/client/ui/widget/picmap/PicMap.java
HammerGS: Renamed throughout in 5f189037fe: graphics, width, height, element, hotArea, event, drawer, offScreenImage, contentBounds, shape/rectangle/polygon and so on. Loop indices stayed.

megamek/src/megamek/common/units/Crew.java
HammerGS: Fixed in 5f189037fe - Math.clamp, and the deprecated MathUtility import is gone from Crew.

megamek/src/megamek/server/commands/CancelGameMasterCommand.java
HammerGS: Fixed in 5f189037fe - the whole Game Master vote family moved to messages.properties: the four command help texts, the vote announcements, ballots and outcomes.

IllianiBird (approved):

PR #8570, opened 2026-07-23
Implement equipment activation/deactivation (Off modes) for probes, ECM, C3, gauss rifles, improved heavy lasers and heat sinks

HammerGS:
## Summary

The rules let players deactivate certain equipment - and MegaMek mostly didn't. This PR gives every piece of equipment on that list a working Off switch, with the game effects actually stopping while it is off. A switch can be declared at any time but only takes effect in the End Phase, just like Nova CEWS network changes.

What a player can now do:

- **Active Probes** (Beagle, Bloodhound, Clan AP, Light AP, BA probes): switch Off - no hidden-unit or minefield detection, no sensor benefit, and ARAD missiles are no longer attracted. Worked example: a hidden Mek 3 hexes from a stationary Beagle stays hidden while the probe is off; switch the probe on and it is revealed at the end of the next Movement Phase
  without the probe carrier moving.
- **ECM suites** (Guardian, Angel, Watchdog CEWS, BA ECM): switch Off - the ECM bubble is genuinely gone, confirmed by C3 to-hit numbers changing with the ECM state. The mode switcher is now always available for ECM (it used to be hidden unless the ECCM/Ghost Target options were on).
- **C3, C3i, Naval C3, C3 Masters, Nova CEWS**: switch Off - the unit gives and receives no targeting data, but its network membership is deliberately preserved: **switch it back on and the old network simply works again**, no lobby reconfiguration. A Nova CEWS set to Off now silences all three of its functions (ECM, probe, C3) - previously only the jammer stopped.
  The pre-end phase Nova network button grays out while the Nova is off.
- **Gauss rifles**: Powered Up/Down is now always available instead of requiring the TacOps game option. Rules-visible consequence: a powered-down gauss rifle cannot fire and does not explode when critically hit, now reachable in standard games (this matches the core-rules activation/deactivation list; flagging it for reviewer sign-off
- **Improved Heavy Lasers**: On/Off - an activated one explodes on its first critical hit as before; a deactivated one cannot fire and cannot explode.
- **Heat sinks - per individual sink, all types** (single, IS/Clan double, compact, laser, prototype DHS, Freezers): switch specific sinks off (TSM players rejoice). The existing "active sinks" slider still works and now drives the individual mounts; prototype double heat sinks and Freezers become switchable for the first time (they were hard-wired always-on); sinks can be toggled while shut down per TW. Aerospace units, whose sinks are a bare count with no mounts, get a count-based control.
- **Stealth armor guard**: stealth armor requires a running ECM, "ECM Off" while stealth is engaged (or engaging), nor "stealth On" while the ECM is off - in both orders, including queueing both in the same round - with a server-side rejection as the
  backstop.

## Changes

1. Off modes added to the equipment types; behavior gates go through a single shared seam (`Mounted.isModeTurnedOff()`); switches ride the existing pending-mode/`ENTITY_MODE_CHANGE`  pipeline (no new packets) and apply at the round rollover.
2. C3 is a benefit gate, not a membership change: `EquipmentActivation.isC3SwitchedOff` cuts the benefit in `onSameC3NetworkAs` (non-ignoreECM path) and `ComputeC3Spotter`, while net IDs,
   master links and partner UUIDs are untouched - so save files, `wireC3` and `getC3Master`'s
   self-heal keep working and reactivation restores the network.
3. Heat capacity computed from per-mount modes (`Mek.getHeatCapacity`); the classic active-sinks
   counter becomes a derived value; new `ActiveHeatSinkController` interface for Mek + Aero; the
   obsolete Mek counter fields stay declared only for save-game compatibility.
4. New `ImprovedHeavyLaserWeapon` base class (also prevents a spurious "Pulse Off" combination
   from the RISC pulse-module mode injection); deactivated IHLs a
   Blue Shield `isExplosive` pattern; a generic firing gate in `ComputeToHitIsImpossible` blocks
   any weapon switched Off (Princess honors it automatically sinc
5. New stateless helpers instead of mega-class growth: `EquipmentActivation` (eligibility
   queries), mode-change validators in `ServerHelper`; `TWGameManager` keeps parsing+delegation.
6. Systems tab now offers only the modes actually available on the mount (fixes ordinary lasers
   offering RISC "Pulse" modes without a module); lobby unit configuration exposes probe, Nova
   and C3 modes so equipment can begin the game deactivated.
7. Diagnostic feature loggers `[EquipOff]` and `[HiddenUnits]` (c
   same pattern as the AVLB logger).

## Files Changed

- `MiscType`, `GaussWeapon`, `ISC3M/ISC3MBS`, `ImprovedHeavyLaserWeapon` (+3 subclasses) - mode definitions
- `Entity`, `BattleArmor`, `ComputeC3Spotter`, `ComputeToHitIsImpossible`, `EquipmentType` - behavior gates
- `Mounted` (+`EquipmentActivation`) - shared mode-state seam and
- `Mek`, `Aero`, `ActiveHeatSinkController` - per-sink / count-based heat sink activation
- `TWGameManager`, `ServerHelper` - mode-change validation, sink-change handling, diagnostics
- `SystemPanel`, `EquipChoicePanel`, `NovaNetworkDialog`, `PreEnd
- `EquipmentOffModesTest` (new, 9 tests), `messages.properties`, `mmconf/log4j2.xml`

## Testing

- New `EquipmentOffModesTest` covers all gates including the End-Phase timing; full test suite
  green (re-run after merging current main).
- Playtested in-game, all confirmed: ECM off/on changing real to-hit numbers via C3; stationary
  probe revealing a hidden Mek at end of Movement Phase (and not while off); C3i deactivate ->
  benefit gone -> reactivate -> same network restored; gauss powe
  option; improved heavy laser off; per-sink heat dissipation; stealth-ECM guard both ways
  including the same-round pending loophole; Nova button graying; a save game from several days
  before this branch loads correctly.


megamek/src/megamek/common/units/Mek.java
```diff
@@ -4682,27 +4679,80 @@ public boolean hasLaserHeatSinks() {
...
+     * equipment mode.
+     *
+     * @param sinks the number of heat sinks that should be active next round
+     */
     public void setActiveSinksNextRound(int sinks) {
-        sinksOnNextRound = sinks;
+        int remainingActive = sinks;
+        for (MiscMounted mounted : getMisc()) {
+            if (!isCountedHeatSink(mounted) || mounted.isDestroyed() || mounted.isBreached()) {
+                continue;
```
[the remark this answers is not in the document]
HammerGS: Valid point on the asymmetry with `Aero#setActiveSinksNextRound` - fixed in 9517540d20. `Mek.setActiveSinksNextRound` now clamps explicitly (`Math.max(0, sinks)`), with the contract documented in the JavaDoc and a new test covering both out-of-range directions. Worth noting the prior behavior already degraded safely - a negative count switched every sink off (identical to a request of 0) and an oversized count simply ran out of sinks to switch on - so this makes the intent explicit rather than changing an outcome, but explicit is better for a packet-supplied value.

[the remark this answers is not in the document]
HammerGS: Addressed in 9517540d20.

`setActiveSinksNextRound` now clamps the packet value with `Math.max(0, sinks)`, so a negative count no longer switches every sink off. The upper bound is handled by the loop itself, which can only activate as many sinks as it iterates over. This matches `Aero#setActiveSinksNextRound`.

megamek/src/megamek/common/equipment/MiscType.java
```diff
@@ -5590,7 +5612,10 @@ public static MiscType createGECM() {
               .andNot(F_JS_EQUIPMENT)
               .andNot(F_SS_EQUIPMENT);
         misc.bv = 61;
-        misc.setModes("ECM");
+        misc.setModes("ECM", "Off");
+        // Deactivation is declared at any time but takes effect in the End Phase; other mode
+        // switches (ECM/ECCM/Ghost Targets) remain instant
+        misc.addEndTurnMode("Off");
         misc.setInstantModeSwitch(false);
```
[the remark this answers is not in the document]
HammerGS: Good catch, and confirmed. `isNextTurnModeSwitch` is only ever read from `Mounted#canInstantSwitch` (and the `InfantryWeaponMounted` override), both of which short-circuit on `hasInstantModeSwitch()`. With that false, the end-turn mode list is never consulted, so `addEndTurnMode("Off")` was dead on arrival.

Two corrections to the scope:

1. `setInstantModeSwitch(false)` is not new here. It is already on `main` at every one of these sites -- this PR only added `setModes`, the comment, and the `addEndTurnMode` call. So the comment was not just wrong about the new code, it described behavior that was never true. ECM mode switches have always been deferred.

2. It was 10 sites, not 1: `createGECM`, `createGECMPrototype`, `createCLECM`, `createISAngelECM`, `createISTHBAngelECM`, `createWatchdogECM`, `createBAISAngelECM`, `createBACLAngelECM`, `createISSingleHexECM`, `createCLSingleHexECM`.

Went with your option (b) in bbc8f022e2 -- dropped all 10 calls and corrected the comment. Zero behavior change: deactivation already lands in the End Phase via the pre-existing `setInstantModeSwitch(false)`, so the PR goal is met without them.

Option (a) would have made ECM/ECCM/Ghost Target switching instant for the first time. That changes long-standing behavior and is a rules question (TW ECM timing) unrelated to Off modes, so it does not belong in this PR.

megamek/src/megamek/client/ui/dialogs/unitDisplay/SystemPanel.java
```diff
@@ -893,26 +893,52 @@ public void valueChanged(ListSelectionEvent event) {
...
+                    // Every ECM suite now carries at least the "ECM"/"Off" pair (activation/deactivation
+                    // rules), so the mode switcher is always offered - the ECCM and Ghost Target game
+                    // options merely add further modes.
+                    // Iterate the PER-MOUNT mode count rather than the raw type mode list: it filters
+                    // modes that need linked equipment to be available, such as the "Pulse ..." laser
+                    // modes that require a RISC Laser Pulse Module (LaserWeapon keeps them last and
+                    // reduces the count when no module is linked).
+                    int visibleModeCount = mounted.getModesCount();
+                    for (int modeIndex = 0; modeIndex < visibleModeCount; modeIndex++) {
+                        EquipmentMode equipmentMode = mounted.getType().getMode(modeIndex);
```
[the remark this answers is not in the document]
HammerGS: Confirmed, and fixed in 38404dfb20. This one was a real regression from this branch.

Traced it through:

- `Mounted#getModesCount()` delegates to `getType().getModesCount(this)`, but `InfantryWeaponMounted` overrides it to return `getModes().size()` -- the merged primary+secondary list built by `rebuildModeList()`.
- `EquipmentType#getMode(int)` is `modes.elementAt(n)` on a `Vector`, so an over-range index throws rather than returning null.
- `Mounted#getMode(int)` is `protected`, which is why the loop was reaching through `getType()` in the first place.

Concrete repro: infantry weapons only gain modes when `isFlameBased()`, via `adaptToGameOptions` adding `MODE_FLAMER_DAMAGE`/`MODE_FLAMER_HEAT`. `BLKInfantryFile:144` constructs `InfantryWeaponMounted(infantry, primary, secondary)` when there is at most one secondary per squad, so `getType()` is the primary. A platoon with a non-flame primary (Auto Rifle) and a flamer secondary therefore has 0 modes on the type and 2 on the merged list. `hasModes()` reads the merged list, so the block is entered, and the very first `getType().getMode(0)` hits an empty Vector. Selecting that platoon under "All Weapons" in the unit display throws.

Fix taken:

```java
int visibleModeCount = mounted.getType().getModesCount(mounted);
```

`LaserWeapon` is the only override of `getModesCount(Mounted<?>)`, so the RISC Laser Pulse Module filtering this loop depends on is preserved. For every mount except `InfantryWeaponMounted` this is precisely what `Mounted#getModesCount()` already delegated to, so it is a no-op everywhere else.

I did not take the second suggestion (exposing a public per-mount `getMode(int)`). It widens the API and would newly surface merged secondary modes for infantry -- a behavior change beyond fixing the crash. This keeps the pre-branch behavior for that platoon: the mode list shows nothing instead of throwing. Whether infantry ought to display merged flamer modes is a separate, pre-existing question.

megamek/src/megamek/common/units/Entity.java
```diff
@@ -13830,6 +13845,10 @@ public void setGameOptions() {
                     }
                 }
 
+                // ECM suites can be deactivated (activation/deactivation rules); the switch is
+                // registered as an end-turn mode on the type, so it takes effect in the End Phase
```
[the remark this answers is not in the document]
HammerGS: Correct, and fixed in dc18514079. This one was fallout from bbc8f022e2, which removed the ten dead `addEndTurnMode("Off")` calls after your earlier comment -- the explanatory comment here was left pointing at a mechanism that no longer exists (and never actually did anything).

Verified the rest of your reasoning holds: `setModes()` replaces the mode list only, it does not touch `instantModeSwitch`. These ECM types are built with `setInstantModeSwitch(false)` in their `MiscType` factory methods, so the End Phase timing survives the rebuild here. Nothing in the ECM path registers an end-turn mode any more.

New comment:

```java
// ECM suites can be deactivated (activation/deactivation rules). These types are built
// with setInstantModeSwitch(false), so every mode switch here -- deactivation included --
// is declared now and takes effect in the End Phase. Nothing registers an end-turn mode.
modes.add("Off");
```

I also swept the rest of the tree for the same stale wording (`grep -ri "end-turn mode|end turn mode|endTurnMode"`). This was the only remaining one. The surviving `addEndTurnMode("AAA")` in `Weapon.java:174` is unrelated and untouched by this branch.

Comment only, no behavior change.

megamek/src/megamek/client/ui/dialogs/unitDisplay/SystemPanel.java
```diff
@@ -893,26 +893,56 @@ public void valueChanged(ListSelectionEvent event) {
...
-                          && !(client.getGame().getOptions().booleanOption(OptionsConstants.ADVANCED_TAC_OPS_ECCM)
-                          || client.getGame().getOptions()
-                          .booleanOption(OptionsConstants.ADVANCED_TAC_OPS_GHOST_TARGET))) {
-                        return;
+                    // Heat sinks can be activated or deactivated even while the unit is shut down (TW,
+                    // Deactivating Heat Sinks)
+                    boolean isHeatSinkMount = (mounted.getType() instanceof MiscType)
+                          && (mounted.getType().hasFlag(MiscType.F_HEAT_SINK)
+                          || mounted.getType().hasFlag(MiscType.F_DOUBLE_HEAT_SINK)
+                          || mounted.getType().hasFlag(MiscType.F_IS_DOUBLE_HEAT_SINK_PROTOTYPE));
```
IllianiBird: This chained conditional is hard to follow, let's break it up into component booleans and pass those in.

megamek/src/megamek/client/ui/dialogs/unitDisplay/SystemPanel.java
```diff
@@ -893,26 +893,56 @@ public void valueChanged(ListSelectionEvent event) {
...
                               && !client.getGame().getOptions()
                               .booleanOption(OptionsConstants.ADVANCED_COMBAT_HOT_LOAD_IN_GAME)) {
                             continue;
                         }
+                        // Hide the Off mode for ECM suites while the stealth armor system is engaged or
+                        // queued to engage this round - the server rejects that switch (stealth armor
+                        // requires an operating ECM). Off is always the LAST mode in every ECM mode list, so
+                        // skipping it keeps the combo indices aligned with the equipment type's mode indices
+                        // (the selection is applied by index).
+                        if (equipmentMode.equals("Off")
```
IllianiBird: If we're going to be attributing special properties to these Strings they should be defined as consts and then reused. If possible we should avoid string comparison as it's _really_ fragile. Finally, can I verify what happens if the player is using a non-English language gui?

megamek/src/megamek/client/ui/panels/phaseDisplay/PreEndDeclarationsDisplay.java
```diff
@@ -653,13 +653,24 @@ private int localPlayerId() {
...
     /**
-     * Checks if the local player has any Nova CEWS units.
+     * Checks if the local player has any Nova CEWS units whose CEWS is currently functioning. A Nova the player has
+     * switched to "Off" cannot be reconfigured (the button grays out), though it keeps its network membership and
+     * rejoins when switched back on.
      */
     private boolean hasNovaUnits() {
         int localPlayerId = localPlayerId();
-        return game.getEntitiesVector().stream()
+        boolean hasActiveNova = game.getEntitiesVector().stream()
```
IllianiBird: We don't need to use a stream here. Streams are less performative than traditional loops and generally should be reserved for instances where we're handling thousands of objects within the stream.

megamek/src/megamek/client/ui/panels/phaseDisplay/PreEndDeclarationsDisplay.java
```diff
@@ -653,13 +653,24 @@ private int localPlayerId() {
...
               .filter(entity -> entity.getOwnerId() == localPlayerId)
-              .anyMatch(Entity::hasNovaCEWS);
+              .anyMatch(Entity::hasActiveNovaCEWS);
+        if (!hasActiveNova) {
+            boolean hasSwitchedOffNova = game.getEntitiesVector().stream()
```
IllianiBird: Same here

megamek/src/megamek/client/ui/panels/phaseDisplay/PreEndDeclarationsDisplay.java
```diff
@@ -653,13 +653,24 @@ private int localPlayerId() {
...
+                  .filter(entity -> entity.getOwnerId() == localPlayerId)
```
IllianiBird: In fact, as an aside right now you're streaming through the same array of objects twice and performing very similar checks. It'd be far more performative to loop through the objects (using a traditional loop) and performing all relevant checks on each object, rather than streaming twice. Means each object is only being parsed once, rather than twice.

megamek/src/megamek/common/battleArmor/BattleArmor.java
```diff
@@ -1327,11 +1327,10 @@ public boolean hasImprovedSensors() {
      */
     public boolean hasActiveProbe() {
         for (Mounted<?> equip : getMisc()) {
-            if (equip.getType().hasFlag(MiscType.F_BAP)
-                  && !(equip.getType().getInternalName()
-                  .equals(Sensor.IS_IMPROVED)
-                  || equip.getType()
-                  .getInternalName().equals(Sensor.CL_IMPROVED))) {
+            boolean isImprovedSensors = equip.getType().getInternalName().equals(Sensor.IS_IMPROVED)
+                  || equip.getType().getInternalName().equals(Sensor.CL_IMPROVED);
```
IllianiBird: `Mounted.getType()` can return `null` (it isn't marked `Nullable` so I double checked before leaving this comment) so we can't call `equip.getType()` without null protections else we risk an NPE every time we parse this. We're also getting internal name twice, so should really be only calling that once and then reusing the return.

megamek/src/megamek/common/units/Mek.java
```diff
@@ -274,8 +274,15 @@ private void clear() {
     // for Harjel II/III
     private final boolean[] armorDamagedThisTurn;
 
+    /**
+     * No longer used: heat sink activation is tracked per mount via equipment modes. The two fields are retained only
+     * so that pre-existing save games (which contain them) still deserialize; XStream rejects unknown elements.
+     */
+    @SuppressWarnings("unused")
     private int sinksOn = -1;
```
IllianiBird: Is this used outside of MegaMek games? In MekHQ, for example, or is it just held at the Entity level (which isn't stored in MekHQ)? Unless we're explicitly passing this back to MekHQ and then storing it in Unit, we don't need to keep these values.

We don't have any cross-version compatibility guarantees for scenarios.

megamek/src/megamek/common/units/Mek.java
```diff
@@ -2363,24 +2367,17 @@ public int getHeatCapacity(boolean radicalHeatSink) {
...
                 capacity++;
-                activeCount--;
-            } else if ((activeCount > 0)
-                  && mounted.getType().hasFlag(MiscType.F_DOUBLE_HEAT_SINK)) {
-                activeCount--;
-                capacity += 2;
-                isDoubleHeatSink = true;
-            } else if (mounted.getType().hasFlag(
-                  MiscType.F_IS_DOUBLE_HEAT_SINK_PROTOTYPE)) {
+            } else if (mounted.getType().hasFlag(MiscType.F_DOUBLE_HEAT_SINK)
```
IllianiBird: We're calling `mounted.getType` a lot here. In addition to this being an NPE risk we really should just call this once and then pass around the return value, instead of repeatedly fetching the value.

IllianiBird (changes requested):
There's some relatively minor issues here, but they do need to be actioned before the PR can be merged.

The biggest issue is the volume of magic Strings that are being used. `On` and `Off` are used throughout this PR, as are other states. It would be _significantly_ cleaner to define them as a const somewhere (maybe `Mounted`) and then call that. This will give us a single point of truth and should we ever need to change the Strings we only then need to change in a single place, rather than 20+.

The other one is the `getType` NPE risk. `getType` needs to be marked with `@Nullable` (Jakarta version, please) and then we need to put null protections throughout this PR. Sorry for the pain, but with this codebase we need to be _really_ paranoid when it comes to NPEs.

megamek/src/megamek/client/ui/dialogs/unitDisplay/SystemPanel.java
HammerGS: Done in 40f06026ea.

Split into one boolean per concern, combined by name:

```java
boolean isSingleHeatSink = isMiscEquipment && mountedType.hasFlag(MiscType.F_HEAT_SINK);
boolean isDoubleHeatSink = isMiscEquipment && mountedType.hasFlag(MiscType.F_DOUBLE_HEAT_SINK);
boolean isPrototypeDoubleHeatSink = isMiscEquipment
      && mountedType.hasFlag(MiscType.F_IS_DOUBLE_HEAT_SINK_PROTOTYPE);
boolean isHeatSinkMount = isSingleHeatSink || isDoubleHeatSink || isPrototypeDoubleHeatSink;
```

While in here I also hoisted the two things this block repeated: `isMiscEquipment` (the `instanceof MiscType` test appeared in five separate conditions) and a single `mountedType` local for the equipment type. That second one doubles as the fix for your `getType()` NPE point in this file - the whole block is now entered only when the type resolves.

megamek/src/megamek/client/ui/dialogs/unitDisplay/SystemPanel.java
HammerGS: Constants added in 40f06026ea, and every mode-string site this PR touches now goes through them:

- `Mounted.MODE_ON` / `MODE_OFF` / `MODE_NONE` - the generic activation states, so they sit next to the code that reads them (`isModeTurnedOff()`, `modeNextRound()`, `curMode()`).
- `MiscType.MODE_ECM` - the ECM suite's operating mode.
- `Weapon.MODE_GAUSS_POWERED_UP` - added next to the `MODE_GAUSS_POWERED_DOWN` that was already there.

I put them in those three classes rather than all in `Mounted` to match the existing precedent: `Weapon.java` already carries a block of ~30 `MODE_*` constants (`MODE_AMS_ON`, `MODE_AC_RAPID`, `MODE_GAUSS_POWERED_DOWN`, ...). Happy to consolidate if you would rather they all live in one place.

On avoiding string comparison entirely - I agree it is fragile, but it cannot be fixed inside this PR. The mode system is string-keyed all the way down: `EquipmentMode` is constructed from a name and interned in a `Hashtable<String, EquipmentMode>`, `EquipmentType.setModes(String...)` / `addMode(String)` build the list, `Mounted.setMode(String)` looks up by name, and the localization key is literally `"EquipmentMode." + name`. Converting that to an enum touches every piece of equipment in the codebase and would swamp this PR. The constants at least give the single point of truth you are asking for.

On the non-English GUI - it is safe, and the split is deliberate in the existing design:

- `EquipmentMode.equals(String)` compares `name`, the **internal** identifier. It never touches a resource bundle.
- `EquipmentMode.getDisplayableName()` is the only localizing method - it looks up `EquipmentMode.<name>` in `EquipmentMessages`.

Everything that makes a decision (this file's mode filtering, `isModeTurnedOff()`, the server's stealth/ECM validation) compares internal names. The combo box shows `getDisplayableName()`, but the selection is applied by **index** into the equipment type's mode list, not by parsing the displayed text. So a German or Russian client filters and switches modes identically; only the label changes.

One consequence worth stating since it is the load-bearing assumption here: because the filtering skips a mode by index, `"Off"` has to remain the **last** mode in every ECM mode list (and `"On"` last for stealth), or the combo indices would drift out of alignment. That is noted in the comments on both `continue` blocks.

megamek/src/megamek/client/ui/panels/phaseDisplay/PreEndDeclarationsDisplay.java
HammerGS: Done in 40f06026ea - both streams in this method are gone, replaced by a single traditional loop.

megamek/src/megamek/client/ui/panels/phaseDisplay/PreEndDeclarationsDisplay.java
HammerGS: Same fix, same commit (40f06026ea) - see the reply on the first one. Both streams are replaced by one loop.

megamek/src/megamek/client/ui/panels/phaseDisplay/PreEndDeclarationsDisplay.java
HammerGS: Agreed, and fixed in 40f06026ea. The method now walks the entity list once and does both checks per entity:

```java
private boolean hasNovaUnits() {
    int localPlayerId = localPlayerId();
    boolean hasSwitchedOffNova = false;
    for (Entity entity : game.getEntitiesVector()) {
        if (entity.getOwnerId() != localPlayerId) {
            continue;
        }
        if (entity.hasActiveNovaCEWS()) {
            return true;
        }
        if (entity.hasNovaCEWS()) {
            hasSwitchedOffNova = true;
        }
    }
    if (hasSwitchedOffNova) {
        LOGGER.debug("[PreEnd] Nova network button disabled - all of the player's Nova CEWS are switched off");
    }
    return false;
}
```

In practice it is better than one pass: it returns on the first active Nova, so the common case does not even finish a single traversal. The old version always completed one full stream and, in the disabled case, a second one.

The `hasSwitchedOffNova` flag only exists to make the debug line meaningful - it distinguishes "you have Novas but they are all switched off" (the diagnostic worth logging) from "you have no Novas at all" (nothing to say).

megamek/src/megamek/common/battleArmor/BattleArmor.java
HammerGS: Both points fixed in 40f06026ea, and you were right to double check `getType()` - it can genuinely return null. It lazily resolves through `EquipmentType.get(typeName)`, which returns null when a mount is restored from a save or over the network naming equipment this build does not know. I have annotated it (`jakarta.annotation.Nullable`, per your review note) with the contract documented, and gone through every call site in this PR's scope.

Here:

```java
for (MiscMounted equip : getMisc()) {
    MiscType miscType = equip.getType();
    if ((miscType == null) || !miscType.hasFlag(MiscType.F_BAP)) {
        continue;
    }
    String internalName = miscType.getInternalName();
    boolean isImprovedSensors = internalName.equals(Sensor.IS_IMPROVED)
          || internalName.equals(Sensor.CL_IMPROVED);
    if (!isImprovedSensors && !equip.isModeTurnedOff()) {
        return true;
    }
}
```

Three changes: the type is fetched once and null-guarded, `getInternalName()` is called once instead of twice, and the `F_BAP` test moved up into the guard clause so the improved-sensors strings are only computed for equipment that is actually a probe. The loop variable also went from `Mounted<?>` to `MiscMounted`, which is what `getMisc()` returns - that removes the need to re-check the type at all.

megamek/src/megamek/common/units/Mek.java
HammerGS: Good question, and the answer is no - nothing outside MegaMek used them, so they are now deleted (40f06026ea).

What I checked before removing them:

- **MekHQ**: grepped the whole repo for `sinksOn`, `getActiveSinks` and `setActiveSinksNextRound` - no hits at all. MekHQ persists units as MUL-style entity XML, which has no concept of these fields.
- **Scenarios / MUL export**: after the removal the only references to the names left anywhere in MegaMek are the two `omitField` calls below and the regression test. Nothing writes them.

So the single thing that carried them was the **MegaMek save game**, because this XStream setup rejects elements it does not recognise - a save written before this PR contains `<sinksOn>` and would fail to load against a build without the field. That is the only reason I had kept them, and it is better handled explicitly:

```java
// in SerializationHelper.getLoadSaveGameXStream()
xStream.omitField(Mek.class, "sinksOn");
xStream.omitField(Mek.class, "sinksOnNextRound");
```

Fields gone from `Mek`, old saves still load. I added a regression test that serializes a Mek, splices the two legacy elements back into the XML, and asserts it still deserializes - so if someone later removes those `omitField` calls, the test says why they exist.

megamek/src/megamek/common/units/Mek.java
HammerGS: Fixed in 40f06026ea - the type is fetched once per iteration into a local and null-guarded:

```java
for (MiscMounted mounted : getMisc()) {
    MiscType miscType = mounted.getType();
    if ((miscType == null) || mounted.isDestroyed() || mounted.isBreached() || mounted.isModeTurnedOff()) {
        continue;
    }
    if (miscType.hasFlag(MiscType.F_HEAT_SINK)) {
    ...
```

That drops this method from six `getType()` calls per mount to one. I gave the same treatment to the other heat sink methods in this class that had the pattern: `sinksUnderwater()`, `resetSinks()` and the new `isCountedHeatSink()` helper. The loop variables also moved from `Mounted<?>` to `MiscMounted` (what `getMisc()` actually returns), so no re-checking of the type is needed.

On the NPE risk generally - you were right, `getType()` resolves lazily via `EquipmentType.get(typeName)` and returns null for equipment names this build does not know (restored saves, network mounts). It is now marked `@Nullable` (Jakarta) with the contract in its JavaDoc, and I worked through every call site in this PR's scope: `Entity` (probe/ECM/Nova loops), `Mek`, `BattleArmor`, `EquipmentActivation`, `ServerHelper`, `SystemPanel`, `EquipChoicePanel` and the `TWGameManager` mode-change handler.

One extra thing that fell out of this pass, since you mentioned equipment flags: in `TWGameManager.receiveEntityModeChange` the BattleArmor missile-dump branch was calling `mounted.getType().hasFlag(WeaponType.F_MISSILE)` on equipment that had not been checked to be a weapon. `MiscType`, `WeaponType` and `AmmoType` each override `hasFlag` to log a warning **with a stack trace** and return false when handed a flag of the wrong class, so that line was quietly writing warnings to the log for any body-mounted non-weapon. It now binds the type first:

```java
(equipmentType instanceof WeaponType weaponType) && weaponType.hasFlag(WeaponType.F_MISSILE)
```

Pre-existing, not something this PR introduced, but it was in a method I was already touching.

HammerGS:
Thanks for the review - all eight comments are addressed in 40f06026ea, with individual replies on each thread. Summary of the two headline items:

**Magic strings.** Constants added and used at every site this PR touches: `Mounted.MODE_ON` / `MODE_OFF` / `MODE_NONE`, `MiscType.MODE_ECM`, and `Weapon.MODE_GAUSS_POWERED_UP` (next to the `MODE_GAUSS_POWERED_DOWN` already there). That covers the 34 `setModes` calls in `MiscType`, both C3 masters, the gauss and improved heavy laser weapons, `Entity`'s two mode-list rebuilds, `SystemPanel`, `ServerHelper`, `Mek`, `EquipChoicePanel` and the tests.

I split them across three classes to match the existing precedent (`Weapon.java` already holds a block of ~30 `MODE_*` constants) rather than putting everything in `Mounted` - say the word if you would rather they all live in one place.

Going further and eliminating string comparison altogether is not something I can do inside this PR: the mode system is string-keyed end to end (`EquipmentMode` is interned by name, `setModes(String...)`, `setMode(String)`, and the i18n key is `"EquipmentMode." + name`). An enum conversion would touch every piece of equipment in the codebase.

**`getType()` null safety.** You were right that it can return null - it resolves lazily through `EquipmentType.get(typeName)` and yields null when a mount is restored naming equipment the build does not know. It is now `@Nullable` (Jakarta, as asked) with the contract documented, and I worked through every call site in this PR's scope: `Entity` (probe, ECM and Nova loops), `Mek` (all four heat sink methods), `BattleArmor`, `EquipmentActivation`, `ServerHelper`, `SystemPanel`, `EquipChoicePanel`, and `TWGameManager.receiveEntityModeChange`. In each the type is fetched **once** into a local and guarded, which also removes the repeated `getType()` calls you flagged separately.

**Two things that fell out of the pass, worth flagging:**

1. `TWGameManager.receiveEntityModeChange` was calling `hasFlag(WeaponType.F_MISSILE)` on equipment not yet known to be a weapon. Since `MiscType` / `WeaponType` / `AmmoType` each override `hasFlag` to log a warning **with a stack trace** and return false on a wrong-class flag, that line was quietly writing warnings for any body-mounted non-weapon on a BattleArmor. Now guarded with an `instanceof` binding. Pre-existing, but it was in a method I was already in.

2. On the flip side of the same coin: the `instanceof` checks in these gates are **load-bearing**, not redundant. Because of those overrides, testing a `MiscType` flag against an `EquipmentType`-typed reference does not just return false quietly - it logs. Where a loop is already over `getMisc()` (statically `MiscType`) no check is used, but wherever the reference is `EquipmentType` the binding stays.

**Non-English GUI** (your question on the mode filtering): safe. `EquipmentMode.equals(String)` compares the internal name and never touches a resource bundle; only `getDisplayableName()` localizes. The combo shows localized text but applies the selection by index, so a German client behaves identically. Detail on the thread.

Also removed the two legacy `Mek.sinksOn` fields entirely rather than keeping them - see that thread for what was actually depending on them (nothing outside MegaMek save games, now handled by `omitField` plus a regression test). Full suite green, `EquipmentOffModesTest` now at 11 tests.

IllianiBird (approved):

PR #8578, opened 2026-07-24
fix: player team name retrieval normalization (MM #8569)

### LINKED ISSUE #8569: Lobby crashes (ArrayIndexOutOfBoundsException) after Reset Game to lobby when a player has no team assigned
## What happened

Loaded a saved game (created a couple of days earlier), then used Reset Game to return to the lobby. The client crashed with an uncaught exception, and the lobby player table then throws the same exception on every repaint, making the lobby unusable.

## Error

```
java.lang.ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 6
	at megamek.client.ui.panels.phaseDisplay.lobby.PlayerTable$PlayerRenderer.getPlayerDescription(PlayerTable.java:224)
	at megamek.client.ui.panels.phaseDisplay.lobby.PlayerTable$PlayerRenderer.getTableCellRendererComponent(PlayerTable.java:203)
	at javax.swing.JTable.prepareRenderer(JTable.java:5818)
	... (repeats on every table repaint on the AWT event thread)
```

## Root cause

`PlayerTable.getPlayerDescription` renders the team name with:

```java
result.append(Player.TEAM_NAMES[player.getTeam()]);   // PlayerTable.java:224
```

`Player.getTeam()` can legitimately return `Player.TEAM_UNASSIGNED = -1` (Player.java:82) - other code such as `Player.isEnemyOf` (Player.java:568) explicitly handles that value - but `TEAM_NAMES` only has entries for teams 0-5, so any unassigned player in the lobby crashes the renderer. After resetting to the lobby from a loaded save, at least one player (apparently a ghost/bot restored from the save) came back with `TEAM_UNASSIGNED`, which triggers the crash.

There is a second latent copy of the same unguarded indexing at `Player.java:1019` (`TEAM_NAMES[team]` in the HTML description builder), which will fail the same way for any unassigned player.

## Suggested fix

Guard both sites so `TEAM_UNASSIGNED` (or any out-of-range team) renders as e.g. "Unassigned" instead of indexing the array blindly. Separately it may be worth checking why Reset Game leaves a restored player without a team, but the renderer should never be able to crash the client over a player state the model explicitly allows.

## Environment

- MegaMek: current main (development build, main @ ff6476f8aa, 2026-07-22)
- OS: Windows 11 Pro
- Java 21
- Reproduced in a hotseat/single-process game after loading a .sav from a build a few days older

NickAragua:
Added a "getTeamName" function to the Player class which returns the team name with lower bound safety check and no upper bound. So now you can just call that instead of calling "Player.TEAM_NAMES[player.getTeam()]"

Also a minor optimization to the RoundsInAirDialog, which was refreshing many times per round even when the dialog wasn't visible.

Fixes #8569

### DIFF
--- a/megamek/src/megamek/client/ui/clientGUI/MapMenu.java
+++ b/megamek/src/megamek/client/ui/clientGUI/MapMenu.java
@@ -423,7 +423,7 @@ public class MapMenu extends JPopupMenu {
     }
 
     private JMenu createBotCommands(Player bot) {
-        JMenu menu = new JMenu(bot.getName() + " (" + Player.TEAM_NAMES[bot.getTeam()] + ")");
+        JMenu menu = new JMenu(bot.getName() + " (" + bot.getTeamName() + ")");
 
         JMenu prioritizeTargetUnitMenu = new JMenu(Messages.getString("Bot.commands.priority"));
         JMenu ignoreTargetMenu = new JMenu(Messages.getString("Bot.commands.ignore"));
--- a/megamek/src/megamek/client/ui/dialogs/RoundsInAirDialog.java
+++ b/megamek/src/megamek/client/ui/dialogs/RoundsInAirDialog.java
@@ -129,6 +129,10 @@ public class RoundsInAirDialog extends JDialog {
      * landing time, with the target hex and warhead listed as "Unknown". Safe to call repeatedly (e.g. on phase change).
      */
     public void refresh() {
+        // if we're not actually showing this control, there's no need to actually do a refresh
+        if (!isVisible()) {
+            return;
+        }
         tableModel.setRowCount(0);
         Game game = client.getGame();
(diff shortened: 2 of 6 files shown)


HammerGS (approved):

IllianiBird (approved):
This looks good. I think we should review Copilot before merging, but I don't _think_ Copilot is in the right here (especially not for it's `Player.java` comment as I think that'll just reintroduce the bug you're fixing).

NickAragua:
I'll double check that the 'rounds in flight' window updates properly if it's open; if it's closed it obviously doesn't matter if it updates.

NickAragua:
Yeah, it updates fine.

NickAragua:
Oof, there's one workflow where it didn't. I'll create a separate PR to deal with it.

PR #8580, opened 2026-07-25
Fix: Flotation Hull and Fire Control Flags and Format MiscType

WellBattle6:
Fix Flotation Hull and Fire Control flags
Format MiscType


WellBattle6:
@HammerGS Looks like CoPilot failed.

PR #8587, opened 2026-07-26
Fix trailer ammo selection never reaching the server 

### LINKED ISSUE #8520: [Issue] Trailers not passing ammunition to vehicle
### Brief Description *

*Tested with Mobile Long Tom*

Trailers passing ammunition seems to be broken. If the vehicle still has ammo in its body it will fire from the first available ton instead of whatever ammo is selected in the dropdown from the trailer. If the body has no ammo, the weapon simply will not fire (though a to-hit number is given).

### 3. Steps to Reproduce

1. Load a vehicle and a trailer carrying ammo for a weapon the vehicle mounts (For example a Mobile Long Tom and an Ammunition Carriage Trailer).
2. During an appropriate firing phase, change the ammo selection to a ton that resides in the trailer (this works particularly well with something like normal artillery rounds and smoke where the difference is immediately apparent at impact).
3. Observe that when the weapon fires, it fires the ammo from the first available body ton instead of the selected trailer ton.
4. Use GM mode to remove all applicable ammo from the vehicle body.
5. Next applicable firing phase, select a ton of ammo from the body. Observe that a to-hit number will be generated when a target is selected.
6. Observe that the weapon fails to fire during attack resolution.

### Attach Files

[megamek.log](https://github.com/user-attachments/files/30048816/megamek.log)

### Severity *

### MegaMek Suite Version *

0.51.01-nightly-2026-07-11

### Operating System *

Windows 11 10.0 (amd64)

### Java Version *

Adoptium 21.0.10

### Final Verification

HammerGS:
Summary

A vehicle towing an ammo trailer could not use the trailer's ammo. Picking a [TR] bin in the ammo dropdown did nothing — the weapon fired the first body ton instead. With the body emptied, the player got a to-hit number but the weapon did not fire.

The dropdown has offered trailer ammo since 2020 (8bfd94c1e4). The packet that tells the server about the choice was never extended to carry it, so the feature has never worked end to end.

Bug Fixed

- The packet could not name another unit's bin. WeaponPanel sent entity.getEquipmentNum(ammo), which returns -1 for a bin owned by a trailer, and the packet had no field for the carrying unit. The server logged does not have ammo #-1 and left the weapon linked to whatever it had before.
- To-hit and attack resolution disagreed. WeaponAttackAction already carries ammoId plus ammoCarrier, and ComputeToHit resolves both. So the player got a to-hit number for ammo the server was never going to fire.
- Cross-entity links leaked into packets. An ammo link to another unit puts an Entity reference inside the tractor's serialized form, so the receiving side would hold a detached duplicate of the trailer.

Changes

1. TrainAmmoSharing.java (NEW) — stateless helper holding the whole rule: which units may supply each other, the shared bin list, and the post-transfer relink.
2. Client.java — sendAmmoChange takes the carrying unit's id.
3. WeaponPanel.java — resolves the bin's equipment number against its own carrier and sends that carrier's id. Builds the dropdown from TrainAmmoSharing.getSharedAmmo instead of its own inline copy of the rule.
4. TWGameManager.java — resolves the bin against the named carrier, after checking the carrier is the firing unit or a directly connected trailer. The equipment number comes from the client, so it is not taken on trust.
5. Entity.java — setGame repoints a cross-entity ammo link at the canonical bin, mirroring the fixup already there for carriedObjects.
6. FireControl.java / ArtilleryTargetingControl.java — updated for the new signature.

Files Changed

- megamek/src/megamek/common/equipment/TrainAmmoSharing.java — NEW: the s and server
- megamek/src/megamek/client/Client.java — carrier id added to sendAmmoChange
- megamek/src/megamek/client/ui/dialogs/unitDisplay/WeaponPanel.java — both call sites resolve the bin against its carrier; dropdown uses the shared rule                        - megamek/src/megamek/server/totalWarfare/TWGameManager.java — carrier re
- megamek/src/megamek/common/units/Entity.java — one-line relink call in setGame                                                                                                 - megamek/src/megamek/client/bot/princess/FireControl.java — new signaturhe carrier to match the setAmmoCarrier two lines above it
- megamek/src/megamek/client/bot/princess/ArtilleryTargetingControl.java — new signature
- megamek/unittests/megamek/common/equipment/TrainAmmoSharingTest.java — NEW: 8 tests
- megamek/unittests/megamek/server/totalWarfare/TWGameManagerTrailerAmmoTest.java — NEW: 7 tests
                                                                                                                                                                                 Testing

- Unit tests: 15 new, all passing.                                                                                                                                                 - TWGameManagerTrailerAmmoTest (7) drives real ENTITY_AMMO_CHANGE packetwo tow-linked Bulldog Medium Tanks: draw from the trailer behind, drawfrom the tractor ahead, own ammo still changes, plus rejection of an unconnected carrier, an unknown carrier id, and an unresolvable bin.
  - TrainAmmoSharingTest (8) covers the sharing predicate, the shared bin list and its order, and the relink — duplicate carrier repointed, own ammo untouched, departed carrier cleared.
- Full ./gradlew test and checkstyleMain green.
- Manual: confirmed in game with a Mobile Long Tom and an Ammunition Carro bin now fires that bin.

Fixes #8520

### DIFF
--- a/megamek/src/megamek/client/bot/princess/FireControl.java
+++ b/megamek/src/megamek/client/bot/princess/FireControl.java
@@ -3290,7 +3290,7 @@ public class FireControl {
 
             // If the selected ammo would cause the shot to miss, skip loading it.
             final WeaponAttackAction cloneWAA = new WeaponAttackAction(info.getAction());
-            cloneWAA.setAmmoId(shooter.getEquipmentNum(mountedAmmo));
+            cloneWAA.setAmmoId(mountedAmmo.getEntity().getEquipmentNum(mountedAmmo));
             cloneWAA.setAmmoMunitionType(mountedAmmo.getType().getMunitionType());
             cloneWAA.setAmmoCarrier(mountedAmmo.getEntity().getId());
             if (cloneWAA.toHit(owner.getGame(), owner.getPrecognition().getECMInfo()).getValue() > 12) {
@@ -3327,8 +3327,10 @@ public class FireControl {
             info.getAction().setAmmoMunitionType(cloneWAA.getAmmoMunitionType());
             info.getAction().setAmmoCarrier(cloneWAA.getAmmoCarrier());
 
+            Entity ammoCarrier = mountedAmmo.getEntity();
             owner.sendAmmoChange(info.getShooter().getId(), shooter.getEquipmentNum(currentWeapon),
-                  shooter.getEquipmentNum(mountedAmmo), mountedAmmo.getSwitchedReason());
+                  ammoCarrier.getEquipmentNum(mountedAmmo), ammoCarrier.getId(),
+                  mountedAmmo.getSwitchedReason());
         }
     }
 
--- a/megamek/src/megamek/server/totalWarfare/TWGameManager.java
+++ b/megamek/src/megamek/server/totalWarfare/TWGameManager.java
@@ -27560,36 +27560,51 @@ public class TWGameManager extends AbstractGameManager {
         int entityId = packet.getIntValue(0);
         int weaponId = packet.getIntValue(1);
         int ammoId = packet.getIntValue(2);
-        int reason = packet.getIntValue(3);
-        Entity e = game.getEntity(entityId);
+        int ammoCarrierId = packet.getIntValue(3);
+        int reason = packet.getIntValue(4);
+        Entity shooter = game.getEntity(entityId);
 
         // Did we receive a request for a valid Entity?
-        if (null == e) {
+        if (null == shooter) {
             LOGGER.error("Could not find entity# {}", entityId);
             return;
         }
         Player player = game.getPlayer(connIndex);
-        if ((null != player) && (e.getOwner() != player)) {
-            LOGGER.error("Player {} does not own the entity {}", player.getName(), e.getDisplayName());
+        if ((null != player) && (shooter.getOwner() != player)) {
+            LOGGER.error("Player {} does not own the entity {}", player.getName(), shooter.getDisplayName());
+            return;
+        }
+
+        // The ammo may be carried by a directly connected trailer rather than by the firing unit itself. Resolve
+        // the bin against its own carrier, but never trust the client about which unit that is allowed to be.
+        Entity ammoCarrier = (ammoCarrierId == entityId) ? shooter : game.getEntity(ammoCarrierId);
+        if (null == ammoCarrier) {
+            LOGGER.error("Could not find ammo carrier# {}", ammoCarrierId);
+            return;
+        }
+        if (!TrainAmmoSharing.canShareAmmoWith(shooter, ammoCarrier)) {
+            LOGGER.error("Entity {} may not draw ammo from {}",
+                  shooter.getDisplayName(),
+                  ammoCarrier.getDisplayName());
             return;
         }
 
         // Make sure that the entity has the given equipment.
-        WeaponMounted weaponMounted = (WeaponMounted) e.getEquipment(weaponId);
-        AmmoMounted mAmmo = (AmmoMounted) e.getEquipment(ammoId);
+        WeaponMounted weaponMounted = (WeaponMounted) shooter.getEquipment(weaponId);
+        AmmoMounted selectedAmmo = ammoCarrier.getAmmo(ammoId);
         AmmoMounted oldAmmo = (weaponMounted == null) ? null : weaponMounted.getLinkedAmmo();
-        if (null == mAmmo) {
-            LOGGER.error("Entity {} does not have ammo #{}", e.getDisplayName(), ammoId);
+        if (null == selectedAmmo) {
+            LOGGER.error("Entity {} does not have ammo #{}", ammoCarrier.getDisplayName(), ammoId);
             return;
         }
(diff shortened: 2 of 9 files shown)


megamek/src/megamek/server/totalWarfare/TWGameManager.java
```diff
@@ -27560,59 +27560,74 @@ private void receiveEntityAmmoChange(Packet packet, int connIndex) throws Invali
...
+                  shooter.getDisplayName(),
+                  ammoCarrier.getDisplayName());
             return;
         }
 
         // Make sure that the entity has the given equipment.
-        WeaponMounted weaponMounted = (WeaponMounted) e.getEquipment(weaponId);
-        AmmoMounted mAmmo = (AmmoMounted) e.getEquipment(ammoId);
+        WeaponMounted weaponMounted = (WeaponMounted) shooter.getEquipment(weaponId);
+        AmmoMounted selectedAmmo = ammoCarrier.getAmmo(ammoId);
```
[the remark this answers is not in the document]
HammerGS: Verified and fixed in `167f069f8b`.

The mechanism holds. `receiveEntityAmmoChange` casts the client-supplied `weaponId` straight to `WeaponMounted`, and the cast runs before the null check that was meant to reject bad input. `AmmoMounted` and `WeaponMounted` are sibling types, so a packet naming an ammo bin as the weapon throws immediately.

The severity claim also holds, which I did not expect. `TWGameManager.handlePacket` and `Server.handle` each catch only `InvalidPacketDataException`, and `PacketPump.run()` has no catch around `handle()` at all, only `InterruptedException` around the queue wait. An escaping `ClassCastException` ends the Packet Pump thread and the server stops processing packets.

Fixed with `shooter.getWeapon(weaponId)`, which applies the same `instanceof` check the ammo side already gets from `getAmmo(int)`.

Two corrections to the framing. The cast is pre-existing, not introduced here: `git diff main...HEAD` shows the line only because this PR renamed `e` to `shooter`. And the reason it stood out is this PR's own doing, since the sibling line used to be `(AmmoMounted) e.getEquipment(ammoId)` and is now the null-safe `ammoCarrier.getAmmo(ammoId)`, which left the weapon cast as the only raw one in the method. Worth fixing on both counts.

Scope checked: this is the only raw cast on a client-supplied equipment index in any `receive*` method in `TWGameManager`.

Added `nonWeaponEquipmentIndexIsRejectedWithoutThrowing`, and confirmed it fails with `ClassCastException` when the fix is reverted.

megamek/src/megamek/client/bot/princess/FireControl.java
```diff
@@ -3290,7 +3290,7 @@ void loadAmmo(final Entity shooter,
 
             // If the selected ammo would cause the shot to miss, skip loading it.
             final WeaponAttackAction cloneWAA = new WeaponAttackAction(info.getAction());
-            cloneWAA.setAmmoId(shooter.getEquipmentNum(mountedAmmo));
+            cloneWAA.setAmmoId(mountedAmmo.getEntity().getEquipmentNum(mountedAmmo));
```
IllianiBird, line 3293: This looks suspicious, please double-check that `getEntity()` can't return `null` before merging.

IllianiBird (approved):
One comment that needs investigating before this is merged.

megamek/src/megamek/client/bot/princess/FireControl.java
HammerGS: Checked, and I believe this one is safe as written.

The strongest evidence is two lines down: `cloneWAA.setAmmoCarrier(mountedAmmo.getEntity().getId())` is pre-existing on main and does the identical dereference on the identical object. It shows up as context in this diff rather than an addition. If `getEntity()` could return null here, main would already be throwing on that line.

On whether it can be null at all:

- `Mounted.entity` is `private final`, assigned in the constructor, and no construction site in `megamek/src` passes null. Grepping `createMounted(null`, `new AmmoMounted(null`, `new WeaponMounted(null`, `new MiscMounted(null` returns nothing.
- `getEntity()` carries no `@Nullable`.
- There are 95 unguarded `.getEntity().get...()` dereferences across `megamek/src`, so this is a settled assumption rather than a local shortcut.

Provenance in this specific method is also concrete: `mountedAmmo` comes from `getPreferredAmmo(...)`, which sources its candidates from `shooter.getAmmo()`. Those bins are constructed with the shooter as their entity, so `getEntity()` is the shooter.

Worth saying that the concern is reasonable as a diff-level observation, since the line does introduce a second dereference. What it changed is the resolution target, from `shooter.getEquipmentNum(mountedAmmo)` to the bin's own carrier, so that the ammo id and the `setAmmoCarrier` call two lines down agree. Before this PR they could disagree, which is the class of bug the PR exists to fix. Princess only selects its own ammo today, so the two happened to agree in practice.

If you would still prefer belt and braces, the tidiest form is hoisting `Entity ammoCarrier = mountedAmmo.getEntity();` and using it for both lines, which is what this PR already does further down in the same method for the `sendAmmoChange` call. Happy to make that change if you want the assumption stated in one place rather than twice. Annotating `Mounted.getEntity()` as non-null would be the broader fix, but that belongs in its own PR.

Issue #8599, opened 2026-07-28
[Issue] Jump and Paratrooper Infantry cannot drop at above altitude 8


agrazela14:
### Brief Description *

Strategic Operations page 20 states that Jump, Paratrooper, and Mechanized VTOL infantry can make atmospheric drops so long as the carrier is not in a space / atmosphere interface hex. Megamek has an upper limit of altitude 8 included in MovementDisplay.getDroppedUnits which prevents dropping from infantry carriers above altitude 8, but includes no rules reference.

### 3. Steps to Reproduce

1. Start Megamek
2. Load an Aerospace unit with infantry bays and / or compartments (eg. Delta Air Cruiser MA-80)
3. Load an infantry unit with jump (eg. Jump Platoon (Laser))
4. Right Click the infantry > Load Onto your aerospace carrier
5. Start Megamek, deploying the aerospace unit at altitude 9 or 10 anywhere
6. In the movement phase, press the 'Drop Troops' Panel button and Observe that no dialog appears
7. Go Down to altitude 8 or less, and press the 'Drop Troops' button again, Observe that this time it will pop up the dropping dialog

### Attach Files

<img width="1542" height="1022" alt="Image" src="https://github.com/user-attachments/assets/1cbfb758-20a7-47cf-8256-068ae862a373" />
<img width="1573" height="1017" alt="Image" src="https://github.com/user-attachments/assets/b84f3c37-df5e-4b49-b2e4-a5b86a464847" />

### Severity *

Medium (Gameplay Limitation): Non-core functionality is impaired, providing a suboptimal but playable experience.

### MegaMek Suite Version *

0.51.01

### Operating System *

Linux (Debian)

### Java Version *

21.0.11

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have included all necessary information and files to help reproduce this issue
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

IllianiBird:
I thought infantry could only be deployed at Alt 1?

radar6510:
> I thought infantry could only be deployed at Alt 1?

In Total Warfare, yes but after going above Alt 1, it uses StratOps Drop rules.

gsparks3:
Yeah, TW reference is TW page 225 (and does indeed require altitude 1), but the SO drop rules are separate, and work somewhat differently.

PR #8600, opened 2026-07-28
Fix Issue #8599 Jump and Paratrooper Infantry cannot drop at above altitude 8

agrazela14:
Removing the upper altitude limit of 8 for infantry transports dropping troops


IllianiBird:
My understanding was that dropped infantry required the transport unit to be at Altitude 1 (Nap of Earth)?

agrazela14:
> My understanding was that dropped infantry required the transport unit to be at Altitude 1 (Nap of Earth)?

I also thought that was true for anything but paratroopers, but Strategic ops specifically overrules it for jump infantry and VTOL mechanized infantry. What I cannot find is anything mentioning an altitude 8 limit

IllianiBird (approved):

Issue #8608, opened 2026-07-28
canTow ignores movement mode, so railcars can be towed by tracked vehicles


HammerGS:
## Problem

`Entity.canTow` does not look at movement mode, so a rail unit can be hitched behind a vehicle that has no track under it. A player can put a TGV Omni Railcar behind a Destrier Siege Vehicle and the game accepts it.

`canTow` checks the tractor and trailer flags, that the trailer is not already towed, ownership, elevation, the tractor's towing capacity, and that a free hitch exists at the tail of the train. Movement mode is never considered.

## Example

Rail units in the game:

| Unit | Role | UUID |
|---|---|---|
| Adelante Passenger/Cargo Train | tractor | `019f583e-eaba-77cc-bbc6-e0ce02cb8ced` |
| TGV Omni Railcar Primary Configuration | trailer | `019f583e-eadc-75eb-992c-2dc58ce46ff1` |
| TGV Omni Railcar Primary (Luxury) | trailer | `019f583e-ead9-788f-9051-cd4db6b63371` |
| TGV Omni Railcar A (Passengers) | trailer | `019f583e-eabe-7fb3-a5fd-7786e7c0c045` |
| TGV Omni Railcar B (Military Support) | trailer | `019f583e-eac1-7290-a30c-99fa10f3558c` |
| TGV Omni Railcar C (Cargo) | trailer | `019f583e-eac4-72c9-be12-f0862f962a4e` |
| TGV Omni Railcar D (MASH) | trailer | `019f583e-eac8-7be2-a548-c275acaa1e23` |
| Flatbed Railcar | trailer | `019f583e-ebb6-7835-9731-d012361147b1` |

The Adelante is the only Rail tractor in the game. Take a Destrier Siege Vehicle, which is Tracked, put a TGV Omni Railcar C (Cargo) behind it in the lobby, and it hitches. The train then moves cross country on tracks the railcar does not have.

## Proposed fix

Add a movement-mode rule to `canTow`, so a Rail trailer can only be towed by a Rail tractor.

The reverse case needs a decision too: should a Rail tractor be allowed to tow a non-Rail trailer? A locomotive hauling a wheeled trailer is as odd as the case above, so the same rule probably applies in both directions.

## Implementation notes

- One change in `Entity.canTow` covers every caller: the lobby "Connect as Train" action, the single-unit tow menu, the bot connecting its own trains, and MekHQ.
[...]

Issue #8607, opened 2026-07-28
Read a towedBy tag so trailers can be paired with their tractor


HammerGS:
## Problem

Nothing in a unit file says which tractor a trailer belongs to. Anything that has to pair them up has to guess from the name, and for most trailers the name gives nothing away.

A bot has to make this pairing on its own, because it never uses the lobby's "Connect as Train" action. The name heuristic in `BotTrainPlanner` covers the Mobile Long Tom family, where a carriage extends its gun's model, and almost nothing else. A Ballista Artillery Trailer does not read as belonging to a Destrier Siege Vehicle, and no amount of string matching will make it.

There are 45 trailers in mm-data, split across two mechanisms that do not overlap:

- 22 combat vehicles carrying a `<trailer>` block, read in `BLKTankFile.java:269`
- 23 support vehicles carrying `SV Chassis Mod [Trailer]`. `BLKFile.java:1283` writes the `<trailer>` flag only when `!t.isSupportVehicle()`, so support vehicle trailers never have it

## Example

A player gives a bot a Destrier Siege Vehicle and a Ballista Artillery Trailer. The bot has to work out that the Ballista goes behind the Destrier. Nothing in either file connects them, and the names share no text, so the pairing can only fall back to "hitch it to whatever has capacity". With a second tractor on the field it can hitch the Ballista to the wrong one.

Name matching fails outright on a real case: the Teppō Artillery Vehicle carries a macron in its chassis name. A search for "Teppo" does not find it.

## Proposed fix

Add an optional tag to trailer files naming the tractors they belong to. Each line is the tractor's UUID, then a pipe, then its chassis and model:

```
<towedBy>
019f583e-e7a9-7a3a-bc7b-1d5353eb104a|Destrier Siege Vehicle
</towedBy>
```

One line per tractor:

```
<towedBy>
019f583e-e8f6-7734-b7fd-8bec3772bda2|Generic Expandable Services Vehicle (Command Post)
019f583e-e8f7-7f4d-b7a0-d300976b8c70|Generic Expandable Services Vehicle (Field Kitchen)
019f583e-e8f9-7e9f-8ef5-1ba56493b1fe|Generic Expandable Services Vehicle (MASH)
</towedBy>
```

**The UUID is the link. The name is for people.** The parser splits on the first pipe, uses the UUID, and ignores the rest. That way a unit can be renamed without breaking any tag, but anyone opening the file or reading a diff can see what it points at instead of an opaque string.

[...]

Issue #8606, opened 2026-07-28
Counter-battery fire misses trailers and does nothing with special munitions


HammerGS:
## Problem

Counter-battery fire against an off-board artillery unit has two bugs.

**1. Only the aimed unit takes damage.** A counter-battery salvo damages the one entity that was targeted and nothing else. Trailers hitched to that unit are never checked, so an ammunition carriage sitting in the same off-board hex takes no damage at all. `ArtilleryWeaponIndirectFireHandler.java:506-520` builds a blast shape, then reads exactly one key from it: the target's own elevation and position. The code already says so at line 504:

> `// Note: currently this only damages the target and does not deal blast damage to "nearby" off-board units.`

**2. Special munitions do nothing off-board.** Fuel-air rounds report "Direct hit!" and deal zero damage. The `M_FAE` branch at line 409 runs `processFuelAirDamage` and returns before the off-board damage code at line 506 is ever reached. `processFuelAirDamage` walks its blast shape calling `gameManager.artilleryDamageHex(...)`, which looks up board hexes. The center is a virtual off-map coordinate, so it finds no hexes and no units.

The same early-return shape applies to every other special munition: `M_FLARE` (416), `M_FASCAM` (453), `M_INFERNO_IV` (464), `M_VIBRABOMB_IV` (469), `M_SMOKE` (474), `M_LASER_INHIB` (478). `M_DAVY_CROCKETT_M` (430) is the only one that branches on `target.isOffBoard()`.

Note that off-board position is not the problem. `Entity.deployOffBoard()` (`Entity.java:11833`) puts every off-board unit at one computed virtual hex per direction and distance, and `DeploymentProcessor.followTractorsOffBoard()` copies the tractor's distance and direction onto its trailers (RFE #8506). An off-board train does sit in a single shared hex. Bug 1 just never looks past the named target.

## Example

Off-board, a Mobile Long Tom Artillery LT-MOB-25 towing an LT-MOB-25F (Ammunition Carriage) and an LT-MOB-25F (Support Carriage). An enemy Mobile Long Tom fires counter-battery at the tractor with a 2 turn flight time.

Standard Long Tom round, direct hit:

```
Long Tom (Long Tom) at Mobile Long Tom Artillery LT-MOB-25 needs 7, rolls 7
    Direct hit!
    Mobile Long Tom Artillery LT-MOB-25 hit for 25 damage.
```

The tractor takes 25 damage. Both carriages take nothing.

[...]

PR #8614, opened 2026-07-29
Connect, deploy and move tractor-trailer trains as one unit

HammerGS:
## Summary

A tractor and its trailers had to be hitched one pair at a time, deployed one unit at a time, and
each trailer took its own movement turn. This makes a train act like one unit from the lobby through
to the end of movement.

Over half the diff is tests: 15 test files, about 2,400 lines.

## Changes Made

**Build a train**

- Select a tractor and several trailers in the lobby, then "Connect as Train". One packet, one server handler, all or nothing.
- A dialog sets the trailer order, because order fixes the hitch chain and so where each trailer ends up.
- Bots hitch their own trains in the lobby. Matching tries force and design first, so a Long Tom 25 does not grab a Long Tom 95's carriage.

**See it**

- The lobby draws carried units under their carrier, one step deeper per level of nesting. This covers all carried units, not just trailers: DropShips in a JumpShip, fighters in a bay, trailers behind a tractor.
- Sorting keeps a carrier and its load together, in every sort column.
- Trailer bins in the ammo dropdown are labelled by position in the train (TR, TL1, TL2), so identical carriages can be told apart.
- Ammo trailers show what they carry in the tooltip. An ammunition carriage is nothing but bins and hitches, so the per-weapon lines never ran for it and it showed no ammo at all.

**Deploy it**

- The tractor decides where the whole train deploys. Setting the tractor off board takes its trailers with it.
- A trailer on its own can deploy on board or off board.
- A hooked-up train cannot uncouple while off board. Blocked in the lobby, at deployment, and in movement.

**Move it**

- Trailers no longer take their own movement turns. The tractor moves the train and takes the movement penalty.
- Skipped trailers are logged.

**Fixes**

- Detaching more than one trailer at once threw `ConcurrentModificationException`. Four separate defects; see `TrainDisconnectTest`.
- Towing capacity was measured against the attach point's own weight, not the powered tractor's. TM says the combined weight must be at most the tractor's own weight.
- A weapon linked to a trailer's ammo bin kept that link after the train uncoupled, so a tractor went on firing a detached trailer's ammo from any distance.

## Files Changed

New source:

- `megamek/src/megamek/common/units/TrainLayout.java` - which hex each unit in a train sits in, and its position label
- `megamek/src/megamek/server/totalWarfare/TrainBuildHandler.java` - validates and builds a train server side
- `megamek/src/megamek/client/bot/BotTrainPlanner.java` - which of a bot's units should hitch to which
- `megamek/src/megamek/client/ui/dialogs/lobby/TrainOrderDialog.java` - sets trailer order

Main edits:

- `megamek/src/megamek/common/units/Entity.java` - tow and uncouple bookkeeping, `canTow` weight rule
- `megamek/src/megamek/server/totalWarfare/DeploymentProcessor.java` - train deployment and off-board rules
- `megamek/src/megamek/client/ui/panels/phaseDisplay/lobby/sorters/MekTableSorter.java` - keeps carriers and loads together
- `megamek/src/megamek/client/ui/panels/phaseDisplay/lobby/LobbyMekCellFormatter.java` - the nesting glyph
- `megamek/src/megamek/common/equipment/TrainAmmoSharing.java` - drops ammo links on uncouple
- `megamek/src/megamek/client/ui/clientGUI/tooltip/UnitToolTip.java` - carried ammo line
- Plus `Client`, `BotClient`, `Tank`, `PacketCommand`, `TWGameManager`, `MovePathHandler`, `ServerLobbyHelper`, `ChatLounge`, `LobbyActions`, `LobbyErrors`, `LobbyMekPopup`, `CustomMekDialog`, `DeploymentDisplay`, `MovementDisplay`, `WeaponPanel`, `messages.properties`

Tests: 15 files. `TrainLayoutTest`, `TrainBuildHandlerTest`, `TrainDisconnectTest`, `EntityCanTowTest`,
`TractorMovementTest`, `TrailerDeploymentTurnTest`, `OffBoardTrailerDeploymentTest`,
`BotTrainPlannerTest`, `CarrierStackSortingTest`, `CarriedBranchDepthTest`, `TrainPositionLabelTest`,
`CarriedAmmoToolTipTest`, `TrainAmmoSharingTest`, `TWGameManagerTrailerAmmoTest`,
`DeploymentDisplayTest`.

## Testing

- Full suite green: 568 classes, 14,451 tests. `checkstyleMain` clean.
- The crash fix and the uncouple ammo fix were both checked by reverting the fix and watching the new
  tests fail.
- Manual testing ran alongside the work, with QA feedback folded in as it came: building trains of
  several carriages, off-board artillery trains, detaching mid-train, bot-built trains, and the
  Mobile Long Tom battery.

## Rules clarification: trailer ammo sharing

Xotl answered four questions about trailer ammunition in the official rules Q&A
([forum topic 74296](https://www.battletech.com/forums/index.php/topic,74296.msg1742770.html#msg1742770)).
The code was checked against all four.

| Ruling | Status |
|---|---|
| "Immediate use only. No top off." | Already correct. A weapon links to the trailer's bin and fires from it. Nothing moves rounds between bins. |
| "Only vehicles directly coupled can share ammo." | **Fixed here.** See below. |
| "No limit" on rounds drawn at once. | Already correct. Links are per weapon and nothing caps rounds per turn. |
| "The tractor may pull from either trailer, but one trailer may not pull from the other." | Already correct for a chain. The trailer-tractor-trailer layout cannot be built in MegaMek at all; see Out of scope. |

Two changes came out of ruling 2:

- This branch had widened sharing from the coupled pair to the whole train, arguing from the Mobile
  Long Tom convoy and TM's "trailers act as part of the Tractor for movement, stacking and firing".
  One commit does that and a later one reverts it, so the net change to the sharing rule is nothing.
  RAW the gun reaches only the 25 rounds in the carriage hitched to it, not the convoy's 75.
- A weapon linked to a trailer's bin kept that link after the train uncoupled, so a tractor went on
  firing a detached trailer's ammo from any distance. Nothing on the firing path re-checked the
  coupling. Now dropped when the train splits.

The javadoc on `canShareAmmoWith` records the ruling and says not to widen it again on the strength
of the convoy description, because that is the reading this branch already tried once.

Note that "adjacent" in the rule means coupled, not adjacent on the map. A hex-distance check would
have been the wrong reading.

## Out of scope

- Counter-battery fire against a train (issue #8606).
- Reading a `towedBy` tag so trailers can be paired with their tractor (issue #8607, with the data
  half at MegaMek/mm-data#491).
- Letting rail units tow (issue #8608).
- A trailer on each end of a tractor cannot be built. `Entity.towing` is a single id and `towUnit`
  always appends to the tail, so every train is a linear chain. This is why ruling 4's layout is
  untestable rather than supported. Not changed here.


megamek/src/megamek/client/bot/BotClient.java
```diff
@@ -553,6 +575,41 @@ public List<Entity> getFriendEntities() {
...
+                continue;
+            }
+
+            Entity tractor = getGame().getEntity(plannedTrain.getKey());
+            LOGGER.info("[Train] {} connecting {} + {} trailer(s)",
+                  getName(),
+                  (tractor == null) ? plannedTrain.getKey() : tractor.getDisplayName(),
+                  plannedTrain.getValue().size());
+            sendBuildTrain(plannedTrain.getKey(), plannedTrain.getValue());
+        }
```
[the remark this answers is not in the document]
HammerGS: Valid, fixed in d7f032c3b9.

Verified: `requestedTrains` was cleared only on phase change, so within the lounge a tractor got exactly one request for the whole lobby. A tractor handed more trailers after that first request was skipped and those trailers stayed loose all game.

The dedupe now keys on the planned trailer list rather than the tractor id. An identical plan is still suppressed, which is what stops the loop (the server's reply arrives as another lobby update and calls this again), but a plan that has grown or been reordered is a different request and gets through.

Extracted the decision as a static `recordTrainRequest` so it is testable without a live BotClient, following the pattern `wantsStealthHeatForTsm` already sets in this class. `BotTrainRequestTest` has six tests; three fail under the old tractor-keyed behaviour, including your case.

megamek/src/megamek/client/ui/dialogs/lobby/TrainOrderDialog.java
```diff
@@ -0,0 +1,179 @@
...
+        connectButton.addActionListener(event -> {
+            confirmed = true;
+            setVisible(false);
+        });
+        JButton cancelButton = new JButton("Cancel");
+        cancelButton.addActionListener(event -> setVisible(false));
+        JPanel buttonPanel = new JPanel();
+        buttonPanel.add(connectButton);
+        buttonPanel.add(cancelButton);
+
```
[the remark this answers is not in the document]
HammerGS: Valid, fixed in d7f032c3b9.

Checked the convention before changing anything: comparable dialogs do route everything through the bundle (`CustomMekDialog` calls `Messages.getString` 101 times). All seven user-visible strings in this dialog now live in `messages.properties` under a `TrainOrderDialog.` prefix: title, order header, the `(tractor)` label, both reorder buttons, Connect and Cancel.

One related site left alone on purpose. The `Connect as Train...` item in `LobbyMekPopup` is also hard-coded, but so is every item around it (`Disembark / leave from carriers`, `Offload all carried units`, and the rest). Localising only the one line this PR added would be inconsistent with the file, and localising the whole popup is outside this PR's subject. Worth its own change if the team wants that file done.

megamek/src/megamek/client/ui/panels/phaseDisplay/lobby/sorters/PlayerTransportIDSorter.java
IllianiBird, line 1: This file hasn't been edited and so shouldn't have a year update

megamek/src/megamek/client/ui/panels/phaseDisplay/lobby/LobbyMekPopup.java
```diff
@@ -278,6 +279,17 @@ static ScalingPopup getPopup(List<Entity> entities, List<Force> forces, ActionLi
...
 
+        // Connecting several units at once. Offered whenever the selection could plausibly form a train; the exact
+        // ordering is chosen in the dialog and the server has the final say on legality.
+        boolean anyFreeTrailerSelected = joinedEntities.stream()
+              .anyMatch(entity -> entity.isTrailer() && (entity.getTractor() == Entity.NONE));
+        boolean anyFreeTractorSelected = joinedEntities.stream()
+              .anyMatch(entity -> entity.isTractor() && (entity.getTractor() == Entity.NONE)
+                    && (entity.getTowing() == Entity.NONE));
+        if ((joinedEntities.size() > 1) && anyFreeTrailerSelected && anyFreeTractorSelected) {
+            popup.add(menuItem("Connect as Train...", LMP_CONNECT_TRAIN + NO_INFO + seIds, true, listener));
```
IllianiBird: Hardcoded user-facing String

megamek/src/megamek/common/units/TrainLayout.java
```diff
@@ -0,0 +1,266 @@
...
+     * <p>
+     * The position is what a player actually needs to know. Ordering fixes the hitch chain and so which hex each
+     * unit occupies, and with several identical carriages the name alone does not say which is which.
+     * </p>
+     *
+     * @param unit the unit to label
+     *
+     * @return the label, or an empty string when the unit is not part of a train
+     */
+    public static String trainPositionLabel(@Nullable Entity unit) {
```
IllianiBird: If this is player facing it shouldn't be using hardcoded Strings

IllianiBird (approved):
Some minor adjustments needed, otherwise good to go.

n.b. I cannot speak to the rules as I'm not familiar. We should be open to the possibility that rules fixes may need to follow if there are any areas where we've misunderstood RAW.

megamek/src/megamek/client/ui/panels/phaseDisplay/lobby/sorters/PlayerTransportIDSorter.java
HammerGS: Correct, fixed in `d4c1ffc646`. The file now matches `main` exactly and is out of the diff.

It was a trail left by my own rework rather than a stray edit. I added train-grouping logic to this sorter and bumped the year for it, then moved that logic to a wrapper on `MekTableSorter` so every sorter benefits instead of just this one. Reverting the logic left the year behind, and I did not re-check the file afterwards.

megamek/src/megamek/client/ui/panels/phaseDisplay/lobby/LobbyMekPopup.java
HammerGS: Fair, fixed in `d4c1ffc646`. It now reads from the bundle as `ChatLounge.ConnectAsTrain`.

I had argued the other way on the Copilot thread for this same line: every item around it in `LobbyMekPopup` is hardcoded too, so localising one looked inconsistent. That reasoning was wrong in the way that matters. The line is one this branch adds, so it should be right regardless of what its neighbours do, and "the file is already like this" is how a file stays like that.

The rest of the file is untouched, so this does not turn into a popup-wide localisation pass inside a trailer PR.

megamek/src/megamek/common/units/TrainLayout.java
HammerGS: Yes, it is player facing: the label shows in the lobby unit list and in the weapon panel's ammo dropdown. Fixed in `d4c1ffc646`.

Localising it where it stood would have traded one problem for another, since `TrainLayout` is in `megamek.common.units` and the strings a player reads do not belong in common code. Split by layer instead:

- `TrainLayout.trainPosition(Entity)` returns a position: `NOT_IN_TRAIN`, `TRACTOR_POSITION`, or the trailer number.
- `UIUtil.trainPositionLabel(Entity)` turns that into words, from `ChatLounge.trainPositionTractor` and `ChatLounge.trainPositionTrailer={0}`.

Both callers were already client UI classes, so nothing else moved. `TrainPositionLabelTest` becomes `TrainPositionTest` and pins the position rather than the wording, which is the part that should not change when someone translates it.

PR #8643, opened 2026-08-04
Fix network desync

exeea:
This PR improves network packet ordering and authoritative state reconciliation to prevent client desync during lag. Adds validation and regression tests for packet handling, deployment, movement, and attacks.

- Keeps network packets in the correct order during lag.
- Resyncs clients when the server rejects an action.
- Prevents deployment loading from leaving units in a broken state.
- Validates SBF movement, attacks, ranges, and sprinting server-side.
- Handles malformed packets safely.
- Adds regression tests for packet ordering and state recovery.

gameplay fixes:
- fixes Jump modifier wrongly assigned (attacker modifier was applied to target)
- fixed sprinting: now properly round-up and added -1 TN for sprinting


HammerGS:
I'm going to hit this with Claude, but the real job keeps getting in the way today.

HammerGS:
# Review of #8643

First off: this is a strong PR. The packet-pump rewrite fixes a real lost-wakeup race (packets added between the `isEmpty()` check and `wait()` could sit unprocessed), the `SBFClient` missing-`break` fall-through was a genuine desync bug, the `loadUnit` turn-removal-before-load fix stops a failed load from eating a unit's turn, and the correction/resync design is thoughtful - including the double-blind safety of only ever returning entities owned by the requesting connection, and sending `prevPlayerId = connId` so simultaneous-phase displays actually re-init (verified against `MovementDisplay.gameTurnChange`). Test coverage is excellent. Everything compiles and all touched test classes pass locally on Windows.

Below is a checklist of items to address or double-check before merge.

## Must address (mechanical)

- [ ] **Indentation errors in shipped code** (checkstyle didn't catch these, but they'll trip readers):
  - `Server.java:285-286` - misindented closing braces at the end of `receivePacket`
  - `TWGameManager.java:4336` - `sendTurnSubmissionCorrection(...)` indented ~32 spaces
  - `TWGameManager.java:9971-9973` - `if (!isAttackSubmissionPhase()) {` at 10 spaces, the correction call at 24
  - `TWGameManager.java:27729` - same issue in `receiveEntitySystemModeChange`
  - `DeploymentProcessorTest` - the new test methods use 6-space indentation instead of the project's 4
- [ ] **Fully-qualified class names mid-code** - project convention is to add an import instead:
  - `SBFGameManager.java:615` - `megamek.common.actions.sbf.SBFAttackAction` inline in the stream filter
  - Tests: inline `java.util.List.of(...)`, `org.mockito.Mockito.when(...)`, `org.mockito.ArgumentMatchers.eq/any` in `DeploymentProcessorTest`, `SBFGameManagerCorrectionTest`, `SBFGameManagerPendingPacketTest`, `TWGameManagerTrailerAmmoTest`
- [ ] **Revert noise-only edits**:
  - `DropShipMekLoadTest` - the only change is swapping the order of two import lines
  - `TWGameManagerTrailerAmmoTest` / `DeploymentProcessorTest` - several untouched argument lists and the board text block were re-indented without a functional change

## Please double-check (substantive)

- [ ] **SBF rules changes are embedded in this networking PR - please confirm against IO:BattleForce RAW.** Three gameplay-affecting changes ride along:
  - Range modifiers changed: distance 1 was **+3** ("long range"), now **+2** (`LONG`); same-hex attacks previously had **no modifier**, now the attacker chooses SHORT/MEDIUM/LONG at **0/+1/+2**.
  - Damage previously used the client-submitted range - which the UI hardcodes to `ASRange.LONG` - so every attack did long-band damage regardless of distance. Damage now follows the authoritative effective range. Almost certainly a fix, but it changes battle outcomes.
  - Sprinting (floor 1.5x MP) is now legal server-side and sprinted formations are barred from the firing phase.

  These all look plausible, but none cite a rulebook page. Can you confirm the 0/+1/+2/+3 band table and the same-hex band selection match the published SBF rules?
- [ ] **`TWGameManager` grows by ~209 lines** - the project actively tries not to expand this class. Most of the wiring into the existing `receive*` methods is unavoidable, but the new helpers (`completeEntitySetting`, `sendEntitySettingCorrection`, `sendTurnSubmissionCorrection`, `correctInvalidPacket`, `isAttackSubmissionPhase`) could live in a small helper class with one-line delegators. Worth considering, not necessarily blocking.
- [ ] **`DeploymentProcessor.stageLoads` catches only `IllegalArgumentException`** - if `loader.load()` throws any other `RuntimeException`, the rollback is skipped and staged state leaks (the server survives via the outer guard in `Server.handle`, but entities can be left half-loaded). Consider `catch (RuntimeException)` for the rollback path.
- [ ] **`FighterSquadron.cancelLoad` partial-match edge case** - when only some fighters of a merged squadron are found (`success` ends up `false`), the found ones are still removed but weapon groups/skills are never recomputed, leaving the squadron inconsistent. Theoretical today (cancel always follows a successful staged load), but a comment or an all-or-nothing check would make it safe.
- [ ] **`SBFMovementProcessor.rebuildPath` rebuilds every step as `SurfaceSBFMoveStep`** - fine today since that's the only step type, but this will silently convert any future step type (e.g. aero). A comment noting the assumption would help.

## Minor / take or leave

- [ ] `SBFToHitData`: `effectiveRange.get().name().toLowerCase()` - consider `Locale.ROOT` to avoid locale-sensitive case mapping (and the report string is built from the enum name rather than a message resource, though that matches the file's existing style).
- [ ] `ServerPacketPumpTest` binds real server sockets per test - works, but slightly heavyweight; noting in case CI ever runs sandboxed.
- [ ] `ServerPacketPumpTest`: `values.get(values.size() - 1)` can be `values.getLast()` (Java 21).

Note: the earlier CodeQL comment about a missing `@Override` on `Entity.cancelLoad` is already resolved (verified present at `Entity.java:10122`).

PR #8657, opened 2026-08-05
Ask for the called blow when declaring a club attack instead of on target selection (Fixes #8636)

### LINKED ISSUE #8636: [Issue] [51.0] Hatchet-type aimed attack window pops up even when hatchet attack not possible
### Brief Description *

During the melee phase, hatchet-equipped units trigger the hatchet-type punch/kick table selection whenever they select an adjacent enemy, before even selecting an attack mode, and even when using the hatchet is not possible. The window correctly does not allow using the hatchet or the aiming, and so is not usable - you have to close it to do something else, like kicking.

This is a minor annoyance, and easy to work around. It may reveal an underlying issue, though?

### 3. Steps to Reproduce

Use the provided save. Click on an enemy with the Axman active.

### Attach Files

[hatchet.sav.gz](https://github.com/user-attachments/files/30635076/hatchet.sav.gz)

### Severity *

### MegaMek Suite Version *

51.0

### Operating System *

w10

### Java Version *

21

### Final Verification

HammerGS:
## What this changes for players

In the physical phase, clicking an enemy while piloting a hatchet/sword/mace unit no longer pops the punch/kick "Call blow?" window. Selecting a target is now always silent - the buttons just enable or disable. The called-blow question is asked at the moment it matters instead: when you actually declare the club attack (the Club button or the right-click menu), a compact dialog offers the three real options with their to-hit numbers, e.g. for an Axman swinging at an adjacent Mek:

> Aim the blow from the Hatchet:
> - No called blow (full body) - To Hit: 5  *(pre-selected)*
> - Aim high (Punch table, +4) - To Hit: 9
> - Aim low (Kick table, +4) - To Hit: 9

Cancel aborts the attack. The dialog is resizable and remembers its size and position between uses. When the choice cannot apply - the weapon can't aim its blows, the attack is impossible, attacker/target aren't Meks at equal elevation, or the TacOps `clubs_punch` option is on (its elevation-based tables override any called blow) - the question is skipped entirely and each skip reason is logged at DEBUG with a `[CalledBlow]` tag.

## Root Cause

`PhysicalDisplay.target()` opened the aimed-shot window on every target selection, and its `canAim` gate was set purely from the club's *type flags* - it never checked whether the club attack was actually possible. In the reporter's save the adjacent Rifleman is prone, and per BMM a prone Mek can only be clubbed from one level below - so the window popped for an attack that was impossible (the exact complaint in #8636). The window's selection also lived in sticky handler state, so a table picked once silently applied +4 to later club attacks declared from the right-click menu.

## Changes

1. `PhysicalDisplay.target()` - no longer opens any dialog.
2. `PhysicalDisplay.club(MiscMounted)` - new `chooseCalledBlowTable()` asks the called-blow question at declaration time and threads the chosen table through the confirm dialog, zweihander to-hit, and both Melee Master attacks. Guard clauses skip the question (with a logged reason) whenever the choice cannot apply.
3. Deleted `PhysicalDisplay`'s private inner `AimedShotHandler` (the modeless dialog machinery and its sticky aim state). `FiringDisplay`'s standalone `AimedShotHandler` class is a different class and is untouched.
4. New `CalledBlowDialog` extending `AbstractButtonDialog` - radio buttons with live to-hit numbers, resizable, size/position persisted via the existing `JWindowPreference` mechanism.
5. Replaced the hardcoded `{"punch", "kick"}` strings with i18n'd `PhysicalDisplay.CalledBlowDialog.*` keys; removed the orphaned `AimedShotDialog` keys from the English, Spanish and Russian files.
6. Renamed abbreviated locals in the touched method (`en` -> `attacker`, `clubDmg` -> `clubDamage`, `dmgString` -> `damageString`) and replaced a check-then-cast with an `instanceof` pattern binding.

## Files Changed

- `megamek/src/megamek/client/ui/panels/phaseDisplay/PhysicalDisplay.java` - dialog lifecycle moved from target selection to attack declaration; inner handler deleted
- `megamek/src/megamek/client/ui/dialogs/phaseDisplay/CalledBlowDialog.java` - new compact, resizable, bounds-persisting chooser
- `megamek/resources/megamek/client/messages.properties` - new CalledBlowDialog keys, orphaned AimedShotDialog keys removed
- `megamek/resources/megamek/client/messages_es.properties`, `messages_ru.properties` - orphaned keys removed

## Testing

Manual (UI-only change), using the reporter's save from #8636 plus edited variants covering each configuration:

- Reporter's save (adjacent Rifleman **prone**, hatchet impossible): clicking the enemy shows no popup and Club is disabled - the reported bug, confirmed fixed in-game.
- Variant with the Rifleman standing and `clubs_punch` **off**: Club opens the chooser showing To Hit 5 uncalled / 9 called (+4 applying correctly); dialog size and position confirmed persisting across uses.
- Variant with `clubs_punch` **on**: chooser correctly suppressed, normal confirm only.
- Compile, test-compile and checkstyle green; megamek.log clean of feature-related errors.

Fixes #8636


megamek/src/megamek/client/ui/dialogs/phaseDisplay/CalledBlowDialog.java
```diff
@@ -0,0 +1,112 @@
...
+        this.choices = choices.clone();
+        initialize();
+        setTitle(Messages.getString("PhysicalDisplay.CalledBlowDialog.title", weaponName));
+    }
+
+    @Override
+    protected Container createCenterPane() {
+        JPanel panel = new JPanel();
+        panel.setLayout(new BoxLayout(panel, BoxLayout.PAGE_AXIS));
+        panel.setBorder(new EmptyBorder(10, 15, 10, 15));
```
IllianiBird: Needs to use scaleForGUI

IllianiBird (approved):

megamek/src/megamek/client/ui/dialogs/phaseDisplay/CalledBlowDialog.java
HammerGS: Fixed in ce63b31c6c - the border and strut padding now go through UIUtil.scaleForGUI (stored in named locals since the values repeat).

PR #8659, opened 2026-08-06
Fix water pricing and add Attack/Defend combat posture for CASPAR (Fixes #8648)

### LINKED ISSUE #8648: Bots stop in deep water and are reluctant to cross rivers
## Problem

Bots handle rivers badly in two separate ways.

They **stop in the middle of the water and stay there**. A unit that halts in a river is immobile, cannot fight well, and blocks the crossing point for everyone behind it. It is the worst place on the board to stand still.

They are also **very reluctant to cross at all**, so a force that has to cross a river to reach the enemy can spend the game on its own bank.

Both live in `PathRanker` / `BasicPathRanker`, which every bot shares, so Princess and CASPAR both inherit them.

## Example

Map `32x34 WWE2019 ARequiemForABlueStar`, which has a north-south river running down columns 2-13. Two 12-unit companies, one deployed on each bank.

A **Dragonfly (Viper) D** - 12 movement points, 100% armour, never damaged - moved into depth 2 water at (11, 6) on round 5 and was still standing in the same hex on round 12. Eight rounds. It spent **zero** of its twelve movement points every single turn. Its armour never changed, so it was neither shooting nor being shot at.

The ranking was rewarding it for staying:

| Option | Utility |
|---|---|
| Stand still, in the water | -29 to -122 |
| Best dry hex, one hex away | -175 to -209 |

Getting out of the river scored 62 to 167 **worse** than staying in it.

Counts from that one game, taken from the chosen path of every unit that was in water or had a crossing available:

| | CASPAR | Princess |
|---|---|---|
| Refused an available crossing | 58 | 17 |
| Decisions made while in water | 10 | 11 |
| Halted in the water | 0 | 5 |
| Mean margin against crossing | 39.8 | 104.6 |
| Worst margin against crossing | 153.3 | 233.3 |

## Why it happens

Depth 2 water gives partial cover, so the ranker reads the river as a good fighting position. In practice a crossing point is the one piece of ground the enemy expects you to use. Real armoured units cross without stopping, because a halted vehicle in a gap is a stationary target that blocks the route behind it.

The bot banks the cover the water gives and ignores what it costs: no running, piloting rolls to move, weapons underwater, and no way to withdraw quickly.

The reluctance has a different cause. The bot weighs the hazard of entering the water against nothing at all on the far side. A real force crosses because there is an objective over there worth the risk. Nothing in the ranking says the far bank is worth reaching.

## Proposed fix

In rough priority order.

- **Do not end a move in deep water when there is a way out.** Penalty scaled by water depth and by how much movement the unit did not use. A unit with 12 movement points ending in depth 2 having spent none of them is the case to kill outright.
- **Stop treating deep water as a fighting position.** Offset the cover it gives against what it costs in movement and weapons.
- **Give the far bank a value.** Hexes just beyond the exit should be worth something, so crossing has a purpose and not only a cost.
- **Use the near bank as a firing line.** Units waiting to cross should be covering the far bank. The set-friend cover bonus in `MutualSupportPathRanker` already does this shape of thing and has been measured as inert; a crossing is the case it was written for.

The first two are the ones the example above proves. The last two are reasoning from how crossings actually work, and should be measured before being kept.

## Implementation notes

- Both problems are in the shared ranker, so the fix lands for every bot at once.
- Follow the usual approach for this area: land it behind CASPAR first, measure against Princess as the unchanged control, then backport to Princess once the numbers hold.
- `PathRanker` already has water handling from #7627 - `WATER_FALL_FORGIVENESS_DEPTH_1`, `WATER_FALL_FORGIVENESS_PER_DEPTH`, `WATER_FALL_FORGIVENESS_FLOOR`. That work stopped Princess pacing at the bank instead of wading a shallow ford. This issue is the next layer: what happens once she is in the water, and whether she values what is on the other side.
- East-west river crossing scenarios exist for measuring this, with the two companies deployed on opposite banks so the river falls between them rather than behind the meeting point.
- The data above came from temporary instrumentation in `PathRanker.getBestPath` that logs the chosen path whenever a unit is in water or has a crossing available, along with the best alternative it passed over. Worth re-adding when this is picked up.

## Background

Found in manual play on 2026-08-05 while testing formation behaviour on an east-west map. Deliberately kept out of the Mutual Support work, which is about formations and should not also carry water rules.

Related: #7627.

HammerGS:
## Summary

The bots misplay water. A submerged unit is scored as if all its guns still work, and standing still in water costs nothing, so damaged units sat in lakes as free bunkers. The bot also has no idea whether it is attacking or defending, so it prices a river the same from both sides.

This PR fixes the water pricing for both bots and adds a combat posture - Attack, Defend, or Auto - that only CASPAR acts on. Princess is unchanged and stays the control arm. About a third of the diff is tests: 5 files, ~500 lines.

## Changes Made

*Water pricing (both bots)*

- The to-hit estimate had no water logic at all - a decade-old `todo test water` marks the spot. `UnderwaterFire` predicts which weapons are underwater at a considered position and applies the same restrictions and range tables the server enforces (TW p.107-109). A submerged unit stops counting guns it cannot fire, and dry units stop shooting at targets they cannot hit.
- Hazards were priced per path step, so a stationary path paid nothing for the water it stood in. Standing in water now pays breach risk. A sealed, undamaged Mek still pays nothing, which is correct.

*Combat posture (CASPAR only)*

- `CombatPosture` is a `BehaviorSettings` field with XML round-trip. A defending unit is charged a full turn of advance for entering or crossing water. Fords count: a river the enemy can cross anywhere is still the line. Withdrawing units, units with a destination or waypoint, and units already in the water pay nothing.
- "Same side" means walkable connectivity (`BankRegions`, one flood fill per round), not a straight line. The line test charged defenders for moving along their own bank wherever the river bent, and pushed them out of their firing positions.
- Auto resolves once a round. A destination edge means attack. An enemy closing at 0.5 hexes a round means defend. Leaving defend takes the advance actually stopping (0.1 or less), so the stance does not flip-flop during a mutual approach. The bot announces posture changes in chat with its reason.
- Set it in the lobby bot config, with `/posture` in chat, in the Bot Commands Panel, or with a scenario `posture:` key. Every ranked path logs its posture and charge as TSV columns.

## Files Changed

*New source*

- `megamek/src/megamek/client/bot/princess/UnderwaterFire.java` - NEW: underwater fire prediction
- `megamek/src/megamek/client/bot/princess/CombatPosture.java` - NEW: the posture enum
- `megamek/src/megamek/client/bot/princess/PostureResolver.java` - NEW: per-round Auto resolution
- `megamek/src/megamek/client/bot/princess/BankRegions.java` - NEW: walkable-bank labeling
- `megamek/src/megamek/client/bot/princess/commands/PostureCommand.java` - NEW: `/posture`

*Main edits*

- `FireControl.java` - water check wired into the to-hit estimate
- `BasicPathRanker.java` - stationary water pricing
- `MutualSupportPathRanker.java` - the posture charge, chat announcements, TSV columns
- `BehaviorSettings.java`, `PrincessSettingsBuilder.java`, `ChatCommands.java` - posture plumbing
- `BotConfigDialog.java`, `BotCommandsPanel.java`, two `messages.properties` - controls

*Tests* - 5 files, ~500 lines

## Testing

- The underwater-fire test was written first and failed before the wiring. The ford, bank-connectivity and hysteresis tests each pin a failure seen in a real game before its fix.
- Full suite: 14,830 tests, one failure. `SettingsContentHostTest.widePageKeepsCappedViewport...` fails on unmodified main too when a local `mmconf/clientsettings.xml` is present. Not this branch.
- 30-game headless arms on the river map:
  - Attack: water entries halved against the pre-fix run (337 to 161), worst stay down from 30 rounds to 23 - and the 23 is an ejected pilot swimming, not a Mek.
  - Defend: 93% of unit-games end on the defender's own bank. The defender keeps 49% BV to the attacker's 40%.
  - Auto: flips both ways on real closing rates. The first run exposed the flip-flop the hysteresis commit fixes.
  - Defend against defend (two CASPARs): no stalemate on this map - both banks are in gunnery range and these games ended fastest of all four arms.
- Manual games: an attack game crossed cleanly and won. The first defend game exposed the ford leak, the second held the bank in firing positions.

## Rules clarification

| Ruling | Status |
|---|---|
| Underwater weapons only hit targets in water; most cannot fire underwater; the rest use shortened ranges (TW p.107-109) | Already enforced by the server. **Fixed here** for the bot's own estimate, which ignored all of it. |

## Out of scope

- Fire and magma have the same stationary-path gap; only water is priced here. Issue to follow.
- Terrain positioning (high ground, partial cover, woodline edges) - next PR.
- Mission tone over Auto and crossing-site selection - the PR after.
- MekHQ carries the posture in a separate small PR that depends on this one.


megamek/src/megamek/client/bot/princess/UnderwaterFire.java
```diff
@@ -0,0 +1,238 @@
...
+        }
+
+        int partialWaterDepth = ((shooter instanceof Mek) && shooter.isSuperHeavy()) ? 2 : 1;
+        if ((shooter instanceof Mek) && !shooterState.isProne() && (depth <= partialWaterDepth)) {
+            return isLegLocation(shooter, weapon.getLocation());
+        }
+
+        // Fully in the water: every location is wet unless the unit's top still clears the surface.
+        int topElevation = shooterState.getElevation() + (shooterState.isProne() ? 0 : shooter.height());
+        return topElevation < 0;
```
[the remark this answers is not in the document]
HammerGS: Verified - valid, fixed in e729a6de. `Entity.height()` does read the entity's current prone state. One nuance from tracing it: the cited stand-up case (prone Mek, depth 1) was already handled, because the partial-water branch reads the hypothetical pose and catches a standing Mek in shallow water before height is consulted. The full-submersion branch did read the current height though, and misclassified standing poses on raised footing in deep water. Height is now computed from the pose being evaluated (`hypotheticalHeight`), and a test pins it: a mock whose current height is 0 evaluates a standing pose at elevation -1 in depth 3 as dry.

megamek/src/megamek/client/bot/princess/MutualSupportPathRanker.java
```diff
@@ -311,12 +333,122 @@ protected double calculateMutualSupportMod(Coords friendsCoords, MovePath path)
...
+     * the answer changes - a flip of the auto-resolution or a new explicit order taking effect - the bot says
+     * so in the chat, with its reason, so an observer can follow the force's intent without reading logs.
+     */
+    private CombatPosture resolvePosture(Game game) {
+        int round = game.getCurrentRound();
+        if (round != postureResolvedRound) {
+            postureResolvedRound = round;
+            posture = postureResolver.resolve(getOwner().getBehaviorSettings(), round,
+                  deployedPositions(getOwner().getEntitiesOwned()),
+                  deployedPositions(getOwner().getEnemyEntities()));
```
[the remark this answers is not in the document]
HammerGS: Verified - valid, fixed in e729a6de. `getEnemyEntities()` is game-wide, so a multi-board game would have blended boards into a meaningless closing rate. Posture is now resolved per board per round from only that board's deployed units (`deployedPositions` filters on `getBoardId()`), cached per board, with a separate resolver distance history per board so one board's approach cannot flip another board's stance. The crossing charge already keyed off the moving unit's own board. A test pins the filter: a unit on another board has no say in this board's posture.

megamek/src/megamek/client/bot/princess/BankRegions.java
```diff
@@ -0,0 +1,150 @@
...
+     */
+    static BankRegions of(Board board, int minimumDepth) {
+        int width = board.getWidth();
+        int height = board.getHeight();
+        int[] regions = new int[width * height];
+        java.util.Arrays.fill(regions, WATER);
+
+        int nextRegion = 0;
+        Deque<Coords> frontier = new ArrayDeque<>();
+        for (int x = 0; x < width; x++) {
```
IllianiBird, line 91: So this is o(n) which I can see being expensive. Is there any way we could cache the values at game start? Boards don't really change mid-scenario, so no need to re-evaluate every time we check. Would be far better to have a map with coords as the key and a boolean (isBankRegion) as the value.

Better yet, if you expect to have multiple of these special hex values (bank, road, I dunno, burger shack) you could create a HexProperties record. As you add new special properties they can all be stored in that record.

The map would then be keyed to the cords with a HexProperties record as the value. That would turn per-instance checks from n(O) to `1` as all you'd need to do is: `hexPropertiesMap.get(COORD).bank()`

megamek/src/megamek/client/bot/princess/BasicPathRanker.java
```diff
@@ -2448,6 +2463,26 @@ private double calcIceHazard(Entity movingUnit, Hex hex, MoveStep step, MovePath
...
+     * @param elevation  the unit's elevation in the hex
+     * @param prone      whether the unit is prone
+     * @param endsInHex  whether the unit ends its turn in this hex
+     * @param movePath   the path entering the hex, or {@code null} for a unit that is standing still. A unit
+     *                   that does not move makes no water-entry roll, so it runs no fall-contingent breach
+     *                   risk.
+     *
+     * @return the hazard value for being in this water hex in that pose
+     */
+    private double waterHazard(Entity movingUnit, Hex hex, int elevation, boolean prone, boolean endsInHex,
```
IllianiBird, line 2484: I wonder how this logic needs to be changed for Core?

megamek/src/megamek/client/bot/princess/BehaviorSettings.java
IllianiBird: Mind changing this (and the handful of other instances in this class) to Math.clamp()? Saves CodeQL whining about it constantly.

IllianiBird (approved):
Approved, but there are a couple areas I think we can improve. Especially the `o(n)` processing when evaluating banks.

megamek/src/megamek/client/bot/princess/BasicPathRanker.java
HammerGS: I've left this for now, but I suspect in general Princess will pick up the changes.

megamek/src/megamek/client/bot/princess/BehaviorSettings.java
HammerGS: Done in 8afb285 - all three clamp sites in the class (`validateIndex`, `setNumberOfEnemiesToConsiderFacing`, `setAllowFacingTolerance`) now use `Math.clamp`, and the `MathUtility` import is gone. Also added to our pre-PR checklist so it stops arriving in review.

megamek/src/megamek/client/bot/princess/BankRegions.java
HammerGS: The per-check cost is already what you're describing: `regionOf()` is one flat-array index - the map-with-coords-key structure, just cheaper than a HashMap. The O(n) flood fill runs once per round per board (not per check), which is ~1,100 cells on this map, well under a millisecond - and it replaced a straight-line hex walk that ran on *every ranked path*, so the aggregate cost went down substantially with the change.

I kept per-round rather than per-game invalidation deliberately: it costs nothing measurable and means any mechanic that ever changes water mid-game can't leave the labels stale. Happy to move it to game-start if you'd rather - it's a one-line change - but the round recompute is noise next to ranking a single path.

The HexProperties record is a good design and I'd rather build it where it earns its keep: the terrain-positioning PR that follows this one needs several per-hex properties (cover, woodline edges, elevation lines), and that's where a single precomputed per-hex record beats one-off structures like this. Noted in the plan for that PR.

megamek/src/megamek/client/bot/princess/BasicPathRanker.java
HammerGS: Agreed with your follow-up - the stationary-water pricing lives entirely inside BasicPathRanker's hazard pass and reads only the hex and the unit's pose, so it rides along wherever the ranker goes. Nothing in it binds to the current package layout.

HammerGS:
Need to merged and had passed prior checks

PR #8672, opened 2026-08-07
Added MegaMek Support for MekHQ #9744

IllianiBird:
# Required by https://github.com/MegaMek/mekhq/pull/9744


HammerGS:
Chased down the "units aren't being picked up" problem on MegaMek/mekhq#9744. The date fields added here are fine - `Entity.getPrototypeDate()` and friends delegate to `compositeTechLevel` and populate correctly, and `getAllMeks()` blocks until the cache is initialised so there's no load-order race with the MekHQ side.

The break is in how the MekHQ side filters by tech base, but the clean fix needs one more field here.

## The gap

`MekSummary.getTechBase()` holds the **localized display string** - it's set at `MekSummaryCache.java:952` from `e.getTechBaseDescription()`, which builds "Inner Sphere" / "Clan" / "Mixed Inner Sphere" / "Mixed Clan" out of `messages.properties`. That's fine for display, but it can't be used for filtering: it's locale-dependent, and Mixed is two separate strings.

`MekSummary` already carries `isClan()` (set at `MekSummaryCache.java:886`), so the only thing missing to identify a tech base without going through prose is a mixed-tech flag. Since this PR is already adding fields to `MekSummary`, it seems like the natural home for it:

```java
private boolean mixedTech;

public boolean isMixedTech() {
    return mixedTech;
}

public void setMixedTech(boolean mixedTech) {
    this.mixedTech = mixedTech;
}
```

and in `MekSummaryCache.getSummary()`, alongside the existing `ms.setClan(e.isClan())`:

```java
ms.setMixedTech(e.isMixedTech());
```

That lets MekHQ resolve `isMixedTech()` / `isClan()` into a real `TechBase` instead of comparing display strings. Full detail on the MekHQ side is over on MegaMek/mekhq#9744.

## Testing gotcha

Worth flagging because it'll bite during testing: `serialVersionUID` is now pinned at `2L`, so adding `mixedTech` will **not** invalidate a `units.cache` built from the current state of this branch. Java will happily deserialize it and default the new field to `false`, making every unit look non-mixed. Deleting `units.cache` after adding the field avoids chasing a ghost.

(That pinning is the right call, to be clear - it just means field additions now need a manual cache clear during development rather than healing themselves.)

VicenteCartas (approved):

PR #8680, opened 2026-08-10
Fix test that fails if the GUI scaling is not exactly 1.0

Sleet01:
This test will fail for any dev that changes their GUI scaling, or if any other test changes the scaling option.
Fix by explicitly scaling the expected width by the current GUI scale value.


IllianiBird (approved):
headscratch

PR #8694, opened 2026-08-12
fix: disallow loading of superheavy meks into standard mekbays (MM-8686)

NickAragua:
What it says in the description.


IllianiBird (approved):

NickAragua:
Unit tests pass locally, so merging anyway as the error isn't a test failure.

PR #8702, opened 2026-08-13
Fixes #8693 - Possible places a bombast laser charge does not reset

### LINKED ISSUE #8693: [Issue] [CORE Nightly] Bombast laser not discharging on a miss
### Brief Description *

With the new implementation, the bombast laser is not switching from "charged" to "discharged" when a shot is missed in the charged state. Upon a miss, the weapon will remain charged the following turn and can make another attack at any selected damage level with no to-hit penalties. If the shot hits, the weapon discharges correctly.   

### 3. Steps to Reproduce

1. Select a unit with a bombast laser 
2. During the Ranged Attack Phase, select the bombast laser and hit "Charge Laser" button.
3. Observe in the following phase, the weapon's status in the unit display changes to "Charged" and no penalties are added to the attack calculation for firing at 12 or 16 damage.
4. Fire the charged weapon at a target, but miss.
5. Observe in the following phase that the bombast laser still says "Charged" in the unit display, and no penalties are added to the attack calculation for firing at 12 or 16 damage despite the weapon being fired the previous turn.
6. Fire the charged weapon at a target, but hit.
7. Observe in the following phase that the bombast laser says "No Charge" in the unit display, and there are penalties added to the attack calculation for firing at 12 or 16 damage.

### Attach Files

Image
Image
Image
Image

### Severity *

### MegaMek Suite Version *

v0.51.01-nightly-2026-08-12

### Operating System *

Windows 11 10.0 (amd64)

### Java Version *

Adoptium 21.0.10

### Final Verification

TenkawaBC:
Fixing https://github.com/MegaMek/megamek/issues/8693

There are two early exits in the weapons handler. I have added the bombast laser check and reset code to each early exit, in case it is going down those paths.


HammerGS:
I traced this on `main` to work out why the charge survives a miss. Short answer: the discharge code is at the very bottom of `WeaponHandler.handle()`, and a normal miss returns out of that method about 165 lines before it gets there. A hit walks all the way to the bottom, which is why a hit discharges correctly.

## The code path

A Warhammer charges its Bombast Laser, then next turn shoots at a Griffin standing in an open field and misses.

`handle()` reaches the miss branch at line 1097 and calls `handleSpecialMiss(...)`. For this shot that returns `false` — it only returns `true` when the target is inside a building and the attacker is adjacent, so the missed shot damages the building instead (`WeaponHandler.java:552`). The shot loop variable `i` is `1` for any weapon that isn't a two-roll Ultra AC, so `i < 2` is true as well. Both halves true means:

```java
if (!handleSpecialMiss(entityTarget, bldgDamagedOnMiss, bldg, vPhaseReport) && (i < 2)) {
    returnedReports.addAll(vPhaseReport);
    return false;          // <- WeaponHandler.java:1112, method exits here
}
```

The discharge block lives at `WeaponHandler.java:1276-1279`, after the shot loop closes. On a hit `bMissed` is false, that whole branch is skipped, execution falls out of the loop, and the charge is dropped. On a miss it is never reached. That is the entire hit/miss asymmetry.

## This is not a server/client sync problem

Worth stating explicitly, since it came up: the server pushes the full entity list to every client twice a round via `entityAllUpdate()`, once at `FIRING_REPORT` and once in the `END` phase (`TWPhasePreparationManager.java:259` and `:273`). `chargeState` is an ordinary field on `Mounted`, so it rides along in that packet. Once the server's copy is right, the unit display and the to-hit numbers follow on their own. The loop is already closed — the server simply never changed its own value.

## On this PR

The second insertion here, the one just above the `handleSpecialMiss` call, **is in the right place** for the case in #8693. That path is reachable and unconditional for a laser missing a target in the open. If it still reproduces after this patch, I would suspect the build before the logic — dropping a `LOGGER.debug` into that block and confirming it appears in `megamek.log` would settle it quickly.

I would change the shape of the fix, though, for the reason Copilot raised. Three copies of the same block in a method with four exits will rot, and `WeaponHandler` is shared by every weapon in the game — it should not know what a Bombast Laser is at all.

Suggestion: move the whole thing into `BombastLaserWeaponHandler` and wrap the parent.

```java
@Override
public boolean handle(GamePhase phase, Vector<Report> returnedReports) {
    try {
        return super.handle(phase, returnedReports);
    } finally {
        dischargeIfCharged();
    }
}

/**
 * Drops the charge a Bombast Laser was holding once it has been fired. The charge is spent whether the shot
 * hits or misses, so this has to run on every exit from {@link WeaponHandler#handle(GamePhase, Vector)} -
 * a normal miss returns early and never reaches the end of that method.
 */
private void dischargeIfCharged() {
    if (weapon.getChargeState() != ChargeLevel.CHARGED) {
        LOGGER.debug("[Bombast] {}: nothing to discharge after firing, charge state is {}",
              attackingEntity.getShortName(), weapon.getChargeState());
        return;
    }
    weapon.setChargeState(ChargeLevel.CHARGE_NONE);
    LOGGER.debug("[Bombast] {}: charge spent, laser discharged", attackingEntity.getShortName());
}
```

All three `F_BOMBAST_LASER` blocks then come out of `WeaponHandler`. This covers every early exit in the parent — the impossible-to-hit exit at line 981, the `specialResolution` exit at 1094, the miss exit at 1112, and the normal path — without touching shared code, and it survives whatever early return someone adds to `handle()` next year.

One alternative I looked at and would not take: overriding `setDone()`, which is tempting because it runs exactly once from the constructor via `useAmmo()` and so cannot be missed. But the constructor runs at declaration time, and a charged Bombast Laser is explosive until it fires (`ISBombastLaser.isExplosive`, line 154). Discharging at declaration would change whether it cooks off when an enemy crits it earlier in the same firing phase. Resolution time is the correct moment.

## Testing

`EnergyWeaponHandlerRiscEdgeTest` in `megamek/unittests/megamek/common/weapons/handlers/` is the closest existing pattern — it drives an energy handler with a controlled roll, which is what you need to force a miss and then assert `getChargeState()` comes back `CHARGE_NONE`.

## What is not proven here

This is read from the code on `main`; I have not built this branch or run a game. The claim that the discharge block is unreachable on a miss is solid — the exit at line 1112 and the return value of `handleSpecialMiss` are both plainly there. The claim that this PR as written already fixes the reported case is an inference from those same lines, and it contradicts what you are seeing in testing, so that part deserves the debug-log check before anyone leans on it.

TenkawaBC:
updated this fix to use the better method of using the bombast handler. Also used a check for isFired in the discharge, to confirm it was actually fired.

TenkawaBC:
Used the AI code as a bit of a template on this, then tested and updated. The return code it suggested didn't work properly, but I was able to do it with similar code, and tested it working with a pair of grasshoppers mounting 2 bombast lasers each.

Each charged both, each shot one of them, one hit and one missed. Both fired Bombast lasers reset to not charged, the other two remained charged. Fixed.

IllianiBird (approved):

PR #8710, opened 2026-08-14
Add an optional build note under the version on the main menu

HammerGS:
## Summary

Open MegaMek today and the start-up screen tells you the version and nothing else. There is no way to
say "this one is the Core Rules release" or "this is a development build, don't take it to a
tournament" without changing code and rebuilding.

This adds one optional line of text under the version number, controlled from `Version.properties`.

Here is the whole of it. Put these three lines in `megamek/resources/Version.properties`:

```properties
notes=Core Rules Release
notesColor=Red
notesFontSize=14
```

Start MegaMek, and "Core Rules Release" appears in red, just under "MegaMek v0.51.01" and just above
the "Start a New Game" button. Comment the three lines out and the start-up screen goes back to
exactly what it looks like now.

**This PR ships that note switched on**, labelling the build as the Core Rules Release. That is
deliberate, not a leftover test value.

The colour is a plain name: red, orange, yellow, green, blue, cyan, magenta, pink, white, black, gray,
lightgray or darkgray. Capitals and the "grey" spelling are both fine. The size is a point size
between 6 and 72. Leave either one out and the note uses MegaMek's usual warning colour at the usual
menu size.

## Changes

1. **Three new optional entries in `Version.properties`** - `notes`, `notesColor`, `notesFontSize` -
   each documented in the file itself, next to the existing `revision` entry.
2. **`Version` reads them.** The note describes the build, not any particular version number, so it
   stays out of a version's identity: it does not appear in `toString()`, it plays no part in
   comparing or equalling two versions, and it is never sent to another player. Everything added here
   is `static`, so nothing changes about how a `Version` travels in a packet or a save game.
3. **`MegaMekGUI` puts it on the screen**, working out the colour and size and placing the note under
   the version label.
4. **A typo cannot break the start-up screen.** Write `notesColor=puce` or `notesFontSize=200` and the
   note falls back to the default and writes a line in `megamek.log` naming what was wrong. The menu
   still comes up.

### Two things that did not work, in case they look odd in review

**The note could not simply be given its own row in the menu.** That was the obvious approach and it
looked wrong: the note ended up floating in the middle of the empty space, nowhere near the version
number it belongs to. The cause is that every row in that column is set to take an equal share of the
leftover height, so a row of its own became a tall row with the text centred in it. The note and the
version label now sit together in a single row, which keeps them tucked up against each other. When
there is no note, that same code hands back the plain version label, which is why an ordinary build
comes out byte-for-byte as before.

**The obvious way to make long text wrap made the note too wide.** Asking for a 250-pixel wrap width
in the usual way produced a label 325 pixels wide - visibly wider than the buttons above and below it.
Swing quietly enlarges those width values by about a third. Sizing a table cell instead gives exactly
the width asked for; measured at 100, 250 and 400 pixels, the label came out 100, 250 and 400.

The point size is put through `UIUtil.scaleForGUI`, so a player running MegaMek scaled up or down gets
a note in proportion with the rest of the menu.

## Files Changed

- `megamek/resources/Version.properties` - the three new entries, documented, with the Core Rules
  Release note filled in
- `megamek/src/megamek/Version.java` - reads and sanity-checks them
- `megamek/src/megamek/client/ui/clientGUI/MegaMekGUI.java` - builds the note and places it under the
  version label
- `megamek/unittests/megamek/VersionTest.java` - covers an entry that is missing, blank, or padded with
  spaces

## Testing

- `VersionTest` passes, 31 of 31. `compileJava`, `compileTestJava` and checkstyle are clean.
- Played through it on this branch with current `main` merged in: the note comes up in the right place,
  all seven menu buttons still fit, and `megamek.log` has no ERROR, FATAL or null-pointer entries for
  the session.
- Filled all three entries in for real and read them back to confirm the file is being picked up
  (`notesColor=Red` came back as `Red`, `notesFontSize=14` as `14`), then set `notesFontSize=200` and
  confirmed it falls back to the normal size instead of producing a giant note.
- The wrapping was measured rather than eyeballed, which is the only reason the width problem above was
  caught.

## What is NOT proven yet

- **Only run at the default GUI scale.** The scaling call is the same one the rest of the interface
  uses, but I have not started MegaMek at 1.5x or 0.75x to look at it.
- **Only run with the default skin.** The colours are fixed values, so on a pale skin a note set to
  white or yellow would be hard to read. Nothing stops a bad choice.
- **Only run with a short note.** Longer text wraps correctly when measured, but I have not looked at a
  two- or three-line note on screen, and every extra line pushes the menu buttons further down.
- **No automated test covers the note as it is drawn.** The tests cover reading the entries out of the
  file. The colour lookup and the label itself were checked by running MegaMek and looking at it, not
  by an assertion, so a future change could move the note without failing a test.

### DIFF
--- a/megamek/resources/Version.properties
+++ b/megamek/resources/Version.properties
@@ -43,3 +43,27 @@ patch=01
 # versions the whole suite. Only set it on the release branch the point release is cut from.
 #
 #revision=1
+#
+# Optional note about this build, shown on the main menu directly under the version number.
+#
+# Leave the line below commented out for an ordinary release: the main menu then looks exactly as it
+# always has. Set it to say something about a particular build - for example that it is a development
+# build, or that it is a hotfix for a specific problem - and that text appears under the version.
+#
+# Keep it to a short sentence or two: the text wraps to the width of the menu buttons, so a long note
+# pushes the buttons down the screen. Use plain ASCII only.
+#
+notes=Core Rules Release
+#
+# Colour of the note. One of: red, orange, yellow, green, blue, cyan, magenta, pink, white, black,
+# gray, lightgray, darkgray (grey is accepted as well). Leave it out and the note uses the same
+# warning colour the rest of MegaMek uses for things the player should notice. An unrecognised name
+# falls back to that warning colour and is reported in megamek.log.
+#
+notesColor=Red
+#
+# Point size of the note, between 6 and 72. Leave it out and the note is drawn at the same size as
+# the rest of the menu. A size outside that range falls back to the menu size and is reported in
+# megamek.log. Remember that a large size makes the note taller and pushes the buttons down.
+#
+notesFontSize=14
--- a/megamek/src/megamek/Version.java
+++ b/megamek/src/megamek/Version.java
@@ -79,6 +79,42 @@ public final class Version implements Comparable<Version>, Serializable {
     /** Key of the optional revision entry in {@code Version.properties}. Absent for ordinary releases. */
     private static final String REVISION_BUNDLE_KEY = "revision";
 
+    /** Key of the optional build note entry in {@code Version.properties}. Absent for ordinary releases. */
+    private static final String NOTES_BUNDLE_KEY = "notes";
+
+    /** Key of the optional colour name the build note is drawn in. Absent unless a build note sets one. */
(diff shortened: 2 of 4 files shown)


HammerGS:
Need this for this release. Going to merge since co-pilot has no feedback.

PR #8719, opened 2026-08-15
Fixing shield checks to use F_SHIELD. 

TenkawaBC:
Correcting tests that still referenced F_CLUB for shields

### DIFF
--- a/megamek/src/megamek/client/ui/dialogs/unitDisplay/SystemPanel.java
+++ b/megamek/src/megamek/client/ui/dialogs/unitDisplay/SystemPanel.java
@@ -63,21 +63,11 @@ import megamek.client.ui.widget.picmap.PicMap;
 import megamek.common.Configuration;
 import megamek.common.CriticalSlot;
 import megamek.common.battleArmor.BattleArmor;
-import megamek.common.equipment.AmmoType;
-import megamek.common.equipment.BridgeLayerState;
-import megamek.common.equipment.EquipmentActivation;
-import megamek.common.equipment.EquipmentMode;
-import megamek.common.equipment.EquipmentType;
-import megamek.common.equipment.GunEmplacement;
-import megamek.common.equipment.MiscMounted;
-import megamek.common.equipment.MiscType;
-import megamek.common.equipment.Mounted;
-import megamek.common.equipment.WeaponType;
+import megamek.common.equipment.*;
 import megamek.common.equipment.enums.MiscTypeFlag;
 import megamek.common.game.Game;
 import megamek.common.interfaces.ILocationExposureStatus;
 import megamek.common.options.OptionsConstants;
-import megamek.common.rules.core.CoreRulesManager;
 import megamek.common.units.ConvInfantry;
 import megamek.common.units.Entity;
 import megamek.common.units.Mek;
@@ -538,7 +528,7 @@ class SystemPanel extends PicMap
                 sb.append(')');
             }
 
-            if ((m instanceof MiscMounted) && ((MiscMounted) m).getType().isShield()) {
+            if ((m instanceof MiscMounted) && ((MiscMounted) m).getType().hasFlag(MiscType.F_SHIELD)) {
                 sb.append(" ").append(((MiscMounted) m).getDamageAbsorption(en, m.getLocation())).append('/')
                       .append(((MiscMounted) m).getCurrentDamageCapacity(en, m.getLocation())).append(')');
             }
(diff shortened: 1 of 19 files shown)


megamek/src/megamek/client/ui/dialogs/unitDisplay/SystemPanel.java
```diff
@@ -538,7 +528,7 @@ private String getMountedDisplay(Mounted<?> m, int loc, CriticalSlot cs) {
                 sb.append(')');
             }
 
-            if ((m instanceof MiscMounted) && ((MiscMounted) m).getType().isShield()) {
+            if ((m instanceof MiscMounted) && ((MiscMounted) m).getType().hasFlag(MiscType.F_SHIELD)) {
```
IllianiBird, line 531: This feels like an anti-pattern. Are we going to have situations where a shield won't have F_SHIELD? Unless that's true there's no reason to lose the convenience boolean.

megamek/src/megamek/client/ui/panels/phaseDisplay/PhysicalDisplay.java
```diff
@@ -1231,24 +1233,20 @@ private boolean isAimablePhysicalWeapon(MiscMounted club) {
               MiscTypeFlag.S_MACE,
               MiscTypeFlag.S_LANCE,
               MiscTypeFlag.S_CHAIN_WHIP,
-              MiscTypeFlag.S_RETRACTABLE_BLADE,
-              MiscTypeFlag.S_SHIELD_LARGE,
-              MiscTypeFlag.S_SHIELD_MEDIUM,
-              MiscTypeFlag.S_SHIELD_SMALL);
```
IllianiBird, line 1237: Can I ask you to confirm that shield bashes are no longer amiable

megamek/src/megamek/common/battleValue/BVCalculator.java
IllianiBird, line 509: add `miscType` to the end of this line, then on line 526 you can just call `miscType.isShield()`

megamek/src/megamek/common/equipment/MiscType.java
```diff
@@ -395,13 +396,7 @@ public int getBaseDamageCapacity() {
     public boolean isBoobyTrap() {
         return hasFlag(F_BOOBY_TRAP);
     }
-
-    public boolean isShield() {
```
IllianiBird, line 399: I feel like we want to keep this boolean

megamek/src/megamek/common/rules/core/CoreRulesPhysical.java
```diff
@@ -74,7 +74,7 @@ public int getShieldDamageBoost(Entity entity, int armLoc) {
 
                 Mounted<?> m = cs.getMount();
                 EquipmentType type = m.getType();
-                if ((type instanceof MiscType) && ((MiscType) type).isShield()) {
```
IllianiBird, line 77: Change to `if (type instanceof MiscType miscType && miscType.isShield()) {`

IllianiBird (changes requested):
My recommendation is to keep `isShield`, but to add it to `ITechnology` In Technology we add `isShield` and have it default to `false`, then MiscType implements isShield with the flag test. This will remove a bunch of casts as well as being far cleaner.

If we know we're going to be frequently testing for a specific flag/s then we want to include utility methods like this to avoid boilerplate. As it is the sheer volume of `(MiscMounted) miscMounted).getType().hasFlag(MiscType.F_SHIELD)` feels like an anti pattern.

megamek/src/megamek/client/ui/panels/phaseDisplay/PhysicalDisplay.java
TenkawaBC: It was replaced in core with a damage increase to punches with that arm. (Similar to claws)

Shield bashes are no longer a separate physical attack.

megamek/src/megamek/common/rules/core/CoreRulesPhysical.java
TenkawaBC: change in latest commit

megamek/src/megamek/common/equipment/MiscType.java
TenkawaBC: it adds extra calls and not a lot used it. it was a hack for shields not having a flag of their own

megamek/src/megamek/common/battleValue/BVCalculator.java
TenkawaBC: no need on this.

IllianiBird (approved):

Issue #8721, opened 2026-08-15
Client never finishes shutting down after a local save is refused in a double-blind game


HammerGS:
## Brief Description

In a double-blind game with **Disable local saves when using double blind** turned on, pressing `File > Save Game` leaves the client unable to quit. The window will not close, and the only way out is to kill the process from Task Manager.

Worked example:

1. Host a game. In Game Options, enable **Double Blind** and **Disable local saves when using double blind**.
2. Start the game with any two units.
3. Press `File > Save Game`, pick a filename, and confirm.
4. The chat prints `Local Save only outside double blind games.` and no save is written. So far this is correct and intended.
5. Now try to quit, or `File > Load Game`, or quick-load. The client never shuts down.

The refusal in step 4 is deliberate and not the bug. The bug is that the client is never told the save was refused, so it waits for it forever.

## Root cause

Saving is asynchronous. `ClientGUI.saveGame()` sends `/localsave` and sets `client.setAwaitingSave(true)`, expecting the server to serialize the game and stream it back as a `SEND_SAVEGAME` packet.

`LocalSaveGameCommand.run()` refuses the save by sending a chat message and returning, without sending any packet:

```java
if (server.getGame().getOptions().booleanOption(OptionsConstants.ADVANCED_DOUBLE_BLIND) &&
      server.getGame().getOptions().booleanOption(OptionsConstants.BASE_DISABLE_LOCAL_SAVE)) {
    server.sendServerChat("Local Save only outside double blind games.");
} else {
    ...
    server.sendSaveGame(connId, fileName, localPath);
}
```

`awaitingSave` is only ever cleared inside the `SEND_SAVEGAME` handler (`Client.java:1349` and `Client.java:1363`), so when no such packet arrives it stays `true` for the rest of the session.

`ClientGUI.die()` then never completes, because it defers itself for as long as a save is outstanding:

```java
@Override
public void die() {
    if (client.isAwaitingSave()) {
        SwingUtilities.invokeLater(this::die);
        return;
    }
    ...
}
```

Each call re-posts itself, so shutdown spins on the event dispatch thread indefinitely. This affects every path that goes through `die()`: quitting, `File > Load Game`, and quick-load.

## Suggested fix

Have the server tell the client the request was refused rather than staying silent, so the client can clear `awaitingSave`. Failing that, `die()` needs a bound so a never-answered save cannot block shutdown forever.

## Steps to Reproduce

See the worked example above.

## Severity

[...]

Issue #8718, opened 2026-08-15
[Issue] 0.51.00 Pilots equipped with Enhanced Imaging rolling for damage even when EI turned off.


Ursur1minor:
### Brief Description *

Enhanced Imaging has an effect similar to VDNI that whenever they take internal damage the pilot must make a PSR to not take a hit.
Except Enhanced Imaging has the benefit of being able to turn itself off, which disables most effects, including the risk for damage.
Currently this does not appear to be working in MegaMek as a pilot with Enhanced Imaging turned off still has to roll for damage upon any internal damage suffered as seen in the following image.

<img width="1457" height="726" alt="Image" src="https://github.com/user-attachments/assets/8ab9c5be-7c36-4407-bcaf-42f898f517e9" />

<img width="334" height="115" alt="Image" src="https://github.com/user-attachments/assets/0e130521-a673-425f-bfc5-54a1b83f79b6" />

### 3. Steps to Reproduce

1. Add a Clan BattleMech to your force, let's say the Warhawk C.
2. Give the pilot the Enhanced Imaging Implant in the Configuration Menu.
3. Edit Damage to remove armour from all locations other than head to make testing easier.
4. Add a Princess Bot and give her a unit that can deal small amounts of damage to test this out, for example a Locust LCT-1L.
5. Start up the game and before finishing deployment make sure to turn Enhanced Imaging to "Off" in the Systems Tab, confirm that it is off.
6. Skip phases until the enemy unit hits you on a non-head location.
7. See that you have to make rolls for each internal hit suffered despite EI being offline.

### Severity *

Medium (Gameplay Limitation): Non-core functionality is impaired, providing a suboptimal but playable experience.

### MegaMek Suite Version *

0.51.00

### Operating System *

Windows 10

### Java Version *

    Java Vendor: Eclipse Adoptium     Java Version: 21.0.11

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
- [x] I have included all necessary information and files to help reproduce this issue
- [x] I have asked on MegaMek Discord about this issue
- [x] I have confirmed this issue is being opened on the correct repository: MegaMek, MegaMekLab, or MekHQ

Issue #8709, opened 2026-08-14
[RFE] Allowing Advanced Structures armed with Artillery to be placed Off-Map


Ursur1minor:
### Enhancement Type *

Improvement to Existing Feature

### Brief Description of Enhancement *

Currently Advanced Structures can not be placed off-map even when loaded with artillery as can be seen in the supplied images:

<img width="564" height="806" alt="Image" src="https://github.com/user-attachments/assets/e683bdf8-6d0b-4c9d-9a27-81d1e11d11c2" />

<img width="880" height="755" alt="Image" src="https://github.com/user-attachments/assets/8c8732a7-1a83-4587-aca2-c3cd374982cf" />

### Use Cases or Rationale

1. Would allow Advanced Structures to represent off-map artillery assets.
2. Would align them with the rules for Artillery units.

### MegaMek Suite Version *

0.51.00

### Final Verification

- [x] I confirm this request hasn't already been submitted (checked the tracker)
- [x] I've discussed or asked about this enhancement on MegaMek Discord
- [x] I’m opening this on the correct repo (MegaMek, MegaMekLab, or MekHQ)

Issue #8433, opened 2026-07-04
[Issue] Infantry vs Infantry Combat MP Casualties are calculated incorrectly


Rethid:
### Brief Description *

<img width="1032" height="233" alt="Image" src="https://github.com/user-attachments/assets/3f813297-1a83-460d-97eb-7b421bc0179b" />
It appears that Infantry vs Infantry Action is currently calculating MP damage incorrectly, currently calculating casualties using the percentage as a ratio of the infantry's own current strength rather than their enemy's. i.e. in this example the attacker has taken 45% of their own strength (27*.45 = 12.15) as casualties. Additionally, these casualties appear to be rounded down rather than up, which may be due to TO using the term casualties in two separate locations, one of which is correctly rounded up (MP damage during an Infantry vs Infantry Combat) and the other is correctly rounded down (The conversion of MP back to original personnel at the end of an Infantry vs Infantry Combat). There was some discussion of a separate issue relating to the latter on the Discord so it felt important to mention that value is meant to be rounded down.

Per TO:AR, Pg172, second paragraph under the Infantry v Infantry Action Roll heading,

> For each infantry vs. infantry action roll, roll 2D6 and crossreference the result with the appropriate column on the Infantry
Vs. Infantry Action Table (see p. 173). **The number to the left of the
slash is the percentage of the defender’s total strength subtracted
from the attacker’s strength as casualties, while the number to the
right of the slash is the percentage of the attacker’s total strength
subtracted from the defender’s strength.** However, the defender
in an infantry vs. infantry action has a decisive advantage and
suffers only half damage for as long as defending troops remain
in full control of the unit or building (until the attack results in a P
or an E; see Letters on the Infantry Vs. Infantry Action Table; p. 173).
**Round all fractions up.**

(Emphasis mine)

### 3. Steps to Reproduce

1. Initiate an Infantry vs Infantry action and allow it to resolve a turn.

### Attach Files

[Infantryvinfantryreport.zip](https://github.com/user-attachments/files/29651946/Infantryvinfantryreport.zip)

### Severity *

Medium (Gameplay Limitation): Non-core functionality is impaired, providing a suboptimal but playable experience.

### MegaMek Suite Version *

51.00

### Operating System *

Windows 11 10.0 (amd64)

### Java Version *

Java Vendor: Eclipse Adoptium Java Version 17.0.11

### Final Verification

- [x] I confirm this is a single, unique issue that hasn't been reported before
[...]

PR #8722, opened 2026-08-15
Add one-click bug report bundle and prefilled issue forms

HammerGS:
## Summary

Adds a **Report a Bug** button that saves the current game, collects the relevant logs, and writes a single zip the player can attach to an issue. The MegaMek, MegaMekLab and MekHQ buttons now open the issue form with the version, operating system and Java version already filled in.

This is the MegaMek counterpart of MekHQ #8433. MegaMek already had the dialog itself (`BugReportDialog`, the four repo links, the Discord button), so this fills in the part that was missing: the player still had to find their save, work out which of the files in `logs/` mattered, and zip them by hand.

Reporting is also now offered where a player actually is when something goes wrong, rather than only under the Help menu: on the error dialog raised for an uncaught exception, on the Commands button above the board, and on a button that sits in the same place in the phase display all game.

## What the player sees

Before: the dialog told them to "save your game... best make a ZIP of it", then the issue form asked for MegaMek Suite Version, Operating System and Java Version - all required - and instructed them to open `megamek.log` in a text editor and copy them out of the header, with a screenshot showing what to look for.

After: one button produces `MegaMek-BugReport-<timestamp>.zip` containing the save, the logs that matter, and a `system-info.txt`. Pressing "MegaMek" opens the issue form with those three fields already populated. In the dialog that button is larger than its neighbours and framed in yellow and red hazard stripes; in game it keeps the ordinary phase display skin and sits at the bottom of the Done column in every phase, so it does not have to be hunted for in a menu.

image

https://github.com/user-attachments/assets/0f9ee990-a4e2-40fb-840c-5a2f932940de

## Changes

1. **`BugReportBundle`** decides what goes in the archive and writes it. It uses an explicit manifest rather than scanning the log folder, because that folder also accumulates one `gamelog*.html` combat report per game and a `Bot_*.mul` unit list per bot per game - a working install can hold hundreds of megabytes there. A plain `*.log` filter has the opposite problem: it would skip the combat report, which is usually the most useful single artifact in a MegaMek bug report. The archive is capped at the 25 MB GitHub allows for attachments, and anything dropped is named back to the player rather than silently omitted.

2. **`IssueReportUrl`** builds the prefilled URL. All three suite repositories use issue forms whose field ids are identical, so one builder serves them all and only the repository differs. `mm-data` has no template and is left alone.

3. **Save-completion callback on `AbstractClient`.** Saving is asynchronous: `/localsave` goes to the server, which serializes the game and streams it back as a `SEND_SAVEGAME` packet, so the archive cannot be built until that arrives. There is a 30 second timeout after which the archive is written with logs alone - see "Known limitation" below.

4. **Reporting offered from three new places.** The error dialog gains a Report a Bug button, gated on the error carrying a `Throwable`; the Commands menu gains a Report a Bug entry; and every phase display carries one at the bottom of the Done column, added in `StatusBarPhaseDisplay` after the phase's own contents so it lands below Skip where there is one and below Done where there is not. It is in the same place regardless of which button group is showing, so it is never a "More..." away.

5. **Copy System Data is now offered only where nothing gathers the files automatically.** In MegaMek the archive carries a `system-info.txt` and the repository buttons fill the same three details into the issue form, so the button duplicated work the player no longer has to do. MegaMekLab and MekHQ open this dialog without a packaging action and keep it, and the Help menu keeps the item in its own right.

## Two decisions worth reviewing

**No customs export.** MekHQ #8433 threaded an `isBugReportPrep` flag through `Campaign.writeToXML` so custom units would be baked into the save. MegaMek needs no equivalent: `GameManagerSaveHelper` serializes the whole `Game` with XStream and `Server.loadGame` restores it without ever consulting `MekSummaryCache`, so customs are already inside the save.

**The `labels` query parameter is deliberately not set, and should stay that way.** GitHub requires the visitor to hold permission for any action a query parameter performs, and serves a **404 page** when they do not. Ordinary players have no triage permission here, so adding a `labels` parameter would replace the issue form with a 404 for exactly the people this feature is for - while working fine for every maintainer who tested it. There is a regression test named for this (`IssueReportUrlTest.neverSetsTheLabelsParameter`) and the reasoning is in `IssueReportUrl`'s Javadoc.

**Separately, and not caused by this PR:** while testing the above I noticed that bug reports filed through the web form arrive with **no label at all**. `bug_report.yml` declares `labels: [bug]`, but the repository's label is named `Bug`, and issue-template label matching is case sensitive - so it silently applies nothing. The RFE template, which declares the exactly-matching `(RFE) Enhancement`, does work. Compare #8718 (bug form, unlabelled) with #8709 (RFE form, labelled). That is a one-character fix in `.github/` and is up as #8724.

## Known limitation (not introduced here)

The 30 second timeout exists to work around #8721: the server refuses a local save in a double-blind game with local saves disabled by sending a chat message and no packet, so `awaitingSave` is never cleared and the client can never finish shutting down. This PR works around it for the packager only; the underlying bug is filed separately and is not fixed here.

## Layering note

`MMLogger` takes the error-dialog button as an installable hook rather than calling the dialog directly, because `megamek.logging` sits below the user interface and is shared with MegaMekLab, MekHQ and the headless dedicated server. Those install nothing and keep the plain OK dialog they have today. The hook carries both button labels, already localized, so no player-facing text lives in `megamek.logging`.

## Files Changed

- `megamek/common/util/BugReportBundle.java` - new; manifest and zip writer, Swing-free so it is headless-testable
- `megamek/common/util/IssueReportUrl.java` - new; prefilled issue-form URL builder
- `megamek/client/ui/PackageBugReportAction.java` - new; the Swing action - chooser, save request, timeout, result dialog
- `megamek/client/AbstractClient.java` - one-shot save-completion callback
- `megamek/client/Client.java` - fire the callback on both `SEND_SAVEGAME` exit paths
- `megamek/logging/MMLogger.java` - installable error-dialog button, offered only for errors carrying a `Throwable`
- `megamek/client/ui/clientGUI/MegaMekGUI.java` - install the hook; resolve the running client on click
- `megamek/client/ui/clientGUI/BugReportDialog.java` - button order, the enlarged reporting button and its hazard stripe border; prefilled repo links
- `megamek/client/ui/clientGUI/GameCommandsMenu.java` - Report a Bug entry
- `megamek/client/ui/panels/phaseDisplay/StatusBarPhaseDisplay.java` - the always-present button in the Done column
- `megamek/client/ui/ShowBugReportDialogAction.java`, `CommonMenuBar.java`, `ClientGUI.java` - wiring
- `BugReport.properties`, `messages.properties` - new keys; step 1 of the dialog text no longer says to zip by hand
- `BugReportBundleTest.java`, `IssueReportUrlTest.java` - new; 15 tests

## Testing

Unit tests: 15 new across the two test classes, covering the empty and missing log directory, manifest selection against a directory seeded with 40 `Bot_*.mul` files and three game logs, newest-game-log-only, save-at-archive-root, the size cap, and the labels regression. Full suite green at 15,072 tests, 0 failures.

Playtested and confirmed:

- The crash dialog. A deliberate `NullPointerException` on both the event dispatch thread and a background thread produces the "Uncaught Exception" dialog with the Report a Bug button, and it opens the helper.
- Packaging in game. The archive was produced and its contents were complete, which exercises the asynchronous save round trip, the completion callback, and the manifest against a real log directory.
- **The archived save loads back into MegaMek correctly.** The save the packager writes is a normal, complete save; being routed through the archive does not damage it.
- Copy File to Clipboard, on Windows 11.
- **The prefilled issue form, end to end against this repository.** A test report was filed from the in-game button and arrived with MegaMek Suite Version, Operating System and Java Version already populated, and with the generated archive attached. That test issue has since been closed.

## What is NOT proven yet

- **The phase-display button has been compiled, not seen.** The Done column now asks for three button heights instead of two; whether that costs board space or fills the gap that was already under Skip needs a look in game.
- The prefilled form has only been opened from an account with triage permission. The URL sets no permission-gated query parameter, so it should behave identically for everyone, but that has not been confirmed from a plain account.
- The **double-blind save timeout** has not been exercised. It needs a game deliberately configured with Double Blind plus "Disable local saves when using double blind".
- **Copy File to Clipboard is untested on macOS and Linux.** Where the desktop does not support it the button simply does nothing and Open Folder remains the guaranteed route.
- The dedicated server path is unchanged by inspection - the hook is never installed there - but has not been run.


megamek/src/megamek/client/ui/PackageBugReportAction.java
```diff
@@ -0,0 +1,372 @@
...
+     * <p>The save is written beside the chosen archive rather than into the player's save folder, so a bug report
+     * never leaves a stray save behind among their real ones.</p>
+     *
+     * @param client      the connected client to save through
+     * @param archiveFile the archive to write
+     */
+    private void requestSaveThenBuildArchive(Client client, File archiveFile) {
+        File archiveDirectory = archiveFile.getParentFile();
+        String temporarySaveName = stripExtension(archiveFile.getName());
+
```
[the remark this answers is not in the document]
HammerGS: Good catch on both, fixed in e44fa8b6f9.

The space issue is real and reachable: the file chooser lets the player type "Bug Report.zip", `temporarySaveName` came straight off that name, and `LocalSaveGameCommand` reads `args[1]` as the filename and `args[2]` as the directory. So the save would have been written as "Bug" into a directory called "Report". The path already survives because it is escaped to pipes here and unescaped in `GameManagerSaveHelper`, but there is no equivalent mechanism for the filename, so whitespace is now collapsed to underscores in a new `temporarySaveNameFor`.

On the null parent: `chooseArchiveFile` now returns `getAbsoluteFile()`, so the containing directory is always resolvable rather than being guarded at each use.

megamek/src/megamek/common/util/BugReportBundle.java
```diff
@@ -0,0 +1,345 @@
...
+     *
+     * @return {@code true} if the file is a rolled archive of a collected log
+     */
+    private static boolean isCollectableRolledLog(File file) {
+        if (!file.isFile() || !file.getName().endsWith(ROLLED_LOG_EXTENSION)) {
+            return false;
+        }
+        List<String> allBaseNames = new ArrayList<>(ESSENTIAL_LOG_BASE_NAMES);
+        allBaseNames.addAll(BOT_LOG_BASE_NAMES);
+        return allBaseNames.stream().anyMatch(baseName -> file.getName().startsWith(baseName + "_"));
```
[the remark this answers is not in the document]
HammerGS: Valid, fixed in e44fa8b6f9. The combined list is now a precomputed `ALL_LOG_BASE_NAMES` constant rather than being rebuilt per file.

Worth noting this was a regression I introduced: the list used to be built once in the enclosing method, and moving the predicate into its own method to keep the stream out of the filter expression pushed the allocation inside the per-file call. Thanks for catching it.

megamek/src/megamek/client/ui/PackageBugReportAction.java
```diff
@@ -0,0 +1,372 @@
...
+            @Override
+            protected void done() {
+                // The temporary save has been copied into the archive, so the loose copy is no longer wanted. On
+                // failure it is deliberately left behind, so the player still has something to attach.
+                try {
+                    BugReportBundle.BundleResult result = get();
+                    deleteTemporarySave(saveGameFile);
+                    showResult(result, cautionNotice);
+                } catch (Exception exception) {
+                    LOGGER.error(exception, "[BugReport] Could not write the archive to {}", archiveFile.getPath());
```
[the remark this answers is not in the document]
HammerGS: Declining this one - I believe it is a false positive.

This is an `MMLogger` call, not `String.format`. `MMLogger.parametrizedStringAnyway` checks `message.contains("{}")` and formats via log4j's `ParameterizedMessage`; `String.format` is only the fallback used when the message has no brace placeholder:

```java
if (message.contains("{}")) {
    message = new ParameterizedMessage(message, args).getFormattedMessage();
} else {
    message = String.format(message, args);
}
```

So the `{}` here does consume the argument, and the message formats correctly at runtime.

Braces are also the preferred form in this codebase. `MMLogger`'s class Javadoc documents both styles and says of the `%s` form: *"grandfathered exclusively for logs already existing, and it is not recommended for any new log."*

The analyzer looks to be counting `%` specifiers only, which is why it reports zero for a brace-formatted message. The same applies to the other three instances flagged on this PR.

megamek/src/megamek/client/ui/PackageBugReportAction.java
```diff
@@ -0,0 +1,372 @@
...
+    private static void openContainingFolder(File archiveFile) {
+        File containingFolder = archiveFile.getParentFile();
+        if ((containingFolder == null) || !Desktop.isDesktopSupported()) {
+            LOGGER.warn("[BugReport] Cannot open the containing folder on this platform");
+            return;
+        }
+        try {
+            Desktop.getDesktop().open(containingFolder);
+        } catch (Exception exception) {
+            LOGGER.error(exception, "[BugReport] Could not open {}", containingFolder.getPath());
```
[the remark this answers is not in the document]
HammerGS: Declining this one - I believe it is a false positive.

This is an `MMLogger` call, not `String.format`. `MMLogger.parametrizedStringAnyway` checks `message.contains("{}")` and formats via log4j's `ParameterizedMessage`; `String.format` is only the fallback used when the message has no brace placeholder:

```java
if (message.contains("{}")) {
    message = new ParameterizedMessage(message, args).getFormattedMessage();
} else {
    message = String.format(message, args);
}
```

So the `{}` here does consume the argument, and the message formats correctly at runtime.

Braces are also the preferred form in this codebase. `MMLogger`'s class Javadoc documents both styles and says of the `%s` form: *"grandfathered exclusively for logs already existing, and it is not recommended for any new log."*

The analyzer looks to be counting `%` specifiers only, which is why it reports zero for a brace-formatted message. The same applies to the other three instances flagged on this PR.

megamek/src/megamek/client/ui/PackageBugReportAction.java
```diff
@@ -0,0 +1,372 @@
...
+            public Object getTransferData(DataFlavor flavor) {
+                return List.of(archiveFile);
+            }
+        };
+
+        try {
+            Toolkit.getDefaultToolkit().getSystemClipboard().setContents(fileTransferable, null);
+            LOGGER.debug("[BugReport] Copied {} to the clipboard", archiveFile.getName());
+        } catch (Exception exception) {
+            LOGGER.error(exception, "[BugReport] Could not copy {} to the clipboard", archiveFile.getName());
```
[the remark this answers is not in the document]
HammerGS: Declining this one - I believe it is a false positive.

This is an `MMLogger` call, not `String.format`. `MMLogger.parametrizedStringAnyway` checks `message.contains("{}")` and formats via log4j's `ParameterizedMessage`; `String.format` is only the fallback used when the message has no brace placeholder:

```java
if (message.contains("{}")) {
    message = new ParameterizedMessage(message, args).getFormattedMessage();
} else {
    message = String.format(message, args);
}
```

So the `{}` here does consume the argument, and the message formats correctly at runtime.

Braces are also the preferred form in this codebase. `MMLogger`'s class Javadoc documents both styles and says of the `%s` form: *"grandfathered exclusively for logs already existing, and it is not recommended for any new log."*

The analyzer looks to be counting `%` specifiers only, which is why it reports zero for a brace-formatted message. The same applies to the other three instances flagged on this PR.

megamek/src/megamek/common/util/BugReportBundle.java
```diff
@@ -0,0 +1,345 @@
...
+          throws IOException {
+        zipOutputStream.putNextEntry(new ZipEntry(entryName));
+        try (InputStream fileInputStream = new FileInputStream(source)) {
+            byte[] buffer = new byte[COPY_BUFFER_SIZE];
+            int bytesRead;
+            while ((bytesRead = fileInputStream.read(buffer)) != -1) {
+                zipOutputStream.write(buffer, 0, bytesRead);
+            }
+        } catch (IOException exception) {
+            LOGGER.warn(exception, "[BugReport] Could not read {}; it will be empty in the archive", source.getName());
```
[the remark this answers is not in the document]
HammerGS: Declining this one - I believe it is a false positive.

This is an `MMLogger` call, not `String.format`. `MMLogger.parametrizedStringAnyway` checks `message.contains("{}")` and formats via log4j's `ParameterizedMessage`; `String.format` is only the fallback used when the message has no brace placeholder:

```java
if (message.contains("{}")) {
    message = new ParameterizedMessage(message, args).getFormattedMessage();
} else {
    message = String.format(message, args);
}
```

So the `{}` here does consume the argument, and the message formats correctly at runtime.

Braces are also the preferred form in this codebase. `MMLogger`'s class Javadoc documents both styles and says of the `%s` form: *"grandfathered exclusively for logs already existing, and it is not recommended for any new log."*

The analyzer looks to be counting `%` specifiers only, which is why it reports zero for a brace-formatted message. The same applies to the other three instances flagged on this PR.

HammerGS:
### On the four "Unused format argument" findings

Collecting this here rather than repeating it in each thread.

All four are the same false positive, and the rule is flagging this codebase's preferred logging style rather than anything new in this PR.

**Why the argument is not unused.** These are `MMLogger` calls, not `String.format`. `MMLogger.parametrizedStringAnyway` branches on the message itself:

```java
if (message.contains("{}")) {
    message = new ParameterizedMessage(message, args).getFormattedMessage();
} else {
    message = String.format(message, args);
}
```

`String.format` is only the fallback for messages with no brace placeholder. A `{}` message is formatted by log4j's `ParameterizedMessage`, which consumes the argument. The check appears to count `%` specifiers only, so it sees zero for a brace-formatted message.

**How common this is.** The exact flagged shape - `logger.<level>(throwable, "...{}...", arg)` - appears **69 times** in files this PR does not touch:

```
WeaponDataEncoder.java:87    logger.error(e, "Error while trying to serialize Weapon {} data for {}", weapon, entity);
BehaviorSettings.java:461    logger.error(ex, "Add Priority Unit - Invalid unit ID: {}", id);
PathRanker.java:175          logger.error(e, "{} while processing {}", e.getMessage(), path);
Client.java:1361             LOGGER.error(ex, "Unable to save file {}", sFinalFile);
```

That last one sits four lines above a change in this PR. It was not flagged, because it is not part of the diff. Brace-style logging appears roughly 1,449 times across `megamek/src` in total.

**What the convention actually is.** `MMLogger`'s own class Javadoc documents both forms and says of the `%s` style: *"grandfathered exclusively for logs already existing, and it is not recommended for any new log."* So converting these four would make them inconsistent with their neighbours and contrary to the documented guidance.

Declining all four on that basis. The two Copilot findings on this PR were separate and both valid - they are fixed in e44fa8b6f9.

**Worth noting for the future:** this will fire on every PR that adds logging, since the `security-and-quality` suite is enabled in `.github/workflows/codeql-analysis.yml`. If the noise becomes tiresome, that single query can be excluded through a CodeQL config while keeping the rest of the suite. Happy to open that separately if maintainers want it.

megamek/resources/megamek/client/BugReport.properties
```diff
@@ -29,10 +29,25 @@
...
+mainText=While we work hard to keep the program suite stable and running smoothly, things can still go wrong. Because this is a volunteer project without the funding or team for large-scale testing, we rely on players to report any issues they find. To do this, follow these steps:</p><blockquote><b> 1)</b> Press 'Package Bug Report' below. This saves your game and collects the logs into a single ZIP file for you. If you are reporting a unit or campaign problem instead, save that and make a ZIP of it yourself.<br><b> 2) (Optional)</b> If you are on our Discord, it can help to ask whether the bug has already been reported. This may save you the effort of opening a report that ends up closed because the issue was already fixed. Duplicate reports are fine. We just want to save you time. Not a member of our Discord community? Consider pressing the 'Discord' button to join.<br><b> 3)</b> Press the appropriate button for the type of issue. You will need a GitHub account. We used to allow anonymous reporting, but it caused problems and had to be removed .<br><b> 4)</b> Fill out the issue report form, then upload the ZIP you created in step 1.</blockquote>
```
IllianiBird: > If you are reporting a unit or campaign problem instead, save that and make a ZIP of it yourself.

Probably best to mention that for campaign problems the player should use the report bug feature in mekhq. We don't want players making their own zips, where possible, as that invariably leads to missing information.

megamek/src/megamek/client/ui/clientGUI/MegaMekGUI.java
IllianiBird: From JavaDocs

> Finalization is deprecated and subject to removal in a future release. The use of finalization can lead to problems with security, performance, and reliability. See JEP 421   for discussion and alternatives.
> 
> Subclasses that override finalize to perform cleanup should use alternative cleanup mechanisms and remove the finalize method. Use java.lang.ref.Cleaner and java.lang.ref.PhantomReference as safer ways to release resources when an object becomes unreachable. Alternatively, add a close method to explicitly release resources, and implement AutoCloseable to enable use of the try-with-resources statement.
> 
> This method will remain in place until finalizers have been removed from most existing code.

megamek/src/megamek/client/ui/PackageBugReportAction.java
```diff
@@ -0,0 +1,390 @@
...
+ * logs-and-system-information archive rather than an error.</p>
+ *
+ * <p>Saving in MegaMek is asynchronous - see {@link megamek.client.AbstractClient#setSaveCompletionCallback} - so the
+ * archive cannot be built until the server has returned the save. This action therefore requests the save, waits for
+ * the completion callback, and falls back to a logs-only archive if the save does not arrive within
+ * {@link #SAVE_TIMEOUT_MILLISECONDS}. The player always ends up with a file.</p>
+ *
+ * @see BugReportBundle
+ */
+public class PackageBugReportAction extends AbstractAction {
```
IllianiBird, line 83: Needs a Clone() method

megamek/src/megamek/client/AbstractClient.java
```diff
@@ -46,6 +48,7 @@
 import megamek.client.generator.RandomUnitGenerator;
 import megamek.client.ui.Base64Image;
 import megamek.common.Player;
+import megamek.common.annotations.Nullable;
```
IllianiBird: Should be Jakarta Nullable

megamek/src/megamek/logging/MMLogger.java
```diff
@@ -275,9 +307,36 @@ public void errorDialog(String title, String message, Object... args) {
...
+        ErrorDialogButton extraButton = errorDialogButton;
         try {
-            JOptionPane.showMessageDialog(null, message, title, JOptionPane.ERROR_MESSAGE);
+            if (!offerBugReport || (extraButton == null)) {
+                JOptionPane.showMessageDialog(null, message, title, JOptionPane.ERROR_MESSAGE);
+                return;
+            }
+
+            String okLabel = UIManager.getString("OptionPane.okButtonText");
+            Object[] options = { extraButton.label(), (okLabel == null) ? "OK" : okLabel };
```
IllianiBird: Hardcoded player-facing String

IllianiBird (commented):
Some points of feedback that I think should be reviewed before this can be merged.

megamek/resources/megamek/client/BugReport.properties
HammerGS: Agreed - fixed in f79b87b806.

Step 1 now reads:

> **1)** Press the 'Report a Bug' button below. This saves your game and collects the logs into a single ZIP file for you. For a campaign problem, use 'Report a Bug' in MekHQ instead, so that the right campaign files are gathered for you. For a unit design problem, simply attach the unit file itself.

So nobody is told to build their own archive: campaigns go to MekHQ's own reporter, and a unit is a single file that needs no zipping at all.

megamek/src/megamek/client/ui/clientGUI/MegaMekGUI.java
HammerGS: Good catch - removed in f79b87b806.

The line was `System.runFinalization()` in `unlaunch()`, which is deprecated for removal since Java 18 for the reasons you quote. It also does nothing useful here: it only nudges finalizers that are already scheduled, and the `client = null` above it is what actually lets the game be collected. The `System.gc()` on the preceding line is left as it was.

Pre-existing rather than something this PR added, but it is a one-line deletion in a file the PR already touches, so it is done.

megamek/src/megamek/client/ui/PackageBugReportAction.java
HammerGS: Added in f79b87b806, following the same shape as `CancelAction`, `CloseAction` and `OkayAction`:

```java
@Override
public @Nullable PackageBugReportAction clone() {
    try {
        return (PackageBugReportAction) super.clone();
    } catch (CloneNotSupportedException cloneNotSupportedException) {
        LOGGER.error("Failed to clone PackageBugReportAction. State of the object: {}", this,
              cloneNotSupportedException);
        return null;
    }
}
```

Both fields (the parent component and the client supplier) are final references the copy is meant to share, so there is nothing to deep-copy after `super.clone()`; the JavaDoc says as much so a future reader does not wonder whether an assignment was forgotten.

megamek/src/megamek/client/AbstractClient.java
HammerGS: Switched in f79b87b806. `jakarta.annotation.Nullable` ships in `jakarta.annotation-api:3.0.0`, which is already a dependency, and it declares no `@Target`, so it is a drop-in for the megamek annotation (which is restricted to `METHOD` and `PARAMETER`).

Four files are converted: `AbstractClient`, `BugReportBundle`, `IssueReportUrl` and `PackageBugReportAction` - the three new classes plus the one where this PR added the import in the first place.

The other six changed files (`Client`, `ClientGUI`, `MegaMekGUI`, `GameCommandsMenu`, `BugReportDialog`, `ShowBugReportDialogAction`) already imported `megamek.common.annotations.Nullable` before this PR. Java will not take two imports of the same simple name, so converting those means rewriting `@Nullable` usages this PR never touched - up to 15 in `MegaMekGUI` alone - which would put unrelated lines in the diff. I have left them as they are.

Worth flagging that the codebase is currently 513 files on `megamek.common.annotations.Nullable` against exactly one on jakarta (`Mounted.java`), so this is not a convention the tree follows yet. Happy to do the sweep as its own PR if that is the direction you want, and to deprecate the megamek annotation at the same time - it is a big mechanical diff that should not ride along inside a feature PR.

megamek/src/megamek/logging/MMLogger.java
HammerGS: Fixed in f79b87b806 - there is no player-facing literal in `megamek.logging` any more.

Offering a choice of buttons means naming both of them, and this package sits below the UI layer, so it cannot reach `Messages`. The label now comes in on the hook that the UI installs, alongside the one it was already supplying:

```java
public record ErrorDialogButton(String label, String dismissLabel, Runnable action) {}
```

`MegaMekGUI` fills it from `Messages.getString("Okay")` when it installs the hook, and `showErrorDialog` just uses what it was given. The `UIManager` lookup and the `"OK"` fallback are both gone.

HammerGS:
Second round pushed - `f79b87b806`, `21a77ef922` and `98fb2f030a`. All five review points are addressed and answered inline. Full suite green, checkstyle clean.

There are also changes here that came out of testing rather than the review, so they are worth calling out rather than leaving to be found in the diff:

**The button is now called Report a Bug everywhere.** It was "Package Bug Report" in the dialog while every other route into the same tools said "Report a Bug", which made it look like two different features. One name, one thing.

**It is marked out with a black and yellow hazard border,** in the dialog and in the phase display, from a single shared `attentionBorder()` so the two cannot drift apart. In the dialog it is also larger and bolder than its neighbours: it shares that dialog with nine buttons that only open a link, and at the default size it read as one option among many rather than the step the instructions begin with.

**Step 3 no longer recounts the history of anonymous reporting.** It now just says a GitHub account is needed because we don't allow anonymous reports.

**There is now a Report a Bug button in every phase,** at the bottom of the Done column, in the same place all game. The menu bar is not somewhere a player looks when something has just gone wrong on the board, and by the time they have found it the game has often moved on.

**The dialog's buttons follow its own instructions.** Gathering the files is step 1 of the text but was the last row of buttons; it is now the first row after Discord, with the repository buttons below it.

**Copy System Data is only offered where nothing gathers the files for you.** In MegaMek the archive carries a `system-info.txt` and the repository buttons put the same three details into the issue form, so the button asked the player to do work that had already been done. MegaMekLab and MekHQ open this dialog without a packaging action and still get it, and the Help menu item is untouched.

The one thing still unproven is layout: the Done column now asks for three button heights instead of two, and I have not yet seen in game whether that costs board space or simply fills the gap that was already sitting under Skip.

Next step after this merges is to rebuild the dialog as a wizard - pick what went wrong, gather the files, check it is not already reported, then file it - so that the wall of instructions and the eight buttons stop competing for the player's attention. That wants its own PR, not least because MegaMekLab and MekHQ open the same dialog and need a different path through it.

IllianiBird (approved):

PR #8728, opened 2026-08-16
Add atmospheric aerospace doctrine for CASPAR (fighters, small craft, DropShips) (Fixes #8712)

### LINKED ISSUE #8712: BotClient drops a multi-unit bot's later turns in simultaneous phases
## Problem

With any simultaneous phase option on, a bot silently drops every turn after its first in that phase. `BotClient.gameTurnChange` has a guard (`ignoreSimTurn`) meant to ignore other players' turn-change echoes, but it keys on "already calculated once this phase" — so a multi-unit bot's second and later turns are skipped even when the server explicitly says the turn is the bot's. The bot fires one unit per firing phase; the rest never act.

## Example

Headless 2v2+2 game, `simultaneous_firing` on (inherited from the user's saved `gameoptions.xml`). CASPAR's Cheetah selects an 83%-to-hit dive bomb plan (96.6 utility). Its firing turn arrives and the audit log records, verbatim:

```
BotClient$1.gameTurnChange - Strike Flight: turn change - phase FIRING, turnIndex 1,
prevPlayer 2, me 1, myTurn true, ignoreSimTurn true, alreadyCalculatedThisPhase true -> skipping
```

`myTurn true` — and skipped anyway. Across ~50 benchmark games both bots fired at roughly half rate; dive bombing measured zero because the bomber's turn was the usual casualty.

## Proposed fix

- Suppress duplicate events by turn INDEX, not by a once-per-phase flag: remember the last turn index a calculation was spawned for, and spawn whenever `isMyTurn()` and the index is new. Duplicate echoes of the same pending turn still spawn once; a genuinely new turn is never dropped, whoever's event carried it.
- The once-per-phase flag (`calculatedTurnThisPhase`) is removed.

The turn-order logic and non-simultaneous behavior are NOT modified (index dedupe is a no-op there: each turn already arrives once with a fresh index).

## Implementation notes

- `BotClient.gameTurnChange` (~line 186) and the `calculatedTurnThisPhase` field; reset the remembered index on phase change.
- Verified by reverting: with the old guard and `simultaneous_firing` on, a 2-fighter bot resolves 0 dive bombs in a bombing scenario; with the fix, the same scenario resolves the full 15-bomb payload (79 turn spawns, duplicates still suppressed).
- Unit test included: duplicate event for the same turn index does not spawn; a new index always spawns.

## Background

Found 2026-08-14 during headless aerospace benchmarking on the `Implement-Aerospace-Doctrine` branch — a turn-dispatch audit log added the night before caught the skip verbatim. The trigger environment (simultaneous options leaking into headless games from the user's saved `gameoptions.xml`) is the same one behind #8698. Fix implemented and verified on that branch; will arrive with its PR.

HammerGS:
## Summary

Stock bots cannot fly. They hold one altitude, treat unmoved enemies as committed, plan shots the
server refuses, and leave the board for free. This PR gives CASPAR a full atmospheric aerospace
doctrine. Princess is unchanged: she resolves the same object instances as before, pinned by
`AerospaceSeamTest`. About a third of the diff is tests (15 files, ~3,400 lines of 10,600).
Space maps are out of scope by prior agreement; the `isSpaceborne()` guards are the boundary.

## What the bot does now

*Fly* - Ground map and low altitude are separate venues (`AerospaceVenue`). Altitude candidates,
honest reaction to unmoved enemies, velocity and edge pricing, climb doctrine.

*Attack the ground* - Attack runs priced along the flown line (TW p.242-243). Guns pick one
victim by approach arc (rear x1.5). Bombs aim at formation seams, rationed to the victim, and
auctioned by value: a 4-bomb Locust kill outbids a 10-bomb Atlas dent. Strafing works for the
first time: straight runs up to 5 hexes, energy only, heat once per weapon, announced by toast.

*Fight the air* - Dead-zone-honest gunnery. TW p.85 maneuvers behind odds floors and EV pricing;
elite pilots use them 15x more. An intercept credit hunts laden bombers while friendly ground
units exist.

*Leave and return* - Winchester: a lone bombless fighter that cannot decide the fight in 30
rounds flies off. Grounded units take off again (runway first, vertical above a 0.5 roll floor,
gun turret otherwise). Includes a rules fix: `IAero.canTakeOffVertically` contradicted SO and the
engine's own roll table; fighters and aerodyne small craft may now lift vertically.

*Take orders* - Two chat commands, panel UI in a later PR. `af : AEROSPACE|GROUND|AUTO` biases
the air/ground credit sets x2 / x0.25. `al : LIFT_OFF|LAND|AUTO` moves DropShips and small craft
between domains; under AUTO they hold station, and a DropShip defers liftoff while friendlies sit
in the blast.

## Measured

Headless CASPAR-vs-Princess games on identical forces. The ratio is surviving Battle Value at
game end, CASPAR : Princess - so 4.08:1 means CASPAR ended with about four times the BV Princess
had left. "25/29" = CASPAR ahead in 25 of 29 games; p is a sign test on those pairs.

| Test | Surviving BV, CASPAR : Princess |
|---|---|
| A2A ground map, n=29 pairs | 4.08:1, ahead 25/29, p<.0001 |
| A2A low altitude, n=30 pairs | 2.14:1, ahead 28/30, p<.0001 |
| Mixed combined arms | 1.62:1 headless; live game won |
| Intercept credit, 30 games | enemy fighter kills 9 to 15, BV ratio unchanged |
| Focus order vs AUTO | 3.23:1 vs 2.41:1; own fighter losses 11 to 7 |
| Elite maneuver rate | 0.2 to 3.0 maneuvers per game |
| CPU (JFR, live game) | all bot AI: 1.7% of client CPU |

Both A2A arms were flown from the weaker deployment slot (mirrors show a 9/10 team-2 edge).
Every feature has a live-game sighting. Key tests verified by reverting the fix. Known boundary:
constants are tuned against gunnery 4; an all-elite mirror flips to 0.72:1 (documented follow-up).

## Fixes included

- #8712: BotClient dropped bot turns under simultaneous options. Turn dedupe by index,
  revert-verified.
- Bots no longer self-dismiss mid-game.
- Maneuver and strafe announcements with toasts (reports 9604/9610), double-blind safe.
- Done-ledger logging: report-phase ready-acks and blockers are now named in the log.
- TSV kill ledger and per-turn DEBRIEF decision lines.

## Files changed (55 files, +10,639 / -64)

- New bot source: `AerospaceVenue`, `AerospaceGeometry`, `AerospacePathRanker`,
  `AerospaceFireControl`, `AeroGroundDoctrinePathFinder`, two order enums, two chat commands.
- Main edits: `Princess` (seams, order state), `Caspar` (registration, takeoff), `BotClient`
  (turn dedupe), `IAero` (liftoff rule), `TWGameManager` (announcement), TSV classes.
- Tests: one suite per new class, `AerospaceSeamTest` (Princess-unchanged proof), engine pins.
- Scenarios: 8 test setups with game options pinned.

## Out of scope

- Space maps and vector movement - next PR.
- Bot commands panel UI - separate PR; both commands work from chat now.
- Landing approach doctrine, altitude bombing, Princess promotion.

## Notes for reviewers

- This is the large PR agreed in advance. Review by commit: seams, movement, gunnery, maneuvers,
  attack runs, orders, strafing, grounded recovery, SC/DS.
- One addition to `TWGameManager`: `announceStrafingRun`, ~20 lines of report plumbing at
  `receiveAttack`. Happy to extract to a handler if preferred.

[Generated with Claude Code](https://claude.com/claude-code)


megamek/src/megamek/client/bot/BotClient.java
```diff
@@ -164,8 +175,21 @@ public AIType getAIType() {
...
+                  getName(), getGame().getPhase(), game.getTurnIndex());
+            try {
+                calculateMyTurn();
+                flushConn();
+                LOGGER.info("{}: calc thread finished for phase {}, turnIndex {}",
+                      getName(), getGame().getPhase(), game.getTurnIndex());
+            } catch (Throwable t) {
+                LOGGER.error(t, "{}: calc thread DIED for phase {}, turnIndex {} - the server is now "
+                      + "waiting on a move this bot will never send",
+                      getName(), getGame().getPhase(), game.getTurnIndex());
```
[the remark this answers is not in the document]
HammerGS: Verified and fixed in 5fb28ff58d. MMLogger's varargs overload does substitute {} at runtime (parametrizedStringIfEnabled), so the call worked - but this is a cold exception path, so it is now plain concatenation, which is clearer and matches CodeQL's model.

megamek/src/megamek/client/bot/princess/Princess.java
```diff
@@ -4019,6 +4118,35 @@ private MovePath performPathPostProcessing(final RankedPath path) {
...
+        // has forgotten an entity mid-calculation, which the bot's own turn thread can race into.
+        //
+        // So an embellishment that fails now costs only itself. The path the ranker chose is still a legal,
+        // fully-formed move, and moving is always better than standing still because a searchlight failed.
+        try {
+            return embellishPath(path, expectedDamage);
+        } catch (Exception exception) {
+            LOGGER.error(exception,
+                  "Post-processing failed for {}; using the un-embellished path rather than losing the turn",
+                  path.getEntity().getDisplayName());
```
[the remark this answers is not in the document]
HammerGS: Verified and fixed in 5fb28ff58d - same resolution as the BotClient call: runtime behavior was correct ({} substitution), converted to concatenation for clarity on the cold path.

megamek/unittests/megamek/client/bot/BotClientTest.java
```diff
@@ -0,0 +1,164 @@
...
+    private static final class RecordingBotClient extends BotClient {
+
+        private boolean dismissedItself = false;
+
+        RecordingBotClient() {
+            super("tester", "localhost", 0);
+        }
+
+        @Override
+        public void die() {
```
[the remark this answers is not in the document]
HammerGS: Fixed in 5fb28ff58d - the test stub's die() override is now synchronized, matching AbstractClient's contract.

megamek/src/megamek/client/bot/princess/AerospaceFireControl.java
```diff
@@ -0,0 +1,660 @@
...
+    }
+
+    /**
+     * Whether this is one airborne aerospace unit shooting at another, in an atmosphere and on the same board.
+     *
+     * <p>Space has no altitude levels and therefore no dead zone, and a cross-board shot has its positions
+     * substituted by the engine in ways this geometry does not model.</p>
+     */
+    private boolean isAtmosphericAirToAir(Entity shooter, EntityState shooterState, Targetable target,
+          EntityState targetState, Game game) {
```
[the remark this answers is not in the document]
HammerGS: Verified and fixed in 5fb28ff58d - the parameter was left over after the dead-zone check moved to pose-based geometry; removed, both callers updated.

megamek/unittests/megamek/common/scenario/ScenarioBombLoadingTest.java
```diff
@@ -0,0 +1,80 @@
...
+            List<BombMounted> bombs = entity.getBombs(AmmoType.F_GROUND_BOMB);
+            if (bombs.isEmpty()) {
+                continue;
+            }
+            sawBomber = true;
+            int shots = bombs.stream().mapToInt(BombMounted::getBaseShotsLeft).sum();
+            assertTrue(shots > 0, entity.getShortName() + " mounts " + bombs.size()
+                  + " ground bomb(s) but has " + shots + " shots - the racks are empty");
+        }
+        assertFalse(!sawBomber, "the scenario's fighters must mount their declared bombs at all");
```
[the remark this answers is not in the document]
HammerGS: Accepted, fixed in 5fb28ff58d - assertTrue(sawBomber, ...).

megamek/unittests/megamek/common/units/VerticalLiftoffRulesTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+        assertTrue(spheroid.canTakeOffVertically(), "spheroids always lift vertically");
+    }
+
+    @Test
+    void liftingOffSpendsTwoThrustPoints() {
+        AeroSpaceFighter winded = new AeroSpaceFighter();
+        winded.setOriginalWalkMP(1);
+        assertFalse(winded.canTakeOffVertically(),
+              "one thrust point cannot pay the two the liftoff costs");
```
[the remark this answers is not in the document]
HammerGS: Accepted, fixed in 5fb28ff58d - message now reads 'one thrust point cannot pay the two-point liftoff cost'.

megamek/src/megamek/common/pathfinder/AeroGroundDoctrinePathFinder.java
```diff
@@ -0,0 +1,491 @@
...
+        // NoE is deliberately NOT offered. The strafe window candidate above covers gun work at
+        // altitude 3; dropping to 1 buys nothing but the NoE +2 to-hit, the strafing dead-zone
+        // rules, and a ground that catches every failed control roll.
+        // A live Cheetah loitered at altitude 1 for five rounds on this candidate, where any failed
+        // control roll is the ground, and died there.
+
+        List<Integer> candidates = new ArrayList<>(altitudes);
+        if (candidates.size() > MAXIMUM_CANDIDATE_ALTITUDES) {
+            return candidates.subList(0, MAXIMUM_CANDIDATE_ALTITUDES);
+        }
```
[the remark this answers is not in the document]
HammerGS: Rejected with reasoning: the truncation priority is deliberate. When three or more committed enemy air altitudes are live, that turn is an air fight and matching them outranks the ground-attack windows; the windows return as the air thins, and in practice the LinkedHashSet dedupe usually collapses the set (strike altitude 5 is often also an enemy altitude). This has been the strike candidate's behavior since the finder was introduced; the strafe candidate follows the same rule. Happy to revisit if review prefers a reserved slot.

megamek/src/megamek/ai/dataset/UnitState.java
```diff
@@ -173,4 +182,27 @@ public static UnitState fromEntity(Entity entity, Game game) {
...
+    private static String removalConditionName(int removalCondition) {
+        return switch (removalCondition) {
+            case IEntityRemovalConditions.REMOVE_IN_RETREAT -> "IN_RETREAT";
+            case IEntityRemovalConditions.REMOVE_PUSHED -> "PUSHED";
+            case IEntityRemovalConditions.REMOVE_CAPTURED -> "CAPTURED";
+            case IEntityRemovalConditions.REMOVE_SALVAGEABLE -> "SALVAGEABLE";
+            case IEntityRemovalConditions.REMOVE_EJECTED -> "EJECTED";
+            case IEntityRemovalConditions.REMOVE_DEVASTATED -> "DEVASTATED";
+            case IEntityRemovalConditions.REMOVE_NEVER_JOINED -> "NEVER_JOINED";
+            default -> "ACTIVE";
```
IllianiBird: I feel like these should be consts somewhere, rather than magic strings. It'll make maintaining easier.

megamek/src/megamek/client/bot/caspar/Caspar.java
```diff
@@ -83,9 +111,270 @@ public void initializePathRankers() {
...
+            return takeoffPath;
+        }
+        MovePath landingPath = buildLandingPath(entity);
+        if (landingPath != null) {
+            return landingPath;
+        }
+        return super.continueMovementFor(entity);
+    }
+
+    private MovePath buildTakeoffPath(final Entity entity) {
```
IllianiBird: This can return null under normal circumstances so should include the Jakarta Nullable annotation.

megamek/src/megamek/client/bot/caspar/Caspar.java
```diff
@@ -83,9 +111,270 @@ public void initializePathRankers() {
...
+        return false;
+    }
+
+    /**
+     * Rough-in of the LAND order for the bot commands panel: an airborne DropShip or small craft
+     * under a standing LAND order sets down vertically as soon as it is legal - altitude 1, the
+     * capability, and clear ground below. No approach doctrine yet (the ship does not deliberately
+     * descend to reach the window); that ships with the landing doctrine block.
+     */
+    private MovePath buildLandingPath(final Entity entity) {
```
IllianiBird: This can return null under normal circumstances so should include the Jakarta Nullable annotation.

megamek/src/megamek/client/bot/caspar/Caspar.java
```diff
@@ -83,9 +111,270 @@ public void initializePathRankers() {
...
+        if (!(entity instanceof IAero aero) || !entity.isAero() || !entity.isAirborne()
```
IllianiBird, line 251: I don't think `!entity.isAero() || !entity.isAirborne()` is necessary, both should be unreachable if the object isn't an inheritor of IAero

megamek/src/megamek/client/bot/caspar/Caspar.java
```diff
@@ -83,9 +111,270 @@ public void initializePathRankers() {
...
+        double moveIndex = super.calculateMoveIndex(entity, msg);
+        if (!isAtmosphericAerospace(entity)) {
+            return moveIndex;
+        }
+
+        double reactionFactor = reactionFactor(entity);
+        if (reactionFactor != 1.0) {
+            msg.append("\n\t\tx")
+                  .append(String.format("%.2f", reactionFactor))
+                  .append(" (aerospace reaction: enemy fighters committed)");
```
IllianiBird, line 308: Looks like there might be player facing hardcoded strings here

megamek/src/megamek/client/bot/princess/AerospaceFireControl.java
```diff
@@ -0,0 +1,660 @@
...
+        return plan;
+    }
+
+    /**
+     * Considers a strafing run alongside the stock per-target plans and flies whichever bids
+     * higher. The stock chooser never strafes at all; this is the whole of PC-16-01's strafing
+     * half for the bot.
+     */
+    @Override
+    FiringPlan getBestFiringPlan(final Entity shooter, final IHonorUtil honorUtil, final Game game,
```
IllianiBird, line 178: It looks like best can return null under normal circumstances so the method should include the Jakarta Nullable annotation.

megamek/src/megamek/client/bot/princess/AerospaceFireControl.java
```diff
@@ -0,0 +1,660 @@
...
+     * Builds the best legal strafing run over this turn's flown line (TW p.243): the straight
+     * window of at most five passed-through hexes covering the most enemy ground units, every
+     * eligible energy weapon rolling once against every unit under it. Each shot carries the
+     * strafing flags so the server resolves it as a strafe, and heat lands once per weapon
+     * however many targets its run crosses. Nap-of-the-earth strafing is deliberately out of this
+     * first pass - the +2 and its dead-zone rules make altitude 2-3 the honest window.
+     *
+     * @return the strafe plan, empty when no legal run covers anyone
+     */
+    FiringPlan getStrafePlan(final Entity shooter, final Game game) {
```
IllianiBird, line 203: Needs Jakarta nullable

megamek/src/megamek/client/bot/princess/AerospaceFireControl.java
```diff
@@ -0,0 +1,660 @@
...
+        Entity nominalTarget = bestVictims.getFirst();
+        for (Entity victim : bestVictims) {
+            if (victim.calculateBattleValue() > nominalTarget.calculateBattleValue()) {
+                nominalTarget = victim;
+            }
+        }
+        FiringPlan strafePlan = new FiringPlan(nominalTarget);
+        for (WeaponMounted weapon : strafeWeapons) {
+            boolean firstShot = true;
+            for (Entity victim : bestVictims) {
```
IllianiBird, line 257: This is n(o), is there any way to convert it to n(1)?

megamek/src/megamek/client/bot/princess/AerospaceFireControl.java
```diff
@@ -0,0 +1,660 @@
...
+              shooter.getDisplayName(), strafePlan.size(), bestVictims.size(),
+              Math.round(strafePlan.getUtility()));
+        return strafePlan;
+    }
+
+    /** The enemy ground units a strafing window can hit: on it, not airborne, not dug into a building. */
+    private List<Entity> strafeVictims(Entity shooter, Game game, List<Coords> window) {
+        List<Entity> victims = new ArrayList<>();
+        for (Coords hex : window) {
+            for (Entity unit : game.getEntitiesVector(hex, shooter.getPassedThroughBoardId())) {
```
IllianiBird, line 286: How heavy is this? If we can expect to fire this check multiple times a phase we might need to find a way to make it n(1), if not then it's a non-issue as the number of parses should be small

megamek/src/megamek/client/bot/princess/AerospaceFireControl.java
```diff
@@ -0,0 +1,660 @@
...
+     */
+    static double bombPlanUtilityAdjustment(double stockExpectedDamage, double payloadExpectedDamage,
+          int bombsReleased, int targetEffectiveHitPoints, int targetBattleValue,
+          boolean otherBombTargetsRemain) {
+        int effectiveHitPoints = Math.max(1, targetEffectiveHitPoints);
+        double honestDamage = Math.max(stockExpectedDamage, payloadExpectedDamage);
+        double cappedDamage = Math.min(honestDamage, effectiveHitPoints);
+        double killFraction = cappedDamage / effectiveHitPoints;
+        double killCredit = BOMB_KILL_UTILITY * killFraction * killFraction
+              * (targetBattleValue / 1000.0);
```
IllianiBird, line 431: It's not clear why we're dividing `targetBattleValue` by `1000`, can we have a comment clarifying it?

megamek/src/megamek/client/bot/princess/AerospaceFireControl.java
```diff
@@ -0,0 +1,660 @@
...
+        return selection;
+    }
+
+    /**
+     * The hex on the flown line whose blast footprint delivers the most total damage across every
+     * enemy ground unit, or {@code null} when no flown hex delivers anything. Uses the executed
+     * move's own hexes ({@link Entity#getPassedThrough}) in preference to the planning-time path,
+     * because at firing time the flown line is a fact.
+     */
+    private Coords bestFootprintAimHex(Entity shooter, MovePath flightPath, Game game) {
```
IllianiBird: Nullable

megamek/src/megamek/client/bot/princess/AerospaceFireControl.java
```diff
@@ -0,0 +1,660 @@
...
+        if (targets.isEmpty()) {
+            return null;
+        }
+        Coords bestAimHex = null;
+        double bestFootprint = 0;
+        for (Coords aimPoint : flownLine) {
+            double footprint = 0;
+            for (Entity target : targets) {
+                int ring = aimPoint.distance(target.getPosition());
+                for (BombMounted bomb : groundBombs) {
```
IllianiBird: This is n(o)(o), we really shouldn't be nested a for-loop inside a for-loop inside a for-loop. That'll be really expensive. We need to find a cheaper way to get the information you're looking for.

megamek/src/megamek/client/bot/princess/AerospaceGeometry.java
```diff
@@ -0,0 +1,280 @@
...
+     */
+    public static int hexesToNearestEdge(Coords position, Board board) {
+        if ((position == null) || (board == null)) {
+            return Integer.MAX_VALUE;
+        }
+        int fromWest = position.getX();
+        int fromEast = board.getWidth() - 1 - position.getX();
+        int fromNorth = position.getY();
+        int fromSouth = board.getHeight() - 1 - position.getY();
+        return Math.min(Math.min(fromWest, fromEast), Math.min(fromNorth, fromSouth));
```
IllianiBird: This is really clunky and hard to parse. Let's simplify it to...

```
int westToEast = Math.min(fromWest, fromEast);
int northToSouth = Math.min(fromNorth, fromSouth);
return Math.min(westToEast, northToSouth);
```

megamek/src/megamek/client/bot/princess/AerospacePathRanker.java
```diff
@@ -0,0 +1,1990 @@
...
+        return lastWinchester;
+    }
+
+    int lastPostAttackAltitudeForTest() {
+        return lastPostAttackAltitude;
+    }
+
+    void lastPostAttackAltitudeForTest(int altitude) {
+        lastPostAttackAltitude = altitude;
+    }
```
IllianiBird, line 416: We shouldn't be having functions in the middle of the variable declaration block

megamek/src/megamek/client/bot/princess/AerospacePathRanker.java
```diff
@@ -0,0 +1,1990 @@
...
+        // seam hex between two meks delivers full no-falloff cluster damage to BOTH at hex to-hit
+        // odds. The best aim point is a search over the line, not a lookup of enemy positions.
+        double bombRun = 0;
+        if (inDiveBombWindow) {
+            List<BombMounted> groundBombs = mover.getBombs(AmmoType.F_GROUND_BOMB);
+            for (Coords aimPoint : flownLine) {
+                double footprint = 0;
+                for (Entity target : targets) {
+                    int ring = aimPoint.distance(target.getPosition());
+                    for (BombMounted bomb : groundBombs) {
```
IllianiBird: n(o)(o) again :/

megamek/src/megamek/client/bot/princess/AerospacePathRanker.java
```diff
@@ -0,0 +1,1990 @@
...
+        // approach premium - the astern pass over the walking column is the whole play.
+        double strafeRun = 0;
+        double strafeDamage = strafeEligibleDamage(mover);
+        boolean inStrafeWindow = (finalAltitude >= 2) && (finalAltitude <= 3) && (strafeDamage > 0)
+              && !mover.isSpheroid();
+        if (inStrafeWindow) {
+            List<Coords> orderedLine = orderedFlownLine(path);
+            for (List<Coords> window : straightWindows(orderedLine, 5)) {
+                double windowValue = 0;
+                for (Entity target : targets) {
```
IllianiBird: Similar to my earlier comment, I wonder if we need to convert this from n(o) to n(1)

megamek/src/megamek/client/bot/princess/AerospacePathRanker.java
```diff
@@ -0,0 +1,1990 @@
...
+    }
+
+    /**
+     * Credit for exiting the turn astern of the enemy ground force's direction of travel. The
+     * force's dominant drift direction (majority vote of the per-unit movement vectors) points
+     * where the column is going; a fighter whose final pose sits behind the force - the direction
+     * FROM the fighter TO the force agreeing with the drift - is positioned to roll in on rear
+     * arcs next round. Zero when the force is not moving or the fighter is on top of it.
+     */
+    double sternSetupCredit(MovePath path, Game game, List<Entity> enemies) {
```
IllianiBird: I think CodeQL is right, I'm not seeing game used here

megamek/src/megamek/client/bot/princess/AerospacePathRanker.java
```diff
@@ -0,0 +1,1990 @@
...
+     * the legal shapes of a strafing run (TW p.243: consecutive, one straight line, five hexes at
+     * most). Shared by the movement ranker (which steers toward good windows) and the fire control
+     * (which builds the actual run), like the bomb ring tables.
+     */
+    static List<List<Coords>> straightWindows(List<Coords> orderedLine, int maxLength) {
+        List<List<Coords>> windows = new ArrayList<>();
+        for (int start = 0; start < orderedLine.size(); start++) {
+            windows.add(List.of(orderedLine.get(start)));
+            Integer direction = null;
+            for (int end = start + 1; end < orderedLine.size(); end++) {
```
IllianiBird, line 1220: Depending on how often this is called, I can see it getting expensive, especially for fast units traveling in long straight lines.

megamek/src/megamek/common/units/IAero.java
```diff
@@ -739,7 +739,23 @@ default boolean canLandHorizontally() {
...
+            return false;
+        }
+        if (isVSTOL() || isSpheroid()) {
+            return true;
+        }
+        Entity self = (Entity) this;
+        // Conventional fighters need VSTOL gear (handled above); aerospace fighters and aerodyne
+        // small craft may always lift vertically.
+        return (self instanceof AeroSpaceFighter && !(self instanceof ConvFighter))
+              || (self instanceof SmallCraft && !(self instanceof Dropship));
```
IllianiBird, line 760: What about aerodyne DropShips? Are there non-aerodyne SmallCraft?

IllianiBird (changes requested):
I'm seeing a lot of inefficient parsing, some n(o)(o) processing where we're nesting for inside a for inside a for, for example. There's also some missing annotation and a few other minor issues. With a PR this big, it's not going to be possible to catch all the issues, I did my best.

megamek/src/megamek/ai/dataset/UnitState.java
HammerGS: Accepted, fixed in 90d40f8bce - the removal-state tokens are now named constants (REMOVAL_IN_RETREAT etc.) beside the switch that maps them.

megamek/src/megamek/client/bot/caspar/Caspar.java
HammerGS: Accepted, fixed in 90d40f8bce - @Nullable added (used the project's megamek.common.annotations.Nullable, which this package imports throughout; happy to switch if jakarta is the new direction).

megamek/src/megamek/client/bot/caspar/Caspar.java
HammerGS: Accepted, fixed in 90d40f8bce - @Nullable added.

megamek/src/megamek/client/bot/caspar/Caspar.java
HammerGS: Declined with reasoning - both checks are reachable through IAero implementors: a grounded fighter is an IAero with isAirborne() false (that state is what the takeoff feature exists for, and buildLandingPath must only fire for airborne units), and a LandAirMek is an IAero in every mode while isAero() is fighter-mode-only (LandAirMek.isAero() returns getConversionMode() == CONV_MODE_FIGHTER), so the isAero() check keeps a mek-mode LAM out of aero landing paths. Kept both; happy to add a clarifying comment if useful.

megamek/src/megamek/client/bot/caspar/Caspar.java
HammerGS: Declined with reasoning - these strings append to the move-index diagnostic trace (msg) that goes to the bot logger, same pattern as the stock trace lines around it ("x1.5 (is Tank)" etc.); they are never rendered to players. If the trace convention changes to i18n I'll follow it.

megamek/src/megamek/client/bot/princess/AerospaceFireControl.java
HammerGS: Accepted, fixed in 90d40f8bce - @Nullable added (super can return null and so can this override).

megamek/src/megamek/client/bot/princess/AerospaceFireControl.java
HammerGS: Accepted, fixed in 90d40f8bce - @Nullable added, and the javadoc corrected (it said 'empty', the method returns null).

megamek/src/megamek/client/bot/princess/AerospaceFireControl.java
HammerGS: Declined with reasoning - the weapons x victims nesting here IS the output, not a search: TW p.243 resolves a strafe as one attack per weapon per victim, so each iteration constructs one WeaponAttackAction the server requires. Upper bound is about six strafe-eligible weapons x five victims = thirty constructions, once per firing turn for one unit. There is no smaller set of attack actions that represents the same strafe.

megamek/src/megamek/client/bot/princess/AerospaceFireControl.java
HammerGS: Answered with numbers in 90d40f8bce - it runs once per fighter per firing phase (not per candidate path; the movement side prices strafes separately from ranker state). Cost was windows (~200 on a long line) x 5 hexes x one position-map lookup. Now cheaper still: each flown hex is queried once into a map (~48 lookups) and windows read the map. Well under a millisecond per firing turn.

megamek/src/megamek/client/bot/princess/AerospaceFireControl.java
HammerGS: Accepted, fixed in 90d40f8bce - comment added: BV/1000 normalizes battle value into utility-scale units, so a full kill of a 1,000-BV target earns exactly BOMB_KILL_UTILITY points, a 2,000-BV target twice that.

megamek/src/megamek/client/bot/princess/AerospaceFireControl.java
HammerGS: Accepted, fixed in 90d40f8bce - @Nullable added.

megamek/src/megamek/client/bot/princess/AerospaceFireControl.java
HammerGS: Accepted, fixed in 90d40f8bce - the bombs dimension is gone: bombRingProfile precomputes the payload's summed damage per blast ring once (rings never exceed three hexes), and the search is now flown-line x targets with an O(1) array lookup. Same fix applied to the identical triple loop in AerospacePathRanker.scoreAttackRuns and the double loop in enemiesUnderFootprint, since the shape recurred.

megamek/src/megamek/client/bot/princess/AerospaceGeometry.java
HammerGS: Accepted, fixed in fd756ca74b - applied exactly as suggested.

megamek/src/megamek/client/bot/princess/AerospacePathRanker.java
HammerGS: Accepted, fixed in fd756ca74b - the stragglers moved up to join the class's field block; declarations now all sit above the methods.

megamek/src/megamek/client/bot/princess/AerospacePathRanker.java
HammerGS: Already fixed in 90d40f8bce alongside the same shape you flagged in AerospaceFireControl - bombRingProfile precomputes the payload's per-ring damage once, so this site is now line x targets with an O(1) ring lookup; the bombs dimension is gone.

megamek/src/megamek/client/bot/princess/AerospacePathRanker.java
HammerGS: Accepted, fixed in fd756ca74b - the ranker no longer enumerates windows at all: a per-hex target-value map built once, then a one-pass sliding sum over the straight segments of the flown line. O(line length) per candidate path. A new test pins the sliding sum to the enumeration it replaced across straight, bent, and gapped lines.

megamek/src/megamek/client/bot/princess/AerospacePathRanker.java
HammerGS: Accepted, fixed in fd756ca74b - CodeQL was right, the parameter was left over from an earlier draft that read the round number from Game. Removed.

megamek/src/megamek/client/bot/princess/AerospacePathRanker.java
HammerGS: Accepted in effect, fixed in fd756ca74b - the expensive caller is gone: the movement ranker (which ran this per candidate path) now uses the one-pass sliding sum above instead. straightWindows itself survives only in the fire control, once per fighter per firing phase, where a worst-case 48-hex line yields ~240 windows a turn.

megamek/src/megamek/common/units/IAero.java
HammerGS: Answered, comment expanded in fd756ca74b - aerodyne DropShips are deliberately excluded: SO gives them the vacuum-only rule and Dropship has its own canTakeOffVertically override; since Dropship extends SmallCraft, the instanceof exclusion here is what keeps them on that override. Non-aerodyne small craft are the spheroids, and they never reach this line - the isSpheroid() branch above returns true for them first.

IllianiBird (approved):

PR #8729, opened 2026-08-16
F_PHYSICAL_WEAPON

exeea:
F_PHYSICAL_WEAPON for all physical weapons (only Mek, not affecting BA physical weapons, we need to decide what to do there...)

this PR also restores the lance +1 tohit modifier


TenkawaBC (approved):
Looks good to me on these.

PR #8730, opened 2026-08-16
Implement the Oblique Artilleryman range extension (Fixes #8726)

### LINKED ISSUE #8726: [Issue] 0.51.00 Oblique Artilleryman SPA makes Artillery scatter 0 Hexes regardless of degree of failure
### Brief Description *

The Oblique Artilleryman SPA is supposed to decrease scatter by 2 hexes (to a minimum of 0), so one would expect that if on were to miss the target number by example 10 it would decrease the scatter to 8 hexes away from the target hex for an artillerypiece.

In testing the scatter was 0 every time for a unit with the Oblique Artilleryman SPA.

Image

Image

Testing without the Oblique Artilleryman SPA gives expected scatter results.

Image

### 3. Steps to Reproduce

1. Add two units with artillery weapons to your force.
2. worsen both units gunnery to 8 to help ensure missing for scatter tests.
3. Give one (1) of the units the Oblique Artilleryman SPA
4. Load in on a map and spend a number of turns shooting at visible hexes.
5. Notice that the Artillery unit with the SPA will never scatter its shots.

### Attach Files

_No response_

### Severity *

### MegaMek Suite Version *

0.51.00

### Operating System *

Windows 10

### Java Version *

    Java Vendor: Eclipse Adoptium     Java Version: 21.0.11

### Final Verification

HammerGS:
## Root Cause

The 0.51.0 release worked out a missed shot's drift as `Math.max(toHit.getMoS() + 2, 0)`. Margin of
failure is stored as a negative number, so adding 2 left it negative and the clamp flattened every
miss to zero hexes - a shot that missed by 10 landed dead on the target hex. Only the single-weapon
artillery handler was affected; the bay and bomb handlers negated the margin first.

The Advanced Scatter work (5aebecfaf3, 27 June, after the v0.51.0 tag) already rewrote that line as
`Math.max(|MoS| - reduction, 0)`, so the reported bug is fixed on development. The regression tests
here stop it coming back.

The ability's other half was never implemented: CamOps p.78 also grants +10% range. An IS Arrow IV is
rated 8 map sheets = 136 hexes, so an Oblique Artilleryman now reaches 150. The range check used to
compare whole map sheets (`ceil(distance / 17) > maxRange`) and now compares hexes, because 10% is
less than a map sheet; the two are algebraically identical, so nobody else's range changes.

The extension applies to `F_ARTILLERY` weapons, which is exactly the CamOps artillery range table -
Arrow IV, Thumper, Sniper, Long Tom, BA Tube, cruise missiles. Artillery cannons are a separate
weapon class and are excluded, as are capital missiles fired at ground targets.

## Changes

1. `ArtilleryRange` (new) - artillery range rules in one stateless place: rated map sheets, the
   gravity modifier (TO:AR p.155), the ability's extension, and the artillery-piece gate. Shared by
   the engine and both UI call sites.
2. `ComputeToHitIsImpossible` - inline range block replaced by one call; the class ends up 29 lines
   smaller.
3. `WeaponPanel` - an extended range shows its fraction and an asterisk (`8.8*` Arrow IV, `19.8*`
   Sniper) with a tooltip naming the ability. Aerospace displays show attack values and are left alone.
4. `CustomMekDialog` - the off-board slider allows the extra reach: a Sniper battery goes from 289 to
   320 hexes.
5. `ArtilleryWeaponDistantFireHandler` - a miss that does not drift now reports "misses the intended
   hex" and draws no drift arrow, instead of "scatters to" the hex it was aimed at, which read like
   the bug above. The bay handler skips the zero-length arrow too.
6. `SpaCatalog` + `messages.properties` - ability marked FULL, stale "not implemented" note removed,
   description names both effects.

## Files Changed

- `megamek/src/megamek/common/compute/ArtilleryRange.java` - NEW: artillery range rules
- `megamek/unittests/megamek/common/compute/ArtilleryRangeTest.java` - NEW: 5 tests
- `megamek/src/megamek/common/actions/compute/ComputeToHitIsImpossible.java` - calls ArtilleryRange
- `megamek/src/megamek/client/ui/dialogs/unitDisplay/WeaponPanel.java` - extended range display
- `megamek/src/megamek/client/ui/dialogs/customMek/CustomMekDialog.java` - off-board slider maximum
- `megamek/src/megamek/common/weapons/handlers/artillery/ArtilleryWeaponDistantFireHandler.java` - no-drift miss report
- `megamek/src/megamek/common/weapons/handlers/artillery/ArtilleryBayWeaponDistantFireHandler.java` - no zero-length drift line
- `megamek/src/megamek/common/options/SpaCatalog.java` - PARTIAL -> FULL
- `megamek/resources/megamek/client/messages.properties` - range marker and tooltip
- `megamek/resources/megamek/common/options/messages.properties` - ability description; stale note removed
- `megamek/unittests/megamek/common/compute/scatter/ScatterTest.java` - 2 regression tests
- `megamek/unittests/megamek/common/options/SpaCatalogTest.java` - status counts

## Testing

`ArtilleryRangeTest` (5): rated range, extended range, extension under 1.2 gravity, the
artillery-cannon exclusion, the display value. `ScatterTest` (2): a negative margin of failure with
the reduction, which is the exact shape that broke. Both sets were run against the original defects
reintroduced on purpose and failed as they should before passing again. Full `./gradlew build` green.

Played in game with a Sniper battery and a control unit alongside:

- Scatter proven as a pair - both needed 12 and both rolled 6; the Oblique Artilleryman drifted 4
  hexes, the control 6.
- Off-board slider: 320 with the ability, 289 without.
- Weapon panel `3 - 19.8*` with tooltip, nothing on non-artillery weapons.
- Ability description and FULL status in Configure Unit.

## What is NOT proven yet

- The range boundary has not been fired in game - a shot at exactly the extended maximum, and one hex
  past it, rests on the unit tests.
- The new "misses the intended hex" line has not been seen in game.
- The playtest predates the extraction of `ArtilleryRange`; values are unchanged and covered by tests,
  but the dialogs have not been reopened since.
- Artillery cannons and direct-fire artillery are excluded per code and unit test, not in game.
- The panel shows the rated extension and ignores gravity, as it always has for every weapon, so on a
  high-gravity world the displayed figure reads higher than the engine's actual limit.

Fixes #8726


megamek/src/megamek/client/ui/dialogs/customMek/CustomMekDialog.java
```diff
@@ -1395,38 +1396,37 @@ public void actionPerformed(ActionEvent actionEvent) {
...
-                        if (w instanceof ArtilleryBayWeapon) {
-                            // Artillery bays can mix and match, so limit the bay
-                            // to the shortest range of the weapons in it
-                            nDistance = getBayShortestRange(wep);
-                        } else {
-                            // Max TO range in map sheets - 1 for the actual play area
-                            nDistance = (w.getLongRange() - 1);
-                        }
-                    } else if (w.isCapital() || w.isSubCapital()) {
+                for (WeaponMounted weaponMount : entity.getWeaponList()) {
```
IllianiBird, line 1399: Is there any way we can make this o(1) instead of o(n)?

IllianiBird (approved):
Approved, though there is one small thing I'd like looked into

megamek/src/megamek/client/ui/dialogs/customMek/CustomMekDialog.java
HammerGS: Looked into it, and I do not think O(1) is reachable here - but there was a real inefficiency hiding in the block, so thank you for the poke.

The answer this loop produces is "how far back can these units deploy and still reach the board", which depends on which weapons they are carrying. Something has to look at each weapon at least once to know that, so the honest floor is O(weapons). The engine's own equivalent, `Entity.getMaxWeaponRange()`, has the same shape for the same reason, and Princess wraps it in a cache rather than computing it more cheaply.

Caching it on the unit would move the cost rather than remove it, and it would go stale the moment a player changes equipment in this very dialog, so I would rather not.

What I could fix is the constant factor: the capital branch was asking each weapon for its range bracket up to four times, and `getMaxRange(WeaponMounted)` is not a field read - it follows the mount's linked ammo and branches on ATM/MML munitions. That is now one lookup feeding a switch (814c4bfde6).

For scale: this runs once, when the player clicks the Offboard Distance button, over the units currently being customised - usually one - times their weapons. It is not on any per-frame or per-turn path.

PR #8731, opened 2026-08-16
Fix server crash when Semi-Guided missiles are fired at a hex or building (Fixes #8723)

### LINKED ISSUE #8723: [0.51.00.1] TAG appears to cause game to hang
### Brief Description *

Last entry in the log: 
```
19:48:22,071 ERROR [megamek.server.Server] {Connection 4}
megamek.server.Server.handle(Server.java:1432) - Uncaught exception handling CLIENT_NAME packet from connection 4 - the packet was dropped
java.lang.NullPointerException: Cannot invoke "megamek.common.units.Entity.getTaggedBy()" because "entityTarget" is null
    at megamek.common.weapons.handlers.lrm.LRMHandler.calcHits(LRMHandler.java:323)
    at megamek.common.weapons.handlers.MissileWeaponHandler.handle(MissileWeaponHandler.java:963)
    at megamek.server.totalWarfare.TWGameManager.handleAttacks(TWGameManager.java:32636)
    at megamek.server.totalWarfare.TWGameManager.handleAttacks(TWGameManager.java:32554)
    at megamek.server.totalWarfare.TWPhaseEndManager.managePhase(TWPhaseEndManager.java:169)
    at megamek.server.totalWarfare.TWGameManager.endCurrentPhase(TWGameManager.java:2419)
    at megamek.server.totalWarfare.TWGameManager.sendCurrentInfo(TWGameManager.java:955)
    at megamek.server.Server.sendCurrentInfo(Server.java:795)
    at megamek.server.Server.receivePlayerName(Server.java:751)
    at megamek.server.Server.handle(Server.java:1364)
    at megamek.server.Server$1.packetReceived(Server.java:275)
    at megamek.common.net.connections.AbstractConnection.processConnectionEvent(AbstractConnection.java:404)
    at megamek.common.net.connections.AbstractConnection.processPacket(AbstractConnection.java:363)
    at megamek.common.net.connections.AbstractConnection.update(AbstractConnection.java:317)
    at megamek.server.ConnectionHandler.run(ConnectionHandler.java:67)
    at java.base/java.lang.Thread.run(Thread.java:1474)
```

The game was played in core rules mode. The game was unable to proceed past this point.

### 3. Steps to Reproduce

1. Bring tagging mechs
2. Tag enemies
3. Fire at enemies
4. Game hangs? Exact conditions unclear, but this issue reproduced from a previous save with different targets

### Attach Files

[autosave.sav.gz](https://github.com/user-attachments/files/31107814/autosave.sav.gz)
[megamek.log](https://github.com/user-attachments/files/31107812/megamek.log)
[tagfail.sav.gz](https://github.com/user-attachments/files/31107813/tagfail.sav.gz)

### Severity *

### MegaMek Suite Version *

0.50.00.1

### Operating System *

Fedora Linux 44

### Java Version *

    Java Vendor: Red Hat, Inc.     Java Version: 25.0.4 (Fedora 44 packaged openjdk)

### Final Verification

HammerGS:
## What was broken

Firing Semi-Guided LRM or MML ammunition at a building - or at a hex you picked for clearing - froze the
game. The weapons fire phase never finished, and reconnecting just made the server hit the same error
again. TAG turned out to be a red herring: no TAG unit needs to be on the board, and it happens under
both rulesets.

The Semi-Guided branch in `LRMHandler` read the TAG designation straight off the target unit. Every
neighbouring branch - Artemis, ATM, Narc - checks first that the target *is* a unit, because there is no
target unit when you shoot a building or a hex.

## Changes

- Ask the game for the designation through `Compute.isTargetTagged`, which is what the to-hit code
  already uses for this munition. It reads the designation off units and walks the game's TAG list for
  everything else, so a TAG'd building now gets its designation bonus instead of crashing.
- Split the branch condition and the modifier lookup into two named helpers, so the rule can be tested
  without standing up a whole firing phase. The Artemis/ATM/Narc/Semi-Guided precedence is unchanged.

## Testing

Playtested in-game under both rulesets: the shots that used to freeze the game - Semi-Guided missiles
fired at a building, and at a hex targeted for clearing - now resolve normally and the phase completes.
Testing both rulesets matters here, because the crash was never Core-Rules-specific - Total Warfare
evaluates the same null read before discarding the result.

New `LRMHandlerSemiGuidedTest`, four cases: a hex nobody designated, a hex designated by TAG, a hex under
Total Warfare rules, and a unit designated by TAG. Putting the old expression back makes the three hex
cases fail with the exact `NullPointerException` from the issue report, so the test covers the reported
crash and not just the new code path. Full `./gradlew build` green.

## What is NOT proven yet

The Core Rules missile counts themselves (-1 undesignated, +2 direct-and-designated) come from the
existing `CoreRulesAmmo` implementation and were not re-checked against the book as part of this fix -
this PR changes which targets the rule can be applied to, not what the rule pays out.

Fixes #8723

### DIFF
--- a/megamek/src/megamek/common/weapons/handlers/lrm/LRMHandler.java
+++ b/megamek/src/megamek/common/weapons/handlers/lrm/LRMHandler.java
@@ -161,6 +161,40 @@ public class LRMHandler extends MissileWeaponHandler {
         return false;
     }
 
+    /**
+     * Returns {@code true} if the given ammunition is a Semi-Guided load for one of the long range missile launchers
+     * that can carry it, otherwise {@code false}.
+     *
+     * @param ammoType the ammunition loaded in the firing weapon
+     *
+     * @return {@code true} if this is a Semi-Guided load for a launcher that can carry it
+     */
+    private boolean isSemiGuidedMissileAmmo(AmmoType ammoType) {
+        boolean isSemiGuidedCapableLauncher = ammoType.getAmmoType()
+              .isAnyOf(AmmoType.AmmoTypeEnum.LRM,
+                    AmmoType.AmmoTypeEnum.LRM_IMP,
+                    AmmoType.AmmoTypeEnum.MML,
+                    AmmoType.AmmoTypeEnum.NLRM);
+        boolean isSemiGuidedMunition = ammoType.getMunitionType().contains(AmmoType.Munitions.M_SEMIGUIDED);
+        return isSemiGuidedCapableLauncher && isSemiGuidedMunition;
+    }
+
+    /**
+     * Returns the cluster roll modifier that Semi-Guided ammunition contributes to this attack, which depends on
+     * whether the target is currently designated by TAG. The modifier is zero under Total Warfare rules, where
+     * Semi-Guided ammunition improves the to-hit roll instead of the number of missiles that hit.
+     * <p>
+     * Buildings and hexes can be designated by TAG just like units can, so the designation is looked up through the
+     * game rather than on the target unit, which does not exist for those targets.
+     *
+     * @return the number of missiles to add to or subtract from the cluster roll
+     */
+    protected int getSemiGuidedClusterModifier() {
+        boolean isTargetTagged = Compute.isTargetTagged(target, game);
+        boolean isIndirectFire = weapon.curMode().equals("Indirect");
+        return Game.rulesManager.getRulesAmmo().getSemiGuidedNMissiles(isTargetTagged, isIndirectFire);
+    }
+
     @Override
     protected int calcHits(Vector<Report> vPhaseReport) {
         // Use effective rack size (reduced by 20% when incendiary mixed)
@@ -315,13 +349,8 @@ public class LRMHandler extends MissileWeaponHandler {
                         nMissilesModifier += 2;
                     }
                 }
-            } else if (((ammoType.getAmmoType() == AmmoType.AmmoTypeEnum.LRM)
-                  || (ammoType.getAmmoType() == AmmoType.AmmoTypeEnum.LRM_IMP)
-                  || (ammoType.getAmmoType() == AmmoType.AmmoTypeEnum.MML)
-                  || (ammoType.getAmmoType() == AmmoType.AmmoTypeEnum.NLRM))
-                  && (ammoType.getMunitionType().contains(AmmoType.Munitions.M_SEMIGUIDED))) {
-                nMissilesModifier += Game.rulesManager.getRulesAmmo().getSemiGuidedNMissiles((entityTarget.getTaggedBy() != WeaponAttackAction.UNASSIGNED),
-                      weapon.curMode().equals("Indirect"));
+            } else if (isSemiGuidedMissileAmmo(ammoType)) {
+                nMissilesModifier += getSemiGuidedClusterModifier();
             }
         }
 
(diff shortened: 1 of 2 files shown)


megamek/unittests/megamek/common/weapons/handlers/lrm/LRMHandlerSemiGuidedTest.java
```diff
@@ -0,0 +1,244 @@
...
+        doReturn(attackerId).when(attackAction).getEntityId();
+        doReturn(0).when(attackAction).getWeaponId();
+        doReturn(targetType).when(attackAction).getTargetType();
+        doReturn(targetId).when(attackAction).getTargetId();
+        doReturn("").when(toHit).getDesc();
+
+        AmmoType ammoType = mock(AmmoType.class);
+        doReturn(AmmoType.AmmoTypeEnum.LRM).when(ammoType).getAmmoType();
+        AmmoMounted ammo = mock(AmmoMounted.class);
+        doReturn(ammoType).when(ammo).getType();
```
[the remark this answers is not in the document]
HammerGS: Verified and fixed in 008bc0d67f - both halves of this are correct.

The fixture did claim to be a Semi-Guided load in its JavaDoc while never stubbing `getMunitionType()`, so it was describing something it was not actually set up as.

I also checked the second half rather than assuming it, because Mockito does return empty collections for some collection return types. It does not for this one: a probe against `mock(AmmoType.class).getMunitionType()` returns `null`, not an empty `EnumSet`. So a future test driving the full `calcHits` path through this helper really would have died on `getMunitionType().contains(M_SEMIGUIDED)` instead of on the behaviour under test.

The four existing tests call `getSemiGuidedClusterModifier()` directly and never touch the ammo, so this was not masking a failure today - it was a trap for the next person to extend the fixture. Stub added, tests still green.

megamek/src/megamek/common/weapons/handlers/lrm/LRMHandler.java
```diff
@@ -161,6 +161,40 @@ protected boolean specialResolution(Vector<Report> vPhaseReport,
...
+                    AmmoType.AmmoTypeEnum.NLRM);
+        boolean isSemiGuidedMunition = ammoType.getMunitionType().contains(AmmoType.Munitions.M_SEMIGUIDED);
+        return isSemiGuidedCapableLauncher && isSemiGuidedMunition;
+    }
+
+    /**
+     * Returns the cluster roll modifier that Semi-Guided ammunition contributes to this attack, which depends on
+     * whether the target is currently designated by TAG. The modifier is zero under Total Warfare rules, where
+     * Semi-Guided ammunition improves the to-hit roll instead of the number of missiles that hit.
+     * <p>
```
IllianiBird, line 186: Unclosed `<p>` tag

IllianiBird (approved):

megamek/src/megamek/common/weapons/handlers/lrm/LRMHandler.java
HammerGS: Good catch - fixed in 4641c27e8e.

I also found the same unclosed `<p>` in the `LRMHandlerSemiGuidedTest` class JavaDoc and closed that one in the same commit, so it does not come back at you as a second comment.

Worth noting the codebase is split on this - roughly 330 unclosed `<p>` separators against 88 files that close the tag - so if you want one of those styles as the standard I am happy to follow it on everything I touch from here.

Issue #8705, opened 2026-08-13
[Issue] Units with Bombast Lasers fail to load into the unit cache (NPE in ISBombastLaser.isExplosive)


HammerGS:
### Brief Description *

Units that mount a Bombast Laser and carry nothing explosive are silently dropped from the unit cache at startup. They never appear in the unit selector, so they cannot be added to a game, and the force generator and RAT loader then report them as missing.

Thirteen units fail on my machine:

```
fighters/XTRs/Mercenaries/Seydlitz SYD-45X Starling.blk
fighters/ProtoTypes/Lucifer III  LCR-3.blk
dropships/ProtoTypes/IS/Claymore V3.blk
dropships/XTRs/Steiner/Claymore Interceptor.blk
vehicles/3085u/Cutting Edge/Chalchiuhtotolin Support Tank (Bombast Laser).blk
ge/Archive_Turrets/Inner Sphere/Laser/Bombast/Bombast Turret (Single).blk
ge/Archive_Turrets/Inner Sphere/Laser/Bombast/Bombast Turret (Dual).blk
ge/Archive_Turrets/Inner Sphere/Laser/Bombast/Bombast Turret (Triple).blk
ge/Archive_Turrets/Inner Sphere/Laser/Bombast/Bombast Turret (Quad).blk
meks/XTRs/Mercenaries/Grasshopper GHR-7X.mtf
meks/XTRs/Boondocks/Merlin MLN-SX.mtf
meks/3145/Republic/Uraeus UAE-7R.mtf
meks/3145/NTNU RS/Berserker BRZ-D4.mtf
```

all with the same message:

```
WARN [MegaMekUnitSelectorDialog] Unit loading errors:
  {... = Cannot invoke "megamek.common.equipment.Mounted.getType()" because "mounted" is null, ...}
```

Every one of them mounts a Bombast Laser.

### Root cause

When MegaMek builds the unit cache it runs each unit through an AlphaStrike conversion to decide whether it qualifies for ENE ("no explosive component"). That check asks each weapon *type*, in the abstract, whether it is explosive, passing `null` where a specific installed weapon would normally go — it only cares about the equipment type at that point, not a particular mount:

`ASSpecialAbilityConverter.java:538` — `equipment.getType().isExplosive(null)`

`EquipmentType.isExplosive(Mounted, boolean)` handles that `null` properly (`EquipmentType.java:413`). The new Bombast Laser override does not — it dereferences the parameter immediately:

[...]

Issue #1234, opened 2019-01-26
Feature request : Option to toggle "next target" behaviour when clicked


Question2:
By default, when the next target button is clicked, it selects the nearest target. Shift+click will select the nearest valid target. Would it be possible to make a toggle in options to allow players to choose whether click should be nearest target or nearest valid target?

Issue #111, opened 2016-03-28
The Effects of Fire


scJazz2:
Per this comment from BT Forums and discussion above it.
http://bg.battletech.com/forums/index.php?topic=51280.msg1197736#msg1197736

TO pg 43 on Fire

And the unofficial MM Option fire burns down N TF per turn.

The TO rules state buildings burn at 2 per turn regardless of size/type. The MM Option allows the user to set how much damage Trees take per turn when on fire. There is nothing about Fields or other non-tree burnable stuff.

Fields should burn down like trees since that is obvious, perhaps even faster/slower.

Other terrains should probably burn down as well. If it has a Terrain Factor (except for Rough or including it) then it should be able to burn. In burning it should also be able to burn down.

Also Ref: TO pg 64

scJazz2:
as me and BW have gone over stuff... Rough, Clear, Sand... they have TFs but should not be able to be ignited. Anything else with a TF < 131 should burn at the Option NN rate. (TF <131) covers ultra heavy woods which exist on some maps.) This also avoids burning Pavement and Roads which have TF values.

BLOODWOLF333:
I second this request.

neoancient:
Closing as nobody has shown any interest in implementing this.

Issue #5329, opened 2024-04-07
49.19 Direct Fire Flak Attacks not supposed to receive standard Flak modifiers


RaozSpaz:
### Environment
49.19 MegaMek
Windows
Java 11

### Description
![image](https://github.com/MegaMek/megamek/assets/132851640/c21432ed-ae01-4f37-813e-9136efa13b13)
Was reading various artillery rules for a question on another server and decided to read over the Flak rules on page 153 of the TO:AR. In the second paragraph when it begins to talk about the various modifiers applied to the attack. It states "Apply a +3 to hit modifier. _Ignore the regular modifiers for using type "F" weapons or making a direct fire attack_"
Reading this section I don't think the -2 modifier from the Flak keyword should be applied to the attack.

ADAs specifically seem to state they use the rules for Standard Flak Attacks so I believe they are still supposed to receive the bonus.

### Files
This can be replicated by adding various artillery and VTOLs to any random battle so no saves are really required.

Gribbly1:
Apply a +3 to-hit modifier. Ignore the regular modifier for
making a direct-fire attack; the only other modifiers applied
are for using Type “F” weapons, the firing unit’s movement and
current damage, as well as woods/jungle, if any are intervening.

Which printing do you have?

RaozSpaz:
Yeah. Its 4th edition. I failed to notice its an older version and the rule has since been erratad.

PR #8749, opened 2026-08-17
PR: Add a pull request template

HammerGS:
## What this changes

MegaMek has no pull request template, so the box is empty when you open a PR. This adds one. It is a prompt, not a gate - GitHub has no required fields for pull requests, so anything here can be deleted by the author.

The content is taken from what the project already asks for rather than from a generic checklist. Three things it addresses, each measured against the last 50 merged PRs:

**The title convention is documented but almost unused.** `Commit-Procedures` on the wiki says titles start with `Fix:` for a bug fix or `PR:` for a feature, because release notes are generated automatically from PR titles. Four of the last fifty MegaMek PRs did that. Release notes are currently being built from titles like "F_PHYSICAL_WEAPON". The reminder is the first thing in the template, in a comment, because a template cannot set the title itself.

**Only five of the last fifty linked an issue** with a closing keyword, so issues are not closing themselves. `Fixes #` now sits directly under the summary.

**AI disclosure is project policy.** The wiki's AI tool guidelines require that any PR involving AI tools be labelled, and that the submitting developer be able to explain and verify the result. The checklist says so and links the page.

## Testing

Rendered the Markdown to check the comment blocks disappear and the checklist renders as checkboxes. The template cannot be seen in the pull request composer until it is on the default branch, which is the same constraint the issue templates had.

## What is not proven yet

Whether it actually helps. That is the point of trying it in MegaMek alone before MegaMekLab, MekHQ and mm-data - if it turns out to be noise, or people delete it wholesale, it is one file to revise or remove.

The four checklist items are drawn from what reviewers have been flagging, but the list is deliberately short. A long checklist gets ticked without being read, which is exactly why the mandatory Discord confirmation came out of the bug report form.

Two related corrections are worth making separately, since they are wiki edits rather than repository changes: the AI guidelines name labels called "AI Code" and "AI Data" which do not exist - the real ones are `AI Assisted Development` and `AI ready for Review` - and the outside contributor page still tells people to target the `master` branch.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,26 @@
+<!-- TITLE: start with "Fix:" for a bug fix, or "PR:" for a feature or improvement.
+     Release notes are generated automatically from PR titles. -->
+
+## What this changes
+
+<!-- The effect someone actually sees, in plain language. One or two sentences is fine. -->
+
+Fixes #
+
+## Testing
+
+<!-- What you ran. Unit tests? Did you play a game and exercise this path?
+     Once it has been tested in game, add the "AI ready for Review" label. -->
+
+## What is not proven yet
+
+<!-- What you did not test, or could not. "Nothing" is a valid answer - say so explicitly. -->
+
+---
+
+- [ ] This PR is focused on one issue or RFE. Large refactoring or accessibility work is in its own PR
+- [ ] Every file in the diff has a deliberate change (no stray formatting, no unrelated files)
+- [ ] Tests added or updated, if this implements a rule or changes game state
+- [ ] Javadoc literals use `{@code true}` / `{@code null}` rather than bare or quoted text
+- [ ] If AI tools were used, the **AI Assisted Development** label is applied, and I can explain and have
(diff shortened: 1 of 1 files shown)


HammerGS:
Corrected in 4fce7c62b3 after feedback: the issue number belongs in the title, `Fix #1234:` for a bug fix and `Close #1234:` for anything else. I had it as a bare `Fix:` / `PR:` prefix, which is what the wiki's Commit Procedures page documents - it does not mention the number at all, so the page is incomplete rather than merely ignored.

The template now asks for the number in the title **and** a closing keyword in the description, because they do different jobs and only one of them closes anything. GitHub acts on closing keywords in the description and in commit messages, never in the title, so `Fix #1234:` reads like it should close the issue and does not.

That shows up in the tracker. Of the last 60 merged MegaMek pull requests, 13 carry an issue number in the title but only 5 close an issue - the rest are presumably being closed by hand.

For what it is worth, all three repositories already do this, just differently: MekHQ uses `Fix #9751: Fixed NPE on New Month` and `Close #111, #5329, #9466, #9596: Added Automatic Tracking of Unit History`, MegaMekLab uses `Fix: MML #2216 ...`, and MegaMek tends to trail it as `... (Fixes #8705)`. If the `Fix #` / `Close #` form is the one we want, the wiki page is worth updating to match, and the other three templates should say the same thing when they follow.

VicenteCartas (approved):

Issue #8761, opened 2026-08-18
Objectives and Victory Points for MegaMek (implementation series)


HammerGS:
This is the tracking issue for bringing the printed Objectives rules (Standard Missions: objective counters, control radius, scanning, the counter variants) and objective-based Victory Points to MegaMek. The work lands as a series of small, reviewable PRs rather than one multi-thousand-line change, following the review feedback that large PRs are hard to review and should be broken into stepping stones.

Groundwork already merged: #8536 gave the lobby a server-built battlefield, so pre-game hex selection happens on the real board the game will use. Parts 1 to 3 were managed as a GitHub PR stack targeting main and merged together on 25 August 2026; later parts are opened individually against main.

### The series

- [x] **1. Victory hex designation in the lobby** (#8762, merged) - designate hexes on the lobby board preview (flag in the player's color), markers placed as ground objects at game start, visibility filtered by side (teammates see each other's, a game master sees all, opponents see nothing). Introduces the `ObjectiveMarker` model everything later builds on.
  - Since superseded in part: the lobby preview designation flow was removed again in #8774, because a preview map and a generated board can disagree about what is in a hex. What this part contributed and is still load-bearing is the marker model, the pre-game assignment path (a scenario or MekHQ can hand a player markers before launch), the side-filtered visibility, and the `ObjectiveScoringScheme` model that part 3's schemes are built on.
- [x] **2. Objective control and Victory Point scoring** (#8769, merged) - control radius, End Phase control resolution (strict majority of eligible units), the running VP tally and the VP victory condition, zone outlines on the board, per-flag radius and VP editing, scenario file authoring.
- [ ] **3. Mission setup: scoring schemes and their UI** - the Control Points model: per-point counters (turns held, points to control, points to lose control), per-side signed VP tallies with starting values and win/loss thresholds, per-mission sudden death, and the presets that drive them (Standard and Raid as the rules-as-written schemes; Defend, Hold and Capture as labeled mission options) - set from the flag properties dialog, with a minimal mission-first rework of the victory dialog. Exposes a "what does this point want" API for the later bot parts.
[...]

PR #8774, opened 2026-08-21
Add the Victory Setup phase: control points are placed on the real board (Objectives series 3, #8761)

HammerGS:
## What this changes

Adds a Victory Setup phase between Exchange and artillery pre-sighting: when the Use Objectives victory option is on and the game is on a ground map, each player in turn clicks hexes on the real generated board to place their control points and set each one's radius, victory point value and scoring scheme - objectives are set first because pre-sighted artillery and minefields are planned around them. Clicking your own point reopens its properties, where it can also be removed; a properties pane describes the configured scheme in plain words and re-writes the description live as the numbers change. Points assigned before launch by MekHQ or a scenario file arrive already placed and are adjusted the same way.

The dividing rule of the series: anything that needs a map coordinate is set in this phase, anything that does not stays a lobby game option. On the board, control points draw as player-colored flags with their scheme word and a zone outline, toggleable with the new Show Objects keybind on O (Show Spotting FOV moved to CTRL-O). Sending the ground objects is the phase's turn action - no new packet commands, so no wire-format risk - and bots answer through an overridable `BotClient` hook that Princess and CASPAR can later use to place their own points.

Part of #8761.

## Testing

- 6 phase-flow unit tests: entry gated on the option and a ground map, ordering into artillery pre-sighting, turn-based behavior, and the phase ending when its turns run out instead of hanging.
- The full 97-test objective suite passes at this branch (= the whole stack combined); compile and checkstyle green.
- The phase itself was playtested in-game before the stack restructure: placing, editing and removing points on the real board, with the pane's live descriptions.

## What is not proven yet

- Not re-run in-game since the restructure and since the lobby designation flow was removed - the lobby (board preview, victory conditions dialog) needs a smoke test to confirm nothing regressed from the removal.
- A bot present during the phase has not been playtested (the hook sends the map back unchanged; unit tests cover the server side of turn ending, not a live bot).
- Non-ground and multi-board games skip the phase by design; only the ground-map path was playtested.

---


megamek/src/megamek/client/ui/clientGUI/boardview/spriteHandler/GroundObjectSpriteHandler.java
```diff
@@ -60,17 +80,90 @@ public void setGroundObjectSprites(Map<Coords, List<ICarryable>> objectCoordList
...
+     * filling it. A radius of 0 draws nothing - the flag itself already marks the single hex.
+     *
+     * @param boardView the board view the sprites are displayed on
+     * @param board     the game board, for the zone's edge at the board border; may be {@code null} pre-board
+     * @param center    the objective's hex
+     * @param radius    the control radius in hexes
+     *
+     * @return the perimeter sprites of the zone, empty for radius 0
+     */
+    private static List<FieldOfFireSprite> zoneOutlineSprites(BoardView boardView, @Nullable Board board,
```
IllianiBird, line 116: It this is a static objective we can map these coords at the beginning of the game, then we just need to check whether the unit's current coords are in the objective coords array. This would change o(n) to o(1)

If it's a non-static objective, we might still be able to use this approach, but I leave to your discretion.

megamek/src/megamek/client/ui/panels/phaseDisplay/VictoryHexPropertiesPane.java
```diff
@@ -0,0 +1,240 @@
...
+            }));
+            rateLabel.setText(Messages.getString((preset == SchemePreset.DEFEND)
+                  ? "VictoryHex.gripDrainPerTurn"
+                  : "VictoryHex.progressPerTurn"));
+            thresholdLabel.setVisible(usesThreshold);
+            thresholdSpinner.setVisible(usesThreshold);
+            rateLabel.setVisible(usesRate);
+            rateSpinner.setVisible(usesRate);
+            countingLabel.setVisible(usesCounting);
+            countingCombo.setVisible(usesCounting);
```
IllianiBird: This lambda is far _far_ too big and contains multiple switches. Please create a utility method and have the lambda call that. Consider creating utility methods to handle the switches, too.

megamek/src/megamek/server/totalWarfare/TWPhasePreparationManager.java
```diff
@@ -115,6 +115,22 @@ void managePhase() {
...
+                gameManager.checkForObservers();
+                gameManager.transmitAllPlayerUpdates();
+                gameManager.resetActivePlayersDone();
+                gameManager.setIneligible(phase);
+
+                if (gameManager.getGame().getBoard().isGround()) {
+                    gameManager.getGame().setTurnVector(gameManager.getGame().getPlayersList().stream()
+                          .filter(player -> !player.isObserver() && !player.isGhost())
+                          .map(player -> new GameTurn(player.getId()))
+                          .collect(Collectors.toList()));
```
IllianiBird: This shouldn't be a switch. We should reserve switches for arrays containing thousands of objects

IllianiBird (approved):
Approved pending actioning of the handful of points of feedback

megamek/src/megamek/client/ui/clientGUI/boardview/spriteHandler/GroundObjectSpriteHandler.java
HammerGS: Checked where the per-unit membership test actually happens: the End Phase control resolution and the zone-entry toast both use `center.distance(position) <= radius` - constant-time hex arithmetic, no scan, so gameplay checks are already O(1) without a precomputed coords set. This method is only the drawing of the perimeter: it runs when the overlay refreshes (toggle, zoom, objects changed), and it has to visit each zone hex once to decide which edges to outline, so a precomputed set would not remove the loop - it would just move it. Given points can also be placed and removed during the Victory Setup phase, I would rather not add cache invalidation here for a draw-only path. Leaving as is unless you feel strongly.

megamek/src/megamek/client/ui/panels/phaseDisplay/VictoryHexPropertiesPane.java
HammerGS: Agreed - fixed in cb17b7fba3. The lambda body moved into a private `refreshSchemeRows(SchemeControls)` method (the controls bundled in a small record per the parameter-object convention), and the switches became `describeConfiguredPreset(...)`, `thresholdLabelKey(...)` and `rateLabelKey(...)` helpers - which also collapsed the duplicated label/tooltip switches, since the tooltip key is always the label key plus ".tooltip". The listeners are now one-liners.

megamek/src/megamek/server/totalWarfare/TWPhasePreparationManager.java
HammerGS: Fixed in cb17b7fba3 - I read this as the stream on this line (the switch itself is the class's pre-existing phase dispatch): the turn-vector stream is now a plain loop with a named `canPlaceObjectives` condition, building an ArrayList and setting it once. If you actually meant restructuring the surrounding phase switch, say the word and I'll take that on separately.

megamek/resources/megamek/common/report-messages.properties
```diff
@@ -1286,6 +1286,22 @@
 7105=<data> achieved the BV destroyed victory condition with <data>%.
 7106=<data> achieved the kill count victory condition with <data> kill(s).
 7110=<data> achieved the Enemy commander destroyed victory condition with.
+7115=<data> finishes the game with <data> victory point(s).
+7116=Victory points are tied - the game is a draw.
+7117=The <data> point <data> at <data> is controlled by <data>.
+7118=The <data> point <data> at <data> is uncontrolled.
+7119=<data> receives <data> victory point(s) for holding objectives.
```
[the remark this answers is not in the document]
HammerGS: Valid - fixed in b548ebce38 (on the part-2 branch, which owns the report texts). 7119 now says "for controlling objectives" - it is the standard per-turn scoring line and applies to every scheme. Line 1299 (report 7137, the Defend grip line) also said "hold", which collided with the Hold scheme's vocabulary; it now says "The grip on control point X is weakening", matching the setup pane's own term for the Defend mechanic. 7141 keeps "hold" deliberately - it IS the Hold scheme's broken-streak report.

megamek/resources/megamek/common/options/messages.properties
```diff
@@ -147,6 +147,8 @@ GameOptionsInfo.option.use_kill_count.displayableName=Kill count victory
 GameOptionsInfo.option.use_kill_count.description=If checked, victory is achieved when a team achieves a specified number of kills. In the event multiple teams achieve the condition in the same round victory goes to the team with the highest kill count.
 GameOptionsInfo.option.game_kill_count.displayableName=Number of kills to achieve victory.
 GameOptionsInfo.option.game_kill_count.description=If checked, victory is achieved when a team achieves a specified number of kills. In the event multiple teams achieve the condition in the same round victory goes to the team with the highest kill count.
+GameOptionsInfo.option.use_objectives.displayableName=Use Objectives
+GameOptionsInfo.option.use_objectives.description=If checked, victory points scored during the game (e.g. for controlling objectives) decide the winner when the game ends: the side with the most victory points wins, ties are a draw. Requires a game duration, e.g. the turn limit option.
```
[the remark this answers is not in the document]
HammerGS: Valid - fixed in 6c5574b32a (on the part-1 branch, which owns the option text). The description now reads "The game must have an ending for the points to resolve: the game turn limit option, or a scenario game-end event."

megamek/src/megamek/server/totalWarfare/TWGameManager.java
```diff
@@ -9767,6 +9782,12 @@ private void receiveGroundObjectUpdate(Packet packet, int connId) throws Invalid
 
         // make sure to update the other clients with the new ground objects data structure
         send(packet);
+
+        // in the Victory Setup phase this packet is the player's turn action: storing their control
+        // points ends their turn, exactly as the minefield packet does in the minefield phase
+        if (getGame().getPhase().isVictorySetup()) {
+            endCurrentTurn(null);
+        }
```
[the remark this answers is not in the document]
HammerGS: Valid - fixed in 1b5cb5bcff. I verified the sender inventory first: the only client-side senders of UPDATE_GROUND_OBJECTS are the Deploy Minefields display, the Victory Setup display and the bot's setup-turn hook - the in-game cargo pickup/drop flows are computed server-side and never send it. The server now rejects the packet outside those two phases with a logged warning, exactly the gate you suggested.

megamek/src/megamek/server/totalWarfare/TWPhaseEndManager.java
```diff
@@ -62,6 +62,18 @@ void managePhase() {
...
+                gameManager.getGame().addReports(gameManager.getMainPhaseReport());
+                // objectives are placed before artillery is pre-sighted and mines are laid: both of those
+                // decisions depend on knowing where the objectives are
+                boolean usesObjectives = gameManager.getGame().getOptions()
+                      .booleanOption(OptionsConstants.VICTORY_USE_OBJECTIVES);
+                if (usesObjectives && gameManager.getGame().getBoard().isGround()) {
+                    gameManager.changePhase(GamePhase.VICTORY_SETUP);
+                } else {
+                    gameManager.changePhase(GamePhase.SET_ARTILLERY_AUTO_HIT_HEXES);
+                }
```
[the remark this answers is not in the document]
HammerGS: Valid - fixed in 1b5cb5bcff. The transition now requires a single-board ground game (getBoards().size() == 1 plus isGround), with a regression test asserting a two-board game goes straight to artillery pre-sighting. The underlying reason is the one you identified: the ground-object map is keyed by hex alone with no board id, so objectives cannot address multi-board games yet.

IllianiBird (approved):

PR #8776, opened 2026-08-21
Allow all four C3 configurations in network capacity rules

HammerGS:
## Summary

Players could not build Configurations 3 and 4 from the Core Rulebook's C3 Configuration Diagram (CR p.199): the engine recognized at most two C3 Master computers per unit and capped every master at 3 slaves. This PR generalizes the network capacity rules to all four configurations. It is the engine half of a two-PR stack; the lobby C3 Network Manager UI follows in the next PR.

## Changes

- The three company links are a shared budget (CR p.198): each subordinate master takes one link, each block of 3 direct slaves takes one of the node's own lance computers, and idle computers take none - so a unit with 2-4 C3 Masters heads Configurations 2-4, or a smaller configuration with computers left idle.
- Subordinate and undesignated multi-masters run a single lance; consolidation happens only at the designated company commander (the diagram shows "the only four ways" a network forms).
- All-C3-Master lances can be formed, and lances stay homogeneous (all slaves or all masters).
- Combined joins are validated jointly, so slaves and masters cannot oversubscribe the company links even when each pool alone would allow them.
- The BV network bonus gate checks network membership directly instead of free-node heuristics, which broke for consolidated nodes.
- The existing right-click menu and unit configuration dialog benefit unchanged; free-count displays are capped by the 12-unit network limit.

## Testing

- 14 tests in `C3NetworkConfigurationsTest`: all four configurations built through `C3Util`, every capacity rejection (13th unit, per-lance and per-node limits, a mixed join oversubscribing the company links), All-C3-Master lances, the subordinate single-lance rule, a multi-master heading a smaller configuration, the BV bonus gate for standard C3 and C3i, and a save/load round trip - a Configuration 3 network's masters recorded as UUID references, live ids dropped, `wireC3` re-resolving, full 11-unit hierarchy and company designation asserted intact. Four of the original eight fail with the capacity formulas reverted.
- Existing suites green: `C3MasterBVTest`, `NovaCEWSNetworkTest`, `NovaCEWSBVTest`, `ComputeC3SpotterTest`, `EntityTest`, `EquipmentOffModesTest`. Checkstyle and javadoc clean.
- Manual: Configurations 3 and 4 built in the lobby during development (via the follow-up UI and the existing right-click menu).

## Not proven yet

- The MUL round trip is covered at the UUID/wireC3 level; an actual file write/parse through EntityListFile/MULParser has not been exercised. In-game play (spotter bonus, link lines) of the new shapes has not been re-verified on this branch.
- No multiplayer pass yet; the transport is the unchanged entity-update path.
- Everything else claimed above is covered by the automated tests.


megamek/src/megamek/common/util/C3Util.java
```diff
@@ -496,6 +496,11 @@ public static void joinNh(Game game, Collection<Entity> entities, int masterID,
 
     /**
      * Connects the passed entities to a standard C3M identified by masterID.
+     *
+     * <p>Joining units are classified by their equipment: C3 Master carriers occupy company-level master links
+     * when the target is a company node (a C3MM unit or a designated company commander), while C3 Slave carriers
+     * occupy slave links. A plain lance master accepts a homogeneous lance only: all slaves, or - per the
+     * All-C3-Master rule (CR p.199) - all masters acting in slave roles.</p>
```
IllianiBird: This cites core rules, but do those rules match TW and if not do we need to option gate this?

IllianiBird (approved):
Approved barring comment

megamek/src/megamek/common/util/C3Util.java
HammerGS: Checked against Total Warfare 11th printing, p. 132 ("Designating A C3 Network"). TW has the same two rules as CR p. 199, word for word:

- "A C3 Master can only control one to three C3 Slaves or one to three C3 Masters."
- "While a C3 Master controls another C3 Master, it cannot control a C3 Slave."

and points at the same C3 Configuration Diagram. The All-C3-Master case is not a new CR rule, it is the first TW bullet applied at lance level: a lance master controlling three other Masters is "one to three C3 Masters", and the second bullet is why that lance cannot also hold a Slave - which is exactly what this branch of connect() enforces (all slaves or all masters, never mixed). CR just spells the example out in prose.

So no TW/CR divergence and nothing to option-gate. I will update the javadoc to cite both (TW p. 132 / CR p. 199) in the next push so the reference is not CR-only.

HammerGS: Javadoc now cites TW p. 132 alongside CR p. 199 - 7081fd3061.

PR #8779, opened 2026-08-21
Fix #8778: Scenario factions without a camo get their player color's camouflage

### LINKED ISSUE #8778: Scenario factions without a camo all look identical on the board (Princess v Princess is unreadable)
Run any V2 scenario whose factions do not declare a `camo:` key - for example the shipped Princess evaluation scenarios - with two bot players. Every unit on the board renders with the same default camouflage, so the two sides cannot be told apart while playing. The player colors themselves are assigned correctly (minimap dots, chat names and objective flags differ); only the unit sprites are indistinguishable.

The gap is in scenario loading: lobby-created players automatically receive a color camouflage matching their player color (`Server.addNewPlayer` calls `new Camouflage(COLOUR_CAMOUFLAGE, colour.name())`), but `ScenarioV2.createPlayers` only sets a camouflage when the scenario file provides a `camo:` key, and otherwise leaves the default. Scenario factions should fall back to their player color's camouflage the same way lobby players do.

Found while playtesting the Objectives series test scenario (Princess v Princess).

HammerGS:
## What this changes

Run any V2 scenario whose factions do not declare a `camo:` key - the shipped Princess evaluation scenarios, for example - and every unit on the board renders with the same default camouflage, so the sides cannot be told apart while playing. The player colors themselves were always assigned correctly (minimap, chat names); only the unit sprites were indistinguishable. Scenario factions without a camo now receive their player color's camouflage, exactly the way lobby-created players always have (the same fallback `Server.addNewPlayer` applies), so a Princess-versus-Princess scenario shows blue units against red. Factions that declare their own `camo:` keep it unchanged. The change is one fallback branch in `ScenarioV2.createPlayers`.

Fixes #8778

## Testing

- Two regression tests in `ScenarioV2PlayerCamoTest`: a camo-less fixture scenario must give every faction a distinct color camouflage matching its player color, and a fixture with declared camos must keep them.
- compile and checkstyle green.

## What is not proven yet

- Not yet relaunched in-game on this branch; the failure itself was observed live in a Princess-versus-Princess scenario and the fix path is covered by the tests above.

---

### DIFF
--- a/megamek/src/megamek/common/scenario/ScenarioV2.java
+++ b/megamek/src/megamek/common/scenario/ScenarioV2.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2024-2025 The MegaMek Team. All Rights Reserved.
+ * Copyright (C) 2024-2026 The MegaMek Team. All Rights Reserved.
  *
  * This file is part of MegaMek.
  *
@@ -409,6 +409,11 @@ public class ScenarioV2 implements Scenario {
                 } else {
                     player.setCamouflage(new Camouflage(new File(camoPath)));
                 }
+            } else {
+                // like lobby-created players (Server.addNewPlayer), a faction without a camo of its own
+                // gets its player color's camouflage - otherwise every faction renders with the same
+                // default camo and the sides cannot be told apart on the board
+                player.setCamouflage(Camouflage.of(player.getColour()));
             }
 
             teamId = playerNode.has(PARAM_TEAM) ? playerNode.get(PARAM_TEAM).intValue() : teamId + 1;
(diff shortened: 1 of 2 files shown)


IllianiBird (approved):

HammerGS:
The failing check was pr_build, not CodeQL itself (CodeQL shows pass) - the camo-less regression test loaded a fixture whose map lives in the staged unofficial board packs, which the CI runner does not stage. Switched the test to the strike-run fixture whose board ships inside testresources (the same one the bomb-loading test already uses in CI). Both tests green locally; CI re-running.

PR #8792, opened 2026-08-22
Fix: Escalating failure table started at 2+, corrected so it no longer does this.

Hwatwasthat:
## What this changes

Fixes the new escalating failure table, previously this started at 2 for the first failure check, ensuring it could never fail. Updated the switch statement to start from 3+ as intended.

## What is not proven yet
I had issues getting Gradle to work so I could not test it. But as this is a simple change to a single switch case I feel this could be acceptable.

---

### DIFF
--- a/megamek/src/megamek/common/rules/RulesCharts.java
+++ b/megamek/src/megamek/common/rules/RulesCharts.java
@@ -48,12 +48,11 @@ public abstract class RulesCharts {
     public int escalatingFailure(int count) {
         // For each count after 5, the number is 11+
         return switch (count) {
-            case 0 -> 2;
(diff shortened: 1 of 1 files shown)


TenkawaBC:
Will review on Tuesday August 25.

TenkawaBC:
It is intentional that this chart has an auto success at the 0 level. This is designed for equipment that never fails at 0 usages.

1 passed as the argument returns 3, which is correct. 1 turn usage or 1 con hit is three.

This is not an issue. If you have a specific issue you are trying to fix,  please open an issue.

HammerGS:
Based on Dev feedback, closing this PR

PR #8865, opened 2026-08-31
Add crew personal equipment: armor kits that keep working after ejection

HammerGS:
## What this adds

A crew that ejects has nothing protecting them from the world outside. In a caustic atmosphere a
pilot who survived the hit that threw them out then dies of the air, and there has never been
anything a player could do about it.

Under a new Advanced Rules option, a crew can be issued an armor kit in the Configure Pilots tab
and they keep wearing it once they are outside their unit. That is nearly free, because an ejected
crew arrives as conventional infantry and MegaMek already knows what infantry survives by reading
their kit: a space suit answers vacuum, a snowsuit answers cold, an environment suit answers
several hazards at once. Almost nothing had to be written for any of it.

The MechWarrior Combat Suit is the exception, and it is the reason this started. A Time of War
p.294 pairs it with a combat neurohelmet that "may be sealed in hostile environments" with its own
air supply, but MegaMek's entry for the suit carries a damage divisor and nothing else, so its
flags cannot say what the book says. That one interpretation lives in code: the kit answers a
tainted or toxic atmosphere and extreme heat. It does not answer vacuum, because the ensemble has
no gloves, which the aerospace pilot kit lists as required to seal.

## Changes

- `Crew` carries the kit's internal name per slot, and the MUL carries it as an `armorkit`
  attribute. A name hands over the damage divisor and all twelve flags at once, so a future rule
  reading a different flag needs no new plumbing. This is also the seam MekHQ will hand a purchased
  kit across on.
- The ejected crew is dressed through `ConvInfantry.setArmorKit`, which owns that state. Adding the
  equipment directly leaves the space suit flag, encumbrance, sneak properties and damage divisor
  unapplied.
- Five kits converted from A Time of War by the printed rules for carrying personal armor to the
  Total Warfare scale: the two MekWarrior kits, the Clan kit, the aerospace fighter pilot kit and
  the tanker's smock.
- Availability defers to each kit's own tech progression, so a crew cannot carry one into a year
  before it was built.
- The ejection report says when the kit is what saved them, and stays quiet in fair weather and
  where the kit does not answer the danger.
- Corrects the existing MechWarrior Cooling Suit, whose Inner Sphere dates read 2680 and 2800 where
  the master equipment table gives a Terran Hegemony prototype around 2480 and production in 2500.
  MegaMek's own Clan line already carried 2480, so the right number was on the wrong line.

## Testing

- 17 unit tests across the rules and the ejection report. Six drive `ejectEntity` itself rather
  than checking the condition in isolation, which is deliberate: the report line was first written
  into the neighbouring `ejectSpacecraft` and so never appeared in a game at all.
- The conversion arithmetic is checked against three kits MegaMek already carries and reproduces
  all three exactly, including the round-to-zero floor: the combat spacesuit at 1 and encumbering,
  the marine suit at 2, the cooling vest at 0.5.
- Playtested in a Caustic Toxic atmosphere before the generalisation: a suited crew survives
  ejection, an unsuited one dies, and the report names the kit.
- Full suite 16,162 passing, checkstyle clean on both main and test sources.

## What is NOT proven yet

- The generalisation has not been played since. The behaviour the playtest covered is unchanged,
  but the kit dropdown, the other kits and the vacuum cases have only been exercised by test.
- MekHQ has no purchasing or assignment yet, so the dropdown is currently the only way to issue a
  kit. The MUL attribute is in place for that work to build on.
- Crews abandoning a DropShip into lifeboats or escape pods go through different constructors that
  were left alone. Those are spacecraft rather than ground combat, so they never receive a kit.
- Only Inner Sphere dates were exercised in game. The Clan entries share the same code path and are
  covered by test.

**Supersedes #8836**, which GitHub closed when its branch was renamed. The code, the commits and
the review history are unchanged; the branch is now named to match the MekHQ half that will build
against it, since MekHQ's CI finds its MegaMek counterpart by branch name. Copilot's review and the
replies to it are on #8836.

**Stacked.** This targets `Implement-Tainted-and-Toxic-Atmospheres` (#8829). The tainted-atmosphere
rules it reads are not on `main` yet, which is why it is based there. It does not depend on #8835.


IllianiBird (approved):
Nothing stands out as wrong here

PR #8917, opened 2026-09-07
Fix #8915: a Mek on the top level of a building is still inside it for line of sight

### LINKED ISSUE #8915: A Mek on the top level of a building counts as outside it for line of sight (TW p. 175)
### What happens
A Mek standing on the top level of a building is treated as being outside it for line of sight. In a one-level building that is every Mek inside it. The Mek cannot fire at infantry in the same building ("Attack on infantry crosses building exterior wall/roof"), and none of the Combat Within Buildings rules apply to its shots.

### What the rule says
Total Warfare 11th printing, p. 175, Combat Within Buildings: "Do not add a level to a 'Mech's height when it is inside a building hex for line of sight purposes." A Mek inside a building is inside on the level it stands on, whatever its height.

### Where
LosEffects.losStraight and LosEffects.losDivided decide whether the attacker and target are inside a building with Compute.isInBuilding using the unit's absolute line-of-sight height, which for a Mek is its elevation plus one. A Mek on level 0 of a one-level building, or on level 1 of a two-level building, therefore tests at the roof height and is counted as outside. Units with height 0 (infantry, vehicles) are unaffected.

### Steps to reproduce
1. Put a Locust and an enemy foot platoon inside the same one-level building hex, both at level 0 (or a two-level building with the Locust deployed on level 1 and the platoon on level 0).
2. Firing phase: the Locust targets the platoon.
3. To-hit reads 0% "Attack on infantry crosses building exterior wall/roof". Put the Locust on level 0 of a two-level building instead and the shot is allowed.

A unit test on the LOS calculation gives the same result: the shared-building flag is set only when the Mek's head is below the roof.

HammerGS:
## What this changes

A Mek whose head reaches the roof line of the building it stands in now counts as inside that building for line of sight, as TW p. 175 requires (no extra level is added to a Mek's height inside a building). Until now every Mek in a one-level building, and any Mek on the top level of a taller one, was treated as outside and could not fire on infantry in its own building.

Fixes #8915

## Testing

- Unit tests: a Mek on the only level of a one-level building, on the top level and below it in a two-level building, all inside; a Mek on the roof outside. Existing line-of-sight tests unchanged.
- Spotless and the building test set green; full suite green on the combined test branch.
- Played hotseat 7 September: a Locust and a platoon on level 0 of a one-level building, the shot allowed with the same-level modifiers; before the fix it read "Attack on infantry crosses building exterior wall/roof".

## What is not proven yet

- The roof case was not played; the unit test covers it.

---

### DIFF
--- a/megamek/src/megamek/common/LosEffects.java
+++ b/megamek/src/megamek/common/LosEffects.java
@@ -301,7 +301,7 @@ public class LosEffects {
     public int getHeavySmoke() {
         return heavySmoke;
     }
-    
+
     public int getBAPReduceSmoke() { return bapReduceSmoke; }
 
     public int getScreen() {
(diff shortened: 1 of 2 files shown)


HammerGS:
Closing without merging. #8855 already implements the forum ruling for this case: a Mek with its top at the roof line protrudes from the building, so main's refusal is correct. See the note on #8915.

PR #8945, opened 2026-09-11
refitUUID, engine fix for support vehicles

exeea:
This PR does:
- add a refitFromUUID property (currently passthrough only, is to link an unit to their origin)
- fixes engine calculations for support vehicles
- fixed "rear right" load equipment for superheavy tanks


megamek/src/megamek/common/loaders/BLKFile.java
```diff
@@ -356,7 +360,7 @@ protected void loadEquipment(Entity t, String sName, int nLoc) throws EntityLoad
                         }
                         if (shots > 0) {
                             mount.setOriginalShots(shots);
-                            mount.setShotsLeft(shots);
+                            mount.setShotsLeft(shots);mount.setShotsLeft(shots);
```
IllianiBird: Looks like we might need a new line here

IllianiBird (approved):
Approved pending copilot and the one missing newline

megamek/src/megamek/common/loaders/BLKFile.java
exeea: was just a duplicate, I am fixing now the engine rating that remained in the other PR I missed the migration

PR #8950, opened 2026-09-12
Close #5868, closes #6649, closes #7362: Sensor defaults and missile fire control toggles

### LINKED ISSUE #5868: RFE: Client Setting for Sensor Defaults
Almost all units come with a set of sensors to choose from, with different strengths and weaknesses. Right now, MegaMek always defaults to using a unit's active probe if installed, and radar if not.

Radar is generally a poor choice on military units, especially next to options like magscan and infrared. As a result, I often start each fight tediously tweaking the sensor setting of each unit as I deploy them.

Rather than simply tweaking the default order to match my own preferences, I propose adding a client setting consisting of an ordered list of all sensors known to MegaMek. Each unit can then default to the first sensor it has on the list when deployed.

### LINKED ISSUE #6649: [RFE] Enable configuring sensors in the Lobby
### Enhancement Type *

New Feature

At present, units default to using Radar as their sensor.
While this can be adjusted during the deployment phase, it would be helpful if sensor selection could also be made pre-game, in the Lobby configuration menu.

(This is a different RFE to #5868, which relates to creating a Sensor Preference list in Client Settings, rather than being able to configure it unit-by-unit in the Lobby.)

### Use Cases or Rationale

1. Radar isn't always the best sensor, and some players would like their units to start a battle using a different choice of sensor.

### Attach Files

_No response_

### MegaMek Suite Version *

0.50.03

### Final Verification

### LINKED ISSUE #7362: [RFE] Implement Electronics toggling from BMM (particularly for Apollo FCS)
### Enhancement Type *

Implementation of Missing Official Rule

Pg. 12 of the BattleMech Manual notes: "Machine gun arrays, stealth armor, the auto-eject system, and
any equipment listed in the Electronics section (see p. 109) can be switched on or off. Pg. 109-114 then lists several systems that fall under the Electronics category, including Artemis FCS, Apollo FCS, targeting computers, the various flavors of C3, active probes, and ECM suites. 

Currently the systems listed under Electronics are not toggleable in MegaMek and are always on until destroyed through critical hits. Until ARAD missiles are implemented, the feature does not have much of a game use *except* in the case of the Apollo MRM FCS. The Apollo removes the +1 to hit inherent to MRMs, but instead applies a -1 to the cluster roll. Being able to toggle Apollo on or off allows the player to choose between more accurate shots or higher clusters for shots that hit. 

### Use Cases or Rationale

1. Implements an official rule
2. Enhances gameplay by giving players additional tactical choices.

### Attach Files

_No response_

### MegaMek Suite Version *

MekHQ-0.50.07-nightly-2025-08-01

### Final Verification

HammerGS:
## What this changes

Units no longer always start a battle on radar.

**Pick your sensors once** - A Sensor Preference list in Client Settings ranks nine sensor families. Each unit deploys using the first one on your list that it actually carries. A Reset to Default button puts the ranking back.

**Set a sensor in the lobby** - Unit configuration gains a Sensor section with a Starting Sensor dropdown, so you no longer have to wait for round zero. A probe the unit cannot currently use is listed and marked Disabled.

**Make it stick to the design** - Tick "Remember for this unit type" and every unit of that chassis and model starts on that sensor, in MegaMek and in a MekHQ campaign. This mirrors the existing per-design weapon order, saved to `mmconf/customSensorChoice.xml`.

**Switch Artemis and Apollo off** - Artemis IV, Artemis V, Proto Artemis and the Apollo MRM FCS now have On and Off modes. An MRM 20 with its Apollo on takes no to-hit penalty and rolls better clusters; switched off it takes the +1 MRMs carry and rolls clusters unmodified. The switch is declared now and takes effect in the End Phase, matching ECM suites and machine gun arrays.

A sensor you picked by hand always outranks the preference.

image

https://github.com/user-attachments/assets/f3b9d6b5-491f-4d2e-b318-42809b981d3d

Fixes #5868, fixes #6649, fixes #7362

## Testing

`compileJava`, `checkstyleMain` and `javadoc` all pass.

Full suite: 16,755 tests, 4 failures. All four are pre-existing and fail identically on unmodified `main`, verified by stashing every change and re-running them: `CommonSettingsDialogTest` (3) and `SettingsFormPanelTest` (1), all off-by-one pixel assertions.

New tests, 35 in total:
- `SensorFamilyTest` (10) - the nine families partition all 24 `Sensor` types, so a newly added sensor type cannot be silently unreachable.
- `SensorPreferenceSelectionTest` (8) - which sensor a real Mek gets, including one carrying a mounted Beagle Active Probe and one whose probe is unusable.
- `FireControlToggleTest` (13) - each system stops guiding when switched off, damage still overrides the mode, and a system with no Off mode is never read as off.
- `SensorChoiceHandlerTest` (7) - saving, replacing and forgetting a per-design choice.
- `CommonSettingsDialogTest` (+2) - Reset to Default restores the shipped order, and every settings group is mapped to a section.

Two tests were verified by breaking the fix. `everySettingsPageMapsEverySectionItBuilds` fails when the section mapping is reverted. `javadoc` caught a genuine error the other gates missed, `{@value}` on `Sensor.SIZE`, which is an array length rather than a compile-time constant.

Tested in game: the Client Settings page and Reset to Default.

## What is not proven yet

- The lobby Starting Sensor dropdown and the Remember checkbox have not been exercised in a live game.
- The per-design file has not been round-tripped in game. Writing the file is not unit tested because it would touch the user's real `mmconf` directory; only the in-memory map is covered.
- The Artemis and Apollo toggles have not been fired in a live game. Their effect is covered by unit tests at the helper level, not end to end through an attack.
- BMM p.12 does not say whether an Electronics switch is instant. End Phase was chosen on local precedent, not a located ruling.
- Aerospace, spacecraft and battle armor sensor families are implemented and unit tested but untried in a real game.

## Rules clarification

#7362 asks for BMM p.12 Electronics toggling in general. Most of what it names already works: machine gun arrays, stealth armor, auto-eject, ECM suites, active probes, C3 computers and the targeting computer all have modes today. The RISC Laser Pulse Module is also already toggleable, by a different mechanism: fitting one adds Pulse modes to the linked laser, and the player switches the laser rather than the module. Only the missile fire control enhancements had no modes at all, so that is what this PR adds.

## Out of scope

- Sensor choice for Princess and CASPAR. A client setting does not reach a bot, and a static ranking is a poor substitute for situational logic. Its own RFE.
- Per-unit sensor memory inside a MekHQ campaign. This is per design, like weapon order, so two Marauder MAD-3Rs cannot differ. That is MegaMek/mekhq#4044 and needs a campaign file field.
- `LandAirMek.updateSensorOptions` resets the chosen sensor on every conversion, so a LAM loses it when it changes mode. Pre-existing, untouched here.

---
- [x] This PR is focused on one issue or RFE - three closely related sensor RFEs, kept together because #6649 is a split-off of #5868 and #7362 shares the lobby Electronics section they both touch


megamek/src/megamek/client/ui/clientGUI/ClientGUI.java
```diff
@@ -3164,6 +3165,37 @@ private void setWeaponOrderPrefs(boolean prefChange) {
...
+     * @param prefChange {@code true} when the player just changed the preference, which also re-applies it to units
+     *                   that have already deployed
+     */
+    private void setSensorPrefs(boolean prefChange) {
+        List<SensorFamily> preferenceOrder = GUIP.getSensorPreferenceOrder();
+        for (Entity entity : client.getGame().getEntitiesVector()) {
+            if (!entity.getOwner().equals(client.getLocalPlayer())
+                  || entity.hasCustomSensorChoice()
+                  || (entity.isDeployed() && !prefChange)) {
+                continue;
```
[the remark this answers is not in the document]
HammerGS: Verified and fixed in `e1ce3d80c6`.

You are right, and the tooltip was the honest half. `setSensorPrefs` skipped deployed units only when `prefChange` was false, so a preference change mid-game did reach units already on the board. That came from copying `setWeaponOrderPrefs`, where re-applying is harmless because weapon order is display-only. A sensor is game state, so it is not harmless here.

Deployed units are now always skipped. Changing the preference part-way through a game reaches reinforcements that have not come on yet and nothing already fighting, which is what the tooltip says. With that, `prefChange` had no remaining purpose, so the parameter is gone and both call sites are plain `setSensorPrefs()`.

Added `aDeployedUnitIsSkipped` to `SensorPreferenceSelectionTest` to hold the rule.

megamek/src/megamek/client/ui/clientGUI/ClientGUI.java
```diff
@@ -3164,6 +3165,37 @@ private void setWeaponOrderPrefs(boolean prefChange) {
...
+     * <p>Units belonging to anyone else are left alone, bots included, so this never changes what Princess or CASPAR
+     * do. A unit whose sensor the player picked by hand is left alone as well, in the lobby or in round zero.</p>
+     *
+     * <p>A unit that has already deployed is never touched. Changing the preference part-way through a game switches
+     * the sensors of reinforcements still waiting to come on, and nothing that is already on the board.</p>
+     */
+    private void setSensorPrefs() {
+        List<SensorFamily> preferenceOrder = GUIP.getSensorPreferenceOrder();
+        for (Entity entity : client.getGame().getEntitiesVector()) {
+            if (!entity.getOwner().equals(client.getLocalPlayer())
```
IllianiBird: Cache client.getLocalPlayer instead of fetching it each loop iteration

megamek/src/megamek/common/equipment/MiscType.java
```diff
@@ -6556,6 +6556,12 @@ public static MiscType createISArtemis() {
               F_WS_EQUIPMENT,
               F_SS_EQUIPMENT,
               F_HEAVY_EQUIPMENT);
+        // Missile fire control may be switched on or off (BMM p.12, Electronics). Switching it off gives up the
```
IllianiBird, line 6559: Is this a Core rule as well, or do we need to gate it behind whether Core is enabled? The easiest approach would be to gate it behind a game option.

megamek/src/megamek/common/equipment/SensorFamily.java
```diff
@@ -0,0 +1,275 @@
...
+            boolean alreadySet = (currentSensor != null) && (currentSensor.type() == preferredSensor.type());
+            return alreadySet ? -1 : sensorIndex;
+        }
+        return -1;
+    }
+
+    private static SensorFamily[] buildFamilyBySensorType() {
+        SensorFamily[] families = new SensorFamily[Sensor.SIZE];
+        for (SensorFamily family : values()) {
+            for (int sensorType : family.sensorTypes) {
```
IllianiBird, line 241: This should be a test, not a base code function

megamek/src/megamek/common/loaders/MekFileParser.java
```diff
@@ -814,6 +817,38 @@ else if (ent instanceof ConvInfantry infantry) {
...
+     * <p>Without this a unit always starts on the sensor the block above picked for it: the active probe if one is
+     * fitted, otherwise radar. Radar is usually the weakest sensor a military unit carries, which is the complaint
+     * behind RFE #5868.</p>
+     *
+     * <p>Example: a player saves Infrared for the Marauder MAD-3R. Every MAD-3R then starts on Mek IR rather than
+     * Mek Radar, in a MegaMek game and in a MekHQ campaign alike, because both load units through this method.</p>
+     *
+     * <p>A design that carries nothing from the saved family keeps its normal default.</p>
+     */
+    private static void applySavedSensorChoice(Entity ent) {
```
IllianiBird: Don't use abbreviated, single letter, or truncated param or variable names.

IllianiBird (approved):
Approved pending some minor needed adjustments

megamek/src/megamek/client/ui/clientGUI/ClientGUI.java
HammerGS: Done in `2ffb0cbd03`. `getLocalPlayer()` is now resolved once before the loop.

megamek/src/megamek/common/equipment/SensorFamily.java
HammerGS: Done in `2ffb0cbd03`. You are right, it was only ever there to be asserted on. `allCoveredSensorTypes()` and the package-private `sensorTypes()` accessor are gone, and `SensorFamilyTest` builds the coverage view itself from the public `covers()`. The enum no longer carries anything that exists for the tests.

megamek/src/megamek/common/loaders/MekFileParser.java
HammerGS: Done in `2ffb0cbd03`, `ent` is now `entity`.

This one should not have reached you. It is the first item on the PR checklist and I reported that check as passing. My pass grepped the added lines against a fixed list of known-bad names, so a truncation that was not on the list went straight through. Matching any short identifier instead of a blocklist from now on.

megamek/src/megamek/common/equipment/MiscType.java
HammerGS: Checked with HammerGS: no gate needed, the option exists in both rule sets, so this applies under Core as well as TW.

For completeness on the mechanics: the switch itself is inert until someone uses it. The five fire control types gain On and Off modes, and `EquipmentActivation.isGuidanceActive` only subtracts guidance when a mode is actually set to Off. Equipment left alone reports as on, so a game where nobody touches the setting resolves exactly as it does today under either rule set.

PR #8954, opened 2026-09-12
Fix #8948: Armor kit lookup flooded the log with stack traces

### LINKED ISSUE #8948: Megamek has some sort of logging issue (125mb logfile)
### What went wrong

In the process of issuing [the report on autoeject warnings](https://github.com/MegaMek/megamek/issues/8947) and making an in battle save, MegaMek reported a "could not include MegaMek.log due to file size exceeding GitHub's 25mb attachment limit" 

I checked the file which had ballooned to approx. 1.5 million lines of text. Zipping it up reduced it back to about 1mb of compressed text, file attached.

### What you expected instead

megamek.log being included in the Report A Bug megamek battle save snapshot

### Steps to reproduce

Start battle from campaign save in the autoeject report 

get to turn 1 player deployment phase

huge logfile

### Attach your files

[megamek_log.zip](https://github.com/user-attachments/files/32131490/megamek_log.zip)

### Relevant log lines

```text
Line 912 to about 1527682
```

### MegaMek Suite Version

0.51.01-nightly-2026-09-11

### Operating System

Windows 11 10.0 (amd64)

### Java Version

Eclipse Adoptium 21.0.11

### Before you submit

- [x] I have asked about this on the MegaMek Discord (optional)

HammerGS:
## What this changes

The log stops filling up with tens of thousands of stack traces.

Listing the crew armor kits walked every equipment type MegaMek knows and asked each one whether it was an armor kit. Asking a weapon or an ammo type about a MiscType flag makes it log a warning with a full stack trace, so a single call emitted thousands of them.

In the log attached to the issue, that warning and its stack traces run from line 913 to line 1,527,640 of a 1,527,794 line file. **99.9% of a 125MB log.** That is why Report A Bug could not attach it: the file was over GitHub's 25MB limit.

The list of armor kits is unchanged. The wrong-flag check already returned false, so nothing behaved differently. The bug was purely the logging.

Fixes #8948

## Testing

`checkstyleMain` and `javadoc` pass. Full suite run: 4 failures, all pre-existing and unrelated, in `CommonSettingsDialogTest` and `SettingsFormPanelTest`. Those are off-by-one pixel assertions that fail the same way on unmodified `main`.

No tests ship with this. Two were written and verified, then removed at the reviewer's request as not wanted going forward. One of them attached a counting appender to the ammo and weapon loggers and failed with 2,884 warnings from a single call when the fix was reverted, so the fix is confirmed even though the test is gone.

The 99.9% figure comes from counting lines in the reporter's attached log.

## What is not proven yet

- Not reproduced in a running game. The reporter's path was a MekHQ campaign opening the Issue Equipment dialog, and I worked from their log rather than reproducing it.
- The reporter's session logged 31,760 of these warnings while one call in the test environment produces 2,884. The dialog evidently calls this more than once. I did not chase how many times, because the fix removes all of them either way.
- The wider question this raises is out of scope, see below.

## Out of scope

A single miswritten loop turned into a 125MB log because the wrong-flag diagnostic writes a full stack trace on every call, with no throttling. Fixing this caller removes today's flood, but nothing stops the next one. Whether that diagnostic should rate-limit itself, or log once per call site, is worth deciding separately. I have not filed an issue for it.

---
- [x] This PR is focused on one issue or RFE
- [x] Tests added or updated - written and verified, then removed at the reviewer's request; see Testing


IllianiBird (approved):

megamek/unittests/megamek/common/units/CrewArmorKitRulesTest.java
IllianiBird, line 1: We don't need these unit tests. They're useful for confirming the issue has been fixed, but are not needed moving forward

HammerGS: Removed in 72636bb228. The file is back to exactly its state on main; the fix itself is unchanged.

megamek/unittests/megamek/common/units/CrewArmorKitRulesTest.java
```diff
@@ -192,6 +202,64 @@ void whatEachKitAnswers() {
...
+        appender.stop();
+    }
+
+    /** Counts the "wrong flag type" warnings a class emits while a block of code runs. */
+    private static final class CountingAppender extends AbstractAppender {
+        private int count;
+
+        private CountingAppender() {
+            super("CrewArmorKitRulesTest", null, PatternLayout.createDefaultLayout(), false, null);
+        }
```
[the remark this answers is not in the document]
HammerGS: You were right that the helper never restored the level, so it could have leaked into other tests through ordering. It is moot now: the tests it supported were removed at the reviewer's request in 72636bb228, and the helper went with them.

Issue #8977, opened 2026-09-14
[Bug] spotlessApply breaks BoardView.java compilation (ambiguous System import)


HammerGS:
## Problem

`BoardView.java` will stop compiling the first time a pull request touches it. It imports `java.lang.System` explicitly and also `megamek.common.units.*`, which contains a class named `System`. The CI formatting step runs `./gradlew spotlessApply`, and its `removeUnusedImports` rule deletes `import java.lang.System;` because `java.lang` is implicit. After that, every `System.` call in the file is ambiguous between the two classes.

## Example

This already happened to `MoveStep.java` on the megamek #8589 branch: after `spotlessApply`, compilation failed with

```
MoveStep.java:1017: error: reference to System is ambiguous
        System.arraycopy(tempMv, 0, mv, 0, 6);
```

`BoardView.java` has the same pair of imports (`import java.lang.System;` and `import megamek.common.units.*;`) and four `System.` calls, so it is set up to fail the same way.

## Proposed fix

- In `BoardView.java`, write the four calls as `java.lang.System.` (the fix used for `MoveStep` on the #8589 branch), or replace the `megamek.common.units.*` wildcard with explicit imports.
- Optionally, search for any other file that combines `megamek.common.units.*` with a `System.` call.

## Implementation notes

- Spotless only formats files changed against `origin/main` (`ratchetFrom`), so the break stays hidden until someone edits the file.
- Renaming `megamek.common.units.System` would remove the trap everywhere, but it touches far more code.

## Background

Found on the megamek #8589 branch, 2026-09-13, when `spotlessApply` broke `MoveStep.java`.

### What you expected instead

Running `spotlessApply` on a file that compiles leaves it compiling.

### Steps to reproduce

1. Make any small change to `megamek/src/megamek/client/ui/clientGUI/boardview/BoardView.java`.
2. Run `./gradlew :megamek:spotlessApply` then `./gradlew :megamek:compileJava`.

### MegaMek Suite Version

0.51.01 (development build)

### Operating System

Windows 11 (10.0.26200) amd64

### Java Version

Eclipse Adoptium 21.0.11

### Before you submit

- [x] This report only contains one issue or request, and I haven't already reported it
- [x] I am on the right repository - MegaMek for battles, MegaMekLab for unit design, MekHQ for campaigns, mm-data for unit and planet data

Issue #8972, opened 2026-09-14
[Bug] Dataset logger throws NullPointerException after a unit mounts a DropShip


HammerGS:
## Problem

After a unit mounts a grounded DropShip, the server logs a NullPointerException from the AI dataset logger. The logger replays the finished move to collect piloting rolls, but by then the unit is loaded and has no position, so `SharedUtility.doPSRCheck` starts from a null `lastPos`.

## Example

An Ares Medium Tank walks one hex to stand beside a Triumph (2593) and presses Mount. The mount succeeds, and the log shows:

```
ERROR [megamek.common.game.GameDatasetLogger] {Packet Pump}
megamek.common.game.GameDatasetLogger.append(GameDatasetLogger.java:276) - Error logging MovePath unit action
java.lang.NullPointerException: Cannot invoke "megamek.common.board.Coords.equals(Object)" because "lastPos" is null
	at megamek.client.ui.SharedUtility.doPSRCheck(SharedUtility.java:283)
	at megamek.client.ui.SharedUtility.getPSRList(SharedUtility.java:79)
	at megamek.ai.dataset.UnitAction.fromMovePath(UnitAction.java:158)
	at megamek.common.game.GameDatasetLogger.append(GameDatasetLogger.java:270)
	at megamek.server.totalWarfare.TWGameManager.receiveMovement(TWGameManager.java:4496)
```

The exception is caught and the game carries on. Only the dataset row for that move is lost.

## Proposed fix

- In `UnitAction.fromMovePath`, skip the piloting roll replay when the entity no longer has a position (loaded, or off the board), or seed the replay from the path's starting hex instead of the entity's current position.

The mount itself is NOT changed.

## Implementation notes

- `SharedUtility.getPSRList` seeds `lastPos` from `entity.getPosition()`. `TWGameManager.loadUnit` sets that to null during `MovePathHandler.processMovement`, before `receiveMovement` hands the path to the dataset logger.
- Every successful Mount reproduces it, so a unit test can drive `UnitAction.fromMovePath` with a loaded entity.

## Background

Found while testing megamek #8589 (mounting a DropShip through a friendly Mek hex), 2026-09-13.

### What you expected instead

No error; the move is logged like any other move.

### Steps to reproduce

1. Place a tank next to a friendly grounded DropShip that has a vehicle bay.
2. Press Mount and pick the DropShip.
3. Open `logs/megamek.log`.

### MegaMek Suite Version

0.51.01 (development build)

### Operating System

[...]

PR #8978, opened 2026-09-14
Fix #8589: Mount DropShips through friendly Meks, TW mount rules and crane loading

### LINKED ISSUE #2475: [0.51.00] Incorrect MP expenditure rounding and inability to choose facing for dismounting from Aero units
### Description
When dismounting from an aero unit (e.g. a grounded DropShip), a unit must expend half its Walk/Cruise MP to dismount, and this expenditure should round up (per [rules forum thread](https://bg.battletech.com/forums/total-warfare/dismounting-mp-expenditure-rounding/)). Also per that thread (and Total Warfare p91), a dismounting unit should be able to freely choose its facing as part of this MP expenditure. 

Currently, while the "half Walk/Cruise MP" expenditure is implemented in Megamek, it rounds down rather than up, and the facing of the unit cannot be chosen. The interface for dismounting is also rather clunky. It would be nice if it could be improved - something like the Deployment interface, allowing you to visually pick the hex to dismount to, and the facing of the dismounted unit, would be excellent. However, while I expect the rounding issue to be relatively easy to fix, I suspect the latter issue may be much more difficult, depending on how old and weird the dismounting code is. Good luck.

### Files
A save with a grounded Union loaded with mechs of a variety of walking speeds is provided for convenience in reproduction and testing. 

Playername: Greyscales
[dismountTest.sav.gz](https://github.com/MegaMek/megamek/files/5640744/dismountTest.sav.gz)

### LINKED ISSUE #8589: [Issue] [0.51.0] Mech cannot mount Large Craft through a friendly mech
### Brief Description *

Standard movement rules say a mech cannot end its movement in the same hex as another mech, even if the latter is friendly. However, moving _through_ a hex containing a friendly mech is supposed to be allowed. In the specific case of plotting movement that ends with mounting a Large Craft, this does not seem to be working properly. A test example is shown below:

Image

Nominally the north mech should be able to plot movement into the hex adjacent to the Large Craft (containing a friendly mech), then use its remaining MP to mount the dropship (therefore not ending its movement in the hex of the friendly mech). During the movement plotting stage, we see that this is possible (the Mount button is lit up, and it even asks which of the two possible bays we would like to mount into).

Image

 However, actually attempting to do so does not function properly - on pressing "Ok", the mech stops short of the hex with the friendly mech in it, and fails to board the dropship.
Image

I _was_ able to mount a DS with a vehicle moving through a hex with a friendly mech in it, but was unable to do so through a hex with two friendly units in it (one mech, one non-mech). I assume that this failure would occur for any case of movement where the final step prior to mounting would violate the stacking rules.

### 3. Steps to Reproduce

1. Load the provided savegame
2. Attempt to plot movement to board the DS as shown above.
3. Observe that this fails as shown above.

### Attach Files

[dropshipMountingTest.sav.gz](https://github.com/user-attachments/files/30393997/dropshipMountingTest.sav.gz)

### Severity *

### MegaMek Suite Version *

0.51.00

### Operating System *

Windows 10

### Java Version *

     Java Vendor: Eclipse Adoptium     Java Version: 21.0.11

### Final Verification

### LINKED ISSUE #8981: [Bug] Mounting a DropShip does not use the TW MP cost of half Walking MP, rounded up (TW p.90)
## Problem

Mounting a grounded DropShip or Small Craft does not follow the MP cost in Total Warfare. The unit must spend half its Walking MP, rounded up, on top of the MP it already spent, and that total must fit in its Walking MP. MegaMek rounds the cost down, so a unit with an odd Walking MP can mount after spending one MP too many. It also refuses a unit that starts beside the carrier but is too slow to pay the cost, which TW allows through Minimum Movement.

Total Warfare, 11th printing, p.90 (Mounting):

> The mounting unit must be in an adjacent hex to the Small Craft or DropShip and within two levels of the level of the underlying hex(s) of the aerospace unit, and then expend half its standard Walking/Cruising MP (round cost up) to enter the carrying Small Craft or DropShip. Units cannot have declared Running/Flank in a turn they will mount a Small Craft or DropShip. If a unit enters an adjacent hex using Walking/Cruising MP and then does not have sufficient Walking/Cruising MPs to enter the Small Craft or DropShip hex, it must wait until the following turn's Movement Phase (Ground) to enter the Small Craft or DropShip. If a unit starts its Movement Phase in an adjacent hex and still does not have sufficient Walking/Cruising MP (i.e. for any reason, such as damage, heat and so on), the unit may use the Minimum Movement rule (see p. 49) to enter the Small Craft or DropShip hex.

## Example

- A Mek with Walking MP 5 spends 3 MP to reach the hex beside a grounded Union. Mounting costs 3 more (half of 5, rounded up), for a total of 6, which is more than 5. TW says it must wait a turn; MegaMek lets it mount.
- A Mek reduced to Walking MP 1 by heat starts its turn beside the Union. TW lets it mount using Minimum Movement; MegaMek does not.

## Proposed fix

- Mount is legal when MP already spent + ceil(Walking MP / 2) is no more than Walking MP.
- A unit that has not moved this turn may always mount (Minimum Movement).
- A unit that jumped this turn cannot then mount.

## Implementation notes

- The check belongs in the shared MOUNT step legality (`MoveStep`), so the Mount button and the server agree.
- Dismounting has the same rounding error; that is issue #2475.

## Background

Found while working on megamek #8589 (mounting a DropShip through a friendly Mek hex), 2026-09-13. Rules text confirmed in the TW 11th printing PDF and errata v11.01.

### What you expected instead

The Mount button lights only when the unit can pay half its Walking MP, rounded up, on top of what it spent, or when it has not moved this turn.

### Steps to reproduce

1. Place a Mek with Walking MP 5 three hexes from a friendly grounded Union with a free Mek bay.
2. Walk 3 hexes to the hex beside the Union.
3. The Mount button is enabled, although mounting needs 3 more MP and only 2 are left.

### MegaMek Suite Version

0.51.01 (development build)

### Operating System

Windows 11 (10.0.26200) amd64

### Java Version

Eclipse Adoptium 21.0.11

### Before you submit

### LINKED ISSUE #8982: [Bug] Infantry and battle armor can move and then mount a DropShip (TW p.223, errata v11.01)
## Problem

Infantry and battle armor can move and then mount a grounded DropShip, and can mount from a hex up to two levels away. Total Warfare treats infantry mounting a DropShip like mounting a Large Support Vehicle: the platoon spends all its MP to enter the carrier, and mounts from its own hex or an adjacent hex at the same level.

Total Warfare, 11th printing, p.89 says infantry mount Small Craft and DropShips using the infantry carrier rules. Errata v11.01 changes TW p.223 (Mounting) to:

> To mount a carrying unit during a turn, an infantry unit must spend all its MP to enter a carrier in its hex.

TW p.224 lets infantry mount a Large Support Vehicle from adjacent hexes at the same level.

## Example

A foot infantry platoon walks one hex to stand beside a grounded Triumph (2593), then presses Mount. MegaMek lets it board. By the rules it has already spent MP moving, so it cannot mount this turn.

## Proposed fix

- Infantry and battle armor may mount only if they have not moved this turn.
- They must be at the same level as the carrier's hex; the two-level allowance stays for other units.

Other unit types are NOT changed.

## Implementation notes

- `Compute.getMountableUnits` uses the same two-level difference for every unit type.
- The "has not moved" check fits with the other mount restrictions in the MOUNT step legality, so the button and the server agree.

## Background

Found while working on megamek #8589, 2026-09-13. Rules text confirmed in the TW 11th printing PDF and errata v11.01.

### What you expected instead

Mount stays unavailable for infantry that moved this turn or stand at a different level from the DropShip.

### Steps to reproduce

1. Place a foot infantry platoon two hexes from a friendly grounded Triumph (2593), on level ground.
2. Walk it one hex to the hex beside the Triumph.
3. The Mount button is enabled and the platoon boards.

### MegaMek Suite Version

0.51.01 (development build)

### Operating System

Windows 11 (10.0.26200) amd64

### Java Version

Eclipse Adoptium 21.0.11

### Before you submit

### LINKED ISSUE #8983: [RFE] Crane loading and unloading for VTOLs, fighters and small craft at a grounded DropShip (TW p.90-91)
### What kind of request is this?

Implementation of a missing official rule

## Problem

VTOLs, fighters and small craft can mount and dismount a grounded DropShip or Small Craft under their own power, like a Mek or a tank. Total Warfare says they cannot: the carrier's cranes load them over four turns and unload them over three.

Total Warfare, 11th printing, p.90 (Mounting):

> VTOLs, Small Craft and Fighters: These units cannot expend MP to mount an aerospace unit; i.e. they cannot mount a Small Craft or DropShip under their own power. Instead, they must be in an adjacent hex (within two levels of the level of the underlying hex of the carrying Small Craft or DropShip) during the End Phase of a turn. It then takes four subsequent turns for the DropShip to use cranes to load the units into the carrying aerospace unit; during the End Phase of the fourth turn after the turn in which the units were in an adjacent hex, they are loaded and strapped down. Doors are not occupied when loading units in this fashion.

TW p.91 (Dismounting):

> VTOLs, Small Craft and Fighters: These units cannot expend MP to dismount a grounded aerospace unit; i.e. they cannot dismount an aerospace unit under their own power. Instead, it takes three turns for cranes to unload the units from the carrying aerospace unit; during the End Phase of the third turn after the turn in which the units began to dismount, they are dismounted into an adjacent hex (the hex must be within two levels of the underlying level of the hex occupied by the aerospace unit), with a facing of their choosing. In the turn following, such units may move and fire normally.

TW p.87 also says fighters and small craft cannot be recovered directly by a landed DropShip.

## Example

An Aeron Strike VTOL lands beside a grounded Triumph (2593) and presses Mount; it is aboard at once. By the rules it waits beside the Triumph and is loaded in the End Phase of the fourth turn. The Triumph's plain Load can also take the VTOL straight away.

## Proposed fix

- VTOLs, fighters and small craft can no longer Mount, Unload, or be taken by a carrier's plain Load at a grounded carrier.
- For these units the Mount button reads Load by Crane. The unit is loaded aboard in the End Phase of the fourth turn, if it stays beside the carrier.
- The carrier's Unload offers them by crane: the player clicks a hex beside the carrier and picks a facing, and the unit comes out in the End Phase of the third turn. If that hex is blocked, it waits.
- The player can stop loading or unloading.
- Units already waiting count against the carrier's free bay space.
- Moving away, the carrier lifting off, or either unit being destroyed cancels the work.
- A countdown of turns left shows in the hex, and the unit fades out or in as the cranes work.

Airborne recovery of fighters and small craft (TW p.86-87) is NOT changed.

## Implementation notes

- Keep the crane state and rules out of `Entity` and `TWGameManager`: a small state class on the carrier, a stateless rules class, and an End Phase rule handler.
- The state must save and load with the game.

## Background

Found while working on megamek #8589, 2026-09-13, and agreed with HammerGS. Rules text confirmed in the TW 11th printing PDF and errata v11.01.

### Why is it worth doing?

It is an official rule. Without it, VTOLs and fighters board and leave a DropShip instantly, which changes how quickly a landed force can re-embark.

### MegaMek Suite Version

0.51.01 (development build)

### Before you submit

### LINKED ISSUE #8984: [Bug] Cancelling the bay choice when mounting a DropShip throws a NullPointerException
## Problem

When a unit mounts a DropShip that has more than one bay it can use, MegaMek asks which bay to use. Pressing Cancel on that prompt throws a NullPointerException instead of cancelling the mount. The code reads the chosen bay's number from the prompt's answer without checking for a cancelled (null) answer.

## Example

A Marauder MAD-3R stands beside a grounded Union (2708) with two Mek bays and presses Mount. The bay prompt opens, and the player presses Cancel. The client throws a NullPointerException from `MovementDisplay.getMountedUnit`.

## Proposed fix

- When the bay prompt is cancelled, clear the chosen bay and return no unit, so the Mount step is not added and the plotted move stays as it was.

## Implementation notes

- `MovementDisplay.getMountedUnit` calls `bayString.substring(...)` on the result of `JOptionPane.showInputDialog`, which is null when cancelled.

## Background

Found while working on megamek #8589, 2026-09-13.

### What you expected instead

Cancelling the bay prompt cancels the mount, with no error.

### Steps to reproduce

1. Place a Mek beside a friendly grounded DropShip with two free Mek bays, such as a Union (2708).
2. Press Mount and pick the DropShip.
3. Press Cancel on the bay prompt.

### MegaMek Suite Version

0.51.01 (development build)

### Operating System

Windows 11 (10.0.26200) amd64

### Java Version

Eclipse Adoptium 21.0.11

### Before you submit

HammerGS:
## What this changes

Units can mount a grounded DropShip or Small Craft through a friendly Mek's hex, and boarding and leaving a grounded carrier now follow Total Warfare p.86-91 (11th printing, errata v11.01).

**Mounting**
- Moving through a friendly Mek's hex and pressing Mount now boards the DropShip. Before, the move was cut back and the unit never boarded (#8589).
- Clicking a friendly DropShip plots the path to the hex beside it, so Mount lights without picking that hex by hand.
- Mounting costs half Walking MP, rounded up, on top of the MP already spent, and the total must fit in Walking MP. A unit that has not moved can always mount (Minimum Movement). No unit can jump and then mount (#8981).
- Infantry and battle armor mount only if they have not moved this turn, and only at the same level as the carrier (#8982).
- Cancelling the bay-choice prompt no longer crashes the game (#8984).

**Dismounting**
- The dismount cost of half Walking MP now rounds up.
- A unit dismounting a DropShip or Small Craft is placed the same way as crane unloading: click one of the shaded hexes, then pick a facing (TW p.91). Before, the hex came from a drop-down list of coordinates and the unit always faced away from the carrier. Dismounting from other carriers is unchanged (#2475).

**Crane loading and unloading (#8983)**
- VTOLs, fighters and small craft can no longer Mount or Unload under their own power, and a carrier's plain Load no longer takes them.
- For these units the Mount button reads Load by Crane. The unit is aboard in the End Phase of the fourth turn if it stays beside the carrier. While it waits, the button reads Stop Loading.
- A carrier's Unload offers them by crane: click one of the shaded hexes around the carrier and pick a facing. The unit comes out in the End Phase of the third turn, or waits if the hex is blocked. When crane unloading is all that is left, the button reads Stop Unloading.
- A countdown of turns left shows in the hex. The unit fades out while being loaded and fades in while being unloaded; the marks clear as soon as a waiting unit moves.
- Units already waiting count against the carrier's free bay space, so a Union with 2 fighter slots takes 2 waiting fighters, not 3.
- Moving away, the carrier lifting off, or either unit being destroyed cancels the work, with a report saying why. Lifting off with crane work in progress asks for confirmation first. Firing does not cancel it, as TW p.90 does not say it does.
- Crane work saves and loads with the game.

No buttons were added. Bots are not affected: Princess and CASPAR never mount or use cranes. Airborne recovery of fighters and small craft is unchanged.

Fixes #8589, fixes #2475, fixes #8981, fixes #8982, fixes #8983, fixes #8984

## Testing

- New and updated unit tests: `MountThroughFriendlyMekTest` (2), `MountPathHelperTest` (9), `MountBayMatrixTest` (53), `CraneRulesTest` (12), `CraneOperationTest` (4), `CraneOperationHandlerTest` (15), `CraneOperationSpriteTest` (5), `CraneUnloadFacingCompileTest` (1).
- Verified by reverting the fix: the #8589 regression test fails with the stacking fix removed; `CraneUnloadFacingCompileTest` fails with the `MovePath.compile` fix removed (the chosen unload facing was lost on the server); the destroyed-unit and destroyed-carrier report tests fail with their fixes removed.
- Before merging `main`: all `megamek.common.moves` tests and every crane and DropShip loading test class passed, with `spotlessApply`, `checkstyleMain` and `javadoc` clean.
- After merging `main`: `spotlessApply` leaves no changes, `checkstyleMain` and `javadoc` pass. Full megamek suite: 16946 tests, 16 failed, 8 skipped. None of the failures come from this branch:
  - Four settings-dialog layout tests (`CommonSettingsDialogTest` x3, `SettingsFormPanelTest` x1) fail by one pixel because this machine's `mmconf` sets GUI scale 1.1. They fail the same way on unmodified `main`.
  - All 12 `CrewSidearmEjectionTest` tests (new from `main`) fail because this machine's `mmconf/gameoptions.xml` has `ejected_pilots_flee` on, so ejected crews leave the board. They pass in CI on `main` at the commit this branch merged.
- Played in game by HammerGS on a Triumph (2593), Union (2708) and Union-X (3071), checked against `megamek.log` and the round report: every item on the in-game checklist passed. That covers mounting through a friendly Mek, click-to-board, the MP limits, infantry rules, crane loading of VTOLs and fighters, Stop Loading and Stop Unloading, the plain Load block, bay space, lift-off warning and cancellation, unload hex and facing, blocked unload hexes, dismount rounding, moving away, and saving and loading mid-work.

## What is not proven yet

- Choosing a dismount facing (#2475) was added after the in-game checklist and has not been played yet; `MountPathHelperTest` and `CraneUnloadFacingCompileTest` cover the facing choice and that it reaches the server.
- The Mount MP rule also applies to mounting tractors and trailers, which share the Mount step; that path was not tested.
- The branch is large; the crane work could be reviewed as its own chunk (`CraneRules`, `CraneOperationHandler`, the crane sprites and the `MovementDisplay` crane methods).

---
- [x] Focused on one issue or RFE - #8589 plus the TW DropShip boarding rules on the same code path, as agreed with HammerGS; out-of-scope findings are filed separately (#8972-#8977)


megamek/src/megamek/server/totalWarfare/CraneOperationHandler.java
```diff
@@ -0,0 +1,307 @@
...
+    /**
+     * Records a unit's declaration to be loaded by crane. The step was checked on the client; this repeats the check
+     * so a stale or tampered path cannot start an illegal operation.
+     *
+     * @param unit   the VTOL, fighter or Small Craft to be loaded
+     * @param target the carrier named by the step, may be {@code null}
+     */
+    void declareLoad(Entity unit, @Nullable Targetable target) {
+        String illegalReason = CraneRules.loadByCraneIllegalReason(unit, target, true, getGame());
+        if (illegalReason != null) {
```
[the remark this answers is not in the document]
HammerGS: Verified and fixed in 97dafd91a8. The server did pass isFirstStep = true for both LOAD_BY_CRANE and UNLOAD_BY_CRANE, and it trusts the step types the client sends, so a stale or tampered path could declare crane work after moving. A loading declared that way would have been cancelled at the End Phase (the unit is no longer in its waiting hex), but an unloading by a carrier that moved first was not caught. MovePathHandler now passes previousStep == null from the server's own walk of the path to declareLoad and declareUnload, instead of trusting the client's step flags. Two new tests in CraneOperationHandlerTest check that a load or unload declared after other steps is rejected.

megamek/src/megamek/client/ui/clientGUI/boardview/spriteHandler/CraneOperationSpriteHandler.java
```diff
@@ -0,0 +1,120 @@
...
+
+    /** Rebuilds the crane work sprites from the crane operations of every carrier. */
+    public void updateCraneOperationSprites() {
+        clear();
+        if (clientGUI.boardViews().isEmpty()) {
+            return;
+        }
+        for (Entity entity : game.getEntitiesVector()) {
+            if (entity instanceof SmallCraft carrier) {
+                for (CraneOperation operation : carrier.getCraneOperations().getOperations()) {
```
IllianiBird, line 71: I feel like we can reduce this to n(1) instead of n(o) using a hash map

megamek/src/megamek/client/ui/panels/phaseDisplay/commands/MoveCommand.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2025 The MegaMek Team. All Rights Reserved.
+ * Copyright (C) 2025-2026 The MegaMek Team. All Rights Reserved.
```
IllianiBird: Dirty edit. Nothing changed in this class. I recommend doing legal updates as the last step in a PR, that way if you remove a class edit later you don't need to remember to undo the legal statement changes, too.

megamek/src/megamek/client/ui/panels/phaseDisplay/MovementDisplay.java
IllianiBird: This is an unnecessary stream, the number of objects being parsed will be tiny for most games. A traditional loop would be more performative.

megamek/src/megamek/client/ui/panels/phaseDisplay/MovementDisplay.java
```diff
@@ -4929,14 +5222,27 @@ private Entity getDisconnectedUnit() {
...
+        craneUnits.addAll(unitsBeingUnloaded);
         Entity choice = null;
-        if (unloadableUnits.isEmpty()) {
+        if (unloadableUnits.isEmpty() && craneUnits.isEmpty()) {
             LOGGER.error("No loaded units");
-        } else if (unloadableUnits.size() > 1) {
-            // Only show the units we are not already planning to unload
+        } else if ((unloadableUnits.size() + craneUnits.size()) > 1) {
+            // Only show the units we are not already planning to unload, then the units only the cranes can unload
             List<Entity> filteredUnits = unloadableUnits
```
IllianiBird: Unnecessary stream

megamek/src/megamek/common/units/CraneRules.java
```diff
@@ -0,0 +1,413 @@
...
+
+    /**
+     * Checks whether a unit is a Small Craft or DropShip that can work its cranes: on the ground, on the board and not
+     * destroyed.
+     *
+     * @param carrier the possible carrier, may be {@code null}
+     *
+     * @return {@code true} if the unit is a grounded, operational Small Craft or DropShip
+     */
+    public static boolean isGroundedCarrier(@Nullable Entity carrier) {
```
IllianiBird, line 82: A small utility in the Entity class returning false and then in the SmallCraft class returning the below will allow us to use a visitor pattern instead of using `instanceof`. Not super critical, but worth considering.

megamek/src/megamek/server/totalWarfare/CraneOperationHandler.java
```diff
@@ -0,0 +1,373 @@
...
+     */
+    void checkCraneOperations() {
+        // Copies: loading and unloading change transport state while we walk the units and their operations
+        List<Entity> units = new ArrayList<>(getGame().getEntitiesVector());
+        for (Entity entity : units) {
+            if (!(entity instanceof SmallCraft carrier) || carrier.getCraneOperations().isEmpty()) {
+                continue;
+            }
+            List<CraneOperation> operations = new ArrayList<>(carrier.getCraneOperations().getOperations());
+            for (CraneOperation operation : operations) {
```
IllianiBird, line 174: Can this be made n(1)?

megamek/src/megamek/server/totalWarfare/CraneOperationHandler.java
```diff
@@ -0,0 +1,373 @@
...
+            return;
+        }
+        if (!CraneRules.isGroundedCarrier(carrier)) {
+            cancelLoading(carrier, unit, 5356, true, "the carrier is no longer grounded");
+            return;
+        }
+        boolean hasLeftWaitingHex = !operation.getUnitPosition().equals(unit.getPosition());
+        if (hasLeftWaitingHex || (unit.delta_distance > 0)
+              || !CraneRules.isInReach(unit, unit.getPosition(), carrier, getGame())) {
+            cancelLoading(carrier, unit, 5354, true, "the unit moved or is out of the cranes' reach");
```
IllianiBird, line 211: Is this player facing? If so, it shouldn't be hard coded, if not then ignore this feedback

IllianiBird (approved):
Approved pending feedback being reviewed

megamek/src/megamek/client/ui/panels/phaseDisplay/MovementDisplay.java
HammerGS: Fixed in 9a468bbf98. This walks the units in the hexes a carrier can load from, which is a handful in any normal game, so it is now a plain loop that breaks on the first loadable unit. Same result, no pipeline.

I have also put the rule into our PR checklist in HammerGS's words so it is caught before review: streams are reserved for code that can reasonably be expected to handle thousands of objects regularly, otherwise a traditional loop is more performant.

megamek/src/megamek/client/ui/panels/phaseDisplay/MovementDisplay.java
HammerGS: Fixed in 9a468bbf98. Now a plain loop building the filtered list. This one is the units aboard a single DropShip, so it is tens at most. Spotless then removed the Collectors import, which nothing else in the class used.

megamek/src/megamek/client/ui/panels/phaseDisplay/commands/MoveCommand.java
HammerGS: You are right, reverted in 9a468bbf98. The file is back to its state on main; the only change had been the copyright year. Point taken on doing the legal updates as the last step of a PR.

megamek/src/megamek/client/ui/clientGUI/boardview/spriteHandler/CraneOperationSpriteHandler.java
HammerGS: I looked at this and I do not think a map gets us to O(1) here, so I have left it - happy to be told otherwise.

The cost is finding which carriers have crane work at all. Game exposes the full entity vector and position lookups, but no index by type, so something has to walk the units. A map would only help if we kept a registry of carriers with live operations, and that registry then has to stay correct through loading, unloading, cancellation, destruction and save/load. That is real state to keep in sync for a loop over the entity list on a redraw.

If you would rather have the registry, I will do it, but I would want it in its own PR with the sync points tested rather than folded into this one.

megamek/src/megamek/server/totalWarfare/CraneOperationHandler.java
HammerGS: Same answer as the sprite handler, and same reasoning: there is no type index on Game, so finding carriers with operations means walking the entity list. This one runs once per End Phase rather than per redraw, so it is the cheaper of the two.

The inner work is already skipped immediately for any unit that is not a carrier or has no operations, so the per-unit cost is one instanceof and one isEmpty check.

megamek/src/megamek/server/totalWarfare/CraneOperationHandler.java
HammerGS: Not player facing, so ignoring as you suggested. That string is the last argument to cancelLoading, which only puts it into LOGGER.info for the server log. What the player sees comes from unitReport(reportId, unit), so the wording lives in report-messages.properties with the numbered reports.

megamek/src/megamek/common/units/CraneRules.java
HammerGS: Worth considering, agreed, but I have left it as is for this PR.

The reason is a standing rule from HammerGS on this branch: no new logic on Entity or the other superclasses. A canWorkCranes hook returning false on Entity and the real check on SmallCraft is exactly the shape he asked me to avoid, and CraneRules exists to keep these rules out of the unit classes.

If the team would rather have the visitor, it is a small change and I am happy to do it as a follow-up, but I did not want to make that call inside this PR.

megamek/src/megamek/server/totalWarfare/CraneOperationHandler.java
IllianiBird: It's fine as implemented.

megamek/src/megamek/common/units/CraneRules.java
IllianiBird: It's good how it is.

Issue #8973, opened 2026-09-14
[Bug] Client NullPointerException in GhostEntitySprite after a unit mounts a DropShip


HammerGS:
## Problem

When a unit mounts a DropShip, the client fails to process the entity update that follows. The board view tries to animate the unit's move with a ghost sprite at the unit's position, but the unit is now loaded and has no position, so `GhostEntitySprite` throws.

## Example

An Ares Medium Tank walks one hex to stand beside a Triumph (2593) and presses Mount. The client log then shows:

```
ERROR [megamek.client.AbstractClient] {AWT-EventQueue-0}
megamek.client.AbstractClient.handlePacket(AbstractClient.java:441) - Failed to parse Packet command of ENTITY_UPDATE
java.lang.NullPointerException: Cannot read field "x" because "p" is null
	at java.desktop/java.awt.Rectangle.setLocation(Rectangle.java:472)
	at megamek.client.ui.clientGUI.boardview.sprite.GhostEntitySprite.<init>(GhostEntitySprite.java:60)
	at megamek.client.ui.clientGUI.boardview.BoardView.addMovingUnit(BoardView.java:1071)
	at megamek.client.ui.clientGUI.boardview.BoardView$1.gameEntityChange(BoardView.java:548)
	at megamek.common.event.entity.GameEntityChangeEvent.fireEvent(GameEntityChangeEvent.java:115)
	at megamek.common.game.Game.setEntity(Game.java:1557)
	at megamek.client.Client.receiveEntityUpdate(Client.java:721)
```

The mount still happens, but the rest of that entity update is skipped on the client.

## Proposed fix

- In `BoardView.addMovingUnit`, do not create the ghost sprite or the movement animation when the entity has no position (it was loaded during the move).

## Implementation notes

- `addMovingUnit` checks `isOnThisBord(entity)` and that the move path is not empty, but not that `entity.getPosition()` is set.
- Any successful Mount reproduces it. The same path may be hit by other moves that end with the unit leaving the board.

## Background

Found while testing megamek #8589 (mounting a DropShip through a friendly Mek hex), 2026-09-13.

### What you expected instead

The unit disappears into the DropShip with no error.

### Steps to reproduce

1. Place a tank next to a friendly grounded DropShip that has a vehicle bay.
2. Press Mount and pick the DropShip.
3. Open `logs/megamek.log`.

### MegaMek Suite Version

0.51.01 (development build)

### Operating System

Windows 11 (10.0.26200) amd64

[...]

PR #8993, opened 2026-09-16
Fix #8977: spotlessApply broke files that import both System classes

### LINKED ISSUE #8977: [Bug] spotlessApply breaks BoardView.java compilation (ambiguous System import)

HammerGS:
## What this changes

Three files would stop compiling the first time a pull request touched them. They import `java.lang.System` explicitly and also `megamek.common.units.*`, which holds a class called `System`. CI runs `spotlessApply`, which deletes the `java.lang.System` import because `java.lang` is implicit, and every `System.` call in the file is then ambiguous. The calls are now written as `java.lang.System.`, so the files keep compiling after spotless.

Files fixed: `BoardView.java` (4 calls), `MoveStep.java` (1 call), `BulkUnitFileTest.java` (2 calls). `BLKFile.java` and `TROView.java` already qualify their calls and were left alone. Nothing changes in play.

Fixes #8977

## Testing

- `spotlessApply` then `compileJava` pass on all three files. Before the change, spotless removed the import from `BoardView.java` and the compile failed with four "reference to System is ambiguous" errors.
- `checkstyleMain` and `javadoc` are clean.
- `BulkUnitFileTest` passes.
- This already bit two branches: `MoveStep.java` broke on the #8589 branch, and `BoardView.java` broke on the #8973 branch. Both needed this same fix to build.

## What is not proven yet

- No in-game test was run, and none applies: the change only qualifies a class name, so the compiled behaviour is identical.
- Spotless also dropped some unused imports and trailing whitespace in the three files, which CI would have done anyway.

---
- [x] Focused on one issue or RFE
- [x] Tests added or updated, if this implements a rule or changes game state - not applicable, no rule or game state changes; the check is that spotless then compile succeeds


IllianiBird (approved):
Approved, but do we need to adjust spotless to remove this error in future?

HammerGS:
On adjusting spotless: I do not think we should. Spotless is right here. `import java.lang.System;` really is redundant under Java's rules, so `removeUnusedImports` removing it is correct behaviour, and turning that rule off would let genuinely dead imports build up across the codebase.

The actual trap is the collision: `megamek.common.units` contains a class called `System`, and any file that wildcard-imports that package and also calls `java.lang.System` breaks the moment spotless touches it. Three ways to stop it recurring, in increasing order of scope:

1. What this PR does: qualify the calls as `java.lang.System.`, so there is nothing for spotless to remove. Fixes the three files that were already primed to break.
2. Replace the `megamek.common.units.*` wildcard with explicit imports in files that call `java.lang.System`. Our checkstyle allows a wildcard once a file imports more than 10 types from a package, which is what admits these.
3. Rename `megamek.common.units.System`. That removes the trap everywhere and is the only durable fix, but it touches a lot of code and deserves its own PR.

I would leave spotless alone and keep option 1 for now. Happy to open an RFE for option 3 if you want it on the list - say the word and I will write it up.

PR #8994, opened 2026-09-16
Fix #8972: Mounting a DropShip threw a NullPointerException in the dataset logger

### LINKED ISSUE #8972: [Bug] Dataset logger throws NullPointerException after a unit mounts a DropShip

HammerGS:
## What this changes

Mounting a DropShip no longer throws a NullPointerException in the server log. The AI dataset logger replays a finished move to collect its piloting rolls, but a unit that just boarded has no position, so the replay started from a null hex and threw. It now skips the replay for a unit with no position and records the same value it already uses for a move that needs no rolls.

The mount itself is unchanged. Only the dataset row for that move was affected, and the game carried on either way.

Fixes #8972

## Testing

- New `UnitActionFromMovePathTest` (2 tests). `fromMovePathWithLoadedUnitDoesNotThrow` fails with the fix reverted, throwing the exact error from the issue: `Cannot invoke "Coords.equals(Object)" because "lastPos" is null` at `SharedUtility.java:283`.
- `megamek.ai.dataset` package: 53 tests, 0 failures. `spotlessApply` leaves no changes; `checkstyleMain` and `javadoc` are clean.
- Played in game by HammerGS: an Agrotera AGT-1A mounted a grounded Leopard (2537). The move ended with the unit loaded, the log shows `Piloting roll replay skipped: Agrotera AGT-1A has no position (loaded or off board)`, and there is no `GameDatasetLogger` error. The only errors in that log are `MekFileParser: Unsupported file suffix` from the unit cache at startup, which happen on unmodified `main` too.

## What is not proven yet

- Other routes that end a move with no position (carrier recovery, fleeing the board) were read but not played; they reach the same guard.
- A mounting move now records "no piloting rolls" in the dataset. Any roll such a move should have needed is not captured, so AI training data for mounting moves is thin rather than wrong.

---
- [x] Focused on one issue or RFE


megamek/src/megamek/ai/dataset/UnitAction.java
```diff
@@ -153,12 +155,20 @@ public static UnitAction fromMovePath(MovePath movePath) {
...
+        if (entity.getPosition() == null) {
+            LOGGER.debug("Piloting roll replay skipped: {} has no position (loaded or off board)",
+                  entity.getShortName());
+            map.put(Field.CHANCE_OF_FAILURE, 1.0);
+        } else {
+            map.put(Field.CHANCE_OF_FAILURE,
+                  SharedUtility.getPSRList(movePath)
+                        .stream()
+                        .map(psr -> psr.getValue() / 36d)
+                        .reduce(1.0, (a, b) -> a * b));
```
IllianiBird: A stream doesn't seem warranted here. We should be reserving streams for when we have thousands of objects, otherwise it's less performative than a traditional loop.

IllianiBird (approved):
Approved, but I think we should revise the stream usage.

megamek/src/megamek/ai/dataset/UnitAction.java
HammerGS: Agreed and fixed in d981be58b3. A move produces a handful of piloting rolls, nowhere near thousands, so the stream is now a plain loop that multiplies each roll's target number over 36 into the running total. Same value, no pipeline setup.

I have also written the rule into our PR checklist in your words, so it gets checked before the PR rather than at review: streams are reserved for code that can reasonably be expected to handle thousands of objects regularly, otherwise a traditional loop is the more performant choice.

megamek/unittests/megamek/ai/dataset/UnitActionFromMovePathTest.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    @Test
+    void fromMovePathWithUnitOnBoardStillReplaysPilotingRolls() {
+        MovePath movePath = createTankPath();
+
+        UnitAction unitAction = assertDoesNotThrow(() -> UnitAction.fromMovePath(movePath));
+
+        assertNotNull(unitAction.get(UnitAction.Field.CHANCE_OF_FAILURE));
+    }
```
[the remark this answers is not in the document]
HammerGS: Good catch, and correct: both tests passed on a value the code always sets, so the on-board test would not have failed if the replay were removed. Fixed in d981be58b3.

Both tests now use Mockito static mocking of SharedUtility, as you suggested:
- The loaded case verifies getPSRList is never called, and asserts the recorded chance is 1.0. Without the guard that call happens and throws, which is the bug this PR fixes.
- The on-board case stubs a known two-roll list (targets 8 and 9), then asserts the recorded value equals (8/36) * (9/36), so the replay result is actually checked rather than just its presence.

PR #8998, opened 2026-09-16
Fix #8997: Modified the check for capped explosion damage to be inclusive of cap variable

### LINKED ISSUE #111: The Effects of Fire

### LINKED ISSUE #1234: Feature request : Option to toggle "next target" behaviour when clicked

### LINKED ISSUE #5678: RFE: ability too use multiple QTY with the tier system in <awards>
This post ties into that my previous RFE for rank and would effectively make rank images a thing in mhq IMO. Perhaps I should of put both together I guess but they are separate improvements too me and this one will probably be harder too implement than just <range>implicit</range>, so i decided it best too make 2 separate post
==
Add the ability too use multiple quantities with the excellent tier system , for award. It would probably be useful too have this trigger automatically after the rank is changed though

Now, this might prove useful for other awards as well, but im mainly thinking of the <item>ranks</ranks>. 

Not sure how too put this in words, so an example
<award>
        <name>RANKS</name>
        <description></description>
        <group>SLEEVES_RANKS</group>
        
		<misc>RANKS_E0.png</misc>
		<misc>RANKS_E2.png</misc>
		<misc>RANKS_E4.png</misc>
		<misc>RANKS_E6.png</misc>
		<misc>RANKS_E8.png</misc>
		<misc>RANKS_E10.png</misc>
		<stackable>false</stackable>
		
		<qty>0,2,4,6,8,10</qty>
		<item>rank</rank>
		<range>implicit</range>
    </award>
    
    KG_BRANDENBURG

### LINKED ISSUE #8997: Incorrect Handling of Armor Blowout When Explosion Damage Is Exactly 20
### What went wrong

An internal explosion causing exactly 20 damage resulted in the "armor blown out" effect despite surviving internal structure.

Image

### What you expected instead

Armor blowout effect only occurs when explosion damage is actually capped by the new rule. An explosion of exactly 20 damage is not capped per p.96 of the core rulebook and armor should not be blown out when internals survive.

### Steps to reproduce

1. Receive an internal explosion of exactly 20 damage which also leaves surviving internal structure.

### Attach your files

[armorBlowoutBug.sav.gz](https://github.com/user-attachments/files/32294662/armorBlowoutBug.sav.gz)

### Relevant log lines

```text

```

### MegaMek Suite Version

0.51.00.1

### Operating System

Windows 10 10.0 (amd64)

### Java Version

Eclipse Adoptium 21.0.12.1

### Before you submit

- [ ] I have asked about this on the MegaMek Discord (optional)

viper-42:
## What this changes
Modifies the damage-to-cap comparison to be inclusive of cap variable.

Armor will no longer blow out if you experience an ammo explosion of 20 damage and have surviving internal structure.

Fixes #8997

## Testing
I reran the same scenario and my armor did not blow out this time.

## What is not proven yet

Nothing that I can think of

---

- [X ] This PR is focused on one issue or RFE. Large refactoring or accessibility work is in its own PR
- [X ] Every file in the diff has a deliberate change (no stray formatting, no unrelated files)
- [ ] Dev team only: if AI tools were used, the **AI Assisted Development** label is applied, and I can
      
P.S. This is my first time contributing so let me know if I left anything out that I need to fix. And if I broke some GitHub etiquette I apologize as I was not aware.


IllianiBird (approved):
Code is approved, but reserving merge action for Roundtop as Core is their project.

IllianiBird:
Tests are failing due to a GitHub issue. Will need to be re-run.

TenkawaBC:
I checked. The code effectively handled this already later on, as the blowout was gated, however it is always good to return early.

if (mek.getInternal(loc) > cap) { // Location survives, blow out armor

Approved to come in, but it needs a .\gradlew spotlessApply

IllianiBird:
Viper, before we can merge we need you to run the 'spotlessApply' megamek gradle action

viper-42:
> I checked. The code effectively handled this already later on, as the blowout was gated, however it is always good to return early.
> 
> if (mek.getInternal(loc) > cap) { // Location survives, blow out armor
> 
> Approved to come in, but it needs a .\gradlew spotlessApply

This addressed #8997 by ensuring that damage exceeds the cap. For example `if (mek.getInternal(loc) > cap) { // Location survives, blow out armor` would blow out the armor of a mech with 21 internal structure on a 20 damage ammo explosion even though that does not exceed the cap. In other words `mek.getInternal(loc) > cap` needs to be true and `damage <= cap` needs to be false for armor blow out to apply. At least that is my interpreation of the rules as written.

PR #9001, opened 2026-09-16
Fix: Modular Armor in Body

WellBattle6:
## What this changes

Prevent modular armor from being mounted in body (since it's not a hit location)

## Testing

Tested in build

---

- [ ] Dev team only: if AI tools were used, the **AI Assisted Development** label is applied, and I can


IllianiBird (approved):
Code-wise there are no issues, however I am reaching out to Mishra to confirm rules compliance before I merge

IllianiBird:
Rules compliance confirmed by Mishra. Merging

PR #9012, opened 2026-09-18
Fix MegaMek/mekhq#10002: Say that generation accuracy varies by faction

HammerGS:
## What this changes

The faction tooltip in the Force Generator now says that generation accuracy varies by faction.

It already explained that the faction's ruleset drives echelon structure, ranks, weight distribution and availability, but not that some factions have far less of that data than others. A player generating a command for a minor, obscure or short-lived faction had no way to know the result leans on generic rules rather than on how that faction was really organised.

The tooltip is shared with the Random Army dialog, where the same caveat applies.

Part of MegaMek/mekhq#10002

## Testing

- `checkstyleMain` and `spotlessCheck` are clean.
- The edited line is a single well-formed entry, with the added text inside the existing `</html>`.
- No test covers message text, so there is nothing further to run.

## What is not proven yet

- The tooltip has not been read on screen at GUI scale, only in the properties file.

---

- [x] Tests added or updated, if this implements a rule or changes game state - not applicable, this is tooltip text
- [x] Javadoc literals use `{@code true}` / `{@code null}` rather than bare or quoted text - not applicable, no javadoc in this diff
- [x] Dev team only: if AI tools were used, the **AI Assisted Development** label is applied, and I can explain and have verified the result


HammerGS:
Superseded by #9014, which carries this commit plus the fix that makes the size default actually work (#9013). Both belong to the same MekHQ report (MegaMek/mekhq#10002) and the same dialog, and keeping them in one PR means CI runs rather than being skipped for a stacked base.

PR #9027, opened 2026-09-20
Gamemaster can set unit states, explode equipment and blow off limbs in Edit Damage

HammerGS:
## What this changes

The gamemaster's Edit Damage dialog can set a unit's states, explode equipment and blow off limbs as play would, and a gamemaster act that kills a unit mid-turn no longer stalls the game.

- Jammed, Fired and Mount Locked switches on the weapons they apply to.
- Breached and Blown Off switches per Mek location.
- Explode button on explosive equipment, also as /explode.
- Damage Summary on the General panel, each line opening its location.
- Restore Unit brings back blown-off locations.
- Ejection settings on the Head panel, overriding the lobby in play.
- Target Modifier spinner and /targetMod: a this-round change to the target movement modifier, floored at what the unit earned and capped at the table top.

## Testing

- Unit tests cover every switch, restore, serialisation, the explode handler, the modifier clamp and the GM guards.
- Playtested: jam, breach, explode with and without ejection, Restore Unit, target modifier up and down.
- Full suite, Spotless and Checkstyle green after merging main.
- QA steps are in the first comment.

## What is not proven yet

- Princess does not see the target modifier when it picks targets.
- HVAC and hyper laser explosions are tested only through the ammo path.

---


HammerGS:
QA steps. Every gamemaster action on this branch logs at INFO, so megamek.log needs no settings change; grep it for `[EquipState]`, `[Explode]` and `[GMTurn]`.

Setup: one human as gamemaster, a Mek with an autocannon, an LRM rack, a one-shot launcher if available, and a lance-mate.

1. Open Edit Damage. The General panel shows Damage Summary: None. Every switch on an equipment row sits level with its crit boxes.
2. Jammed: an autocannon row has a Jammed box, an LRM rack does not. Tick it, Okay, reopen: the row reads `[Jammed]`, the summary lists it, and clicking the line opens that location. The gun cannot fire. Untick it, Okay: the gun fires.
3. Fired: on a one-shot launcher tick Fired, confirm it cannot fire, untick it, confirm it can.
4. Breached: on a torso panel tick Breached, Okay: the summary lists the torso and its weapons show `[Breached]` and cannot fire. Untick it: they fire again.
5. Blown Off: on an arm panel tick Blown Off, Okay: the paperdoll shows the arm gone and its weapons are unavailable. Reopen, untick it, Okay: the arm is back whole with full structure and armor.
6. Explode, lone unit: with the Mek as its player's only unit, during its own movement turn press Explode on a loaded ammo bin, confirm. The dialog closes, chat reports the bang, the turn passes to the next player, and the next report phase shows the damage.
7. Explode, lance-mate: same again with a second unit of the same player. The lance-mate is selected and movable, and the log shows `[GMTurn]` resending the turn.
8. Explode, weapon: press Explode on an HVAC or RISC hyper laser. The report should read like a crit-triggered explosion of the same weapon.
9. Restore Unit after an explosion blew a limb off, Okay: the limb is back on the paperdoll and its weapons usable.
10. /kill a unit during its own turn from the Commands menu: the game moves on without a /skip.
11. Firing phase: repeat step 6 during the unit's firing turn.
12. Ejection: on a Mek's Head panel tick Disable Automatic ejection, Okay, then explode its ammo; the pilot must stay in. With conditional ejection on, the trigger boxes appear too.
13. Target Modifier: in the firing phase pick an enemy Mek that walked 5 hexes and open Edit Damage. The Target Modifier row says the unit earned +2 by moving. Set +3, Okay: the attacker's to-hit breakdown updates without reselecting and shows a "gamemaster target modifier +3" line, +5 in all. Set +10: the total stops at +7 (+8 with MaxTech movement mods). The spinner will not go below -2 on this unit; set -2 and the to-hit drops by exactly 2. A unit that did not move offers no reduction. Next round it is gone.

megamek/src/megamek/server/totalWarfare/GamemasterTurnUpkeep.java
```diff
@@ -0,0 +1,158 @@
...
+            LOGGER.info("[GMTurn] {} can no longer act; its pending turn was dropped", entity.getDisplayName());
+            gameManager.send(gameManager.getPacketHelper().createTurnListPacket());
+        }
+        // The owner's client may still have the dead unit selected for the current turn, and nothing else tells
+        // it to move on: a fresh turn packet makes every client start the turn over and pick a unit that can act.
+        if (currentTurn.playerId() == entity.getOwnerId()) {
+            LOGGER.info("[GMTurn] {} can no longer act during its owner's turn; resending the turn so the owner's"
+                  + " client selects another unit", entity.getDisplayName());
+            gameManager.send(gameManager.getPacketHelper().createTurnIndexPacket(entity.getOwnerId()));
+        }
```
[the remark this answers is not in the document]
HammerGS: Agreed. The current turn was null-checked for the skip path and then dereferenced for the resend. Guarded in 24897ff940.

megamek/src/megamek/server/totalWarfare/GamemasterTurnUpkeep.java
```diff
@@ -0,0 +1,158 @@
...
+        if (!entity.isDone() && (specificTurnsRemoved == 0)) {
+            turnsChanged = dropOwnersLastTurnFor(game, entity);
+        }
+        if (turnsChanged) {
+            LOGGER.info("[GMTurn] {} can no longer act; its pending turn was dropped", entity.getDisplayName());
+            gameManager.send(gameManager.getPacketHelper().createTurnListPacket());
+        }
+        // The owner's client may still have the dead unit selected for the current turn, and nothing else tells
+        // it to move on: a fresh turn packet makes every client start the turn over and pick a unit that can act.
+        if (currentTurn.playerId() == entity.getOwnerId()) {
```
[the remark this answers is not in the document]
HammerGS: Fixed in 24897ff940: the resend now checks the current turn for null as the skip path already did.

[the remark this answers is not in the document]
HammerGS: Fixed in 24897ff940: the resend now checks the current turn for null as the skip path already did.

megamek/src/megamek/common/units/DamageEditApplier.java
```diff
@@ -479,6 +491,172 @@ private void applyBuildingCriticalState() {
...
+                continue;
+            }
+            // allowChange: a breached location is otherwise pinned, which is what a gamemaster is here to undo
+            entity.setLocationStatus(location,
+                  breached ? ILocationExposureStatus.BREACHED : ILocationExposureStatus.NORMAL, true);
+            for (Mounted<?> mounted : entity.getEquipment()) {
+                if (mounted.getLocation() == location) {
+                    mounted.setBreached(breached);
+                }
+            }
```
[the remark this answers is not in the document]
HammerGS: Taken. The engine's own breach resolution marks the primary location only, but the crit slots here were already breached from either location, so the mount now follows the same split-aware rule. 24897ff940.

megamek/src/megamek/server/totalWarfare/TWGameManager.java
```diff
@@ -421,6 +422,29 @@ public HexEditHandler hexEditHandler() {
      * @return the handler that applies a gamemaster's edits to a building, created on first use for the same reason as
      *       {@link #lobbyBoardHandler()}
      */
+    /**
```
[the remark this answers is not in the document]
HammerGS: Correct, and my miss: the delegators went in between the handler's Javadoc and its method. The Javadoc is back on buildingEditHandler() in 24897ff940.

megamek/unittests/megamek/server/commands/ExplodeEquipmentCommandTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+    /** The connection the command is treated as arriving on. */
+    private static final int SENDING_CONNECTION = 1;
+
+    private final ExplodeEquipmentCommand command = new ExplodeEquipmentCommand(null, null);
+
+    @Test
+    void itIsAGamemasterCommand() {
+        // GamemasterServerCommand is what refuses a sender who does not hold the role; being one is the guard
+        assertTrue(command instanceof GamemasterServerCommand,
```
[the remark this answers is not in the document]
HammerGS: Removed in 24897ff940. The guard itself is covered by the two preRun tests in the same class, which exercise a gamemaster and a non-gamemaster sender.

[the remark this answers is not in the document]
HammerGS: Removed in 24897ff940. The guard itself is covered by the two preRun tests in the same class.

x-Lanceman-x:
These all look good. The only thing I noticed, and it might be intentional, is that hitting Restore Unit doesn't uncheck the Breached or Blown Off checkboxes for locations.  

1 - Pass
2 - Pass
3 - Pass
4 - Pass
5 - Pass
6 - Pass
7 - Pass
8 - Pass
9 - Pass
10 - Pass
11 - Pass
12 - Pass

HammerGS:
Thanks for the full pass. The Restore Unit gap was not intentional: it reset the spinners and crit boxes and never learned about the new switches. As of 30145ef566 Restore also unticks Breached, Blown Off, Jammed, Fired and Mount Locked (and a building's turret lock and gunners). The settings switches, ejection, burst fire, hot-loading and modes, are deliberately left alone since they are how the unit is set up rather than damage.

megamek/src/megamek/client/ui/dialogs/unitEditor/UnitDamagePanelBuilder.java
```diff
@@ -237,6 +256,72 @@ public void addSkillModifiersColumn() {
...
+        startNewColumn(panel);
+        JLabel header = new JLabel("<html><b>" + Messages.getString("UnitEditorDialog.ejection") + "</b></html>");
+        header.setToolTipText(UIUtil.formatSideTooltip(Messages.getString("UnitEditorDialog.ejection.tooltip")));
+        addRow(panel, header, new JLabel());
+
+        boolean isConditionalEjectionInPlay = entity.getGame().getOptions()
+              .booleanOption(OptionsConstants.RPG_CONDITIONAL_EJECTION);
+        String disabledTooltip = "UnitEditorDialog.ejection.autoDisabled.tooltip";
+        String conditionalTooltip = "UnitEditorDialog.ejection.conditional.tooltip";
+        if (entity instanceof Mek mek) {
```
IllianiBird, line 280: This will return true for ProtoMeks, is that intentional?

megamek/src/megamek/client/ui/dialogs/unitEditor/UnitDamagePanelBuilder.java
```diff
@@ -237,6 +256,72 @@ public void addSkillModifiersColumn() {
...
+                controls.chkConditionalEjectEngine = ejectionRow(panel,
+                      "CustomMekDialog.labConditional_Ejection_Engine", mek.isCondEjectEngine(), conditionalTooltip);
+                controls.chkConditionalEjectCenterTorso = ejectionRow(panel,
+                      "CustomMekDialog.labConditional_Ejection_CT_Destroyed", mek.isCondEjectCTDest(),
+                      conditionalTooltip);
+                controls.chkConditionalEjectHeadshot = ejectionRow(panel,
+                      "CustomMekDialog.labConditional_Ejection_Headshot", mek.isCondEjectHeadshot(),
+                      conditionalTooltip);
+            }
+        } else if (entity instanceof Aero aero) {
```
IllianiBird, line 301: This will return true for things like DropShips, JumpShips, and even space stations. Is this intentional?

megamek/src/megamek/client/ui/dialogs/unitEditor/UnitDamagePanelBuilder.java
```diff
@@ -640,14 +809,197 @@ private void initEquipCrits() {
...
+     * it sits in. The location panels show one location at a time, so without this list a gamemaster looking for
+     * the jammed gun would have to page through every location to find it.
+     */
+    private void initDamageSummary() {
+        if (!offersStateSwitches()) {
+            return;
+        }
+        JPanel summary = new JPanel();
+        summary.setLayout(new BoxLayout(summary, BoxLayout.PAGE_AXIS));
+        if (entity instanceof Mek) {
```
IllianiBird, line 960: Same query as before, regarding Protos

megamek/src/megamek/common/equipment/WeaponMounted.java
```diff
@@ -87,6 +88,33 @@ public void setDisposableWeapon(boolean disposableWeapon) {
...
+            return true;
+        }
+        WeaponType weaponType = getType();
+        boolean isAmmoFed = weaponType.getAmmoType() != AmmoType.AmmoTypeEnum.NA;
+        if (isAmmoFed && hasQuirk(OptionsConstants.QUIRK_WEAPON_NEG_AMMO_FEED_PROBLEMS)) {
+            return true;
+        }
+        boolean isPrototype = weaponType.hasFlag(WeaponType.F_PROTOTYPE);
+        return switch (weaponType.getAmmoType()) {
+            case AC, LAC, AC_IMP, PAC, AC_ULTRA, AC_ULTRA_THB, AC_ROTARY, HYPER_VELOCITY -> true;
```
IllianiBird, line 112: I think we need to have a cleaner implementation here. Can we add a CAN_JAM flag to all relevant weapons and call that instead? I'd be happy for that to be in a follow-up PR, as it'll touch on other code, but I do think it'd be a really good idea.

IllianiBird (commented):
I have some concerns about this.

megamek/src/megamek/client/ui/dialogs/unitEditor/UnitDamagePanelBuilder.java
HammerGS: Not for ProtoMeks: ProtoMek extends Entity directly, not Mek, so `entity instanceof Mek` is false for them. The guard above it, `AutomaticEjectionRules.hasEjectionSystem`, is also false for a ProtoMek, so no ejection column is added. c919aa5661 adds a test pinning that down.

megamek/src/megamek/client/ui/dialogs/unitEditor/UnitDamagePanelBuilder.java
HammerGS: Good catch, that was not intentional. SmallCraft and Jumpship both override `isAutoEject()` to always return false, and the lobby Configure dialog excludes them from the ejection settings, so a DropShip was getting a column for a switch that cannot be turned on. c919aa5661 changes the shared rule, `AutomaticEjectionRules.hasEjectionSystem`, to say no for small craft, DropShips, JumpShips, WarShips and space stations, which also stops the deployment warning listing them. Tests added for a fighter (yes) and the three large craft (no).

megamek/src/megamek/client/ui/dialogs/unitEditor/UnitDamagePanelBuilder.java
HammerGS: Same answer: ProtoMek is not a Mek subclass, so this branch does not run for them. That is what we want here, since a ProtoMek has no breached or blown-off location state to summarise; its equipment still goes through the loop below.

megamek/src/megamek/common/equipment/WeaponMounted.java
HammerGS: Agreed, a CAN_JAM flag on the weapon types would be cleaner than deriving it from the ammo type here, and it would let the attack handlers use the same source of truth. I will do that as a follow-up PR so this one stays to the GM editor; this method becomes a one-line read of the flag when it lands.

IllianiBird (approved):

PR #9030, opened 2026-09-20
Fix generator messages

AlexOldenburger:
## What this changes

When generating a formation that has reinforced formation options, the dropdown text is garbled. This PR fixes this.

## Testing

Manual testing:

Before:
ForceSelection

After:
ForceSelectionFixed

## What is not proven yet

Testing if generating other reinforced/understrength formations works as intended now
---

- [X] This PR is focused on one issue or RFE. Large refactoring or accessibility work is in its own PR
- [X] Every file in the diff has a deliberate change (no stray formatting, no unrelated files)


megamek/resources/megamek/client/messages.properties
```diff
@@ -2637,7 +2637,7 @@ FiringDisplay.OverheatNag.title=Allow overheating?
 FiringDisplay.OverheatNag.message=Do you really wish to overheat? This will require a control roll.
 HonorNag.title=Dishonorable action?
 HonorNag.intro=These attacks would break the honor rules:
-HonorNag.bullet=  - {0}
+HonorNag.bullet=- {0}
```
IllianiBird: This doesn't look intentional

IllianiBird (commented):
One small issue that needs to be resolved before this can be merged

AlexOldenburger:
This was not intentional, I think it was due to automatic whitespace editing? It should be gone now.

IllianiBird (approved):

